PR #8, opened 2016-04-28
PEP 8 on hopper.py

iandanforth:
Opening this more as a question around how strict the project intends to be and if there are any non PEP 8 style things you want to enforce.


iandanforth:
Build failure is tied to mujoco key retrieval.

gdb:
Answer: not strict at all. 
In my experience, it's tempting to enforce style since it's easy to measure, but it results in a higher barrier to ensure and the readability + quality improvements aren't significant enough. That being

gdb:
(I will take this patch though, since it makes things more readable!)

PR #13, opened 2016-04-28
Fix readme example, documentation and error messages

dommueller:
Small fixes that I noticed while setting up gym and going through the readme.

Probably, the version tag is also needed for the MuJoCo example. However, I couldn't verify since I don't have a license.


gdb:
Thank you!

PR #20, opened 2016-04-29
Fix victory conditions

aforr:
Old victory condition was "player is white and white wins." Added "player is black and black wins" and "opponent resigned."


hojonathanho:
Thanks! I think we also need to check that the agent resigned too. I'll add that.

PR #22, opened 2016-04-29
Python 3 compatiblity

jonasschneider:
I made a first pass over the code and also fixed compat-related issues in our dependencies.
Our tests pass on this, but they don't actually cover every little thing.
So any testing and review is much appreciated!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,5 @@ mujoco-bundle
 rllab_mujoco
 
 tutorial/*.html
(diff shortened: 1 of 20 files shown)


MadcowD:
Can I ask what happened to this?

jonasschneider:
We merged this! \o/
Are you having problems running Gym under Python 3? If so, please open an issue with the error!

PR #24, opened 2016-04-29
Add CartPoleHard: fork of CartPole without velocity observations

sherjilozair:
CartPoleHard makes the observation non-Markovian by removing the velocity components. Thus, the ideal action distribution is now a function of the history of states, making this task more interesting. Solutions now needs to incorporate memory.

### DIFF
--- a/gym/envs/__init__.py
+++ b/gym/envs/__init__.py
@@ -58,6 +58,13 @@ register(
 )
 
 register(
+    id='CartPoleHard-v0',
+    entry_point='gym.envs.classic_control:CartPoleHardEnv',
+    timestep_limit=200,
+    reward_threshold=195,
+)
+
+register(
     id='MountainCar-v0',
     entry_point='gym.envs.classic_control:MountainCarEnv',
     timestep_limit=200,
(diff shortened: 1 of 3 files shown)


dommueller:
I agree that if there is a normal cart pole there should be a non-Markovian one. I already thought about making a similar pull request. And I would also like to see double pole carts. However, if I may say so, I think CartPoleHard is not a very descriptive name and maybe something like CartPoleNonMarkovian would be better. Hard could describe any dimension (more poles, more steps needed to solve, no delta information, etc.).

sherjilozair:
I agree. I think people also use the term partially observed to refer to such problems. [1] has some interesting partially observed tasks.

@joschu @gdb, any interest in partially observed versions of tasks like CartPole? If yes, how should we name them?

[1]: http://arxiv.org/abs/1512.04455

joschu:
Hi Sherjil, I think it would be great to add some partially observable tasks. But let's try to choose tasks that require non-trivial state estimation. I think it'll take some experimentation to find good ones. Maybe we could have another "partially observed control" category of environments. If you're interested in this, want to create a repo for those experiments? I'll be happy to contribute to it and run my algs to test out the environments.

You might want to also check out the partially observed tasks in the recent benchmark by my colleagues and me: http://arxiv.org/pdf/1604.06778.pdf

sherjilozair:
Got it. Thanks! I'll close this PR for now.

PR #25, opened 2016-04-29
Fix warning being raised incorrectly

sherjilozair:
The warning is being raised at the function call which returns `done=True`. This fixes it to raise the warning if the user calls `step` _after_ `done=True` is returned.


gdb:
Whoops, thanks!

PR #26, opened 2016-04-29
add possibility of using stateless custom policy for GoEnv

sergii-gavrylov:
For self-play training, it is necessary to have a possibility to set custom opponent policy. It could be done just by reassigning attribute `opponent` in the GoEnv instance, but it is better to do this explicitly rather than implicitly.


hojonathanho:
Thanks for the patch. I think the best way to implement this is to have two distinct env modes: self play, and play against an opponent. I'll get started on implementing this.

gdb:
(Closing this PR for now.)

PR #29, opened 2016-04-29
Add quotes to pip install packages

wb14123:
Since these package names include special chars like ".[]", many shells
will try to match files, like this:

zsh: no matches found: .[all]


gdb:
Thanks!

PR #39, opened 2016-05-01
update mujoco and bump version

joschu:
- I reorganized things a bit to reduce code duplication, and now qpos/qvel are only set through set_state, which also calls `forward` and `_update_subtree`.
- resetData seems to make a numerical difference. I'm doing the ctypes stuff in mujoco_env for now, but we can later push that code into an update to mujoco-py
- I added some tooling for checking if the environments have changed. See the misc directory. All you have to do is run check_envs_for_change.py and it will compare HEAD against current code

@hojonathanho @dementrock could you sanity check mujoco changes?

### DIFF
--- a/gym/envs/mujoco/mujoco_env.py
+++ b/gym/envs/mujoco/mujoco_env.py
@@ -1,24 +1,38 @@
-import os.path
+import os
 
+from gym import error, spaces
 import numpy as np
+from os import path
 import gym
 import six
-from gym import error, spaces
 
 try:
     import mujoco_py
 except ImportError as e:
     raise error.DependencyNotInstalled("{}. (HINT: you need to install mujoco_py, and also perform the setup instructions here: https://github.com/openai/mujoco-py/.)".format(e))
 
-BIG=10000
+
+# move this into mujoco-py next time we upgrade it!!! 
+# ---------------------------------------------------
+from mujoco_py.mjlib import mjlib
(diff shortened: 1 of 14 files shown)


gym/envs/mujoco/mujoco_env.py
```diff
@@ -30,6 +44,57 @@ def __init__(self, model_path, frame_skip):
...
+        Optionally implement this method, if you need to tinker with camera position
+        and so forth.
+        """
+        pass
+
+    # -----------------------------
+
+    def _reset(self):
+        ob = self.reset_model()
+        # mjlib.mj_resetData(self.model.ptr, self.data.ptr)
```
hojonathanho: Is this supposed to be commented out? Also, I can't say for sure, but I would think that `mj_resetData` should go before `reset_model`

Also, take a look at http://www.mujoco.org/book/file/simulate.cpp
It contains the following code for resetting a simulation:

```
case GLFW_KEY_BACKSPACE:            // reset
        mj_resetData(m, d);
        mj_forward(m, d);
        break;
```

Maybe the canonical way to reset is to call these two functions in sequence.

joschu:
OK, I've moved resetData before the robot DOFs are set. 
So now it's
1. resetData
2. reset_model() => sets degrees of freedom, which automatically calls forward()
3. update viewer

PR #42, opened 2016-05-01
add agentnet demos

justheuristic:
So, here was a 7-line description of the AgentNet library... that somehow crashed the build?


gdb:
Sorry, our Py3 tests currently break non-deterministically :(. We'll fix or temporarily silence it soon.

PR #44, opened 2016-05-02
added missing import

lukemetz:
The error checking causes a odd error due to `error` not being imported.


jonasschneider:
Whoops! Thanks, merged!

PR #47, opened 2016-05-03
Set restriction on selected actions for MountainCar-v0

JKCooper2:
Restricts accepted actions for MountainCar-v0 to be from the set (0, 1, 2) throwing a ValueError if the submitted action is outside of that range

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -78,6 +78,9 @@ class Env(object):
         done (boolean): whether the episode has ended, in which case further step() calls will return undefined results
         info (dict): contains auxiliary diagnostic information (helpful for debugging, and sometimes learning)
         """
+        if not self.action_space.contains(action):
+            raise error.InvalidAction('Invalid action selected: {}'.format(action))
+
         self.monitor._before_step(action)
         observation, reward, done, info = self._step(action)
         done = self.monitor._after_step(observation, reward, done, info)
--- a/gym/envs/classic_control/mountain_car.py
+++ b/gym/envs/classic_control/mountain_car.py
@@ -31,8 +31,6 @@ class MountainCarEnv(gym.Env):
 
     def _step(self, action):
         # action = np.sign((self.state[0]+math.pi/2) * self.state[1])+1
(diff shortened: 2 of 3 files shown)


gdb:
Cool. Rather than throwing a `ValueError` though, I think we should add an `InvalidAction` exception to error.py, and throw that here.

jonasschneider:
This check would be generic enough to do in `core.py`. Thoughts on that? We discussed oddly-shaped numpy arrays before, but I'd consider it undefined behavior to `step` with actions not `contain`ed in the `action_space`.

Maybe we can add a generic warning to `core.py`, and raise as a special case for discrete action spaces (where passing in something illegal ~always indicates breakage)?

gdb:
Good point, I'd forgotten we now have hooks there. I think yes, what you suggest in the right way to go.

JKCooper2:
I've moved the action validation into core.py and added a specific exception to error.py. Was this the kind of thing you were meaning?

As long as all action spaces implement .contains() which is currently required then it will be valid for all scenarios

jonasschneider:
This looks good! Though I think we'll change the raised exception to just be a warning, for backwards compatibility. Thanks!

JKCooper2:
Would warnings require logger to be imported into core? I can only seen warn used with logger.warn()
Also if it's a warning should it still pass the invalid action through to the environment? It doesn't feel right to make it randomly choose a valid action to ensure the environment can operate effectively

jonasschneider:
Yeah, we definitely don't want to silently change the values that are passed through. Therefore we'll just try to "keep trucking on" with the invalid value -- if it breaks, the warning will guide the user to the fix.
https://github.com/openai/gym/commit/4519a6ffe377ccf370e507c705709bc673c904d5 shows how I imagine this -- does this look reasonable?

JKCooper2:
Yes that seems fine. Most agents will break from invalid values in any case and can provide customised reasons in there. At least this enforces awareness of invalid action selection over all environments.

PR #48, opened 2016-05-03
Close window (Alt-F4) and one trivial bug in render

olegklimov:
I'm enjoying the possibility to close window for some time now :)  No problems so far, want to share it with others.

And a trivial bug in `rendering.py`.

### DIFF
--- a/gym/envs/classic_control/rendering.py
+++ b/gym/envs/classic_control/rendering.py
@@ -104,7 +104,7 @@ class Viewer(object):
 
 def _add_attrs(geom, attrs):
     if "color" in attrs:
-        geom.set_color(attrs["color"])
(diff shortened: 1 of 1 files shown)


jonasschneider:
Thanks for bringing this up. I understand that `SystemExit` is the behaviour you want if you're just playing around with e.g. new environments or agents. However, consider that many environments also open windows during normal video recording during long-time training. In that case, raising SystemExit is not really a nice thing to do within any library code. (Maybe you can set up the exit-on-close in your other branch's `example/agents/keyboard.py` instead?)

I think the best way to move forward here is to make sure that the window can be closed without side effects, while at the same time making sure that `get_array` still works (and is up-to-date) after the window has been closed.

PR #51, opened 2016-05-03
Minor bugs

ppaquette:
#1 - Mujoco tests were triggered even if mujoco key is not present (e.g. on development environment)
#2 - FrozenLake directions on top of page do not match directions printed when not slippery


ppaquette:
#1 is likely incorrect. It's probably better to test for the value of MUJOCO_KEY than to check for a pull request

Scenarios:
- Main repo, not PR, with MUJOCO_KEY -> don't skip
- PR with no MUJOCO_KEY -> skip
- Local env (no PR) with no MUJOCO_KEY -> skip

PR #52, opened 2016-05-03
Added ViZDoom environments

ppaquette:
Added 9 doom environments, based on ViZDoom

Depends on 'doom-py' (https://github.com/ppaquette/doom-py) which is not yet published


gdb:
Awesome! Thanks so much for doing this.

We'll poke around and get back to you shortly.

wojzaremba:
This stuff looks sooooooo good. We will definitely have a look on it very soon.

JKCooper2:
I'm getting an error when running: pip install -e '.[doom]' inside the gym folder:
>  Could not find a version that satisfies the requirement doom-py>=0.0.1 (from gym==0.0.7) (from versions: ) No matching distribution found for doom-py>=0.0.1 (from gym==0.0.7).
Where should I install doom-py from?
VizDoom looks amazing. I wonder if it could be used for things like model generalisation over different screen dimensions

gdb:
I've broken this into a separate plugin package for now: https://github.com/openai/gym-doom, so it's easy for people to play with until. Once people feel good about the tasks, we'll either bring it back into core, or maybe decide that plugins are actually the way to go.

gdb:
Closing in favor of #92. We should have these out soon. cc/ @jietang

Issue #14, opened 2016-04-28
Add Hex


Botev:
Since this as also a connection game, but a simpler than Go, yet hard enough on bigger boards (hard enough so no efficient algorithm exists (up to my knowledge) for boards on 11+ size.

wojzaremba:
This sounds great.
Feel free to create a new environment. We are happy to merge it.

Botev:
could you provide a bit more documentation of what is the required interface for the environment API? I've checked the Go version, but it heavily relies on Pachi, which does not make it a good example of a standalone environment.

wojzaremba:
Environment has to implement three functions: 
def _step(action)
def _reset()
def _render()

You can have a look on Cartpole implemention, which is self contained: 
https://github.com/openai/gym/blob/master/gym/envs/classic_control/cartpole.py

Botev:
Could someone take a look, as I'm working on weird environment, and have some restrictions, which is why I don't have the Gym yet installed, but the general implementation should be ok:
https://github.com/Botev/gym

joschu:
Feel free to make a PR when you have something working, but the Issues list isn't for works in progress, so I'm closing this.

gdb:
(If you re-open this instead as a pull request, it'll be much easier for us to review!)

PR #53, opened 2016-05-03
Add Hex 

Botev:
As suggested in #14 . Currently support only 'ansi' interface (e.g. command line).
Things to discuss:

* Is the board representation reasonable? Better suggestions?
* More efficient checks whether the game is finished?
* Human rendering. 

Further extensions:
* Integrate MoHex and other code from the researchers from [University of Alberta Computer Hex Research Group ](https://webdocs.cs.ualberta.ca/~hayward/hex/).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,6 @@ mujoco-bundle
 rllab_mujoco
 
 tutorial/*.html
+
+# IDE files
+.idea/
--- a/gym/envs/__init__.py
+++ b/gym/envs/__init__.py
@@ -207,3 +207,15 @@ register(
         'board_size': 19,
     },
 )
+
+register(
+    id='Hex9x9-v0',
+    entry_point='gym.envs.board_game:HexEnv',
+    kwargs={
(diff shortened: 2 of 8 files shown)


tlbtlbtlb:
I'll work on merging this. Do you have an agent? At least one that plays well enough to see that the environment works correctly.

tlbtlbtlb:
Merged into master, with a few fixes for Python 2. Thanks Botev!

Botev:
I actually don't. However, as mentioned the guys from University of Alberta have released and the code for NeuroHex: https://github.com/kenjyoung/Neurohex, which has two players - Wolve and resistance agent (also and a human like gui). However, all their code depends on Fuego heavily, which has some weird comparability issues (they seem to use a very exact commit). For this I have dropped them an email if they would in fact port their code base to the Gym interface (which would be really amazing), since I don't have really time to delve deep in their code. If someone get some free time to look into it, that is probably a way better alternative then my simple code, bit it could be rather painful to do it.

PR #54, opened 2016-05-03
Box2d environments

olegklimov:
"Close window" patch is not here, stripped.

After clean install, you can run:

```bash
python examples/agents/keyboard_agent.py CartPoleSwingUp-v0
python examples/agents/keyboard_agent.py LunarLander-v0
python examples/agents/random_agent.py BipedalWalker-v0
python examples/agents/random_agent.py BipedalWalkerHardcore-v0
```

`random_agent.py` is painfully slow. Most likely because it records video. Keyboard agent cannot be run on `BipedalWalker` as it is continuous actions environment.

NONE OF THIS ENVIRONMENTS ARE TESTED in a sense that exists a learning agent that can learn them successfully.

I did my best to debug them using keyboard and `print state` and `print reward`.

`CartPoleSwingUp` retuned to work 4x faster, 1/2 of episode can be kept upright.

`LunarLander` you may find easier now, because action is now applied before `world.step`.

### DIFF
--- a/examples/agents/keyboard_agent.py
+++ b/examples/agents/keyboard_agent.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python
+import sys, gym
+
+#
+# Test yourself as a learning agent! Pass environment name as a command-line argument.
+#
+
+env = gym.make('LunarLander-v0' if len(sys.argv)<2 else sys.argv[1])
+
+ACTIONS = env.action_space.n
+ROLLOUT_TIME = 1000
+SKIP_CONTROL = 0    # Use previous control decision SKIP_CONTROL times, that's how you
+                    # can test what skip is still usable.
+
+human_agent_action = 0
+human_wants_restart = False
+human_sets_pause = False
+
+def key_press(key, mod):
+    global human_agent_action, human_wants_restart, human_sets_pause
+    if key==0xff0d: human_wants_restart = True
+    if key==32: human_sets_pause = not human_sets_pause
+    a = key - ord('0')
+    if a <= 0 or a >= ACTIONS: return
+    human_agent_action = a
+
+def key_release(key, mod):
+    global human_agent_action
+    a = key - ord('0')
+    if a <= 0 or a >= ACTIONS: return
+    if human_agent_action == a:
+        human_agent_action = 0
+
+env.render()
+env.viewer.window.on_key_press = key_press
+env.viewer.window.on_key_release = key_release
+
+def rollout(env):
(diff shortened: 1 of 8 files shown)


olegklimov:
cartpoleswingup
lunarlander
bipedalwalker
bipedalwalkerhardcore

Yes, `BipedalWalker` has a 10-ray laser range finder to help with hardcore version (ladders, pitfalls, stumps). Normal version has slightly uneven terrain, range finder is working in normal version too.

olegklimov:
I ran out of holidays for now, I can answer on comments/questions/suggestions but I can't code. Possible things to do:

1) Make `Racing` environment from [this tutorial](http://www.iforce2d.net/b2dtut/top-down-car), it's awesome.

2) Show motor torques for Walker and action for CartPole, helps to debug agent (how noisy it is, what it tries to do even if it doesn't work, etc).

3) Learn to walk (that was my original intent LOL).

jonasschneider:
Amazing! This is really good. I especially like the uneven terrains. I'll take a look at these over the next few days and see if I can find any obvious bugs.

joschu:
I think that cartpole swingup should go in the classic_control directory and use similar code to the current cartpole.py, where the dynamics equations are already implemented. That way people can use it without the box2d dependency.
But the lander and the terrain walker look great. And it would also be great to have car physics included.
There's lots of potential here.

olegklimov:
> There's lots of potential here.

Can't stop thinking about this... 

Besides `Racing`, which is nice to have, but there's no real challenge here, I'm thinking about what is the most challenging for agents to learn, what they cannot learn now. Here's what I have in mind at the moment:

`UnwieldyTennis` there are 3 different things to do with incoming tennis ball, if robot learns one, it'll be very hard to switch and search for another two. Hopeless local optima. Demonstrations may help (or may not).

In `MarsBuilder` agent should have internal state, or it will be unable to adapt to weight of various building blocks. Judging from state alone will not do. Memory of at least one float will be required to solve this.

Any thoughts?

gdb:
Ok, merged and deployed these environments experimentally: https://gym.openai.com/envs#box2d. Should be enough to let people to play around with them!

denizs:
@gdb  
I just checked out the environment list and couldn't find a cart pole swingup env. I'm currently working on my master's thesis covering the cartpole swingup task and would be glad to contribute to the project. Let me know if I should create a PR on this 👍

benelot:
There is one in the new roboschool bullet physics project. Check it out
here:
https://github.com/openai/roboschool/blob/master/roboschool/gym_pendulums.py

PR #57, opened 2016-05-04
Issue 45 - Easily return environment score values

JKCooper2:
Added a score function to environment which returns a dictionary of the current scoring attributes used by Open AI. Called using env.score() in your agent. Because the json file is written to on env.reset() it means env.score will show the score as at t-1, essentially the same as reward.
I couldn't see an easy way to get around this without manually adding/deleting values from the scoring lists which seems to be more hassle than was worth

Stats storage location automatically pulled from already set variables
Had to split up stats_recorder.close() into two calls because flush() was adding unwanted data if called outside of it's normal schedule.

Also uses json package which I'm 99% sure is a default package but I could be wrong 

Dictionary returns values which can be access by running something like print env.score()['mean']
{
+            'episode_t_value': None,
+            'timestep_t_value': None,
+            'mean': None,
+            'error': None,
+            'number_episodes': None,
+            'number_timesteps': None,
+            'seconds_to_solve': None,
+            'seconds_in_total': None,
+        }

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,6 @@ rllab_mujoco
 tutorial/*.html
 .eggs
 .tox
+
+# PyCharm project files
+.idea
--- a/README.rst
+++ b/README.rst
@@ -30,7 +30,7 @@ that part is left to you. The following are the ``Env`` methods you
 should know:
 
 - `reset(self)`: Reset the environment's state. Returns `observation`.
-- `step(self, action)`: Step the environment by one timestep. Returns `observation`, `action`, `reward`, `done`.
+- `step(self, action)`: Step the environment by one timestep. Returns `observation`, `reward`, `done`, `info`.
 - `render(self, mode='human', close=False)`: Render one frame of the environment. The default mode will do something human friendly, such as pop up a window. Passing the `close` flag signals the renderer to close any such windows.
 
 Installation
--- a/docs/environments.md
+++ b/docs/environments.md
@@ -1,7 +1,10 @@
 # Environments
 
-The gym comes prepackaged with many many environments. Its this common api around many environments that makes the gym so great. Here we will list additional environments that do not come prepacked with the gym. Submit another to this list via a pull-request. 
+The gym comes prepackaged with many many environments. It's this common api around many environments that makes the gym so great. Here we will list additional environments that do not come prepacked with the gym. Submit another to this list via a pull-request. 
 
 _**NOTICE**: Its possible that in time OpenAI will develop a full fledged repository of suplimental environments. Until this this bit of markdown will suffice._
 
-Coming soon…
+## PGE: Parallel Game Engine
+
+PGE is a FOSS 3D engine for AI simulations, and can interoperate with the Gym. Contains environments with modern 3D graphics, and uses Bullet for physics.
+Learn more here: https://github.com/222464/PGE
--- a/examples/agents/random_agent.py
+++ b/examples/agents/random_agent.py
@@ -14,7 +14,7 @@ class RandomAgent(object):
 if __name__ == '__main__':
     # You can optionally set up the logger. Also fine to set the level
     # to logging.DEBUG or logging.WARN if you want to change the
-    # amount of outut.
+    # amount of output.
     logger = logging.getLogger()
     logger.setLevel(logging.INFO)
 
--- a/gym/core.py
+++ b/gym/core.py
@@ -2,7 +2,6 @@ import logging
 import numpy as np
 
 from gym import error, monitoring
-from gym.scoreboard import scoring
 
 # Env-related abstractions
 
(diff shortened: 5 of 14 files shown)


gdb:
Thanks! We definitely need something like this.

One note is that `monitor` supports multiple parallel environments. To do this properly, I suspect we should be periodically flushing all manifests to disk (probably by writing to a tempfile and then replacing the old manifest), and then reading them from the `score` process.

I don't think this should be a property of Env, but instead this should be another top-level function we add to scoring.py for now.

JKCooper2:
I not sure what you mean with the parallel environments? Each env can only hold 1 monitor and each monitor can only hold 1 stats_recorder, which in turn only outputs to 1 file. It should run fine with multiple environments because the environment to output file is 1:1 and the scoring is only based on things inside the individual environment.

It depends on whether you want to duplicate some of the code from stats_recorder into another class. Are you looking for something like:

import gym.scorer
scorer = Scorer('file_name')

for i in range(episodes):
    while not done:
        scorer.save(reward, done)

if i % 10 == 0:
    scorer.display_score(i)  # Display most recent score

So scorer.save just saves the values in memory, potentially as a list of lists (episodes with rewards inside), and scorer.display_score(i) using scoring.py to return the dictionary of stats for scorer lists[:i]

JKCooper2:
Sorry didn't see that issue 7 was referring to the same thing. 

It would be pretty straight forward to move the score call to be env.monitor.score() rather than env.score() so you run: env.monitor.start(outdir, force=True, score=True) to initialise and call env.monitor.get_score() which returns a dictionary of the same scoring metrics the OpenAI server uses

I think it's important to keep the scoring system flexible though, as future environments might want to implement specific metrics or solved states that would ideally retain local copies.

For example CartPole-v0 is solved when the running mean reward over 100 episodes is greater than 195, so the score metric for that environment would be max_running_mean(100) and 'solved' be 195. The environment 'scorer' would store the results and a range of score metrics inside the scorer class could be called like max_running_mean(), running_mean(), max_episode_score(), variance() where variance() gives variance over all eps, variance(100) gives variance of last 100 eps, and variance(10, 20) gives variance from episode 10-19

Environments then implement a solved set of score metrics like:
if max_running_mean(100) > 195 and variance(100) < 5: solved=True
You could probably put it in a dictionary or list or something (to also allow it to pass as a function for the agent to call), but the end result of being able to implement multiple requirements for being solved at an environment level call means you've got a massive amount of flexibility for future problems sets.

gdb:
> I not sure what you mean with the parallel environments? Each env can only hold 1 monitor and each monitor can only hold 1 stats_recorder, which in turn only outputs to 1 file. It should run fine with multiple environments because the environment to output file is 1:1 and the scoring is only based on things inside the individual environment.

Ah, so the reason the scoring stuff is so complicated is because it supports pointing multiple monitors at a single directory. The results then get merged post-hoc. This allows you to have multiple environment instances, whether in a single process or across multiple processes. (Both are useful for collecting a bunch of parallel rollouts for you to train on.)

I also think we should keep the surface area here small for now. I want to build out the right score interface over time, but I don't want to turn it into a free-for-all of different metrics just yet — I'd rather wait for a handful of use-cases that can't be solved any other way.

I'm picturing a function like this: `scoring.score(outdir)`, which is just a thin wrapper on the existing scoring functions, and also adding an auto-flush (possibly configurable, probably by enabling or disabling but maybe also with frequency if there are perf issues) to the existing monitor. You shouldn't have to pass the `env_id`; that's contained within the manifest files.

JKCooper2:
The file name the stats are outputted to are still unique though for each monitor is it contains both the monitor id and process id. 

Do you think the relevant score history can just be kept in memory? The json dump from Stats Recorder is just to make it simple to send to the server. You could attach a reference to the stat_recorder object like scorer = Scorer(env.monitor.stats_recorder) and then scorer.score() just uses the reference to the in-memory objects of that monitor to generate the relevant scores

Adding env_id was because the env_info doesn't get written until env.monitor.close() and so wasn't available for mid-training reporting

gdb:
> Do you think the relevant score history can just be kept in memory?

Let me clarify: your score on an environment is the **combined** score on all parallel environments you were running. For example, imagine having 100 copies of Cartpole-v0 (in 100 different processes), doing 1 rollout on each in parallel, and then training a network using the 100 trajectories you've just gathered. This is very practical, and easy to do using multiprocessing.

The current monitor would record each environment in its own process, and dump the results to disk. These results would then be merged and uploaded as a single history file to the server.

The intent is that a single monitor directory may have results from multiple environment instances, but should only have results from one actual experiment.

(And in particular, this means a single process does not have enough information to know the score. As well, the process which needs to know the score may not have any env instances at all.)

Make sense?

gdb:
(I have some tweaks to this, but will merge and then do the tweaks post-hoc.)

PR #60, opened 2016-05-05
Blackjack environment from Sutton & Barto textbook

machinaut:
Simple player-vs-dealer game of blackjack.

Based on Example 5.1 in this book: https://webdocs.cs.ualberta.ca/~sutton/book/the-book.html
and used in this paper: http://research.microsoft.com/en-us/um/people/alekha/arxiv_geql.pdf

### DIFF
--- a/gym/envs/__init__.py
+++ b/gym/envs/__init__.py
@@ -83,6 +83,11 @@ register(
 # ----------------------------------------
 
 register(
+    id='Blackjack-v0',
+    entry_point='gym.envs.toy_text:BlackjackEnv',
+)
+
+register(
     id='FrozenLake-v0',
     entry_point='gym.envs.toy_text:FrozenLakeEnv',
     kwargs={'map_name' : '4x4'},
--- a/gym/envs/toy_text/__init__.py
+++ b/gym/envs/toy_text/__init__.py
@@ -1,2 +1,3 @@
+from gym.envs.toy_text.blackjack import BlackjackEnv
 from gym.envs.toy_text.roulette import RouletteEnv
 from gym.envs.toy_text.frozen_lake import FrozenLakeEnv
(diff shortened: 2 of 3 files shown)


JKCooper2:
Should there be an additional check if action = stick for the case of blackjack [1 and 10] that returns 1.5. It wouldn't necessarily change the best policy but it would add stochasticity to the reward sets which could alter the learning performance of the agents. Although it wouldn't be the version of the problem described in the book

machinaut:
Good point @JKCooper2; that would more accurately match standard casino rules[0], but my goal is to match the implementation in the textbook.

My motivation for building this was replicating the results in the book itself and the GEQL paper, so I'd slightly lean towards keeping the rewards as-is.

As a compromise, I can add a constructor argument that switches to casino rules (1.5 reward for a natural blackjack win) and have the default be book rules.

[0] http://www.bicyclecards.com/how-to-play/blackjack/

JKCooper2:
Keeping it following the book is probably best then. I don't think the constructor argument to change the pay-offs is the right way to go because it would make it impossible to tell which version of the rules were used when uploaded to the site. The goal is to have a consistent set of environments to compare algorithms against. There is room for environment versions, but for the moment I think they are more for fixes and expansions where the most recent environment is considered the master versus all environment versions being considered equally valid.

machinaut:
The kwargs passed to the constructor are part of the registration -- so using the same registration will give the same environment.  There's at least a few (e.g. FrozenLake) that register multiple versions of the same environment with different constructor kwargs.

machinaut:
I notice a couple things I don't like about this pull request:

 - It doesn't feel like it fits with `toy_text` but it matches well with `roulette`
 - There isn't any graphics/visualization (my least favorite env on the site is `roulette` for this reason)

I can probably make a visual if that's required for new environments, but not sure what to do with the category.

gdb:
Looks great, also merging.

It'd be nice but not required to have rendering for this (and roulette!). Easy to add later though.

PR #61, opened 2016-05-06
n-Chain environment added

machinaut:
Another new environment, this one challenging agents' ability to explore and accurately model action value.  It's a linear chain of states where there is a small reward to go back to the beginning and no reward to advance, but at the end of the chain there is a large reward.  Confounding estimation is a chance of 'slipping' each action and doing the opposite of what was instructed.

This was introduced in [A Bayesian Framework for Reinforcement Learning](http://ceit.aut.ac.ir/~shiry/lecture/machine-learning/papers/BRL-2000.pdf) by Malcolm Strens (2000).

This was also used as a test of the GEQL algorithm in [Exploratory Gradient Boosting for Reinforcement Learning in Complex Domains](http://research.microsoft.com/en-us/um/people/alekha/arxiv_geql.pdf).

### DIFF
--- a/gym/envs/__init__.py
+++ b/gym/envs/__init__.py
@@ -99,7 +99,7 @@ register(
 register(
     id='NChain-v0',
     entry_point='gym.envs.toy_text:NChainEnv',
-    timestep_limit=200,
+    timestep_limit=1000,
 )
 
 register(
--- a/gym/envs/__init__.py
+++ b/gym/envs/__init__.py
@@ -97,6 +97,12 @@ register(
 )
 
 register(
+    id='NChain-v0',
+    entry_point='gym.envs.toy_text:NChainEnv',
+    timestep_limit=200,
+)
+
+register(
     id='Roulette-v0',
     entry_point='gym.envs.toy_text:RouletteEnv',
     timestep_limit=100,
--- a/gym/envs/toy_text/__init__.py
+++ b/gym/envs/toy_text/__init__.py
@@ -1,2 +1,3 @@
 from gym.envs.toy_text.roulette import RouletteEnv
 from gym.envs.toy_text.frozen_lake import FrozenLakeEnv
+from gym.envs.toy_text.nchain import NChainEnv
--- a/gym/envs/toy_text/nchain.py
+++ b/gym/envs/toy_text/nchain.py
@@ -1,0 +1,51 @@
+import gym
+import random
+from gym import spaces
+
+
+class NChainEnv(gym.Env):
+    """n-Chain environment
+
+    This game presents moves along a linear chain of states, with two actions:
+     0) forward, which moves along the chain but returns no reward
+     1) backward, which returns to the beginning and has a small reward
+
+    The end of the chain, however, presents a large reward, and by moving
+    'forward' at the end of the chain this large reward can be repeated.
+
+    At each action, there is a small probability that the agent 'slips' and the
+    opposite transition is instead taken.
+
+    The observed state is the current state in the chain (0 to n-1).
+
+    This environment is described in section 6.1 of:
+    A Bayesian Framework for Reinforcement Learning by Malcolm Strens (2000)
+    http://ceit.aut.ac.ir/~shiry/lecture/machine-learning/papers/BRL-2000.pdf
+    """
+    def __init__(self, n=5, slip=0.2, small=2, large=10):
+        self.n = n
+        self.slip = slip  # probability of 'slipping' an action
+        self.small = small  # payout for 'backwards' action
+        self.large = large  # payout at end of chain for 'forwards' action
+        self.action_space = spaces.Discrete(2)
+        self.observation_space = spaces.Discrete(n)
+        self.state = 0  # Start at beginning of the chain
+
+    def _step(self, action):
+        assert(self.action_space.contains(action))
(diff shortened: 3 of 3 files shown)


JKCooper2:
Looks good. I was a little concerned that 200 steps wouldn't be enough as the paper uses 1,000 but after some tests with fixed actions it looks like 200 should be plenty. The min of fixed action 0 will be larger than the min of fixed action 1 enough of the time to keep it consistent so that an agent could be reasonably expected to explore enough to learn this is a better long term reward.

Each test is 10,000 episodes of 200 steps:
Always action 0: min: 338 max: 1238 mean: 721.435 std: 124.025121548
Always action 1: min: 278 max: 360 mean: 320.7504 std: 11.2351635431

machinaut:
I think I'd prefer the Strens paper number (1000).  The previous number (200) was largely because many of the other simulations seemed to pick that.

Guidance on how to pick environment parameters would be helpful.  For now I'm leaning towards "keep to the literature".

Comparing via average reward per step makes it easy to see differences invariant to game length.  Here we have a random agent in addition to oracle (always forward) and coward (always backward):

```
agent random min 0.0 max 10.0 mean 1.3099 std 1.84165197309
agent oracle min 0.0 max 10.0 mean 3.614902 std 4.45924741749
agent coward min 0.0 max 10.0 mean 1.603502 std 0.811705202642
```

I like that the random agent does worst.  I think this is a cleverly designed problem.  At some point I'd like to compare various parameterizations (What are the effects on learning of decreasing the large reward? the slip rate? etc...)

tlbtlbtlb:
Agreed that environments should exactly match publications when possible. I
think a key value of gym is being able to compare against previous results
easily.

I'd also like to see optionally re-parameterizable environments. As you
say, it'd be interesting to graph learning performance against variations
in the environment spec, like reward or slip rate. Because the slope of
those curves might reveal as much about different learning algorithms as
the default performance.



On Fri, May 6, 2016 at 9:46 AM, Alex Ray <notifications@github.com> wrote:

> I think I'd prefer the Strens paper number (1000). The previous number
> (200) was largely because many of the other simulations seemed to pick that.
>
> Guidance on how to pick environment parameters would be helpful. For now
> I'm leaning towards "keep to the literature".
>
> Comparing via average reward per step makes it easy to see differences
> invariant to game length. Here we have a random agent in addition to oracle
> (always forward) and coward (always backward):
>
> agent random min 0.0 max 10.0 mean 1.3099 std 1.84165197309
> agent oracle min 0.0 max 10.0 mean 3.614902 std 4.45924741749
> agent coward min 0.0 max 10.0 mean 1.603502 std 0.811705202642
>
> I like that the random agent does worst. I think this is a cleverly
> designed problem. At some point I'd like to compare various
> parameterizations (What are the effects on learning of decreasing the large
> reward? the slip rate? etc...)
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/openai/gym/pull/61#issuecomment-217495072>
>



-- 
Trevor Blackwell   tlb@openai.com    650 776 7870

machinaut:
Here's an example for how to run a custom parameterized environment right now: https://gist.github.com/machinaut/9f1793c4d1f466d3b58966da36b54fad

I think the registry can be extended straightforwardly to enable this kind of experiment.

I've been interested in this kind of parameterization search for a while (trying to better develop intuitions).  Here's a toy example of a parameterization search with simple Multi-Armed Bandits: https://github.com/machinaut/notebooks/blob/master/Multi-arm%20Bandits.ipynb

JKCooper2:
Some thought will need to go into how the site should handle different parameter sets for individual problems. It would be interesting to see if an agent could correctly learn the best policy over parameters where the best policy changes, but it also needs to be displayable in a way that encourages collaboratively improvement over algorithms. If the parameters were able to be set to any value within a range it would require testing each algorithm on their individual problem setups to see if the algorithms scores are a result of the algorithm or as a result of the problem parameters.

tlbtlbtlb:
I don't think we should show results from re-parameterized environments in
the scoreboard. If a particular parameterization is interesting enough, we
should give it its own name in the env registry, and its own scoreboard.



On Fri, May 6, 2016 at 3:16 PM, JKCooper2 <notifications@github.com> wrote:

> Some thought will need to go into how the site should handle different
> parameter sets for individual problems. It would be interesting to see if
> an agent could correctly learn the best policy over parameters where the
> best policy changes, but it also needs to be displayable in a way that
> encourages collaboratively improvement over algorithms. If the parameters
> were able to be set to any value within a range it would require testing
> each algorithm on their individual problem setups to see if the algorithms
> scores are a result of the algorithm or as a result of the problem
> parameters.
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/openai/gym/pull/61#issuecomment-217577069>
>



-- 
Trevor Blackwell   tlb@openai.com    650 776 7870

gdb:
Discussion seems to have converged. I'm cool with doing 1000 steps for now. I think doing 200 is mostly useful for the hello-world task that is CartPole, and less useful for other ones. Thanks for the contribution. (If you add the description as well, it'll appear on the scoreboard! Otherwise, I'll add soon.)

PR #63, opened 2016-05-06
tuple spaces will consider lists for contains check

machinaut:
When evaluating `contains` for tuple spaces, accept lists and check them as tuples.

Before fix:
```
In [32]: d = spaces.Tuple((spaces.Discrete(2), spaces.Discrete(2)))

In [33]: d.contains([0,0])
Out[33]: False

In [34]: d.contains((0,0))
Out[34]: True
```

### DIFF
--- a/gym/spaces/tuple_space.py
+++ b/gym/spaces/tuple_space.py
@@ -11,6 +11,8 @@ class Tuple(Space):
         return tuple([space.sample() for space in self.spaces])
 
     def contains(self, x):
+        if isinstance(x, list):
+            x = tuple(x)  # Promote list to tuple for contains check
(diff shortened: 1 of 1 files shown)


gdb:
Cool. Any reason not to just add a check for `isinstance(x, list)`?

machinaut:
I think in this case it's down to preferences.

The biggest difference between lists and tuples is their mutability, and by casting to a tuple we're making it immutable (so the original list cannot be affected by us as all).  Right now none of the rest of those operations do anything to it but the "defensive coder" in me likes the immutability.

Also that return line seemed too long already.

gdb:
Fair enough :).

PR #64, opened 2016-05-06
Fix Python3 compatibility in Go env

shushen:
Under Python3, user inputs are in string format and needs be converted
into bytes before passing to pachi-py. The returned __repr__ from
pachi-py for the boards are bytes and needs be converted into string for
line breaks to work.

Also fixed raw_input and print functions in play_go under
examples/scripts with six package.


gdb:
Thanks!

Issue #6, opened 2016-04-27
Taxi-v0 isn't returning the right reward for dropoff


chrisranderson:
It returns -1 reward for a successful dropoff. The paper (https://www.jair.org/media/639/live-639-1834-jair.pdf) uses +20.

joschu:
Thanks for pointing that out. It shouldn't affect the problem much because it'll just shift the optimal return by a constant, but we might as well adhere to Dietterich's specification.

joschu:
Closing now, if we upgrade to Taxi-v1, we'll change it.

PR #65, opened 2016-05-06
Remove unused destination in Taxi-v0

nottombrown:
Hi friends!
The "B" location doesn't function as an actual destination, although it looks like one in the rendering. This doesn't affect non-human agents, but it confused my friend when I had him play the game by hand.

taxi

### DIFF
--- a/gym/envs/toy_text/taxi.py
+++ b/gym/envs/toy_text/taxi.py
@@ -11,7 +11,7 @@ MAP = [
     "| : : : : |",
     "| : : : : |",
     "| | :F| : |",
-    "|Y| : |B: |",
+    "|Y| : | : |",
     "+---------+",
 ]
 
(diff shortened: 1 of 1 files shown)


JKCooper2:
From what I can see position (4, 3) is considered a valid passenger pickup point.

locs = [(0,0), (0,4), (4,0), (3,2), (4,3)]
for passidx in range(5) on line 49
if (taxiloc == locs[passidx]) on line 68

You're correct in that it doesn't look like it's a valid destination but is still used in the environment.

joschu:
Thanks for pointing this out, I got rid of that location in 69ef2a7, and I'm also taking this opportunity to fix #6 (terminal reward is different from the paper).

nottombrown:
Awesome! Thanks @joschu 👍

PR #66, opened 2016-05-07
Improve score_from_local implementation

gdb:
cc/ @JKCooper2


gdb:
Important changes:

- Added atomic write for stats files
- Removed `n_monitors` / `episode_id` calculation. (You are correct that our current `episode_id`s are wrong in the multi-monitor case. However, the real fix is a bit more involved — multiple processes could be involved, so it needs to happen at merge time.)

PR #68, opened 2016-05-08
Feature/len

mthrok:
This PR proposes `__len__` hook to `Discrete` space and `Tuple` space, 
which provides intuitive way to get the size of space with built-in `len` function.

`Box` space is not touched since it is less intuitive to define length of array,
which require the decision by admins.

### DIFF
--- a/gym/spaces/discrete.py
+++ b/gym/spaces/discrete.py
@@ -21,3 +21,5 @@ class Discrete(Space):
         return "Discrete(%d)" % self.n
     def __eq__(self, other):
         return self.n == other.n
+    def __len__(self):
+        return  n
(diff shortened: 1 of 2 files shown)


gdb:
Yep, looks good to me.

joschu:
I don't like this one so much. E.g. on Discrete, it's not clear if len should be 1 (since that's the number of elements of the datapoint) or the number of actions.

gdb:
Ok, good point. Reverting.

gdb:
Thanks for the contribution @mthrok. I think John has a good point — maybe we should hold off on this, and see what other spaces we introduce before making a final decision? Always easier to add later than to try to remove it.

mthrok:
@joschu 

I agree that the length of individual data points are 1, but this is the length space, and should be different from size of data point.
I was thinking this as alternative way to access `n` in `Discrete` class. accessing `n` from outside seemed like breaking encapsulation.

mthrok:
@gdb 

fair enough

Issue #75, opened 2016-05-09
Allow arbitrary size of graphics display


mttk:
Some of the games (e.g., MsPacman, MontezumaRevenge) are rendered in very small windows which can't be changed - it would make sense to allow a size parameter when rendering to allow for the graphics to possibly be upscaled (even at the cost of pixelating) for presentation / demonstrative purposes.

gdb:
@joschu what do you think?

joschu:
Seems reasonable, though I'm not sure if it's worth the extra complexity.
You can always get the rgb_array and then render that yourself.

JKCooper2:
I've been running some tests with resizing the SimpleImageViewer used by the atari environment and it looks like it would be reasonably straight forward. It would probably just be an additional function in atari env called set_display_dimensions(width, height) (in order to keep it seperate from the _render function) that would set the dimensions used by SimpleImageViewer to display but not alter the inputs

gdb:
After looking at the implementation, I agree with John that we shouldn't support this right now. If there's lots of demand, we can support something, but otherwise we'll be maintaining something that's a bit complicated and isn't obviously universally needed.

Thanks for raising this, however! If you end up writing some code to render the rgb_array yourself, please make sure to post it somewhere (this issue is fine, or in a Gist writeup).

mttk:
@gdb @joschu is this still relevant? I started typing something up, however I'm not sure if the VideoRecorder allows for some resizing options.

PR #79, opened 2016-05-10
Feature 0075 - Allow resizable display on atari environments

JKCooper2:
PR for https://github.com/openai/gym/issues/75

Run env.set_display_dimensions(width, height) on atari environments to customise display size. 

Atari uses SimpleImageViewer which takes in no constructor arguments whereas classic control uses Viewer which takes in width and height as constructor arguments (but these are defined by the environment rather than the user). 

I'm not sure whether it's worth trying to combine the functionality together so users can define display size for all environments. If so it may be easiest to always display output through SimpleImageViewer and just use it as a resizer for [width, height, rgb] pixel array

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,6 @@ rllab_mujoco
 tutorial/*.html
 .eggs
 .tox
+
+# PyCharm project files
+.idea
--- a/README.rst
+++ b/README.rst
@@ -30,7 +30,7 @@ that part is left to you. The following are the ``Env`` methods you
 should know:
 
 - `reset(self)`: Reset the environment's state. Returns `observation`.
-- `step(self, action)`: Step the environment by one timestep. Returns `observation`, `action`, `reward`, `done`.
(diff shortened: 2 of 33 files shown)


JKCooper2:
I'm not sure whether isopen should be set to True as the display dimensions can be set anytime after the environment is created, not necessarily when the window is first opened. However the only place I could see that being used is in closed so for that usage it will still perform in the same way

gdb:
@joschu mind reviewing?

joschu:
OK, assigned to myself.

gdb:
Thanks for the implementation. This is a pretty reasonable implementation. Unfortunately, I think it's a feature we don't want to support yet. I'm going to close this for now, and revisit if there's a lot of demand.

StuartFarmer:
Can we get this so we're not working with a dinky view?

gdb:
Will wait for more upvotes — I still don't think there's enough demand for us to support it.

PR #80, opened 2016-05-10
Live Rewards Graph Option

TonkWorks:
Found it useful to see a graph of my agents progress as it learned.
I added an option to the monitor to display a live rewards graph if others wanted this as well.

` import gym` 
` env = gym.make('CartPole-v0')` 
`env.monitor.start('/tmp/cartpole-experiment-1',` `live_plot=True)`

screenshot-from-2016-05-10-02-


gdb:
Pretty cool idea!

There's a lot of different ways you might want to graph. I think it might be better for us to allow either plugins or a callback to the monitor, so people can add their own code. I'll try out some options soon.

JKCooper2:
Currently you would be able to implement this without having to alter the monitor by running:

`plt.plot(gym.monitoring.monitor.load_results(outdir)['episode_rewards'], color='blue')` where outdir is the set monitor folder. This also allows for the graph to be displayed when running muliple monitors at once

For other graphs you can run the following (although scipy.stats needs to be included in scoring first so won't work straight away)
```
results = gym.monitoring.monitor.load_results(outdir)
graphs = gym.scoreboard.scoring.compute_graph_stats(results['episode_lengths'], results['episode_rewards'], results['timestamps'], results['initial_reset_timestamp'], CURR_EPISODE)

plt.clf()
plt.plot(graphs['x_episode_y_reward']['x'], graphs['x_episode_y_reward']['y'], color='blue')
plt.pause(0.000001)  # pause so matplotlib will display
```
CURR_EPISODE is the number of bins the graph is reduced to, which is why plt.clf() is required as it can slightly alter the values being displayed

Available plots returned from graphs are all the ones displayed on OpenAI's display (all contain ['x'] and ['y] values:
- x_timestep_y_reward
- x_timestep_y_length
- x_episode_y_reward
- x_episode_y_length
- x_seconds_y_length
- x_seconds_y_reward

I'm not sure this necessarily requires a pull request as it alters the inputs required for monitor.start() but it is probably a good example of something a lot of people are looking to do and there needs to be a way to both show people the available options for how to do it (and to receive feedback on what works/doesn't work)

gdb:
Ok, revisiting this: I'd be happy to see this as an external tool, which works as follows:

```
env = gym.make(...)
env.monitor.start(...)
plotter = LivePlot(env)
env.reset()
plotter.plot()
env.step(...)
plotter.plot()
env.step(...)
plotter.plot()
```

Once we accumulate a few of these tools, we can decide where to go next. It'd be very valuable if you'd be up for writing it. (Note that the actual data should come from `load_results` [i.e. https://github.com/openai/gym/blob/master/gym/monitoring/monitor.py#L281], since there might be multiple processes running instances of the environment for a single training run.)

TonkWorks:
Sure, I can give it a shot.  I'll have something in the next few days.

gdb:
Awesome, thanks!

gdb:
(How's it looking?)

TonkWorks:
I wasn't too sure where I should put the tool so I added it to the monitoring folder.
Also added an example just so people can test that it works.

- gym/monitoring/live_plot.py
- gym/examples/agents/random_agent_live_plot.py

gdb:
Cool, thanks! I'm going to rearrange it and put the LivePlot under examples for now.

Issue #81, opened 2016-05-10
gym.scoreboard.scoring.compute_graph_stats incorrect import


JKCooper2:
gym.scoreboard.scoring.compute_graph_stats has `import scipy` but requires `import scipy.stats`
Function also hard fails if provided input lists are empty

PR #82, opened 2016-05-10
Feature 0081 - Fixes to gym.scoreboard.scoring.compute_graph_stats

JKCooper2:
Fixes required in order to display the same graphs as on OpenAI from monitor data without having to check if something has already been saved every time for https://github.com/openai/gym/pull/80

Raised as issue here: https://github.com/openai/gym/issues/81

This allows people to run the following to plot the graphs regardless of whether episode data has been saved yet. CURR_EPISODE is the number of bins to sort data into
```
results = gym.monitoring.monitor.load_results(outdir)
graphs = gym.scoreboard.scoring.compute_graph_stats(results['episode_lengths'], results['episode_rewards'], results['timestamps'], results['initial_reset_timestamp'], CURR_EPISODE)

plt.clf()
plt.plot(graphs['x_episode_y_reward']['x'], graphs['x_episode_y_reward']['y'], color='blue')
plt.pause(0.000001)  # pause so matplotlib will display
```

Using any of these values:
-     x_timestep_y_reward
-     x_timestep_y_length
-     x_episode_y_reward
-     x_episode_y_length
-     x_seconds_y_length
-     x_seconds_y_reward

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,6 @@ rllab_mujoco
 tutorial/*.html
 .eggs
 .tox
+
+# PyCharm project files
+.idea
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,8 @@ addons:
     packages:
       - cmake
       - cmake-data
-
       - python-dev
+      - swig
 python:
   - "2.7"
 #  - "3.5"
--- a/README.rst
+++ b/README.rst
@@ -30,7 +30,7 @@ that part is left to you. The following are the ``Env`` methods you
 should know:
 
 - `reset(self)`: Reset the environment's state. Returns `observation`.
-- `step(self, action)`: Step the environment by one timestep. Returns `observation`, `action`, `reward`, `done`.
+- `step(self, action)`: Step the environment by one timestep. Returns `observation`, `reward`, `done`, `info`.
 - `render(self, mode='human', close=False)`: Render one frame of the environment. The default mode will do something human friendly, such as pop up a window. Passing the `close` flag signals the renderer to close any such windows.
 
 Installation
--- a/README.rst
+++ b/README.rst
@@ -173,6 +173,18 @@ The board game environments are a variety of board games. If you didn't do the f
 	  env.reset()
 	  env.render()
 
+Box2d
+-----------
+
+Box2d is a 2D physics engine. You can install it via  ``pip install -e '.[box2d]'`` and then get started as follow:
+
+.. code:: python
+
+	  import gym
+	  env = gym.make('LunarLander-v0')
+	  env.reset()
+	  env.render()
+
 Classic control
 ---------------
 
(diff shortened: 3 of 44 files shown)


gdb:
Can we instead return `None` here? Probably shouldn't display anything until there's real data.

JKCooper2:
If someone put the display graph code after the episode run code and it tries to run before env.reset() has been called (to write the data to the file), then using None would still throw an error of: TypeError: 'NoneType' object has no attribute '__getitem__' because it's trying to access dictionary values on a None object. 

The user can put a check in themselves for if episode != 0, but this is an additional check that I'm not sure is beneficial (because it is always needed), as well as if you're running a multi monitor there may be a case where you can display data when the episode of the current env is 0 because others have already written data out

gdb:
Hm, I don't understand — why can't you do this:

```
results = gym.monitoring.monitor.load_results(outdir)
graphs = gym.scoreboard.scoring.compute_graph_stats(results['episode_lengths'], results['episode_rewards'], results['timestamps'], results['initial_reset_timestamp'], CURR_EPISODE)

if graphs is not None:
  plt.clf()
  plt.plot(graphs['x_episode_y_reward']['x'], graphs['x_episode_y_reward']['y'], color='blue')
  plt.pause(0.000001)  # pause so matplotlib will display
```

JKCooper2:
I was just thinking it may be easier to have a default return that's still usable without requiring the additional check.I guess having functions always return a valid output is useful but it does have the side effect of creating a reliance on default values (which are often arbitrarily chosen) which can hide things that need refactoring.

I'll switch the check to just return None and once there's a spot to display useful code snippets then in there it can just highlight that a check that the returned data is in a valid format is required.

It may end up being easier just split out the graph function entirely and leave that it to the user, so `gym.scoreboard.scoring.compute_graph_stats` is just used by the OpenAI server and not included in gym. 

Maybe a function that returns the values from `gym.monitoring.monitor.load_results(outdir)` as well as 
```
x_seconds = [timestamp - initial_reset_timestamp for timestamp in timestamps]
x_episode = range(num_episodes)
```
That way users can graph it however they want and the code is just responsible for handling the raw data

gdb:
Yeah, I think it's much better for us to return clearly distinguished data to the user in each case.

We've included the graph code here less for people to use, and more so that people can reproduce what we're displaying. (Otherwise, it'd be very opaque if we had a bug, or if people are trying to figure out whether something is an artifact or real.)

JKCooper2:
Did you want me to add a function to scoring that takes in the monitors outdir and return dictionaries of the lists of values ordered by timestamp? So you'd just do something like 

```
scores = gym.scoreboard.scoring.merged_scores(outdir)
plt.plot(scores['episode_rewards'], scores['episode_numbers'])
```
Dictionary values: 
- rewards
- timestamps
- episode lengths
- episode numbers
- seconds

gdb:
How would this differ from `gym.monitoring.monitor.load_results`?

JKCooper2:
It means all parameters are available to the user, and it moves the call into the scoreboard.scoring section which is probably a better fit for it

gdb:
I see. I still consider the `scoring` file to be half-baked — we need more time and use-cases before committing to anything in it.

I'd rather see what people write using what's here, and then see what we need to add.

PR #84, opened 2016-05-10
Fix observation_space for BipedalWalker

dommueller:
Previously, the observation_space was `Box(5,)` but the actual returned observations are of shape `Box(22,)`.

Furthermore, I added a warning if the observation doesn't fit the observation_space.

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -102,9 +102,6 @@ class Env(object):
         """
         self.monitor._before_reset()
         observation = self._reset()
-        if not self.observation_space.contains(observation):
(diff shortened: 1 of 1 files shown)


dommueller:
The observation_space was fixed with https://github.com/openai/gym/commit/6bb2d56266ae6679c2fb2336b31769a4133400d7. But leaving this open if you want to add the warning. Maybe the environments should start with initializing their action and observation space so the observation space can be checked on `reset()`.

gdb:
Cool. Let's do this for now, and we can tweak it later if needed. Thanks!

PR #86, opened 2016-05-10
Adds several reward thresholds.

wojzaremba:
@joschu 

Few more reward_thresholds for envs that we can solve.


joschu:
I'll review this one. I can also set the thresholds for the discrete tasks, since I've written an exact solver (policy iteration).

joschu:
Haven't forgotten about this, I've just been busy for the last week. Will do in the next couple days.
Also planning to add thresholds for all of the toy_text environments, for which I've written an exact solver using policy iteration.

joschu:
merged in df515de

Issue #88, opened 2016-05-11
Cartpole observations can occur outside of observation space limits


JKCooper2:
Cartpole is set to fail if the angle of the pole is greater than 0.21, however this value is also used as the observation space limit. This means that the final observation of any episode where cartpole fails from having an angle too high will be outside the observation_space bounds. 

Simple fix is just to set the observation space limit for that value to be twice the fail point. Will have a pull request up shortly

hassanbot:
I get the same problem with the lower (negative) x-coordinate.

rafaelcp:
It's still happening:

`2016-05-26 03:10:56,776] Observation '[-2.41935746 -1.66550681  0.00714052 -0.11863974]' is not contained within observation space 'Box(4,)'.
[2016-05-26 03:10:56,835] Observation '[-2.43129996 -1.78584844 -0.01847349 -0.11857105]' is not contained within observation space 'Box(4,)'.
[2016-05-26 03:10:56,856] Observation '[-2.41582085 -1.13530746  0.07019884  0.22842913]' is not contained within observation space 'Box(4,)'.
[2016-05-26 03:10:56,877] Observation '[-2.42284413 -1.47464028 -0.00405668 -0.15805834]' is not contained within observation space 'Box(4,)'.
[2016-05-26 03:10:56,915] Observation '[-2.42800512 -1.81184147  0.01414837  0.08355538]' is not contained within observation space 'Box(4,)'.
[2016-05-26 03:10:56,993] Observation '[-2.42121197 -1.45672808  0.08031289  0.17295786]' is not contained within observation space 'Box(4,)'.`

It seems the problem is happening to x now, which has a 2.4 limit.

rafaelcp:
Also, CartPole should be updated to v1, since these changes are breaking some algorithms which used boundary information for normalization.

gdb:
From chat: "Artem Yankov @yankov 16:17
anyone ran into this openai/gym#88 looks like it’s still an issue (have just pulled the latest changes from master)"

JKCooper2:
It seems to be the same type of issue as the previous one where the observation space bounds are the same as the environment failure point bounds, so when you check the state after the failure point it is outside the observation space limits. It's easy enough to increase the limit but would it be better to add some code that if an observation is given that is outside the bounds in a box environment that it caps it to being equal to the bounded value before it's returned (but after done is calculated)? Should the observation bounds be a fixed edge state of the environment that the agent can't break out of?

PR #89, opened 2016-05-11
Cartpole observations can occur outside of observation space limits - Issue #88

JKCooper2:
Fix for issue #88

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,8 @@ mujoco-bundle
 rllab_mujoco
 
 tutorial/*.html
+
+# IDE files
 .eggs
 .tox
 
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,6 @@ rllab_mujoco
 tutorial/*.html
 .eggs
 .tox
+
+# PyCharm project files
+.idea
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,8 @@ addons:
     packages:
       - cmake
       - cmake-data
-
       - python-dev
+      - swig
 python:
   - "2.7"
 #  - "3.5"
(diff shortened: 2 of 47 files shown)


jonasschneider:
Yeah, I think that's reasonable. Thanks! CartPole is kind of a special case here; an earlier version actually had a separate angle threshold at which you'd stop getting a reward, but the episode would still continue.

(Also, sorry for the failing tests. Working to fix that.)

JKCooper2:
If you mean the swig install problem for travis-ci and box2d I think I've got a version working of the yaml file here: https://github.com/JKCooper2/gym/blob/bug_swig_travisci/.travis.yml but I didn't make it as a pull request because I can't test the mujoco install (needs to be a push from a computer with the mujoco location available)

rafaelcp:
It's still happening:

`2016-05-26 03:10:56,776] Observation '[-2.41935746 -1.66550681  0.00714052 -0.11863974]' is not contained within observation space 'Box(4,)'.
[2016-05-26 03:10:56,835] Observation '[-2.43129996 -1.78584844 -0.01847349 -0.11857105]' is not contained within observation space 'Box(4,)'.
[2016-05-26 03:10:56,856] Observation '[-2.41582085 -1.13530746  0.07019884  0.22842913]' is not contained within observation space 'Box(4,)'.
[2016-05-26 03:10:56,877] Observation '[-2.42284413 -1.47464028 -0.00405668 -0.15805834]' is not contained within observation space 'Box(4,)'.
[2016-05-26 03:10:56,915] Observation '[-2.42800512 -1.81184147  0.01414837  0.08355538]' is not contained within observation space 'Box(4,)'.
[2016-05-26 03:10:56,993] Observation '[-2.42121197 -1.45672808  0.08031289  0.17295786]' is not contained within observation space 'Box(4,)'.`

It seems the problem is happening to x now, which has a 2.4 limit.

PR #90, opened 2016-05-11
scoring.compute_graph_stats Default return and fix import statement

JKCooper2:
Required import scipy.stats
Set default return of None  for situation where no files saved to directory yet

Issue #81


gdb:
Thanks!

(BTW, I think you should `git reset --hard origin/master` — looks like your local repo has a bunch of merge commits.)

PR #97, opened 2016-05-13
Handle draw in go env

dommueller:
I'm not familiar with the specific Go rules but I think a score of 0 indicates a draw.

Previously one could win (meaning getting a reward of 1) by doing the following steps:
```
import gym
env = gym.make("Go9x9-v0")
env.reset()
env.step(72)
ob, reward, done, info = env.step(81)
```

It seems nice of Pachi to pass and end the game in a draw instead of playing another move and win.


gdb:
cc/ @hojonathanho

PR #98, opened 2016-05-16
Fix box2d tests

gdb:


jietang:
whoa this is old. @jonasschneider, do you remember the context?

jonasschneider:
The Box2D python bindings distribution is very broken. See https://github.com/openai/gym/issues/100.

Dumping what I remember: I think the source is https://github.com/pybox2d/pybox2d, published to https://pypi.python.org/pypi/Box2D. However, that is an old version that has the SWIG version requirement/bug. I forked the source to https://github.com/jonasschneider/box2d-py and pushed https://pypi.python.org/pypi/box2d-py to have the latest version available, which is what `gym` now depends on.

I'm not sure what changes have been merged in the meanwhile and if this patch still applies. If these tests are currently passing, I'd recommend to not merge. However, #100 looks like people are still having problems there, so it might be worth investigating all the of the box2d strangeness more.

gdb:
I never actually got the Travis tests for box2d working. (I thought this PR would do it, but I was wrong, and decided to wait for @jonasschneider to be back from vacation!)

jonasschneider:
A-ha! Yep, then I'll take this up again.

gdb:
@jonasschneider should we actually be able to fix these tests now?

pzhokhov:
travis tests for box2d are working now, closing this

PR #101, opened 2016-05-16
BipedalWalker, LunarLander heuristic, retuned

olegklimov:
I re-tuned `BipedalWalker` and `LunarLander`. I had to write heuristic code to make sure both work this time and there are no more bugs. Hopefully. To run heuristics, type:

```bash
python gym/envs/box2d/bipedal_walker.py
python gym/envs/box2d/lunar_lander.py
```

Changes in `LunarLander`:

 * Landing legs now report contact with ground.
 * Contact with ground give +10 reward points, and -10 if then lose contact.
 * Reward renormalized to 200, tested.
 * Harder initial random push.
 * Heuristic lands 100% of the time.

Changes in `BipedalWalker`:

 * Legs now report contact with ground.
 * Motors have higher torque and speed.
 * Ground has higher friction.
 * Lidar rendered less nervously.
 * Heuristic walks up to the end about 80% of the time.

### DIFF
--- a/gym/envs/__init__.py
+++ b/gym/envs/__init__.py
@@ -85,8 +85,8 @@ register(
 register(
     id='LunarLander-v0',
     entry_point='gym.envs.box2d:LunarLander',
-    timestep_limit=300,
-    reward_threshold=1,
+    timestep_limit=1000,
+    reward_threshold=200,
 )
 
 register(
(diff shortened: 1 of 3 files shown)


gdb:
Awesome, thanks!

gdb:
I also bumped the version on all these tasks. In general, we should do this whenever environments change in a backwards-incompatible way. (I'm considering not requiring this for environments which are still marked as "experimental" though.)

PR #102, opened 2016-05-16
Add close() to Env, add functionality to call close() on exit

jietang:
First pass at adding close-on-exit functionality to envs (the doom tests leave horrible dangling processes otherwise)

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -4,6 +4,7 @@ logger = logging.getLogger(__name__)
 import numpy as np
 
 from gym import error, monitoring
+from gym.utils.atexit_utils import env_close_registry
 
 # Env-related abstractions
 
@@ -17,6 +18,7 @@ class Env(object):
         reset
         step
         render
+        close
 
     When implementing an environment, override the following methods
     in your subclass:
@@ -40,6 +42,10 @@ class Env(object):
     metadata = {'render.modes': []}
     reward_range = (-np.inf, np.inf)
 
+    # Override in SOME subclasses
+    def _close(self):
+        pass
+
     # Set these in ALL subclasses
     action_space = None
     observation_space = None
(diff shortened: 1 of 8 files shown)


gym/core.py
```diff
@@ -155,6 +162,17 @@ def render(self, mode='human'):
 
         return self._render(mode=mode, close=close)
 
+    def close(self):
```
gdb: This should probably be idempotent. Maybe maintain a `closed` variable?

gym/utils/atexit_utils.py
```diff
@@ -0,0 +1,42 @@
+import atexit
```
gdb: (I assume the plan is to DRY this up with the monitor `atexit` stuff?)

jietang: Yup.

gym/core.py
jietang: Good idea. I'll make the change.

gym/monitoring/monitor.py
```diff
@@ -46,20 +39,8 @@ def capped_cubic_video_schedule(episode_id):
...
-
-@atexit.register
-def close_all_monitors():
-    # Explicitly fetch all monitors first so that they can't disappear while
-    # we iterate. cf. http://stackoverflow.com/a/12429620
-    open_monitors = list(monitors.items())
-    for key, monitor in open_monitors:
-        monitor.close()
+def get_running_monitors():
+    return list(monitor_close_registry.close_objects.values())
```
gdb: I don't love the name `monitor_close_registry` BTW — could we do something like `monitor_closer`?

gym/monitoring/__init__.py
```diff
@@ -1,3 +1,3 @@
-from gym.monitoring.monitor import Monitor, load_results, monitors as _monitors
+from gym.monitoring.monitor import Monitor, load_results, get_running_monitors
```
gdb: The intention with `_monitors` is to make clear that this is for use within OpenAI Gym only, and not within user code. I would keep a `_...` name here.

Maybe better to do `get_running_monitors` -> `open_monitors`?

PR #107, opened 2016-05-18
Remove deprecated env file, report different error for deprecated envs

jietang:
More convenient for end-users, but also helpful on the scoreboard side when handling requests from old libraries


gym/envs/registration.py
```diff
@@ -89,7 +89,15 @@ def spec(self, id):
         try:
             return self.env_specs[id]
         except KeyError:
-            raise error.UnregisteredEnv('No registered env with id: {}'.format(id))
+            # Parse the env name and check to see if it matches the non-version
+            # part of a valid env (could also check the exact number here)
+            env_name = match.group(1)
+            matching_envs = [full_env_name for full_env_name in self.env_specs
+                             if full_env_name.startswith(env_name + '-v')]
```
gdb: I would prefer to do the env name parsing in Spec, and then the comparison becomes a straightforward equality comparison (and the semantics are dictated entirely by the regex).

jietang: sorry, do you mean in EnvSpec? This new code is inside registry.spec()

gdb: Yep, exactly. Sorry for being unclear :).

Issue #111, opened 2016-05-20
Error when numpy version isn't of the standard x.x.x format


MarCnu:
When I try to import gym, I get this Traceback

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "dqn/environment.py", line 2, in <module>
    import gym
  File "/usr/local/lib/python2.7/dist-packages/gym/__init__.py", line 32, in <module>
    sanity_check_dependencies()
  File "/usr/local/lib/python2.7/dist-packages/gym/__init__.py", line 18, in sanity_check_dependencies
    if distutils.version.StrictVersion(numpy.__version__) < distutils.version.StrictVersion('1.10.4'):
  File "/usr/lib/python2.7/distutils/version.py", line 40, in __init__
    self.parse(vstring)
  File "/usr/lib/python2.7/distutils/version.py", line 107, in parse
    raise ValueError, "invalid version number '%s'" % vstring
ValueError: invalid version number '1.12.0.dev0+8ccfede'
```

https://github.com/openai/gym/blob/1d414864802bf1e7e4b8beae85cbd5a196e3cdaa/gym/__init__.py#L13-23

```
def sanity_check_dependencies():
    import numpy
    import requests
    import six

    if distutils.version.StrictVersion(numpy.__version__) < distutils.version.StrictVersion('1.10.4'):
        logger.warn("You have 'numpy' version %s installed, but 'gym' requires at least 1.10.4. HINT: upgrade via 'pip install -U numpy'.", numpy.__version__)

    if distutils.version.StrictVersion(requests.__version__) < distutils.version.StrictVersion('2.0'):
        logger.warn("You have 'requests' version %s installed, but 'gym' requires at least 2.0. HINT: upgrade via 'pip install -U requests'.", requests.__version__)
```

In my self-compiled numpy, my version is '1.12.0.dev0+8ccfede', it isn't a recognised format by distutils.version.StrictVersion and it leads to an error.

It would be nice to be less strict in the format of versions.

gdb:
Ah yes we should just skip the check if the version doesn't parse. Up for
writing a patch?

On Friday, May 20, 2016, MarCnu <notifications@github.com> wrote:

> When I try to import gym, I get this Traceback
>
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
>   File "dqn/environment.py", line 2, in <module>
>     import gym
[...]

PR #113, opened 2016-05-20
Replace StrictVersion by LooseVersion

MarCnu:
Replaces the parsing function for the version comparison, to allow non standard versions. Remains unchanged for standard versions. http://epydoc.sourceforge.net/stdlib/distutils.version.LooseVersion-class.html

See https://github.com/openai/gym/issues/111


jietang:
Thanks @MarCnu

PR #117, opened 2016-05-23
Car racing

olegklimov:
This is `CarRacing` environment. It is supposed to be the easiest continuous control environment to learn from pixels. Yes, car physics is fun.

* I had performance issues, ended up using a lot of OpenGL directly from `pyglet.gl`.

* Environment have three resolutions: 1200x1000 for window, 600x400 for video, 96x96 for state. The original intent for `rgb_array` is not entirely obvious, the comment in second commit should save a lot of time for others.

* I'd prefer to return gyroscope, steering wheel position, ABS sensors directly, not through pixels. No way to do this yet.

* `car_dynamics.py` is a separate file, because it can be reused for more car environments.

screen shot 2016-05-23 at 00 38 46

Please someone test.

```
python gym/envs/box2d/car_racing.py
python examples/agents/random_agent.py CarRacing-v0
```

### DIFF
--- a/examples/agents/random_agent.py
+++ b/examples/agents/random_agent.py
@@ -40,6 +40,15 @@ if __name__ == '__main__':
             ob, reward, done, _ = env.step(action)
             if done:
                 break
+            # Note there's no env.render() here.
+            #
+            # It is implied by env.step(), because env.monitor records video and calls env.render('rgb_array').
+            # Environment can show in its window what it renders for video recording, but it can also not show
+            # anything.
+            #
+            # Video is NOT recorded every time, see capped_cubic_video_schedule for details. If you don't see
+            # anything during an episode, that means video recording is off this time. If you want to always
+            # see what's going on, put env.render() back here.
 
     # Dump result info to disk
     env.monitor.close()
--- a/examples/agents/random_agent.py
+++ b/examples/agents/random_agent.py
@@ -40,15 +40,9 @@ if __name__ == '__main__':
             ob, reward, done, _ = env.step(action)
             if done:
                 break
-            # Note there's no env.render() here.
-            #
-            # It is implied by env.step(), because env.monitor records video and calls env.render('rgb_array').
-            # Environment can show in its window what it renders for video recording, but it can also not show
-            # anything.
-            #
-            # Video is NOT recorded every time, see capped_cubic_video_schedule for details. If you don't see
-            # anything during an episode, that means video recording is off this time. If you want to always
-            # see what's going on, put env.render() back here.
+            # Note there's no env.render() here. But the environment still can open window and
+            # render if asked by env.monitor: it calls env.render('rgb_array') to record video.
+            # Video is not recorded every episode, see capped_cubic_video_schedule for details.
 
     # Dump result info to disk
     env.monitor.close()
--- a/gym/envs/__init__.py
+++ b/gym/envs/__init__.py
@@ -100,6 +100,13 @@ register(
     reward_threshold=300,
 )
 
+register(
+    id='CarRacing-v0',
+    entry_point='gym.envs.box2d:CarRacing',
+    timestep_limit=1000,
+    reward_threshold=900,
+)
+
 # Toy Text
 # ----------------------------------------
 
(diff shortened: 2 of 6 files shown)


olegklimov:
Agent sees state as 96x96 image:

test

At the bottom there are some indicators, green bar is a steering wheel position.

jietang:
@olegklimov This looks great, I just played with it in keyboard mode and I'd love to merge it. Could you do the following:
- add the task to the scoreboard registry (gym/scoreboard/init.py) so it shows up on the website with the experimental=True tag
- I like the comment you added but I don't think random_agent.py is the place for it. Could you take a look at the comment for the Env.render() method and the Monitor class and see if there is a concise way to fold it in there?

olegklimov:
Hello @jietang, I've added scoreboard description. Regarding your second point, you're right. This is not a proper place for a descriptive comment like this. But Env.render() is not a good place either :(  There's already a good and clear description of what it does. Making it longer will do no good. It also describes render() from a correct point of view, that is what environment should do when asked to render(). My comment is more about how to write agent in batch mode: no render(), why it still shows something, why not always.

I made the comment much shorter now, maybe it will be appropriate in this shorter form.

jietang:
Thanks for making the changes, I like the new comment. Merging now

arsenious:
Hello everyone

I really like this environment but a few things are puzzling me. What exactly are tiles? In a 96*96 RGB state image that's returned, is it each pixel or are they the only the road pixels i.e. grey ones? Or are they the checkered check-points?

I would like to have the reward function somehow penalize if the car goes off-road (i.e. off the grey patches), or better yet, the width deviation from the centre line of the road. Is one of these already implemented? If not please could you guide me how to implement such a cost function.

Also please could you explain this
# Reward is -0.1 every frame and +1000/N for every track tile visited, where N is
# the total number of tiles in track. For example, if you have finished in 732 frames,
# your reward is 1000 - 0.1*732 = 926.8 points.

If N is the total no. of tiles on the track then in this example, N = 1 ? In 732 frames the car stayed just on the starting tile and then got timed out?

How many tiles are there in each track on an average? 

Also " if you have finished in 732 frames"...Finished the entire game/episode successfully (i.e. the car reached the end) ? 

I'm relatively new to Deep Reinforcement Learning, so sorry if these questions are obvious. Please help.

@olegklimov

olegklimov:
I think most of your questions go away if you play manually for a few minutes. Just run the file, control the car with arrow keys.

There's no point in punishing the car if it's off the center. In fact, it shouldn't be in center of the road, it should be at optimal trajectory, that it discovers for itself. The same for being off the road. If an agent discovers how it can use grass to be faster and still visit all tiles, no problem with that.

1000/N reward for single tile, if you visited N tiles, you get 1000 reward. N is total number of tiles in track. If you visited one tile, you'll get 1000/N (small number).

arsenious:
Thank you for your reply!

I played it a bit. I can see that the road is divided into rectangles, though very faintly. Are these the tiles? Because when going on them I got +ve points while going off them in the grass, I lost points. If so, then there must be a few hundred such tiles. If there were 500 tiles, then going on to a new tile would give me 1000/500 = +2 points per tile.

Also are the checkered artifacts at every bend, the checkpoints? I couldn't see anything else.

Yes you are right, there is no point in penalizing the car from going off centre.
Thanks

olegklimov:
You cannot lose points for anything, other than -0.1 each frame.

Checkered artifacts you mean this? :D

https://en.wikipedia.org/wiki/Chicane#/media/File:Suzuka_Circuit_Chicane_2011_Super_Taikyu_Suzuka_300km.jpg

arsenious:
Oh so that's what they are called! Thanks! But then how does the env know if the car completes the track as intended, rather than just taking a route over the grass (since no penalty for being on the grass)? 

Sorry for the trouble, but I really need to understand this env in and out. I intend to make changes to it and use it for my Masters project on avoiding unsafe road conditions/areas. I'll certain credit you for it and upload it to Gym if it's turns out good enough. Thanks!

olegklimov:
Complete the track == visit all tiles on the road.

You have to understand what you're dealing with:

1. This environment is slow (slower than many 3d games). You may need a cluster to learn anything.

2. This environment is easy to modify.

Issue #114, opened 2016-05-21
Box2d bipedal walker segfault


justheuristic:
This code happens to cause segfault for some reasons i cannot comprehend

```
env_name = "BipedalWalker-v1"
import gym
ll0 = gym.make(env_name)
ll0.reset()
ll1 = gym.make(env_name)
ll1.reset()
ll0.reset()
```

However, doing the same thing with a different box2d env works fine
```
env_name = "LunarLander-v1"
import gym
ll0 = gym.make(env_name)
ll0.reset()
ll1 = gym.make(env_name)
ll1.reset()
ll0.reset()
```

Can anyone reproduce the error? Is there anything i can do to fix it?

olegklimov:
I am able to reproduce everything you wrote. First one crashes, second doesn't. And it is not related with program termination. Scratching my head for now.

olegklimov:
https://github.com/openai/gym/pull/118

jietang:
Fix merged.

justheuristic:
Guys, sorry, but could someone please assert that the fix works at ubuntu 14.04? 
I didn't manage to bypass error for `BipedalWalker-v2` at current version

justheuristic:
@jietang
Could you please reopen

PR #118, opened 2016-05-23
Fix for #114

olegklimov:


olegklimov:
Github says commit not found, but it opens with direct link, hash is the same...
https://github.com/olegklimov/gym/commit/c61f885c5657d1f0a19c27b35b6438a15148d268

jietang:
Thanks @olegklimov

PR #119, opened 2016-05-23
Faster video recording

olegklimov:
Running environments with video recording turned on is painfully slow. Let's investigate. In `random_agent.py`:

```python
        for j in range(max_steps):
            action = agent.act(ob, reward, done)
            import time
            t1 = time.time()
            ob, reward, done, _ = env.step(action)
            t2 = time.time()
            print "step %0.2fms" % ((t2-t1)*1000)
```

Now let's try `python examples/agents/random_agent.py CartPole-v0`

```
step 101.11ms
step 99.19ms
step 98.89ms
step 100.18ms
step 99.97ms
step 99.45ms
step 100.70ms
step 99.77ms
step 100.09ms
```

Why 100ms? My computer, MacBook Pro 13" is connected to 4K monitor. Refresh rate is 30 Hz. Pyglet window is created with vsync=True. Every `viewer.window.flip()` waits for vsync. Then `viewer.get_array()` contains another two `flip()`. 1000ms/30*3 = 100ms. Not only this waits for another two vsync events, but is also jitters image a frame back and it's visible.

It doesn't affect resulting encoded video, but the process doesn't look pretty. 

This PR returns array without extra `flip()`s.

```
step 26.68ms
step 39.91ms
step 26.64ms
step 38.33ms
```

Effect is identical for `CartPole`, `Acrobot`, `MountainCar`, `Pendulum`, `LunarLander`, `BipedalWalker`.

Atari and MuJoCo needs to be checked.

### DIFF
--- a/gym/envs/box2d/bipedal_walker.py
+++ b/gym/envs/box2d/bipedal_walker.py
@@ -468,13 +468,7 @@ class BipedalWalker(gym.Env):
         self.viewer.draw_polygon(f, color=(0.9,0.2,0) )
         self.viewer.draw_polyline(f + [f[0]], color=(0,0,0), linewidth=2 )
 
-        self.viewer.render()
-        if mode == 'rgb_array':
-            return self.viewer.get_array()
-        elif mode is 'human':
-            pass
-        else:
-            return super(BipedalWalker, self).render(mode=mode)
+        return self.viewer.render(return_rgb_array = mode=='rgb_array')
 
 class BipedalWalkerHardcore(BipedalWalker):
     hardcore = True
--- a/gym/envs/box2d/lunar_lander.py
+++ b/gym/envs/box2d/lunar_lander.py
@@ -312,13 +312,7 @@ class LunarLander(gym.Env):
             self.viewer.draw_polyline( [(x, flagy1), (x, flagy2)], color=(1,1,1) )
             self.viewer.draw_polygon( [(x, flagy2), (x, flagy2-10/SCALE), (x+25/SCALE, flagy2-5/SCALE)], color=(0.8,0.8,0) )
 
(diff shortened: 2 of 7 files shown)


tlbtlbtlb:
I prefer this much smaller change to mujoco-py that accomplishes the same thing.
https://github.com/openai/mujoco-py/commit/bda1ac4dc34501eb1cd3518b179e165140a42e0e

olegklimov:
Well `vsync=False` is even smaller, but it will not do the same thing: jittering picture in the window will remain. Solution (I'm certainly not saying I have the best possible solution) should involve removing unnecessary `flip()`s, one way or another.

jietang:
@tlbtlbtlb @olegklimov, what's the status here? is there a use case that requires vsync=False?

tlbtlbtlb:
Oleg's solution seems good to me, let's merge. It makes a huge difference in non-interactive rendering performance.

olegklimov:
I guess @tlbtlbtlb didn't mean `vsync=False` makes a difference in non-interactive setting. But it also true. To set `vsync=False` in batch mode we need some indicator it's a batch mode, in constructor. Pyglet doesn't let to change vsync on the fly.

Thinking of it, we have abstract fabric for environments, right?

```
env = gym.make('CartPole-v0', batch_mode=True)
```

Then any window opened by that environment should have vsync=False. Shouldn't be hard (I don't have a patch, maybe later).

jietang:
@olegklimov I'd rather not add arguments to gym.make(). Perhaps we could add a configure() method you could call on the env to turn on batch mode? e.g.

```
env = gym.make('CartPole-v0')
env.configure(batch_mode=True)
```

PR #126, opened 2016-05-27
Improve closer implementation and docstrings

gdb:
@jietang mind reviewing?


gym/core.py
```diff
@@ -31,13 +33,22 @@ class Env(object):
 
         action_space: The Space object corresponding to valid actions
         observation_space: The Space object corresponding to valid observations
+        reward_range: A tuple corresponding to the min and max possible rewards
```
jietang: It doesn't look like we actually use this anywhere - should we get rid of it?

gym/core.py
```diff
@@ -58,12 +69,6 @@ def _render(self, mode='human', close=False):
             return
         raise NotImplementedError
 
-    # Will be automatically set when creating an environment via
```
jietang: maybe worth keeping this comment in __new__ so it's easier to trace the magic that sets spec

gym/core.py
gdb: This is for use by people's algorithms.

gym/tests/test_core.py
```diff
@@ -0,0 +1,15 @@
...
+
+class ArgumentEnv(core.Env):
+    calls = 0
+
+    def __init__(self, arg):
+        self.calls += 1
+        self.arg = arg
+
+def test_env_instantiation():
+    # This looks like a pretty trivial, but given our usage of
```
jietang: +1

gym/utils/closer.py
```diff
@@ -0,0 +1,67 @@
...
+class Closer(object):
+    """A registry that ensures your objects get closed, whether manually,
+    upon garbage collection, or upon exit. To work properly, your
+    objects need to cooperate and do something like the following:
+
+    ```
+    closer = Closer()
+    class Example(object):
+        def __init__(self):
+        self._id = closer.register(self)
```
jietang: indentation looks off

gym/core.py
gdb: Good call.

jietang:
lgtm

Issue #128, opened 2016-05-28
Can't generate tar file for upload on Windows


rafaelcp:
This happens on scoreboard/api.py. Windows (tested on 8.1) does not allow `tar.add` on opened file, so uploads are not possible. I already have a fix, but wanted confirmation for this bug before sending the PR.

gdb:
Oh, great to know. I don't have a great way of validating Windows issues right now. But please do fix it!

PR #129, opened 2016-05-28
Fixes tar file generation for upload on Windows

rafaelcp:
Temp file cannot be added to the tar while opened on Windows (tested on 8.1). Needs testing on Linux. Fix for issue https://github.com/openai/gym/issues/128.

### DIFF
--- a/gym/scoreboard/api.py
+++ b/gym/scoreboard/api.py
@@ -177,8 +177,13 @@ def write_archive(videos, archive_file, env_id=None):
             tar.add(metadata_path, arcname=metadata_name, recursive=False)
 
         # Actually write the manifest file
-        with tempfile.NamedTemporaryFile(mode='w+') as f:
+        # 'delete = False' avoids deleting the file when closing
+        with tempfile.NamedTemporaryFile(mode='w+', delete = False) as f:
             json.dump(manifest, f)
             f.flush()
-
+            # Closing the file is necessary before adding to tar in Windows
+            f.close()
+            
             tar.add(f.name, arcname='manifest.json')
+            # Manually delete temporary file
+            os.remove(f.name)
--- a/gym/scoreboard/api.py
+++ b/gym/scoreboard/api.py
@@ -178,7 +178,7 @@ def write_archive(videos, archive_file, env_id=None):
 
         # Actually write the manifest file
         # 'delete = False' avoids deleting the file when closing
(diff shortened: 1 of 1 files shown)


gym/scoreboard/api.py
```diff
@@ -177,8 +177,13 @@ def write_archive(videos, archive_file, env_id=None):
             tar.add(metadata_path, arcname=metadata_name, recursive=False)
 
         # Actually write the manifest file
-        with tempfile.NamedTemporaryFile(mode='w+') as f:
+        # 'delete = False' avoids deleting the file when closing
+        with tempfile.NamedTemporaryFile(mode='w+', delete = False) as f:
```
gdb: Please match the style of surrounding code (so do `delete=False` rather than `delete = False` here).

gym/scoreboard/api.py
```diff
@@ -177,8 +177,13 @@ def write_archive(videos, archive_file, env_id=None):
...
             json.dump(manifest, f)
             f.flush()
-
+            # Closing the file is necessary before adding to tar in Windows
+            f.close()
+            
             tar.add(f.name, arcname='manifest.json')
+            # Manually delete temporary file
+            os.remove(f.name)
```
gdb: It'd be nice to remove this file in a `finally`. Given that we're not getting much out of the `with` clause anymore, perhaps it's worth rewriting this as just something like:

```
f = tempfile.NamedTemporaryFile(...)
try:
  json.dump(...)
  f.close()
  tar.add(f.name, ...)
finally:
  f.close()
  os.remove(f.name)
```

gym/scoreboard/api.py
```diff
@@ -177,8 +177,13 @@ def write_archive(videos, archive_file, env_id=None):
             tar.add(metadata_path, arcname=metadata_name, recursive=False)
 
         # Actually write the manifest file
-        with tempfile.NamedTemporaryFile(mode='w+') as f:
+        # 'delete = False' avoids deleting the file when closing
+        with tempfile.NamedTemporaryFile(mode='w+', delete = False) as f:
             json.dump(manifest, f)
             f.flush()
```
gdb: Since we're closing, no need for a `flush` anymore!

rafaelcp:
Yup, I think it is better. Done and tested here.

gdb:
Thanks!

PR #138, opened 2016-05-29
humanoidstandup a little fix

zdx3578:


zdx3578:
pos_after[1] - pos_before[1] > 15.0) or (pos_after[0] - pos_before[0] > 15.0      model.data.qpos[0]  and model.data.qpos[1]  meaning is not clear; so delete.

PR #140, opened 2016-05-29
Support disabling videos by passing video_callable=False

gdb:
I find myself disabling videos enough (it really speeds things up when you're hacking around) that it seems worth having a helper. I could be convinced we still don't need it though. @joschu what do you think?

### DIFF
--- a/gym/monitoring/monitor.py
+++ b/gym/monitoring/monitor.py
@@ -165,9 +165,9 @@ class Monitor(object):
         if not self.enabled:
             return
         self.stats_recorder.close()
-        self.flush()
         if self.video_recorder is not None:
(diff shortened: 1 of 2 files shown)


joschu:
Agreed that it's good to make it convenient to disable video. It's just a little confusing that video_callable argument is sometimes boolean and sometimes a function. An alternative is to have a constant NO_VIDEO which is a function that returns false. I don't have a strong preference.

gdb:
Makes sense. We already allow `video_callable=None`, so I don't think it's that much of a step up on the confusing ladder. I think the NO_VIDEO approach would probably not not result in enough convenience savings.

PR #141, opened 2016-05-30
just standing is reward

zdx3578:
just standing is reward


jietang:
Thanks for the fix @zdx3578, merged.

PR #144, opened 2016-05-30
Switch to a global PRNG for action/observation spaces

gdb:
cf https://github.com/openai/gym/commit/58e6aa95e5af2c738557431f812abb81c505a7cf#commitcomment-17669277


gdb:
@joschu @jietang how's this seem?

Issue #94, opened 2016-05-12
Knowledge transfer for black box optimization environment


iaroslav-ai:
I have a suggestion of environment(s), where I can contribute. 

Consider a certain distribution of black box optimization problems, e.g. model selection / parameter tuning for computer vision applications. When solving new problem from such distribution, one would like to exploit knowledge obtained from solving previous ones. 

AFAIK with most popular approaches for black box optimization like Bayesian Optimization every new problem is solved independently and thus domain knowledge is not exploited. One can expect however that exploiting such knowledge would make a significant difference.

I can imagine solving this with episodic RL, where one episode is solving one random problem from distribution. The feedback to the agent is a feature description of the sample (e.g. description of the dataset) and reward (e.g. accuracy with predicted model parameters). Overall reward for the episode is maximum achieved score (accuracy).

I have a simple code in [this repo](https://github.com/iaroslav-ai/black-box-optimization) which does generation of random datasets (completely / partially artificial) and compares model selection with BO, random and genetic algos. Generation of datasets code could be converted to environment(s), and simple optimization methods used as a baseline.

Does this sounds interesting to you? I think this might be specially well suited for model based reinforcement learning, as it tries to be more data efficient than other RL approaches, and as fitting model is a computationally expensive procedure.

tlbtlbtlb:
I think that's a very interesting kind of environment. 

[David Duvenaud](http://people.seas.harvard.edu/~dduvenaud/) has done interesting work in this area. His paper *Gradient-based Hyperparameter Optimization with Reversible Learning* [[arXiv](http://arxiv.org/pdf/1502.03492.pdf), [slides](http://people.seas.harvard.edu/~dduvenaud/talks/hypergrad-talk.pdf)] describes one approach.

Check out this graph from the above paper, plotting loss vs the learning rate hyperparameter:

![image](https://cloud.githubusercontent.com/assets/219508/15223763/e4bebe46-182a-11e6-9acd-998bf5fb4215.png)

I think that sort of curve is common. If you start on the left, a simple convex optimizer might find its way down to the minimum. But if you start on the right, chaos. That's where prior knowledge is helpful, and where your idea of transfer learning from other domains might be a big win.

One issue is that if the steps are very slow, it may be frustrating to test. So it'd be good to have some nontrivial networks that can converge in a fraction of a second for agent development purposes.

[...]

PR #148, opened 2016-05-31
Environment for architecture and training parameters selection for Deep CNN training

iaroslav-ai:
These commits contain an environment as discussed in #94. In this environment agent can select an architecture of deep CNN classifier and training parameters such that it results in high accuracy. 

It was originally discussed to start with more “classical” methods like SVM etc, but I think it is better to start directly with Deep Learning as parameter selection for the “classical” methods is usually relatively straightforward.

One step in this environment is a training of a deep network for 10 epochs, where architecture and training parameters are selected by an agent.  One episode in this
environment has fixed size of 10 steps. Reward is proportional to the best validation accuracy achieved at current step.

Training parameters that agent can adjust are learning rate, momentum, batch size, l1 and l2 penalty coefficients. Agent can select up to 5 layers of CNN and up to 2 layers of fully connected layers.  As an observation, agent receives # of instances in a dataset and a validation accuracy at every step.

At the beginning of every episode, a dataset to train on is chosen at random. Datasets used are MNIST, CIFAR10, CIFAR100. Number of instances in datasets are chosen at random in range from around 100% to 5% such that adjustment of l1, l2 penalty coefficients makes more difference.

Some of the parameters of the dataset are not provided to the agent in order to make
agent figure it out through experimentation during an episode.

This environment requires Keras with Theano or TensorFlow to run. When run on laptop gpu (GTX960M) one step takes on average 2 min.


jietang:
Thanks @iaroslav-ai, merged.

One potential followup: I think it's likely that current RL algorithms won't be able to handle an env that takes 2 minutes per call to step. Many of them involve taking multiple rollouts (i.e. until done=True) and then computing a single gradient update. Maybe there's a way to deliver more feedback as training progresses?

iaroslav-ai:
Right, RL methods that are popular right now require experimentation with environment to be cheap, and therefore I would also expect that this kind of environment will be tough for these methods. 

There are a few ways I could imagine to address this:

Mix this environment with convergence control, such that agent could both set the architecture of the deep net at the beginning of the training, and control / observe convergence. This could potentially increase number of reward evaluations by an order of magnitude, for around the same time cost. 

Also I could imagine an environment (or set of envs) where you start with smaller amounts of data and as agent becomes better, switch to larger data. I could imagine that this would help speed up agent learning at least in the early training stages, due to training being faster on small data.

Also model based RL approaches can be used; Here is one nice tutorial on these methods that I found http://videolectures.net/nips09_littman_mbrl/ if you are not familiar with those. According to some figures there (55:06), model based RL could be more than order of magnitude more data efficient than model free one. However, model based RL would require to build a model of what training results are given architecture / training parameters; such model would require to have stochastic outputs, due to the stochastic nature of training of NN. 

High cost of experimentation with environment is actually one of the reasons I think this environment is interesting, as it requires data efficient RL methods.

I think the next thing that I will do is I will run model free RL methods on this environment, and see how they behave. Would be also interesting to compare model free vs model based RL methods.

jietang:
Yes I agree with your reasoning. One thing I would add is that while high cost of experimentation is interesting, extremely high cost makes it difficult to experiment. For example CartPole takes ~100 rollouts to solve with ~100 timesteps per rollout; if each step took 2 minutes, that's about 2 weeks of compute time for a single training run, and cartpole is on the low end in terms of timesteps and # of rollouts to solve.

iaroslav-ai:
Right, this is a good point.
One thing I could imagine is to train a neural network, which given architecture, training parameters, and any other parameters which make sense estimates what the outcome of the training would be. To build such network, one could run training many times with different (e.g. randomly sampled) architecture / training params etc, thus collecting a dataset, which is then used to train a neural net. 
Then such network can be used internally instead of actual neural network training, provided that it models the validation accuracy of training outcome sufficiently accurately. The good thing is that such neural net needs to be trained only once. 
I could imagine that one can train first RL agent using such computationally efficient version of environment, and then test on real thing, to see if it generalizes. 
Provided that behavior of training can be modeled accurately enough with nn of reasonable size (or any other class of ML models, whatever works best), I would expect that the step cost would dramatically decrease.

jietang:
Yes, that's an interesting idea; you could view it as a model-based RL algorithm, where you want to make better use of expensive real-world experience by fitting a (NN) model and training your RL agent on that.

iaroslav-ai:
Right. Thanks for your feedback! :)

PR #150, opened 2016-06-01
Make agent examples compatible with python 3

maciejgryka:
This doesn't really change much, but makes it possible to run the example agents using python3 also.

### DIFF
--- a/examples/agents/keyboard_agent.py
+++ b/examples/agents/keyboard_agent.py
@@ -1,11 +1,13 @@
 #!/usr/bin/env python
+from __future__ import print_function
+
 import sys, gym
 
 #
 # Test yourself as a learning agent! Pass environment name as a command-line argument.
 #
 
-env = gym.make('LunarLander-v0' if len(sys.argv)<2 else sys.argv[1])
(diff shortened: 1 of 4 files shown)


examples/agents/keyboard_agent.py
```diff
@@ -40,9 +42,9 @@ def rollout(env):
     human_wants_restart = False
     obser = env.reset()
     skip = 0
-    for t in xrange(ROLLOUT_TIME):
+    for t in range(ROLLOUT_TIME):
```
maciejgryka: This should not make any perf difference in Python 2, unless `ROLLOUT_TIME` is a few orders of magnitude larger than currently. (And in python 3 `range` is already a generator.)

gym/envs/box2d/car_racing.py
```diff
@@ -12,7 +12,7 @@
 import pyglet
 from pyglet.gl import *
 
-from car_dynamics import Car
+from .car_dynamics import Car
```
gdb: This should be the fully qualified path: `gym.envs.box2d.car_dynamics`.

gdb: (I'll fix.)

PR #153, opened 2016-06-02
Only warn once about out-of-bound actions and observations

gdb:


gdb:
@joschu seem reasonable?

joschu:
why not use the warnings module? I think that automatically just warns once

gdb:
Is there a way to use the warnings module with dynamic strings like we have here?

joschu:
Hmm maybe not. Anyway, looks good to me.

PR #155, opened 2016-06-02
use imageio in video_recorder

SimonRamstedt:
Hey, I couldn't install ffmpeg because I'm on a university cluster without root rights. But today I stumbled upon the imageio package (https://github.com/imageio/imageio). It downloads and compiles ffmpeg on its own when used for the first time. As a bonus, it would simplify the video_recorder code a bit.

Should work on all OS but I only tested it on CentOS 7.

Issues:
There is this warning that appears for many environments when using imageio:
```bash
[swscaler @ 0x37f6060] Warning: data is not aligned! This can lead to a speedloss
[2016-06-02 22:57:27,990] IMAGEIO FFMPEG_WRITER WARNING: input image is not divisible by macro_block_size=16, resizing from (500, 500) to (512, 512) to ensure video compatibility with most codecs and players. To prevent resizing, make your input image divisible by the macro_block_size or set the macro_block_size to None (risking incompatibility). You may also see a FFMPEG warning concerning speedloss due to data not being aligned.
```

### DIFF
--- a/gym/envs/mujoco/mujoco_env.py
+++ b/gym/envs/mujoco/mujoco_env.py
@@ -116,8 +116,7 @@ class MujocoEnv(gym.Env):
 
     def _get_viewer(self):
         if self.viewer is None:
-            self.viewer = mujoco_py.MjViewer()
-            #self.viewer = mujoco_py.MjViewer(init_width=512,init_height=512)
+            self.viewer = mujoco_py.MjViewer(init_width=512,init_height=512)
             self.viewer.start()
             self.viewer.set_model(self.model)
             self.viewer_setup()
--- a/gym/envs/mujoco/mujoco_env.py
+++ b/gym/envs/mujoco/mujoco_env.py
@@ -117,6 +117,7 @@ class MujocoEnv(gym.Env):
     def _get_viewer(self):
         if self.viewer is None:
             self.viewer = mujoco_py.MjViewer()
+            #self.viewer = mujoco_py.MjViewer(init_width=512,init_height=512)
             self.viewer.start()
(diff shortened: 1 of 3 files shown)


jietang:
Neat, thanks for working on this.

I'm a little hesitant to replace the existing encoder infrastructure; ffmpeg gives a lot of control over how encoding is done, and we might want that flexibility in the future. Perhaps we could fall back to imageio-based encoding if the other methods fail? Or figure out how they're installing ffmpeg in a way that's compatible with your setup (no root) and incorporate that?

SimonRamstedt:
According to http://imageio.readthedocs.io/en/latest/format_ffmpeg.html#ffmpeg it seems to be possible to pass arbitrary ffmpeg parameters to imageio (also see latest commit). If you still prefer to keep in the manual subprocess call, I'll do that. I'd be happy with both.

gdb:
Yeah, my preference is to keep control of the subprocesses ourselves. I also don't think that we want Gym to download `ffmpeg` from the internet — it's pretty surprising to have your library download arbitrary code. I'd rather have an installation helper or guide which helps people install `ffmpeg` — could we do that instead?

SimonRamstedt:
I see. I wanted to avoid building on my own because that again depended on other libraries that I couldn't  install. However now I found that https://ffmpeg.org/ points to these static builds: http://johnvansickle.com/ffmpeg/

So the ffmpeg install guide for users without root access is:

- download static build from http://johnvansickle.com/ffmpeg/
- unpack to arbitrary location
- create symlink to the ffmpeg binary in ~/.local/bin/

Not that hard after all. Sorry for the troubles.

PR #163, opened 2016-06-07
add get_lives()

ShibiHe:
Add a function get_lives() for agent to get current lives in atari games


gdb:
What do you need this information for? We want to be careful about adding more methods and introspection interfaces. 

(If we need to support it, it should probably come back in the `info` dictionary, though it's also easy enough to just call `.ale.lives()` yourself though we'd reserve the right to break it in the future.)

gdb:
(To be clear, our general plan is to keep the interfaces narrow: the goal is for people to write general RL algorithms.)

ShibiHe:
I am convinced. As you said, it is better to call .ale.lives now and keep interfaces narrow.

PR #169, opened 2016-06-10
Make it possible to step() in a newly created env, rather than throwing AttributeError

catherio:
Under the current implementation of Discrete, if I enter the following commands:
`env = gym.make('FrozenLake-v0')
env.step(0)`
the resulting error is
`AttributeError: 'FrozenLakeEnv' object has no attribute 's'`.

The trouble is that Discrete environments (unlike other simple environments, such as CartPole or Copy) must be first reset before a step can be taken, even when the monitor is not turned on, otherwise the "s" attribute does not get set before it is accessed.

This could be fixed in two ways that I can see - by throwing a more informative error than an AttributeError, or by setting "s" upon creation. Here I do the latter.


catherio:
Er, sorry that this looks like 8 commits, but is actually just 1 line. This should be squashed before merging, which I believe is easy to do on your end (easier than me closing this pull request and re-creating it?)

jietang:
You can also create a new branch and git cherry-pick the relevant commit over.

I'd prefer to call _reset() so that we don't repeat logic.

PR #171, opened 2016-06-10
Monitoring Tests - Added skip_mujoco test

ppaquette:


gym/envs/tests/test_envs.py
```diff
@@ -12,7 +12,12 @@ def should_skip_env_spec_for_tests(spec):
     # We skip tests for envs that require dependencies or are otherwise
     # troublesome to run frequently
 
-    # Skip mujoco tests for pull request CI
+    # Skipping if pull request
+    is_pr = os.environ.get('TRAVIS_PULL_REQUEST') != 'false' and os.environ.get('TRAVIS_PULL_REQUEST') is not None
```
gdb: Why do we need to look for `TRAVIS_PULL_REQUEST`? The check below (looking for `MUJOCO_KEY_BUNDLE` or `~/.mujoco`) should subsume this.

ppaquette: PR builds are all failing with Mujoco not found

gdb: The `skip_mujoco` line below this should handle the case of a PR build. It might be buggy though. Are you sure the issue isn't from the `test_monitor_envs` file? Can you paste me a relevant build URL?

ppaquette: travis.yml has the following line, which invalidates the mujoco test.
```
mkdir -p ~/.mujoco
```
By testing if TRAVIS_PULL_REQUEST is set, we can properly bypass the Mujoco test

ppaquette: Example of failing build:

https://travis-ci.org/openai/gym/jobs/136795259

gdb: Ah ha. I just changed the .travis.yml to only create it if `MUJOCO_KEY_BUNDLE` is set (which is true for non-PR builds).

ppaquette: That should also work. There is no skip_mujoco in test_monitor_envs.py though.

gdb: Yep, that needs to be added.

ppaquette:
I removed the is_pr test. The PR now only has the mujoco test in test_monitor_envs

gdb:
Thank you!

PR #174, opened 2016-06-12
Fixes bug in pendulum.py

RafaelCosman:
Calls `reset()` in pendulum init.


gdb:
I'm not actually a huge fan of adding these `reset` calls at initialization time, since it's technically undefined behavior to `step` before calling `reset`.

What about instead changing `step` to error if you've never called `reset`?

RafaelCosman:
Ok, my bad. I misunderstood the API and thought `step()` should be callable before `reset()`.

PR #175, opened 2016-06-12
Add configure method to Env, and support multiple displays in CartPole

gdb:
This allows people to pass runtime specification which doesn't affect
the environment semantics to environments created via `make`.

Also include an example of setting the display used for CartPole

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -21,6 +21,7 @@ class Env(object):
         reset
         render
         close
+        configure
         seed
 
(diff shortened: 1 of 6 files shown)


tlbtlbtlb:
When would `display` need come from code, rather than from the environment variable `$DISPLAY`?

gdb:
When you have multiple displays :). I have a pretty cool VNC thing coming, where we run multiple Xdummy displays in a single docker container.

PR #177, opened 2016-06-12
Makes off_switch_cartpole use a tuple

RafaelCosman:
Now that `test_determinsm` supports tuple equality.

Also adds the ON state to the observation returned by reset.

### DIFF
--- a/gym/envs/safety/off_switch_cartpole.py
+++ b/gym/envs/safety/off_switch_cartpole.py
@@ -32,15 +32,18 @@ class OffSwitchCartpoleEnv(CartPoleEnv):
         self.observation_space = spaces.Tuple((spaces.Discrete(2), self.observation_space))
 
     def _step(self, action):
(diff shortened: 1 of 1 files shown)


gdb:
Is this the format you prefer for your observations, or would you rather it's flattened?

We can change the test rig pretty easily rather than changing the format...

RafaelCosman:
@gdb, I'm unsure which is better. `spaces.Tuple((spaces.Discrete(2), self.observation_space))` seems to make it pretty clear that it should be a Tuple. But maybe `Tuple` doesn't quite mean what its name implies?

gdb:
I'm happy to do it either way. I think it's just a question of how people are going to write their algorithms. I would assume it's generally a bit nicer to have a simple observation space, so people don't have to do pre-processing to get it into a usable format. We can also change it later.

gdb:
(Merging for now.)

PR #181, opened 2016-06-13
Robotics

tlbtlbtlb:
Adds a proxy environment mechanism, basically opening a websocket and sending step & reset as RPCs over the websocket.

It's used by the Fetch robot environment, where the interface to the robot runs in a separate process.

Also, define an environment FetchRobot-v0 which uses the above proxy.


olegklimov:
Hello @tlbtlbtlb , are you implementing existing protocol or making new? If it is new, I think it should be unified with HTTP interface (i.e. use remote CartPole). See work by @catherio .

If we end up with two ways of calling `step()` `reset()` etc over socket, that can't be good.

(and maybe it's just me, but I once tried to implement WebSocket server side, it's too complicated, overengineered, plain HTTP is so much easier and exist everywhere, on the most simple robot in this case.)

tlbtlbtlb:
For the robotics application, there's a tight real-time constraint on being able to return an observation and get a next action in a millisecond or so. Performance seems adequate with websockets. I doubt that separate HTTP requests can get there, but I'd be interested to look at performance numbers.

An additional issue with robots: there's only one, so we need to know who owns it. The session concept of a websocket seems like a good fit. For example, if you kill the process running the agent, the connection is closed and the server can reset the robot and make available for another session.

tlbtlbtlb:
After some discussion, I'm going back to the drawing board.

Issue #185, opened 2016-06-13
Batch mode / human mode


olegklimov:
We should agree on a common way for agent to say "I want fast batch mode" or "I want slow human mode".

In #168 (Remove sleep statement from DoomEnv render) @ppaquette proposed:

``` python
env = gym.make('DoomBasic-v0')
env.mode = 'human'
env.reset()
```

('fast' 70 fps, 'normal' 35 fps, 'human' human can play)

In #119 @jietang proposed:

``` python
env = gym.make('CartPole-v0')
env.configure(batch_mode=True)
```

And agree on the default, too. My thoughts is for simple programs and tutorial we should have human mode as default. But the paper says batch mode is the default. 

After we agree, I can fix all classic control and box2d.

gdb:
Cool. Let's use the configure method, and pass in a "mode='human'" in
appropriate environments so we leave our options open for future modes, and
have it feel similar to render modes.

On Monday, June 13, 2016, Oleg Klimov <notifications@github.com> wrote:

> We should agree on a common way for agent to say "I want fast batch mode"
> or "I want slow human mode".
>
> In #168 <https://github.com/openai/gym/issues/168> (Remove sleep
> statement from DoomEnv render) @ppaquette <https://github.com/ppaquette>
> proposed:
>
> env = gym.make('DoomBasic-v0')
> env.mode = 'human'
> env.reset()
>
> ('fast' 70 fps, 'normal' 35 fps, 'human' human can play)
>
> In #119 <https://github.com/openai/gym/pull/119> @jietang
> <https://github.com/jietang> proposed:
>
> env = gym.make('CartPole-v0')
> env.configure(batch_mode=True)
>
> And agree on the default, too. My thoughts is for simple programs and
> tutorial we should have human mode as default. But the paper says batch
> mode is the default.
>
> After we agree, I can fix all classic control and box2d.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/openai/gym/issues/185>, or mute the thread
> <https://github.com/notifications/unsubscribe/AAM7kSiWudX4pbv3QPZiOz0rCDNtFQOQks5qLWfmgaJpZM4I0YRJ>
> .
>


-- 
Sent from mobile

ppaquette:
Agree that we need to standardize terminology.

My modes are 'fast', 'normal', 'human'

batch_mode == True  ==> 'fast' mode
batch_mode == False ==> 'normal' mode

[...]

PR #186, opened 2016-06-13
human / batch mode using _display_options #185

olegklimov:
#185

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -54,6 +54,11 @@ class Env(object):
         env._action_warned = False
         env._observation_warned = False
         env._configured = False
+        class DisplayOptions: pass
+        env._display_options = DisplayOptions()
+        env._display_options.display_n    = None
+        env._display_options.display_mode = "human"
+        env._display_options.window_pos   = None
 
         # Will be automatically set when creating an environment via 'make'
         env.spec = None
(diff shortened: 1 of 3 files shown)


jietang:
It seems to me like these options don't apply to all environments, and I'd rather not add complexity to the core interfaces unless it adds functionality that is broadly useful.

If you're concerned about code duplication, what do you think about implementing this as a helper class that can be used by classic_control and box2d envs?

olegklimov:
It is not a real PR at the moment, just an idea about reorganizing the code, to throw something specific to https://github.com/openai/gym/issues/185.

I don't really like interfering with others code, in this case @gdb. He has added display_n recently for some reason (docker container?). The idea is that batch mode and window position look like other display flags. We should hear what he thinks first.

(But of course it applies to most environments! Atari, doom, all benefit from ability to place window at predictable place, batch mode. It applies to any environment that have a window and fps.)

gdb:
Yep, I added `display` for docker container. I agree with Jie's comment completely — we should only add things that are basically universal to the core interfaces.

In this case, I'd recommend just adding a `def configure(display_n, display_mode, ...)` or whatever (I don't love the `display_mode` name since it's kinda generic — what about `framerate` or something more specific) with the set of options as appropriate to each class.

olegklimov:
Okey, I'll do another.

Issue #134, opened 2016-05-28
Add tests to ensure that environments do not change over time


gdb:
I'm picturing starting with a JSON file that contains a simple rollout for each env, and we check that it doesn't change.

jietang:
@arushir is looking at this.

PR #196, opened 2016-06-15
Added test for environment semantics

arushir:
Tests that environments do not change over time  #134


gym/envs/tests/generate_json.py
```diff
@@ -0,0 +1,102 @@
+import gym
```
jietang: nit: this import can go away

gym/envs/tests/generate_json.py
```diff
@@ -0,0 +1,102 @@
...
+import numpy as np
+import datetime
+import os
+import hashlib
+
+import logging
+logger = logging.getLogger(__name__)
+
+from test_envs import should_skip_env_spec_for_tests
+
```
jietang: Let's create a toplevel scripts/ folder in the gym repo and put this file there. That way it won't get shipped as part of the package distribution

gym/envs/tests/generate_json.py
```diff
@@ -0,0 +1,102 @@
...
+  rollout_dict[spec.id] = rollout
+
+  with open(ROLLOUT_FILE, "w") as outfile:
+    json.dump(rollout_dict, outfile, indent=2)
+
+  return True
+
+def create_all_rollouts():
+  environments = [spec for spec in envs.registry.all() if spec._entry_point is not None]
+
```
jietang: How do you feel about moving the reading / writing of the json file here? That way we can avoid a bunch of I/O.

gym/envs/tests/test_envs_semantics.py
```diff
@@ -0,0 +1,85 @@
...
+if not os.path.isfile(ROLLOUT_FILE): 
+  with open(ROLLOUT_FILE, "w") as outfile:
+    json.dump({}, outfile, indent=2)
+
+def test_env_semantics(spec):
+  with open(ROLLOUT_FILE) as data_file:
+    rollout_dict = json.load(data_file)
+
+  if spec.id not in rollout_dict:
+    added = create_rollout(spec)
```
jietang: Let's fail the test here rather than creating a rollout on the fly (it makes less sense e.g. on our CI machines)

gym/envs/tests/test_envs_semantics.py
```diff
@@ -0,0 +1,85 @@
...
+
+      total_steps += 1
+      if total_steps >= ROLLOUT_STEPS: break
+
+      if done: break
+
+  observations_now = hashlib.sha1(str(observation_list)).hexdigest()
+  actions_now = hashlib.sha1(str(action_list)).hexdigest()
+  rewards_now = hashlib.sha1(str(reward_list)).hexdigest()
+  dones_now = hashlib.sha1(str(done_list)).hexdigest()
```
jietang: This logic feels like it's shared across the generate and test methods and could be factored out.

gym/envs/tests/test_envs_semantics.py
arushir: I see, I was doing this for the workflow of generating a rollout for a new environment if a rollout doesn't already exist and the env shouldn't be skipped by this test. Alternatively, I can add a function to generate_json to generate the new rollouts for new environments when explicitly called. Does that sound like a better solution?

gym/envs/tests/test_envs_semantics.py
arushir: Good point, will do!

gym/envs/tests/generate_json.py
arushir: I can actually delete the create_all_rollouts function, since it shouldn't need to be called again. Instead, I can add a add_new_rollouts() function as noted below which requires the I/O to see if a rollout already exists for a given environment.

gym/envs/tests/test_envs_semantics.py
jietang: Yes, that sounds good.

gym/envs/tests/generate_json.py
arushir: Sounds good, I'll move it to scripts/generate_json.py. Is it reasonable to add an __init__.py to the envs/tests folder, since this script requires imports from there? http://stackoverflow.com/questions/4383571/importing-files-from-different-folder-in-python

jietang: Yes, feel free to add an ```__init__.py```

scripts/generate_json.py
```diff
@@ -0,0 +1,87 @@
...
+import sys
+import hashlib
+
+import logging
+logger = logging.getLogger(__name__)
+
+from gym.envs.tests.test_envs import should_skip_env_spec_for_tests
+from gym.envs.tests.test_envs_semantics import generate_rollout_hash, hash_object
+
+DATA_DIR = os.path.join(os.pardir, 'gym', 'envs', 'tests')
```
jietang: nit: I would construct the path relative to __file__ here. Otherwise if you run the script from a different directory you won't find the data file.

arushir:
Note that this test skips all of the Doom environments, because VizDoom generates different observations across platforms.

jietang:
Thanks @arushir

PR #199, opened 2016-06-16
Set up gym logger without touching root logger

mthrok:
With this change, gym will set up module-level logger at import,
not touching root logger, while keeping the expected console
output same as current.

Before this, gym was setting up root logger, which reduces the
flexibility of logging, by requiring user to re-configuring root
logger in non-standard way.

With this commit user can disable gym log at global level.

Case 1. If you want to simply disable gym log.
````python
import logging
import gym
logging.getLogger('gym').handlers = []
```

Case 2. If you want to use root logger so that logs are in the
same format as your root logger. (Import time warnings are not affected.)
```python
import logging
logging.basicConfig(
    level=logging.INFO,
    format='%(pathname)s:%(funcName)s:%(lineno)d:%(message)s',
)
import gym
gym_logger = logging.getLogger('gym')
gym_logger.handlers = []
gym_logger.propagate = True
```


mthrok:
@jietang @gdb Can you verify?

gdb:
Setting up logging in libraries is always really hard. My intention was to:

1. Provide a sane default so most users don't have to think about how to configure Gym logging.
2. Make it hard for unsophisticated users wanting to use logging elsewhere to accidentally double log Gym messages.
3. Make it easy for sophisticated users to just get rid of all our logging config.

I think this patch would lose property 2 — unsophisticated users who want to use logging elsewhere would just add their own handler to the root logger, and then end up with double-logging of Gym messages.

I do agree it's cleaner from a layering perspective, but I think in practice anyone concerned can just `undo_logger_setup()` and configure things from scratch.

gdb:
(We could also mess with double logging by doing `propagate = False`, but that starts to feel more cumbersome to me.)

In practice, this is actually the first suggestion we've gotten to change the logging setup, so I think I'd rather wait for more complaints before changing.

mthrok:
The reason why I made this PR was that while I was debugging my code with logging.DEBUG, because of this, I saw a lot of duplicated logs, and it took a while to figure out why because it is not documented anywhere, and because the way root logger is setup does not indicate where it is emitting the log.

I think this patch actually enforces property 2. Because this way, as long as user does not do anything on gym logger, they always remain untouched. Instead, while I was figuring out, why my code ends up emitting my log twice, I disabled and enabled root logger several times.

gdb:
Huh, how'd you end up with double logging? (And why did it only appear when you were at logging.DEBUG?)

You're right we don't document this, and we should document it regardless of what direction is chosen :).

mthrok:
gym code was not double logging. my code was double-logging

```python

import logging
logging.basicConfig(  # <- this adds a stream handler
    level=logging.DEBUG,
    format='some formatting'
)

import gym  # <- another handler is added here

# in my code
logger = logging.getLogger(__name__)
logger.debug('some message')  # <- emitted by both handlers
```

gdb:
Ok, makes sense.

Anyway, I think I'd be happiest with just adding docs on how to disable our logging config for now, and wait for more complaints before picking a long-term answer. Does that sound reasonable?

mthrok:
By `propagate = False` you can consider it as separate logging module from root logging module.
I think this approach is more general while preserving your intent.

mthrok:
by the way `undo_logger_setup` is broken. Python's logging module does not have `NOTEST` attribute.

gdb:
Hm, the method works for me — it's `NOTSET`, not `NOTEST`.

In any case, another way to describe what's going on is the following:

1. By default, Gym is opinionated about how all your logs work. We ended up there by observing that most ML researchers weren't familiar with the logger and didn't want to spend the time fiddling with it.
2. This is the wrong approach for any sophisticated user. If you care about how your logs work, you should just turn off our configuration altogether.

Anyway, I understand where you're coming from, and I think for many libraries and contexts your perspective would be correct, but I don't think it's a clear win for Gym right now. I appreciate the PR and the detailed thoughts, but I'm going to close this for now.

I would however love a patch which documents our current logger setup. (And if anyone in the community has opinions I'd be happy to hear them.)

mthrok:
Sorry, I was reading it wrong > `NOTEST` it was `NOTSET`

zhousanmao:
I meet a problem 

(tf) zhousanmao@ubuntu:~/Desktop/GA3C-master/ga3c$ cd /home/zhousanmao/Desktop/GA3C-master/ga3c
(tf) zhousanmao@ubuntu:~/Desktop/GA3C-master/ga3c$ sh _clean.sh(tf) zhousanmao@ubuntu:~/Desktop/GA3C-master/ga3c$ sh _train.sh
  File "GA3C.py", line 68, in <module>
    gym.undo_logger_setup()
AttributeError: module 'gym' has no attribute 'undo_logger_setup'

The error code is    gym.undo_logger_setup()

could you please tell me  what is wrong??

PR #201, opened 2016-06-16
humanoidmodelwalk and humanoidstand

zdx3578:
humanoidmodelwalk and humanoidstand;

walkreward= (self.model.data.xipos[6][0]-self.model.data.xipos[9][0]) if (rotateleg % 2) else (self.model.data.xipos[9][0]-self.model.data.xipos[6][0]) 
walk should change leg every time

stand reward = reward - abs(vel).

standup change to   done = bool(qpos_bf[2]>qpos[2]+0.25) : when stanup ,if failed is done

### DIFF
--- a/gym/envs/__init__.py
+++ b/gym/envs/__init__.py
@@ -216,6 +216,10 @@ register(
      id='HumanoidModelWalk-v0',
      entry_point='gym.envs.mujoco:HumanoidModelWalkEnv',
 )
+register(
+    id='HumanoidStand-v0',
+    entry_point='gym.envs.mujoco:HumanoidStandEnv',
+)
 
 
 # Atari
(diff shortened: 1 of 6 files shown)


tlbtlbtlb:
I don't understand how the rotateleg logic makes sense. Your code switches the reward computation every timestep (0.01 seconds), while taking a put-one-leg-in-front-of-the-other step requires 10-30 timesteps.

zdx3578:
sometime  http://weibo.com/p/23044425240a440a2dad2c6a392c7dc778996b; leg not rotation;
http://weibo.com/p/23044406c6232793c33d2137ebc304f5b04882  this look good; train every time rotation result is not good ;  delete this .

PR #203, opened 2016-06-17
walk just tow legs not same time upground

zdx3578:
walk just tow legs not same time upground


tlbtlbtlb:
We're not eager to add new reward functions. Can you demonstrate that these different reward functions allow much easier learning, or that they learn something better? We'd want to see results from your agent and TRPO comparing this environment with the existing humanoid environment before considering adding new ones.

zdx3578:
thanks

PR #205, opened 2016-06-17
IDEA: makes monitor look for true_reward in info

RafaelCosman:
Not requesting a merge. I'd like to discuss if this approach makes sense.


gdb:
Per chat, I think we should just override the `step` (not just `_step`) in the relevant environments directly.

RafaelCosman:
Sounds good. Closing this PR, new one coming up shortly.

PR #206, opened 2016-06-17
overrides step() in safety envs to pass monitor true_reward

RafaelCosman:


RafaelCosman:
```
Ran 548 tests in 115.778s

OK
```

PR #210, opened 2016-06-17
fixes cartpole observation_space

### LINKED ISSUE #195: InterpretabilityCartpoleObservations-v0 environment runtime warning
Taking a single step with a random action in the InterpretabilityCartpoleObservations-v0 environment, generates the following error:

_RuntimeWarning: invalid value encountered in greater_equal
return x.shape == self.shape and (x >= self.low).all() and (x <= self.high).all()
[2016-06-14 23:08:07,331] Action '(0, array([ 0.89130833,         nan,  0.29987182,         nan]), array([ 1.18621149,         nan, -0.16961701,         nan]), array([-2.18249957,         nan,  0.26152186,         nan]), array([-1.02926596,         nan, -0.13622267,         nan]), array([-1.26488122,         nan, -0.30129903,         nan]))' is not contained within action space 'Tuple(Discrete(2), Box(4,), Box(4,), Box(4,), Box(4,), Box(4,))'._

Repro: 
import gym
env = gym.make('InterpretabilityCartpoleObservations-v0')

observation = env.reset()
observation, reward, done, _ = env.step(action)

RafaelCosman:
Fixes #195.

I propose that for all spaces, the following property should hold:

`my_space.contains(my_space.sample())`

Infinite spaces violate this property because `sample()` returns `nan`.

Thus I propose that we either accept this PR or modify `sample()` such that it can sample from infinite spaces.

### DIFF
--- a/gym/envs/classic_control/cartpole.py
+++ b/gym/envs/classic_control/cartpole.py
@@ -33,7 +33,12 @@ class CartPoleEnv(gym.Env):
         self.x_threshold = 2.4
 
         # Angle limit set to 2 * theta_threshold_radians so failing observation is still within bounds
-        high = np.array([self.x_threshold * 2, np.inf, self.theta_threshold_radians * 2, np.inf])
+        high = np.array([
+            self.x_threshold * 2,
+            np.finfo(np.float32).max,
+            self.theta_threshold_radians * 2,
+            np.finfo(np.float32).max])
+
         self.action_space = spaces.Discrete(2)
         self.observation_space = spaces.Box(-high, high)
 


JKCooper2:
It may be worth setting the cartpole values to something more restrictive, if the sampled action predicts and observation of 10^200 then determining the l2 error in PredictObsCartpole-v0
`l2dist = np.sqrt(np.sum(np.square(np.subtract(self.predicted_observations[-(i + 1)][i], observation))))` is going to result in an overflow in any case from squaring the value

The practical range is only something like +/- 120 for the cart's speed (+/- 2.4 for the x space and moving 1 frame per 0.02 seconds) where 1 step results in the cart going from one side to off the other side (which would never be reachable in any case due to the acceleration rate), and I'm guessing it would be similar for the pole acceleration.

I don't see there being any benefit from having such a large range for CartPole. I think it would be better to present unbounded range problems in their own environments like a number guessing game

PR #217, opened 2016-06-18
Guessing Game environments

JKCooper2:
Added two guessing game environments in order to tests agents exploration parameter setting abilities.

There's an explanation of each env at the start of each class. 

I've confirmed an agent is capable of learning (consistently achieving a higher score than random) on HotterColder.

Guessing Game should be possible to learn from what I can tell but may require a different agent structure as improved scores need a decent exploration model capable of using histories

### DIFF
--- a/gym/envs/toy_text/guessing_game.py
+++ b/gym/envs/toy_text/guessing_game.py
@@ -1,0 +1,87 @@
+import gym
+from gym import spaces
+from gym.utils import seeding
+import numpy as np
+
+
+class GuessingGame(gym.Env):
+    """Number guessing game
+
+    The object of the game is to guess within 1% of the randomly chosen number
+    within 200 time steps
+
+    After each step the agent is provided with one of four possible observations
+    which indicate where the guess is in relation to the randomly chosen number
+
+    0 - No guess yet submitted (only after reset)
+    1 - Guess is lower than the target
+    2 - Guess is equal to the target
+    3 - Guess is higher than the target
+
+    The rewards are:
+    0 if the agent's guess is outside of 1% of the target
+    1 if the agent's guess is inside 1% of the target
+
+    The episode terminates after the agent guesses within 1% of the target or
+    200 steps have been taken
+
+    The agent will need to use a memory of previously submitted actions and observations
+    in order to efficiently explore the available actions
+
+    The purpose is to have agents optimise their exploration parameters (e.g. how far to
+    explore from previous actions) based on previous experience. Because the goal changes
+    each episode a state-value or action-value function isn't able to provide any additional
+    benefit apart from being able to tell whether to increase or decrease the next guess.
+
+    The perfect agent would likely learn the bounds of the action space (without referring
+    to them explicitly) and then follow binary tree style exploration towards to goal number
+    """
+    def __init__(self):
+        self.range = 1000  # Randomly selected number is within +/- this value
+        self.bounds = 10000
+
+        self.action_space = spaces.Box(low=np.array([-self.bounds]), high=np.array([self.bounds]))
+        self.observation_space = spaces.Discrete(4)
+
+        self.number = 0
+        self.guess_count = 0
+        self.guess_max = 200
+        self.observation = 0
+
+        self._seed()
+        self._reset()
+
+    def _seed(self, seed=None):
+        self.np_random, seed = seeding.np_random(seed)
+        return [seed]
+
+    def _step(self, action):
+        assert self.action_space.contains(action)
+
+        if action < self.number:
+            self.observation = 1
+
+        elif action == self.number:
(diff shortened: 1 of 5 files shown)


RafaelCosman:
@JKCooper2, you might want to use `git rebase --interactive` to fix the ugly `Merged branch master into master` commits.

gym/envs/toy_text/guessing_game.py
```diff
@@ -0,0 +1,91 @@
...
+    The agent will need to use a memory of previously submitted actions and observations
+    in order to efficiently explore the available actions
+
+    The purpose is to have agents optimise their exploration parameters (e.g. how far to
+    explore from previous actions) based on previous experience. Because the goal changes
+    each episode a state-value or action-value function isn't able to provide any additional
+    benefit apart from being able to tell whether to increase or decrease the next guess.
+
+    The perfect agent would likely learn the bounds of the action space (without referring
+    to them explicitly) and then follow binary tree style exploration towards to goal number
```
RafaelCosman: Taking actions outside of `action_space` is inelegant. Why not just expand the action space to be twice or three times as large as the range that's actually used?

gym/envs/toy_text/hotter_colder.py
```diff
@@ -0,0 +1,66 @@
...
+    The goal of hotter colder is to guess closer to a randomly selected number
+
+    After each step the agent receives an observation of:
+    0 - No guess yet submitted (only after reset)
+    1 - Guess is lower than the target
+    2 - Guess is equal to the target
+    3 - Guess is higher than the target
+
+    The rewards is calculated as:
+    (min(action, self.number) + self.range) / (max(action, self.number) + self.range)
```
RafaelCosman: Doesn't this mean that `self.number` can be computed from the reward?

gym/scoreboard/__init__.py
```diff
@@ -602,6 +602,20 @@
     contributor='machinaut',
 )
 
+add_task(
+    id='GuessingGame-v0',
+    group='toy_text',
+    experimental=True,
+    contributor='jkcooper2',
```
RafaelCosman: Can you add `summary`, `description`, and `background`?

gym/scoreboard/__init__.py
```diff
@@ -602,6 +602,20 @@
...
+)
+
+add_task(
+    id='HotterColder-v0',
+    group='toy_text',
+    experimental=True,
+    contributor='jkcooper2',
```
RafaelCosman: Can you add `summary`, `description`, and `background`?

gym/envs/toy_text/guessing_game.py
RafaelCosman: As currently implemented, agents can't 'learn the bounds of the action space' because you `assert self.action_space.contains(action)`.

RafaelCosman:
Build seems to be failing on this PR for the following reason:

```
Ran 733 tests in 227.444s
OK
ERROR: InvocationError: '/home/travis/build/openai/gym/.tox/py34/bin/nose2'
```

Doesn't seem like the PR's fault.

JKCooper2:
Thank you for the feedback, I'll make some changes and re-push.
I haven't used git a whole lot so will try the rebase and push --force-with-lease and see how that goes

> Taking actions outside of action_space is inelegant. Why not just expand the action space to be twice or three times as large as the range that's actually used? Note: as currently implemented, agents can't 'learn the bounds of the action space' because you assert self.action_space.contains(action).

I'll expand the action_space to be 10x the random number bounds as you suggest (I'd like agents to be able to overshoot by quite a way)

> Doesn't this mean that self.number can be computed from the reward?

Yes. This is mostly to keep the environment simple. I wanted to make some simple environments where the observation alone isn't enough to solve the problem efficiently, and instead require agents that need to use the history including rewards and actions to choose the right next action, as well as exploration algorithms that take into account these histories. While it does mean it's possible for an agent to solve in two steps (one to determine which way to move, the second to move that distance), I can't see it being it an issue because the agent would also needs to learn that the maximum reward is 1 (and not just overshoot by a massive margin). 

The more trivial it is for a human to solve when a computer can't the more interesting I think the environment is :)

I wouldn't mind adding a few more of this type of problem and have them as a challenge set where you submit a single agent and it gets run over 5 or 6 guessing game style envs to highlight weak points of the exploration algorithm chosen by the agent

> Can you add summary, description, and background?
Yep, will do

I'll also add a check to make sure the denominator of reward of HotterColder is not zero as well, just in case.

gdb:
@RafaelCosman thanks for the review! Ready for merge in your opinion?

RafaelCosman:
@gdb @JKCooper2, this looks good to me! Sorry for the slow reply.

PR #226, opened 2016-06-27
adds timestep_limit to safety envs

RafaelCosman:


gdb:
Hm, why so short?

RafaelCosman:
Just matching `cartpole-v0` and `pendulum-v0` so results are comparable.

PR #227, opened 2016-06-28
Add hoodle environment.

baoblackcoal:
An environment for classic control. 

A hoodle is on a one-dimensional wood surface. The goal is to push the hoodle to the goal position on the right, and hoodle will do the slow movement after give a start velocity.. Therefore, the only way to succeed is to give a correct start velocity to move the hoodle to goal position.


baoblackcoal:
https://gist.github.com/baoblackcoal/a26069513780b63acca3b1b619bc503e

The above link is a hill climbing algorithm that performs well with the hoodle environment.

jietang:
I haven't encountered this problem before. Is it described in a paper somewhere?

baoblackcoal:
@jietang yeah, you are right. The hoodle environment is my invention. Is this worth to incorporate it as a gym environment?

baoblackcoal:
@jietang There are two video on youtube, 
https://youtu.be/aQaIJ8Vgg68
https://youtu.be/XyF3EzC-2ik

jietang:
I'm okay with including it. Mind moving it out of classic_control and into a separate category? You can make one specifically for this env.

baoblackcoal:
@jietang  thanks for your reply! I have an idea that gym can make a contributor's category and explain that official will not maintain. is the contributor's category appropriate?

jietang:
You should put an experimental=True tag on the scoreboard registration to indicate that it is not officially maintained. There are already several categories that are contributored.

baoblackcoal:
Hmm, I will put the exprimental=Ture tag. BTW, I have no idea about  making a specifically name for the env, do you have any ideas for my confused.

baoblackcoal:
@jietang  I have created a contrib category, is that ok？

JKCooper2:
I'm not sure the observations provided are actually necessary seeing as all of them are fixed after the first step is taken. It may work better just to provide the current position/velocity of the object.

I would probably also recommend keeping track of the current position/velocity after each step rather than recalculating this information every time. It means that you can easily add another environment expanding the challenge e.g. have a small amount of randomness added to the distance traveled at each step, have it like a game of horse where each episode is a series of 5 shots at different targets lengths (the distance, from the start or from the relative position of the object, would need to be included in the observation), allowing multiple shots to be taken, converting to a 2d surface (essentially extending it into a mini golf game)

The challenge from what I can see is for the agent to learn that only the first action has any effect on the outcome, and that subsequent actions are not responsible for the change in observations. I think this is quite a good challenge because while a basic Q learning agent will eventually reach that point (after an amount of time relative to the number of steps between start and end state), a model based agent should (hopefully) quickly realise that the distributions of change in observations are equal for all actions taken at any point after the first state and then work back from that to recognise that only the first action is relevant. So it can act as a good introduction to model based agents versus non-model agents in tackling the credit assignment issue.

baoblackcoal:
@JKCooper2 
Wooo, it is awesome that you have deep knowledge of the hoodle env , gym env and relevant algorithms. thanks for your advice and ideas. 

It is really enough that just to provide the current position/velocity of the hoodle, this reminds me of
 Kapathy's post http://karpathy.github.io/2016/05/31/rl/ , the agent have just be provided the pixels of Pong.

I have some ideas about observation of the hoodle env, all variables(enable_set_start_v, goal_distance, cur_distance, velocity_start, cur_veloticy) should be linked with reaching the goal distance. I hope to have an agent to summarizes the relationship between the variables and the target, the agent can expresse the relationship in words or formulas that will be better, such agent is so powerful. :)  

example of expanding the challenge is a great idea and the mini golf game is impressive. I have imagined that goal distance will be changed by random after agent have been reached the last goal distance.

jietang:
Apologies for leaving this in limbo for so long.

After consulting with some researchers internally, we want to start being more deliberate about the  environments we include. We're still sorting out exactly what that means, but some examples are simple environments that are frequently tested in published papers, or environments that are visually interesting and challenging for humans.

Please feel free to publish this as a separate add on environment for gym. If you import gym and call gym.envs.registration.register() at the toplevel somewhere in your package, you should be able to call gym.make('Hoodle-v0') and have the interface be the same.

JKCooper2:
Is it possible to have the gym have a function to download/run unofficial environments. Something like
gym.get("https;//site.com/hoodle.tar", update=True) which downloads a tarball containing environment files of a particular format, or uses a cached local copy if it's already downloaded (update=True would force it to redownload), and then registers that environment on the fly to allow gym.make("Hoodle-v0")?

Or you could run it as gym.make("Hoodle-v0", source="https;//site.com/hoodle.tar", update=True)

Opening up the upload function to allow uploading to a location of your choosing as well would be useful. Because combined it would allow for a third party site to manage env's, leaderboards, challenges, displays, etc using the gym as the base processor.

baoblackcoal:
@jietang  It dose not matter.  I really love to create gym env. I  am creating a quantitative trading env  recently , and then test some algorithms.

@JKCooper2 The function to download/run unofficial environments is a great idea, but it may be requires a lot of work.

gdb:
It's actually pretty simple to do this on top of Pypi.

- Create your own Pypi package (`foo`) containing your environments. Make sure the package makes the appropriate `.register(...)` invocation.
- People can install your package as `pip install foo`, and use the environments after doing `import gym; import foo`

baoblackcoal:
@gdb, thanks, I will test, and should I close this PR?

PR #228, opened 2016-06-28
Generic pomdp

fatemi:
An environment for generic POMDPs. Such environments are quite important in testing/comparing performance, stability and robustness of RL algorithms. 
- support for arbitrary underlying MDP
- support for arbitrary clutter added to the MDP state
- support for random confusion (multiplier)
- support for arbitrary partial observability
- support for dot file representation of underlying MDP


tlbtlbtlb:
This looks interesting. Do you have an agent that performs well with it? Does the underlying MDP correspond to something that's been published?

fatemi:
The underlying MDP could be anything and in fact any arbitrary POMDP environment can simply be derived from this class. The one I register in here is a typical one (not yet been used in an official paper). I have tested DQN, DDQN, and policy gradient actor-critic methods with an stand-alone version of this environment. I plan to port the algorithms here as well.

tlbtlbtlb:
Why not wrap an existing Gym environment? That would be allow comparing the partially observable version against the fully observable version. If the table driven environment is useful on its own, it could also be a Gym environment.

A major goal of Gym is to provide standard environments so that performance of agents can be compared against as many other agents as possible, including historical ones. The literature is full of "my agent does great on this environment which nobody else has used before or since" papers, and we want to improve that. So any time we can use a previously described environment we prefer to.

fatemi:
Totally agree. However, to the best of my knowledge, there is no POMDP in existing gym environments (am I missing something?). That's why I have written this one as a generic gym environment. In my view, algorithms must be benchmarked on a complex enough POMDP instead of games. The reason is that a POMDP can represent a complicated real-world scenario with all the clutter/noise, uncertainty, and unobservability. These properties are not necessarily part of an atari game for example (and they are not controllable by the developer as opposed to a set of benchmarking POMDPs).

jietang:
Hi fatemi, I think we're agreed that POMDPs are interesting and that it would be nice to have some in gym. 

When registering new environments I prefer envs where it is easy for a new user to understand what the env is testing (e.g. because there is a paper describing the benchmark, or because it has a clear, succinct description). After reading the name and parameters of the proposed new env it's not clear to me e.g. why the transition matrix is set up the way it is and when I would use it over a POMDP with slightly different parameters. 

Perhaps you could pick a more descriptive (or standard) concrete instantiation?

Implementation-wise having a single underlying generic POMDP class that supports multiple concrete instantiations seems very reasonable. It might be worth looking at FrozenLakeEnv as an example - it derives from a DiscreteEnv base class representing MDPs with discrete state-action spaces, and we have multiple different size maps as different registered envs.

fatemi:
Thank you jietang for your comment. With this, I mainly aimed at having a generic POMDP base class, which is lacking in current gym environments. The registered instance is an example for how to use it. Maybe I should have added a succinct explanation where the example comes from (the underlying MDP for the instantiated env is a simple user-simulator in a "restaurant-seeking" dialogue system).

jietang:
I'd like to wait for more use cases before standardizing on infrastructure for all POMDPs, but I'd be fine merging this if you change the id of the environment in the gym and scoreboard register calls to something descriptive (e.g. RestaurantSeekingDialog) along with a description of the specific environment. I'd also be okay with a patch that commits the generic POMDP class as an example for others to build on without registering any associated environments.

fatemi:
Sorry I was away. I will apply the revisions you mentioned tomorrow and we can go from there.

fatemi:
I have changed the registered id to `RestaurantSeekingDialog`. I've also added both description and a readme. It should be clear and easy to understand for a new user.

jietang:
Apologies for leaving this in limbo for so long.

After consulting with some researchers internally, we want to start being more deliberate about the  environments we include. We're still sorting out exactly what that means, but some examples are simple environments that are frequently tested in published papers, or environments that are visually interesting and challenging for humans.

Please feel free to publish this as a separate add on environment for gym. If you import gym and call gym.envs.registration.register() at the toplevel somewhere in your package, you should be able to call gym.make('RestaurantSeekingDialog-v0') and have the interface be the same.

fatemi:
No problem at all.

That would be understandable. The only point is that POMDPs are quite important and unfortunately missing here. In any case, please let me know if you want me to apply changes or to close this PR.

Also, in the meanwhile I wrote a Tic-Tac-Toe environment which is a basic text book example (see Sutton's book for example). Do you think that one fits in gym (in the board-games)?

danijar:
@fatemi Many of the existing environments are in fact POMDPs. For example, the state of an Atari game is fully described by its main memory, so the RAM versions of the Atari envs are MDPs. But the visual versions are POMDPs as the screen buffer is a function of the current RAM state that hides some information (for example, you can't tell whether an enemy is moving to the left or right from a single frame).

fatemi:
@danijar Being a theorist or algorithm designer, you will need to have full control over the environment to reliably evaluate your algorithm and assess how it works in real-life situations, far beyond game settings. One example is that you may need to gradually increase the difficulty of the problem (e.g., decrease its degree of observability, have more uncertainty, have more underlying states, etc.) and see how your algorithm/implementation reacts to the change. For that reason you will need to have a fully configurable stochastic POMDP. The fact that some of the known games are/could be characterized as hypothetical POMDPs is besides the point here. You may also use generic stochastic POMDPs for the proof of concept, since you can configure them in accordance to the assumptions of your algorithm.

hectorpal:
@fatemi @danijar Any progress on this?

fatemi:
@hectorpal This PR is more than two years old. Back then, I vastly used this environment for sanity-check of my algorithms and thought it might be quite useful for others as well. It is not clear to me if folks at OpenAI are willing to merge it at this point. If so, I should update it and fix the conflicts. Are you using (or wish to use) it?

hectorpal:
@fatemi Some people in the planning community are thinking about creating environment for some planning benchmarks. One important difference is that we cannot list all the possible states. That's the whole point. The space is fully-observable but big.

hectorpal:
@fatemi we also have some benchmarks equivalent to POMDPs over factored spaces. So, yes, we may want this. I think people this days are creating their own environments.

I'm not sure yet if the `observation_space` can be more flexible. A Box is a way to define a huge space, but what we need is factored space.

fatemi:
@hectorpal I see. In a way, this env is quite useful for planning purposes, in particular due to the fact that it also allows for transition uncertainty in terms of noise and confusion. For factored spaces (I assume you mean factored state-spaces and not action), the story is somehow different. If the underlying state-space is too huge, then we may need more efficient implementation than hard-coded python to handle the extremely large graph. It would be interesting though. We may then have a simple embedding layer over the state to build the factored observation space. Currently, the transitions are handled simply with numpy arrays. But I think it should not be too challenging to switch to an efficient c++ based graph library. Is it aligned with what you have in mind?

hectorpal:
@fatemi yes, factored state-spaces. Let's assume the space is too huge to generate.

I don't see a fundamental problem generating states on the fly. 
One option is to represent states as a fixed-length numpy array.
Another, more common in planning, is a sparse representation.
Both are cheap.

However, the current documentation of Env says implementations should 
```
set the following attributes:
        observation_space: The Space object corresponding to valid observations
```

You do so setting the `observation_space` to a `box`,
one of the helpers for creating spaces.

But all the spaces, including `box`, say they should implement
implements `sample()` and `contains(x)`.
In a planning factored spaced sampling and contains are as complex as getting to a goal.

In summary, I hope I don't need to support sample() and contains(x).
What do you think?

pzhokhov:
the timeline of this PR makes me a little uncomfortable... I am going to close this per @jietang 's reasoning.

PR #229, opened 2016-06-29
exposing save, load, clone and restore states at enviroment level

mbz:
mainly for more visibility. 
this is a critical requirement for replicating results which require moving between states such as Deepmind's solution to Montezuma: [Unifying Count-Based Exploration and Intrinsic Motivation](https://arxiv.org/abs/1606.01868)


gdb:
Thanks for the suggestion!

I'd prefer not to add the passthrough methods right now — people can just as easily call the methods on `.ale` directly. Wanna just add these commented-out, and see if people start wanting them?

mbz:
makes sense, will comment them out in case someone was interested.

PR #230, opened 2016-06-30
fixed global variable reference when ssl verification is off

kvfrans:
warned isn't accessed properly, this crash only occurs when SSL verification is off


gdb:
Good catch. Mind putting the `global` declaration directly into the relevant method?

PR #233, opened 2016-07-04
Deallocate opponent and board in _close in GoEnv

### LINKED ISSUE #232: Go environment does not clean up after itself
We should implement a `_close` method on the Go environment, and close all open resources. @hojonathanho mind taking?

hojonathanho:
Fixes #232


gdb:
Is it enough to just drop your reference to them, or do you also have to explicitly close an underlying Pachi process?

hojonathanho:
There's no separate Pachi process; it's just C code invoked via Cython. Dropping references calls the destructor for the Pachi engine.

PR #260, opened 2016-07-23
minor changes to cartpole+acrobot, upgrade to v1

joschu:
cartpole: bump timestep limit to 500, so you need to have a decent balancing policy to solve it
acrobot: change observation to use cos/sin of angles, and increase timestep limit so acrobot gets solved by random exploration

### DIFF
--- a/gym/envs/__init__.py
+++ b/gym/envs/__init__.py
@@ -51,10 +51,10 @@ register(
 # ----------------------------------------
 
 register(
-    id='CartPole-v0',
+    id='CartPole-v1',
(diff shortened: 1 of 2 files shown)


gdb:
Do you think it's worth keeping around the old CartPole-v0 as well?

gdb:
(Might be nice to keep existing code running, even though it's deprecated.)

joschu:
Might as well, since we're not even changing cartpole.py

rafaelcp:
Acrobot-v0 got removed! Can you restore it?

gdb:
That was intentional — why do you want Acrobot-v0?

PR #264, opened 2016-07-26
Update Dockfile

JKCooper2:
Build was failing where folder names weren't ending with /
Set gym to be added to PYTHONPATH automatically

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,15 +21,17 @@ RUN apt-get update \
  && rm -rf /var/lib/apt/lists/* \
  && easy_install pip
 
-WORKDIR /usr/local/gym
+WORKDIR /usr/local/gym/
 RUN mkdir gym && touch gym/__init__.py
-COPY ./gym/version.py ./gym
-COPY ./requirements.txt .
-COPY ./setup.py .
+COPY ./gym/version.py ./gym/
+COPY ./requirements.txt ./
+COPY ./setup.py ./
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -21,15 +21,17 @@ RUN apt-get update \
...
+COPY ./requirements.txt ./
+COPY ./setup.py ./
 RUN pip install -r requirements.txt
 
 # Finally, upload our actual code!
-COPY . /usr/local/gym
+COPY . /usr/local/gym/
 
-WORKDIR /root
+RUN export PYTHONPATH="${PYTHONPATH}:/usr/local/gym"
```
gdb: I don't think this will actually work, since the relevant shell process is just present for that build step.

I'd prefer doing `pip install -e /usr/local/gym` in any case.

Thanks!

JKCooper2: Run just adds another small image layer over the top with the export PYTHONPATH saved

It means you can run the agents with:
`docker run -it openaigym:0.2 python /usr/local/gym/examples/agents/random_agent.py`

I think it'd be good if in the end it was able to download an agent from the internet (git repo or otherwise), based on a key and runs the agent submitting all the results as per normal. This means you could code your file locally, submit it somewhere online, run `docker run -it openaigym:0.2 python /usr/local/gym/load.py %agent_key%` and it'll handle everything else

gdb:
The new Dockerfile should resolve these issues.

PR #265, opened 2016-07-26
Added ability to seed the spaces

smcgregor:
Thanks for all your work on this. It looks like I am one of the edge cases from your [discussion](https://github.com/openai/gym/commit/58e6aa95e5af2c738557431f812abb81c505a7cf#commitcomment-17669277) about sampling the spaces. I am trying to reproduce trajectories from seed values in a web server that doesn't restart between requests, but I can't assign the seed of the space sampler to reproduce stochastic actions.

The spaces [all import](https://github.com/openai/gym/blob/master/gym/spaces/discrete.py#L4) a [PRNG helper](https://github.com/openai/gym/blob/master/gym/spaces/prng.py), but since the PRNG is imported outside the [space subclasses](https://github.com/openai/gym/blob/master/gym/spaces/discrete.py#L6) it is shared among all Space [instances](https://github.com/openai/gym/blob/master/examples/agents/random_agent.py#L33).

Calls to `env.monitor.start` with an assigned seed will not reset the space's seed. This seems reasonable to me, but the way the PRNG is imported into spaces makes it difficult to manually assign its seed. There is currently no way to re-run trajectories based on a seed value without, (a) removing stochastic actions, (b) opening a new Python shell, or (c) resorting to a dirty hack.

Suggested Fix (in this pull request):

Add an instance variable to each of the spaces referencing the shared PRNG so it can be referenced by a `seed_prng` function in the Spaces superclass. See code in pull request for details.

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -265,3 +265,9 @@ class Space(object):
         """Convert a JSONable data type to a batch of samples from this space."""
         # By default, assume identity is JSONable
         return sample_n
+
+    def seed_prng(self, seed):
+        """
+        Seed the random number generator associated with the class.
+        """
+        self.prng.seed_action_space(seed)
--- a/gym/spaces/__init__.py
+++ b/gym/spaces/__init__.py
@@ -1,7 +1,7 @@
 from gym.spaces.box import Box
 from gym.spaces.discrete import Discrete
 from gym.spaces.high_low import HighLow
-from gym.spaces.prng import seed
+from gym.spaces.prng import seed_action_space
 from gym.spaces.tuple_space import Tuple
 
 __all__ = ["Box", "Discrete", "HighLow", "Tuple"]
--- a/gym/spaces/box.py
+++ b/gym/spaces/box.py
@@ -17,6 +17,7 @@ class Box(gym.Space):
             Box(-1.0, 1.0, (3,4)) # low and high are scalars, and shape is provided
             Box(np.array([-1.0,-2.0]), np.array([2.0,4.0])) # low and high are arrays of the same shape
         """
+        self.prng = prng
         if shape is None:
             assert low.shape == high.shape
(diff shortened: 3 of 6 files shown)


smcgregor:
Demonstration of the "bug" and the API to "fix" it.

        import gym.spaces.discrete
        
        #
        # Problem: The random seed is shared and hidden
        #
        d1 = gym.spaces.discrete.Discrete(2)
        print d1.sample()  # >> 0
        
        d2 = gym.spaces.discrete.Discrete(2)
        print d2.sample()  # >> 1
        
        #
        # Fix: Add `seed_prng` function to the Space superclass
        #
        d1 = gym.spaces.discrete.Discrete(2)
        d1.seed_prng(0)
        print d1.sample()  # >> 0
        
        d2 = gym.spaces.discrete.Discrete(2)
        d2.seed_prng(0)
        print d2.sample()  # >> 0

gdb:
Thanks for the suggestion. It's clear you've done your homework on reading the historical discussion here!

Just to be clear, I may be missing something, but in your branch:

```
    d1 = gym.spaces.discrete.Discrete(2)
    d1.seed_prng(0)
    print d1.sample()  # >> 0
```

is equivalent to this in the current code right?

```
    d1 = gym.spaces.discrete.Discrete(2)
    from gym.spaces import prng
    prng.seed(0)
    print d1.sample()  # >> 0
```

So this is effectively just syntactic sugar, rather than a semantics change? If so, I actually don't prefer the new syntax, since it makes it seem like `d1` has its own PRNG rather than in fact having a shared one.

smcgregor:
Your justification for leaving it as-is sounds reasonable to me. I fell down the rabbit hole here and your comment helped me get it sorted. Thanks! Hopefully this PR will help others in the future.

PR #266, opened 2016-07-26
Add Soccer Environments

mhauskn:
Adds 3 soccer tasks based on Half-Field-Offense (https://github.com/LARG/HFO).

Limitations + Warnings:
1) Visualization doesn't support RGB images. Currently only human mode through the SoccerWindow2 visualizer.
2) Reward signal for SoccerEmptyGoal environment should be tested with an agent smarter than random agent. I ported from c++ and it's possible I may have made a mistake.
3) I've only added three tasks: Soccer-v0 is going to be hard as the reward signal is very sparse. SoccerEmptyGoal and SoccerAgainstKeeper should be easier.


tlbtlbtlb:
Wow! I'm excited to play with this one. How close is this to the version you used in http://arxiv.org/pdf/1511.04143v4.pdf?

mhauskn:
The SoccerEmptyGoal environment should be identical to the one used in the paper. However, I've condensed the action space a bit - the gym environment uses two continuous values to represent the power/direction of any discrete action, while in the paper we used a separate continuous value for each parameter of each different discrete action.

tlbtlbtlb:
I think I prefer the action space from the paper, with 6 separate continuous parameters. At least, I'd want to train a policy network to generate the continuous parameters separately. A policy network that generates high probabilities for 2 different discrete actions in some state will have trouble generating sensible values for the multiplexed continuous parameters.

Also, we really like to be able to reproduce published results as closely as possible.

mhauskn:
Thanks for the feedback! I've expanded the action space to mirror what was used in the paper. To the best of my knowledge, the environment should be identical to the paper.

lusob:
Hi, anyone knows about the "tox" error on travis? It seems it is avoiding to enter any pull-request?

gdb:
@jonasschneider is going to fix that error!

Otherwise, this should be ready to merge?

machinaut:
I notice this branch has fallen a bit behind.  Is there a new strategy for incorporating the soccer environment?

If not, I'm happy to do the legwork to get this rebased on master.

catherio:
Hi Alex - thanks for offering to help out; I just dealt with this yesterday actually =). We are moving in the direction of generally having user-contributed envs reside in their own package rather than in gym core; then, the only remaining piece for us to do on the OpenAI side when someone creates a custom environment is to add it to the scoreboard if the user requests it.

I created the https://github.com/openai/gym-soccer package yesterday using the code @mhauskn posted here, as an example of this style, and also added the https://gym.openai.com/envs#soccer section to the website.

catherio:
@mhauskn, if you could take a look at gym_soccer package I threw together, and at the documentation on the website, that would be great. Specifically I don't know exactly what to say about each of the individual environments - your contributions to the documentation would be appreciated.

ceobillionaire:
When trying:

`pip install gym_soccer`

I get:

> Collecting gym_soccer
>   Could not find a version that satisfies the requirement gym_soccer (from versions: )
> No matching distribution found for gym_soccer

Also, the _https://github.com/openai/gym-soccer_ gives a 404 error.

catherio:
Good catch, I accidentally made that git repo private - it should be public now. Also, if you try `pip install gym-soccer` with a hyphen, rather than an underscore, I would expect that to work.

PR #270, opened 2016-08-02
Added PLE environment

lusob:
modified doc with new PLE env

### DIFF
--- a/gym/envs/__init__.py
+++ b/gym/envs/__init__.py
@@ -151,6 +151,18 @@ register(
     reward_threshold=9.7, # optimum = 10.2
 )
 
+register(
+    id='GuessingGame-v0',
+    entry_point='gym.envs.toy_text.guessing_game:GuessingGame',
+    timestep_limit=200,
+)
+
+register(
+    id='HotterColder-v0',
+    entry_point='gym.envs.toy_text.hotter_colder:HotterColder',
+    timestep_limit=200,
+)
+
 # Mujoco
 # ----------------------------------------
 
--- a/gym/envs/toy_text/__init__.py
+++ b/gym/envs/toy_text/__init__.py
@@ -2,3 +2,5 @@ from gym.envs.toy_text.blackjack import BlackjackEnv
 from gym.envs.toy_text.roulette import RouletteEnv
 from gym.envs.toy_text.frozen_lake import FrozenLakeEnv
 from gym.envs.toy_text.nchain import NChainEnv
+from gym.envs.toy_text.hotter_colder import HotterColder
+from gym.envs.toy_text.guessing_game import GuessingGame
--- a/gym/envs/toy_text/guessing_game.py
+++ b/gym/envs/toy_text/guessing_game.py
@@ -1,0 +1,87 @@
+import gym
+from gym import spaces
+from gym.utils import seeding
+import numpy as np
+
+
+class GuessingGame(gym.Env):
+    """Number guessing game
+
+    The object of the game is to guess within 1% of the randomly chosen number
+    within 200 time steps
+
+    After each step the agent is provided with one of four possible observations
+    which indicate where the guess is in relation to the randomly chosen number
+
+    0 - No guess yet submitted (only after reset)
+    1 - Guess is lower than the target
+    2 - Guess is equal to the target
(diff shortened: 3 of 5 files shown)


gdb:
Thanks! Pygame is a pretty heavyweight dependency, and I'd prefer not to bring it into core. We're starting to reach the point where I'd like to start splitting out new environments into their own repositories, which people can install independently.

Would you be up for turning this into an independent package? You can register everything in your `__init__.py` method as you do here.

ppaquette:
@gdb What are your thoughts on:

### Contributed env
- Contributed env needs to be installable through `pip install gym-<name>`
- Skeleton structure is /gym-<name>/{envs,spaces,tests,wrappers}
- Would work as `import gym-<name> as gym; gym.make(...); ...`
- Gym would store a a markdown file in envs/contrib/ that briefly explains each env and how to install deps

- Do you want to keep any of the current contrib envs in the core [box2d, parameter_tuning, toy_text, doom,  safety,  and all the ones waiting in PR]?

Otherwise I'll convert them to the new structure so you can send their contributors an email on what they need to do next.

gdb:
> Contributed env needs to be installable through pip install gym-<name>
Yep.

> Skeleton structure is /gym-/{envs,spaces,tests,wrappers}
Less opinionated here. Reasonable default though, and what I've been doing in my own. I do a bunch of `register` calls in `/gym_foo/__init__.py`.

> Would work as import gym-<name> as gym; gym.make(...); ...
You should do `import gym; import gym_foo; gym.make(...)`.

> Gym would store a a markdown file in envs/contrib/ that briefly explains each env and how to install deps

Cool.

> Do you want to keep any of the current contrib envs in the core [box2d, parameter_tuning, toy_text, doom, safety, and all the ones waiting in PR]?
I think it's fine to leave these for now.

JKCooper2:
Would it be better to have official envs and custom envs that are attached to users? Something similar to how `docker pull` can be passed a straight image-name or user/image-name. 

If the env-name doesn't exist locally then it attempts to get it from the OpenAI server and stores it locally.

`gym.pull('user/env-name', version='v1')` where version is optional and if left off would pull the latest version

So you would call `gym.make('user/env-name:version')` again where if version was left off it would use the latest version found.

Envs are then uploaded by a user to OpenAI through a form filling out the required field (description, version, step limit, requirements, maybe problem type/difficulty) and linking to a Github file like [this](https://github.com/openai/gym/blob/master/gym/envs/classic_control/cartpole.py) (OpenAI can pull the [raw file](https://raw.githubusercontent.com/openai/gym/master/gym/envs/classic_control/cartpole.py) from that and allows other users to submit pull requests). 

This means users can share envs without them having to be accepted by pull request, separates official envs from user envs, means you don't need to install every env, easy tracking of current envs you have installed (`gym.envs()` could display list similar to `docker images`), and means you can more easily expand into having challenge sets of envs where a challenge set is one agent run over a list of environments (otherwise each environment would have to be manually installed before this is possible to do).

gdb:
Yeah, I've had similar thoughts. I think we may want to end up there eventually, but for the short term it's totally fine to do be low-tech here. Once we have an abundance of third-party packages, and manually tracking them isn't enough, we can add more sophisticated tooling.

gdb:
(So there's an important investment right now in building out the actual set of packages and making sure they're high-quality, and the process is well-documented.)

lusob:
Sorry I don't understand how to do it exactly.
Can I just create a /gym_ple folder with a __init__.py registering my env and a envs folder with the env implementation or 
Do I need to create a gym_ple package completly independient of gym (with envs, scoreboard, tests,  spaces...)?

Thanks

gdb:
Sorry for the lack of clarity.

Yes, create a `gym_ple` package. You don't need to include much in there, just:

- Any env specific code
- A basic `setup.py`
- In `gym_ple/__init__.py`, you can import the `register` method to register your environment.

You can then open a pull request to add your environments to the scoreboard on main gym. That should be the only code that enters main Gym right now.

lusob:
Thanks!
Currently there is not pip package for Pygame, could I describe the deps
installation on a Readme file instead of setup.py?

gdb:
Yep, feel free to do however's best — it's your package!

PR #271, opened 2016-08-03
Fixed broken build - nose2 module incorrectly imported from

ppaquette:
Bug is caused by a conflict between 2 gym environments

1) Version being tested, downloaded from git repo (installed in $HOME/openai/gym/)
2) Version installed by tox in virtualenv (with pip install -e .[all])

I simply removed the 2nd version after tox installed gym dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,12 +24,9 @@ addons:
       - libblas-dev
       - liblapack-dev
       - gfortran
-matrix:
-    include:
-        - python: 2.7
-          env: DISPLAY=:12 TOX_ENV=py27
-        - python: 3.4
-          env: DISPLAY=:12 TOX_ENV=py34
+python:
+  - "2.7"
+  - "3.4"
 before_install:
   - Xvfb :12 -screen 0 800x600x24 +extension RANDR &
     # In a pull request, there are no secrets, and hence no MuJoCo:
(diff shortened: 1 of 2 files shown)


jonasschneider:
Hi, Thanks a lot for taking a stab at this, and finding the bug!

Our reason for using `tox` initially was to facilitate running the tests locally under Py2/3, which it does a good job at (you can just run `tox`).

However, the wiring that sets up tox on Travis is actually pretty annoying (that also includes the `tox-travis` package, which is kind of a hack in itself). Your change breaks running the the tests locally via `tox`. 

I think it would be best to just switch away from that setup and instead:
(1) let people "unofficially" run tests locally with tox, with as few config hacks as possible, and
(2) use Travis' build matrix for CI, cutting tox out of there entirely.

The divergence is a bit annoying, but might actually uncover bugs in the Gym code. What do you think?

ppaquette:
I agree with you that tox is pretty annoying, and should be removed from Travis builds.

I've resubmitted a commit that removed it.

I'm also installing Miniconda to avoid building NumPy, and SciPy, which can take 10+ minutes.

jonasschneider:
Great, thanks! I'm running another test now -- in theory, Travis should cache the built packages, so the installation should be much faster. If it works, we can skip installing miniconda. (I'll make that change if so.)

ppaquette:
I'm still getting Timeouts (20 mins) on `travis_wait pip install -r requirements_dev.txt`. This seems to mostly happen on new branches when the cache is not available. (e.g. https://travis-ci.org/ppaquette/gym/jobs/151040342)

Also I'm getting a lot of `exceptions.ImportError: cannot import name gl_info` in the last couple of days. (e.g. https://travis-ci.org/ppaquette/gym/jobs/151040341, https://travis-ci.org/openai/gym/builds/151043722). Not sure why, most likely a broken dependency somewhere.

Issue #267, opened 2016-07-31
Hex9v9-v0 does not behave as it should when there is a full game board.


Icohedron:
Logs:
```
      1  |  2  |  3  |  4  |  5  |  6  |  7  |  8  |  9  | 
     -----------------------------------------------------
  1  |  B  |  B  |  W  |  B  |  B  |  B  |  W  |  W  |  B  |
 --------------------------------------------------------------
     2  |  B  |  W  |  W  |  B  |  W  |  B  |  B  |  W  |  W  |
    --------------------------------------------------------------
        3  |  W  |  W  |  B  |  B  |  B  |  B  |  W  |  W  |  W  |
       --------------------------------------------------------------
           4  |  B  |  B  |  B  |  B  |  W  |  B  |  W  |  W  |  W  |
          --------------------------------------------------------------
              5  |  B  |  W  |  B  |  B  |  B  |  W  |  W  |  W  |  W  |
             --------------------------------------------------------------
                 6  |  W  |  B  |  W  |  W  |  W  |  B  |  W  |  B  |  B  |
                --------------------------------------------------------------
                    7  |  B  |  W  |  W  |  W  |  B  |  W  |  W  |  B  |  W  |
                   --------------------------------------------------------------
                       8  |  O  |  W  |  B  |  B  |  B  |  B  |  W  |  B  |  B  |
                      --------------------------------------------------------------
                          9  |  B  |  W  |  W  |  B  |  B  |  W  |  B  |  W  |  W  |
                         --------------------------------------------------------------
Traceback (most recent call last):
  File "sandbox.py", line 11, in <module>
    observation, reward, done, info = env.step(action)
  File "/home/icohedron/anaconda3/lib/python3.5/site-packages/gym/core.py", line 121, in step
    observation, reward, done, info = self._step(action)
  File "/home/icohedron/anaconda3/lib/python3.5/site-packages/gym/envs/board_game/hex.py", line 117, in _step
    a = self.opponent_policy(self.state)
  File "/home/icohedron/anaconda3/lib/python3.5/site-packages/gym/envs/board_game/hex.py", line 16, in random_policy
    a = np_random.randint(len(possible_moves))
  File "mtrand.pyx", line 1264, in mtrand.RandomState.randint (numpy/random/mtrand/mtrand.c:14508)
ValueError: low >= high
```

Code used:
```
import numpy as np
import gym

env = gym.make('Hex9x9-v0')

for e in range(100):
    observation = env.reset()
    while True:
[...]

PR #274, opened 2016-08-04
Fixed Issue #267 - Hex9v9-v0 does not behave as it should ...

ppaquette:


gdb:
Should we bump the version here?

ppaquette:
I don't think we need to, because it only fixes an error being raised, and does not affect the reward values.

Issue #257, opened 2016-07-20
Doom envs create from threads leave behind running vizdoom processes


danijar:
After some experiments that open multiple Doom environments from different threads, my X server reports this and I cannot open any more windows.

```
Maximum number of clients reached
```

There are several running `vizdoom` instances left after the agent script terminates and I have to run `killall vizdoom -9` to get my machine working again.

![vizdoom](https://cloud.githubusercontent.com/assets/2111293/17001914/51bb25d2-4ec9-11e6-9b78-86deb848585b.png)

ppaquette:
I get the following error while trying to run Doom envs in multiple threads. 

```python
[xcb] Unknown request in queue while dequeuing
[xcb] Most likely this is a multi-threaded client and XInitThreads has not been called
[xcb] Aborting, sorry about that.
python2.7: ../../src/xcb_io.c:179: dequeue_pending_request: Assertion `!xcb_xlib_unknown_req_in_deq' failed.
Aborted (core dumped)
```

This probably means that the envs are not thread-safe, and that's why the vizdoom processes are not killed properly.

If I create a lock on the main thread, and request a lock before any operations, the multiple threads work perfectly, and the vizdoom processes are killed properly (but they run a bit slow)

```python
import gym
from threading import Thread, Lock

def _run_env():
    is_done = False
    actions = [0] * 43
    actions[0] = 1
    with main_lock:
        env = gym.make('DoomBasic-v0')
        env.reset()
    while not is_done:
        with main_lock:
            env.render()
            obs, reward, is_done, info = env.step(actions)
    with main_lock:
        env.close()

# Starting threads
main_lock = Lock()
for _ in range(5):
    env_thread = Thread(target=_run_env)
    env_thread.start()
```

How do you think this bug should be fixed?

ppaquette:
Update:

The issue is caused in the render() function, which is not thread-safe.

The following code works way faster, and cleans up the vizdoom processes properly.

```python
import gym
from threading import Thread, Lock

def _run_env():
    is_done = False
    actions = [0] * 43
    actions[0] = 1
    env = gym.make('DoomBasic-v0')
    env.reset()
    count = 0
    while not is_done:
        count += 1
        if count % 10 == 0:
            with main_lock:
                env.render()
        obs, reward, is_done, info = env.step(actions)
    env.close()

# Starting threads
main_lock = Lock()
for _ in range(5):
    env_thread = Thread(target=_run_env)
    env_thread.start()
```

[...]

Issue #248, opened 2016-07-15
Too large action spaces for most Doom environments


danijar:
In some Doom environments, the action space is restricted using `available_buttons` in the Doom configuration file. However, `env.action_space` is the full actions space for all Doom environments. There is also a member called `allowed_actions` that removes actions not in the list before forwarding the remaining ones to the game.

I think we should dynamically create the action space based on the list of `allowed_actions` so that one bot can be efficiently used on all environments without modifications. I can help with a pull request if wanted, it would be a nice project to get started contributing to gym.

ppaquette:
You should look at the comments on pull request #157 . 

The original PR had a "small" action space (i.e. only the commands used in the env), but it was decided to use the full action space on all environments, because that would only slow down the algos by a constant factor, and would permit transfer learning.

In all cases, if it is an issue, you can probably train an algo on a custom small action space, and map it back to the full action space on each step. 

As for your suggestion, the actions in allowed_actions are different in every doom env, which makes learning across envs hard to do (e.g. the second item in the array might be MOVE_RIGHT in one env, and TURN_RIGHT in another).

jietang:
@danijar any updates?

danijar:
I see the problem and it makes sense to have a common action space for all Doom environments. It's just that with current state-of-the-art, it's hard to learn anything in Doom because it's a comparably complex environment and the large action space doesn't make it easier. The authors of VizDoom also reduce the action space in their experiments.

For my own experiments, I wrapper the Doom envs with my own environment that reduces the action space and unfolds the actions from the agent before forwarding them to the Doom environment. Changing the action space is really part of the environment, not the agent. Otherwise, the agent would not be general purpose anymore. Maybe we want a `env.configure()` option for whether to use the available or full action space?

ppwwyyxx:
The best part of the doom environment is that it is highly customizable. In my research I customize reward & buttons & render options & wad files a lot, and it would be very good to see some of these options available in gym.

danijar:
@ppwwyyxx I agree, it would be nice to have an abstract doom environment class (There is already something similar). Of course, that shouldn't have a scoreboard on the webservice because people might configure it differently. The existing Doom tasks could make use of this class then.

ppaquette:
Submitted PR #282 with action_filter parameter on configure(). You can set action_filter to be one of:

- 'doom-minimal' (Discrete action space with only the level's allowed_actions and noop)
[...]

PR #282, opened 2016-08-09
Fix Issue #257 and Issue #248 - Added params to configure() method for Doom

ppaquette:
Issue #257 - Doom envs create from threads leave behind running vizdoom
- Added lock param, required to run VizDoom across processes
- Added __enter__ and __exit__ on env.core, so envs can be created with a 'with' block, and without needing to call .close() explicitly.

e.g.
```python
with gym.make('DoomBasic-v0') as env:
    env.reset()
    done = False
    while not done:
        obs, rew, done, info = env.step(env.action_space.sample())
```

Issue #248 - Too large action spaces for most Doom environments
- Added screen_resolution and action_filter on configure() method to convert HighLow action space to Discrete

Submitting both fix together, since they both relate to modifying the configure() method

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -48,12 +48,13 @@ class Env(object):
 
     def __new__(cls, *args, **kwargs):
         # We use __new__ since we want the env author to be able to
-        # override __init__ without remebering to call super.
+        # override __init__ without remembering to call super.
         env = super(Env, cls).__new__(cls)
         env._env_closer_id = env_closer.register(env)
         env._closed = False
         env._configured = False
         env._unwrapped = None
+        env._step_count = 0
 
         # Will be automatically set when creating an environment via 'make'
         env.spec = None
@@ -115,7 +116,7 @@ class Env(object):
         """
         self.monitor._before_step(action)
         observation, reward, done, info = self._step(action)
-
+        self._step_count += 1
         done = self.monitor._after_step(observation, reward, done, info)
         return observation, reward, done, info
 
@@ -131,6 +132,7 @@ class Env(object):
 
         self.monitor._before_reset()
         observation = self._reset()
+        self._step_count = 0
         self.monitor._after_reset(observation)
         return observation
 
@@ -276,6 +278,14 @@ class Env(object):
         else:
             return self
 
+    @property
+    def step_count(self):
+        return self._step_count
+
+    @property
+    def wrappers(self):
+        return []
+
     def __del__(self):
         self.close()
 
@@ -314,14 +324,23 @@ class Space(object):
         return sample_n
 
 class Wrapper(Env):
+
+    def __new__(cls, env, *args, **kwargs):
+        # We use __new__ since we want the wrapper author to be able to
+        # override __init__ without remembering to call super.
+        wrapper = super(Wrapper, cls).__new__(cls)
+        wrapper.env = env
+        wrapper.metadata = env.metadata
+        wrapper.action_space = env.action_space
+        wrapper.observation_space = env.observation_space
+        wrapper.reward_range = env.reward_range
+        wrapper.spec = env.spec
+        wrapper._unwrapped = env.unwrapped
+        return wrapper
+
+    # Overridable
     def __init__(self, env):
-        self.env = env
-        self.metadata = env.metadata
-        self.action_space = env.action_space
-        self.observation_space = env.observation_space
-        self.reward_range = env.reward_range
-        self.spec = env.spec
-        self._unwrapped = env.unwrapped
+        pass
 
     def _step(self, action):
         return self.env.step(action)
@@ -340,6 +359,26 @@ class Wrapper(Env):
 
     def _seed(self, seed=None):
         return self.env.seed(seed)
+
+    @property
+    def step_count(self):
+        return self.env.step_count
+
+    @property
+    def wrappers(self):
+        target = self
+        wrappers_list = []
+        while target._unwrapped is not None:
+            wrappers_list.append(target.name)
+            target = target.env
+        return wrappers_list
+
+    @property
+    def name(self):
+        return '{}.{}'.format(
+            '.'.join(self.__module__.split('.')[:-1]).replace('gym.wrappers.', ''),
+            self.__class__.__name__
+        )
 
     def __str__(self):
         return '<{}{} instance>'.format(type(self).__name__, self.env)
--- a/gym/envs/doom/doom_basic.py
+++ b/gym/envs/doom/doom_basic.py
@@ -26,12 +26,6 @@ class DoomBasicEnv(doom_env.DoomEnv):
     Goal: 10 points
         Kill the monster in 3 secs with 1 shot
 
-    Mode:
-        - env.mode can be 'fast', 'normal' or 'human' (e.g. env.mode = 'fast')
-        - 'fast' (default) will run as fast as possible (~75 fps) (best for simulation)
-        - 'normal' will run at roughly 35 fps (easier for human to watch)
-        - 'human' will let you play the game (keyboard only: Arrow Keys, '<', '>' and Ctrl)
-
     Ends when:
         - Monster is dead
         - Player is dead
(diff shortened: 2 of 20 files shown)


gdb:
- I'd like to avoid adding too many extra methods to Env currently, and I don't think that the context manager syntax will be used commonly enough (e.g. doesn't work if you want many envs).
- `configure` should not be used for anything that changes the semantics of the environment — for that, you should add options to the constructor.

So let's remove the `__enter__`/`__exit__` for now, and move most of the configuration options into `__init__`

danijar:
I agree that having this as constructor arguments would be a good idea. This would align well with https://github.com/openai/gym/issues/283. The arguments could default to what is used on the web service. And researchers could still experiment with other versions just by registering a new environment:

```python
register(
    id='SimpleDoomDeathmatch-v0',
    entry_point='gym.envs.doom:DoomEnv',
    timestep_limit=10000,
    kwargs=dict(
        config='deathmatch.cfg',
        # ...
        action_space='minimal',
    ),
)
```

ppaquette:
These are good points.

I do have a couple of comments.

### 1) Context Manager

- Most of the RL implementations I have seen use some sort of worker function that is called by a thread or multiprocessing pool

1. Jie's TRPO parallel version uses a worker function (https://github.com/jietang/modular_rl/blob/parallel_trpo/modular_rl/core.py#L120)
2. A3C LSTM implementation also uses a worker function (https://github.com/muupan/async-rl/blob/master/run_a3c.py#L166)

- Usually the worker function receives either a `make_env` parameter that can create an env, or a local instance of the env that is created for each process
- More complex environments usually have to deal with some inter-process communication, through pipes, shared memory, sockets, and/or queues, and have to explicitly call .close(), otherwise the IPC mechanism will stay open
- It's very easy to forget to explicitly call close(), and this can make the entire computation unstable or even crash

1. Look at issue 257 - The computation crashed because close() was not called explicity
2. Look at Jie's code, his worker function is `parallel_rollout_worker`, which calls `rollout`. The env is reset, but close() is never called anywhere. (https://github.com/jietang/modular_rl/blob/parallel_trpo/modular_rl/core.py#L154)
3. Look at the A3C implementation, same issue, env is created, but never closed. (https://github.com/muupan/async-rl/blob/master/run_a3c.py#L44)

- The destructor (__del__) does not seem to be working effectively in those scenarios. I can easily leak processes if I don't call .close() explicitly
- For example, this gist leaks 30 vizdoom processes, and close() is never called by the destructor. (https://gist.github.com/ppaquette/ef65b7ba019c42e61dd447e6f7626f06)

- The easiest solution is to wrap the env with a 'with' block, and advertise that as the proper way to create an env.
- I can't really come up with a scenario where you wouldn't create envs in a loop or worker function, or where you would use different type of envs, which would make the 'with' syntax difficult to use

### 2) Configure Method

- I agree with you that we need to make the envs as comparable as possible, and we should avoid adding options to change the semantics (obs_space and/or act_space)
- The difficulty with the HighLow action space is that is reproduces a full mouse and keyboard, and has 8.8 e^20 possibilities (2^38 * 20^2 * 200^3).
- A lot of algorithms are not able to handle that many possibilities, so they handicap themselves and make the action space smaller (by wrapping it in a filtered env)

1. That's why Jie added to option to reduce the screen_resolution (https://github.com/openai/gym/commit/bdda59f7a3c85f793f7c6dfa342d8546f45b9b6e)
2. That's what Jie did with TRPO (https://github.com/jietang/modular_rl/blob/parallel_trpo/modular_rl/filtered_env.py#L55)
3. That's also what Dominique did in his writeup on the DoomCorridor level (https://gist.github.com/domluna/a11d5467e8dba3e16f42b591f2b930b3#file-run_cnn-py-L30)

- Rather than having everyone create their own filters, I think it makes more sense to add common filters that everyone can use.
- Currently, the only way to add kwargs to __init__ is to register a new env.
- It's doesn't really make sense to ask people to register their own env, or to add envs like "DoomBasic-minimal", "DoomBasic-constant-8", etc.
- I think environments with these modified settings should not be able to submit their score on the scoreboard (otherwise, it becomes harder to compare)

### 3) Proposed Solution

- I propose to allow kwargs in the make() function (e.g. env = gym.make('DoomBasic-v0', kwargs))
- These kwargs would override the default args registered for the env, and would be passed to the init function
- Envs created this way would be called "customized env", and would have a flag to prevent the monitor from recording them
- This would allow for the creation of fully customizable environments, for researchers looking to use them in their research (e.g. Issue 283)

The full syntax would then be:

```python
with gym.make('env-name', optional-kw-args) as env:
    env.configure()
    done = False
    obs = env.reset()
    while not done:
        obs, rew, done, info = env.step(action)
        env.render()
```

danijar:
Thanks for the many insights in your post. I like the proposed solution of adding `kwargs` to `gym.make()`. However, please keep the monitor from monitoring. Otherwise, everyone has to implement his or her own monitor to use with custom environments. This would also prevent using the online chart. Instead, I would suggest to put the custom `kwargs` in the JSON files and a non-default value prevents the result from being uploaded.

gdb:
Thanks for the detailed writeup. I'm pretty time constrained right now (we have some *really* exciting new Gym stuff coming that I'm spending my time on), so please don't take my shorter reply as a lack of excitement about your work here.

On 1: we recently fixed a bug where environments were being leaked (https://github.com/openai/gym/commit/36b2a58b3f112c829f667029d77d84ab71325b89), and the destructor would not be called until the program exited. So it's now safe to just let your env instance be GC'd. In the interest of simplicity, I really prefer a world where people don't need to even think about closing their environments.

One reasonable pattern is to create a pool of env instances, and then pass each of those into a rollout function as needed (likely in parallel). I don't think the `with` syntax makes sense there.

On 2: I do like people adding filters in a standard fashion. What do you think of instead standardizing on using Wrappers as I've started doing here: https://github.com/openai/gym/commit/21f8312059a795b2bd076761e2344a82a549135a? That way, it's very easy to take the result of a `make` call and Wrap it up; it's also possible to officially register Wrappers for `make` time.

ppaquette:
On 1. I'm fine with it. I'll remove the __enter__, __exit__.

On 2. The underlying goal is 1) standardized filters / modifications for easier reproducibility of results, 2) be able to customize env to fit research. (e.g. Issue 248 says they want a filter for a smaller Discrete action space, Issue 283 says they want a full abstract env to customize as they want).

I can't really register 200 variations of the same env, and put them on the scoreboard.  That's why I proposed the syntax ```gym.make('env-name', optional-kwargs)``` to create a customized env so they can run their research, but can't upload results.

Wrappers are a good way to create standardized variations of an env, but doesn't solve the issue that the numbers of env on the scoreboard is going to be multiplied 10x.

gdb:
Well, rather than tell people to run `gym.make('env-name', optional-kwargs) `, you'd tell them to instead run `Wrapper2(Wrapper1(gym.make('env-name')))`.

I'd be up for adding a setting (perhaps on by default) where wrappers mark their environment as ineligible for upload. I'm not *as* concerned about this aspect right now though — we're going to be doing a major revamp of how the leaderboards work, and I think we'll find this question a bit less important than you'd expect.

ppaquette:
Ok. That's sound good. 

Do you want me to finish your wrapper implementation and re-submit the PR using it, or should I wait until you complete it?

gdb:
Please finish. I'm actually decently happy with my implementation and have been using it for a bit for my use-cases now, so getting additional eyes on it would be helpful.

gdb:
Thanks!

danijar:
I just want to point out a potential problem with wrapping envs that I encountered in my own implementation. Some preprocesses, such as the common frame skipping, expose a different time scale. By wrapping the envs, we can only see the outer timestep, but not the one of the underlying environment anymore. For evaluating different preprocesses, we might want to compare results based on env time.

ppaquette:
Made a couple of changes

```
>>> import gym, gym.wrappers
>>> env = gym.make('DoomBasic-v0', ['doom.Res160x120', 'doom.DiscreteMinimal', 'doom.Skip1'])
[2016-08-12 11:17:46,961] Making new env: DoomBasic-v0

>>> env.monitor.start('outdir')
[2016-08-12 11:17:54,161] Creating monitor directory outdir

>>> env.action_space; env.observation_space; env.wrappers
Discrete(4)
Box(120, 160, 3)
['doom.Res640x480', 'doom.Res160x120', 'doom.DiscreteMinimal', 'doom.Skip1']

>>> obs = env.reset()
[2016-08-12 11:18:25,767] Starting new video recorder writing to /www/ppaquette/gym/outdir/openaigym.video.0.27407.video000000.mp4

>>> done = False
>>> while not done:
...   obs, rew, done, info = env.step(env.action_space.sample())
... 
>>> env.monitor.close()
[2016-08-12 11:19:04,012] Finished writing results. You can upload them to the scoreboard via gym.upload('/www/ppaquette/gym/outdir')

>>> env.step_count; env._unwrapped.step_count
22
43

>>> env.monitor._env_info()
{'wrappers': ['doom.Res640x480', 'doom.Res160x120', 'doom.DiscreteMinimal', 'doom.Skip1'], 'uploadable': True, 'gym_version': '0.1.5', 'env_id': 'DoomBasic-v0'}

>>> env.uploadable
True

>>> humanWrapper = gym.wrappers.doom.HumanPlayer
>>> env = humanWrapper(env)
>>> env.wrappers
['doom.Res640x480', 'doom.Res160x120', 'doom.DiscreteMinimal', 'doom.Skip1', 'doom.HumanPlayer']

>>> env.uploadable
False

>>> skipWrapper = gym.wrappers.doom.Skip1
>>> env = gym.make('DoomBasic-v0', [humanWrapper, skipWrapper])
[2016-08-12 12:11:23,926] Making new env: DoomBasic-v0
>>> env.wrappers
['doom.Res640x480', 'doom.HumanPlayer', 'doom.Skip1']
```

gdb:
Thanks for your work here. However, there are a lot of new features here that I think we can get by without. My top priority is making sure that Gym's interfaces and implementations are very simple, intuitive, and non-magical.

Some concrete thoughts:

- Why is is it useful to allow people to specify wrappers as strings? (It seems better in your case to just have a single wrapper class, which supports a `__call__` method so people can instantiate the desired filter.)
- Why add the wrappers argument to `make`? (I *really* want to last as long as possible without adding new args there. Currently the semantic is very clear: if you `make` an environment it's been created in a standardized fashion. If `make` grows arguments, then things become muddled.)
- Why do we need `step_count` to be tracked internally? (Seems like the user can track that easily on their own.)
- I don't think we need to do the uploadable flag yet. I expect we'll need something like this, but I'd rather wait for it to become an issue.
- You can use `_unwrapped` rather than `hasattr(..., 'env')`

In general, I'd encourage you to think about the way to get by that touches core the least.

ppaquette:
Good points.
1 and 2. I understand your point with ```gym.make```. My reasoning was that you could just copy/paste an array of string from the scoreboard to replicate an environment, without having to manually create instances of every wrapper you need. 

Would ```gym.make('env-name')(['wrapper-1', 'wrapper-2'])``` and/or ```gym.make('env-name')('wrapper1')('wrapper2')``` work instead?

3)  Counting steps might give skewed results if a wrapper skip steps, you might only get the outer wrapper step count, but not the unwrapped's. That was danijar's point on having the ability to properly get the number of steps for each layer to be able to compare experiments.
4)  I see a couple of use cases. e.g. Issue 283 where user wants to load a custom map, or Issue 284 where user is asking for ability to customize the reward function.
5) I'll fix.

danijar:
I think most intuitive would be:

```python
env = gym.make('env-name')
env = WrapperB(WrapperA(env))
```

Regarding the step count, is there any reason the user might want to know the wrapped step count? I think it should always be the one of the original environment. (Also, as @gdb said, the user can count the wrapper step count manually.)

An `uploadable` flag seems necessary to me as well, but this should be another PR.

ppaquette:
Closing and reopening as a PR for wrapper, and one for Doom

PR #286, opened 2016-08-11
Docker - Minor fix for pull requests

ppaquette:
- MUJOCO_KEY_BUNDLE is an empty string in pull requests, rather than being unset.
- Split py27 and py34 to avoid hitting the 50 mins timeout, and for easier debugging.
- COPY takes a trailing slash as per documentation
```
If <dest> does not end with a trailing slash, it will be considered a regular file and the contents of <src> will be written at <dest>.
```

### DIFF
--- a/bin/docker_entrypoint
+++ b/bin/docker_entrypoint
@@ -6,7 +6,7 @@ set -e
 
 path=$(cd $(dirname "$0") && pwd)
 
-[ -z ${MUJOCO_KEY_BUNDLE+x} ] || ( mkdir -p ~/.mujoco && curl https://openai-public.s3-us-west-2.amazonaws.com/mujoco/$MUJOCO_KEY_BUNDLE.tar.gz | tar xz -C ~/.mujoco )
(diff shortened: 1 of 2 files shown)


gdb:
Thanks!

Really need to figure out how to make these builds faster... I was hoping pulling a built image would do it, but doesn't seem to. Maybe next step is to try some explicit export-to-a-cachedir scheme?

gdb:
Ah, good find! Yep, sounds like a plan.

gdb:
Looks like your 3.4 failed. Will merge after fixup!

gdb:
Looks like there's a deeper issue with performance (build caching like this stopped being supported in 1.10): https://github.com/docker/docker/issues/20316#issuecomment-221289631.

I do have an easy trick still... namely, having a much more built-up "base" image. Will see if I can get that running easily.

ppaquette:
3.4 is caused by Segmentation fault. Happens ~ 30% of the time. Not sure what env is causing it.

PR #288, opened 2016-08-13
Wrappers - Added 'wrappers' and 'step_count' property + monitor support

ppaquette:


gym/core.py
```diff
@@ -48,12 +48,13 @@ class Env(object):
...
     def __new__(cls, *args, **kwargs):
         # We use __new__ since we want the env author to be able to
-        # override __init__ without remebering to call super.
+        # override __init__ without remembering to call super.
         env = super(Env, cls).__new__(cls)
         env._env_closer_id = env_closer.register(env)
         env._closed = False
         env._configured = False
         env._unwrapped = None
+        env._step_count = 0
```
gdb: As discussed, don't think we need `step_count`.

gym/scoreboard/api.py
```diff
@@ -54,6 +54,7 @@ def upload(training_dir, algorithm_id=None, writeup=None, api_key=None, ignore_o
         training_episode_batch=training_episode_batch_id,
         training_video=training_video_id,
         env=env_info['env_id'],
+        wrappers=env_info['wrappers'],
```
gdb: We may add something like this in the future, but not yet. For now, I'm ok with just not worrying about the fact that people may be using wrappers. It'll be clear from people's posted code. If it becomes a problem we'll deal with it.

gym/core.py
```diff
@@ -314,14 +324,23 @@ def from_jsonable(self, sample_n):
         return sample_n
 
 class Wrapper(Env):
+
+    def __new__(cls, env, *args, **kwargs):
```
gdb: Good idea.

gym/core.py
```diff
@@ -341,5 +354,21 @@ def _configure(self, *args, **kwargs):
     def _seed(self, seed=None):
         return self.env.seed(seed)
 
+    @property
+    def wrappers(self):
```
gdb: Let's remove `wrappers` and `name` (user can build that list themselves, and the names on their own aren't necessarily meaningful since people can instantiate with various arguments).

PR #289, opened 2016-08-13
Docker - cleans cached code before uploading our code

ppaquette:
If there is already a cached gym on the base image, it needs to be removed before copying the code, otherwise deleted files will never get removed from the cached version.

### DIFF
--- a/test.dockerfile
+++ b/test.dockerfile
@@ -35,8 +35,8 @@ RUN pip install tox
 # Install the relevant dependencies. Keep printing so Travis knows we're alive.
 RUN ["bash", "-c", "( while true; do echo '.'; sleep 60; done ) & tox --notest"]
 
-# Finally, cleans cached code and upload our actual code!
-RUN rm -rf ./gym*
+# Finally, clean cached code and upload our actual code!
+RUN mv .tox /tmp/.tox && rm -rf ./* && mv /tmp/.tox .tox
 COPY . /usr/local/gym
 
 ENTRYPOINT ["/usr/local/gym/bin/docker_entrypoint"]
--- a/test.dockerfile
+++ b/test.dockerfile
@@ -36,7 +36,7 @@ RUN pip install tox
 RUN ["bash", "-c", "( while true; do echo '.'; sleep 60; done ) & tox --notest"]
 
 # Finally, clean cached code and upload our actual code!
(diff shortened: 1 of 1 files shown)


test.dockerfile
```diff
@@ -35,7 +35,8 @@ RUN pip install tox
 # Install the relevant dependencies. Keep printing so Travis knows we're alive.
 RUN ["bash", "-c", "( while true; do echo '.'; sleep 60; done ) & tox --notest"]
 
-# Finally, upload our actual code!
+# Finally, cleans cached code and upload our actual code!
+RUN rm -rf ./gym*
```
gdb: I think better to clear the whole `/usr/local/gym` directory

ppaquette: .tox virtualenv cache is there, it would force tox to reinstall everything

gdb: I see. Can we specifically just save that?

ppaquette: I could do
`RUN mv ./.tox $HOME/.tox && run -rf ./ && mv $HOME/.tox ./.tox`

gdb: Sure, that'll work.

Maybe nicer to move to `/tmp`, but minor detail. (I also prefer to use `.tox` / `rm -rf .`, but minor.)

ppaquette: `rm -rf .` doesn't work, because you can't remove the current directory, and we would need to recreate the directory to move .tox back into it. Replaced ./.tox with .tox, and $HOME with /tmp

gdb: Yeah, I thought that might happen. However, I think `rm -rf ./` won't catch dotfiles. You probably want to `cd` out of the currently directory.

gdb:
Hm, looks like this introduced an error? https://travis-ci.org/openai/gym/builds/152097176#L572

ppaquette:
Strange.  Build was passing with `RUN mv .tox /tmp/.tox && rm -rf ./* && mv /tmp/.tox .tox`.

Do you want to revert to that, even if it doesn't remove dot files?

gdb:
Let's try it for now. I'd like to figure it out properly at some stage though..

ppaquette:
Stack Overflow says to use `rm -rf .[^.] .??*` to also remove dot files. So I'll resubmit with that, and no cd.

http://serverfault.com/questions/47933/how-to-delete-all-hidden-files-and-directories-using-bash)

PR #291, opened 2016-08-14
Tweak Wrapper implementation

gdb:
cc/ @ppaquette

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -240,7 +240,7 @@ class Env(object):
             else:
                 raise
 
-    def build(self):
+    def build(self, extra_wrappers=None):
         """[EXPERIMENTAL: may be removed in a later version of Gym] Builds an
         environment by applying any provided wrappers, with the
         outmost wrapper supplied first. This method is automatically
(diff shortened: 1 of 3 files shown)


ppaquette:
The flip side of not having __ new __ is that you need to call super().__ init __ explicity or copy all the properties in env.

The code below doesn't work, because self.env, self.action_space are not defined.

I think it's easier to remember to use __ init __ (self, env) as the method signature, then to remember to call super().__ init() __ in the method. What do you think?

e.g.
```
class Skip(gym.Wrapper):
    """
        Generic common frame skipping wrapper
        Will perform action for `x` additional steps
    """
    def __init__(self, env, repeat_for_x_steps):
        self.repeat_count = repeat_for_x_steps
        self.record_stepcount = hasattr(env._unwrapped, 'stepcount')
        if self.record_stepcount:
            self.env._unwrapped.stepcount = 0

    def _step(self, action):
        done = False
        total_reward = 0
        current_step = 0
        while current_step < (self.repeat_count + 1) and not done:
            if self.record_stepcount:
                self.env._unwrapped.stepcount += 1
            obs, reward, done, info = self.env.step(action)
            total_reward += reward
            current_step += 1
        return obs, total_reward, done, info

    def _reset(self):
        if self.record_stepcount:
            self.env._unwrapped.stepcount = 0
        return self.env.reset()

    @property
    def stepcount(self):
        return self.env._unwrapped.stepcount if hasattr(self.env._unwrapped, 'stepcount') else 0
```

gdb:
So the main reason we decided to do `__new__` for envs is that your env would mostly work just fine without calling `super`. This might be ok for the short run, but it'd mean that as we added more functionality to the `Env` initialization, many envs would likely not pick it up.

With `Wrapper`, as you say, your wrapper just won't work if you forget to call super. I don't see it being very burdensome to call it; the case to avoid is people silently forgetting to call it.

PR #295, opened 2016-08-17
User envs

ppaquette:
@gdb Support for user envs. 

Test repo is: ppaquette/gym_test with envs (Acrobot and CartPole) ([Config File](https://github.com/ppaquette/gym_test/blob/master/.openai.yml))

I'm out of town until Sunday, with limited access to email. Let me know your thoughts and I'll fix issues when I come back.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,5 @@ tutorial/*.html
 vizdoom.ini
 
 # User envs
-gym/envs/.user_envs.json
-gym/envs/.user_envs.json.bak
+gym/envs/custom/*
 
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,8 @@ tutorial/*.html
 # PyCharm project files
 .idea
 vizdoom.ini
+
+# User envs
+gym/envs/.user_envs.json
+gym/envs/.user_envs.json.bak
+
--- a/gym/__init__.py
+++ b/gym/__init__.py
@@ -32,7 +32,8 @@ del logger_setup
 sanity_check_dependencies()
 
 from gym.core import Env, Space, Wrapper
-from gym.envs import make, spec
+from gym.envs import list, make, spec
 from gym.scoreboard.api import upload
+from gym.versioning import pull
 
-__all__ = ["Env", "Space", "Wrapper", "make", "spec", "upload"]
+__all__ = ["Env", "Space", "Wrapper", "list", "make", "pull", "spec", "upload"]
--- a/gym/envs/__init__.py
+++ b/gym/envs/__init__.py
@@ -1,4 +1,4 @@
-from gym.envs.registration import registry, register, make, spec
+from gym.envs.registration import registry, register, deregister, make, spec, list
 
 # Algorithmic
 # ----------------------------------------
--- a/gym/envs/registration.py
+++ b/gym/envs/registration.py
@@ -7,7 +7,7 @@ from gym import error
 
 logger = logging.getLogger(__name__)
 # This format is true today, but it's *not* an official spec.
-env_id_re = re.compile(r'^([\w:-]+)-v(\d+)$')
+env_id_re = re.compile(r'^([\w/:-]+)-v(\d+)$')
 
 def load(name):
     entry_point = pkg_resources.EntryPoint.parse('x={}'.format(name))
@@ -114,8 +114,19 @@ class EnvRegistry(object):
             raise error.Error('Cannot re-register id: {}'.format(id))
         self.env_specs[id] = EnvSpec(id, **kwargs)
 
+    def deregister(self, id):
+        if not id in self.env_specs:
+            logger.warn('Unable to deregister id: %s. Are you certain it is registered?', id)
+        else:
+            del self.env_specs[id]
+
+    def list(self):
+        return sorted([spec.id for spec in self.all()], key=lambda s: s.lower())
+
 # Have a global registry
 registry = EnvRegistry()
 register = registry.register
+deregister = registry.deregister
 make = registry.make
 spec = registry.spec
+list = registry.list
--- a/gym/scoreboard/__init__.py
+++ b/gym/scoreboard/__init__.py
@@ -8,7 +8,7 @@ http://rst.ninjs.org/
 
 import os
 
-from gym.scoreboard.client.resource import Algorithm, Evaluation, FileUpload
+from gym.scoreboard.client.resource import Algorithm, Evaluation, FileUpload, UserEnvConfig
 from gym.scoreboard.registration import registry, add_task, add_group
 
 # Discover API key from the environment. (You should never have to
@@ -16,6 +16,9 @@ from gym.scoreboard.registration import registry, add_task, add_group
 api_key = os.environ.get('OPENAI_GYM_API_KEY')
 api_base = os.environ.get('OPENAI_GYM_API_BASE', 'https://gym-api.openai.com')
 web_base = os.environ.get('OPENAI_GYM_WEB_BASE', 'https://gym.openai.com')
+github_raw_base = os.environ.get('OPENAI_GITHUB_RAW_BASE', 'https://raw.githubusercontent.com')
+
+base_without_api_key = [ github_raw_base ]
 
 # The following controls how various tasks appear on the
 # scoreboard. These registrations can differ from what's registered in
--- a/gym/scoreboard/__init__.py
+++ b/gym/scoreboard/__init__.py
@@ -8,7 +8,7 @@ http://rst.ninjs.org/
 
 import os
 
-from gym.scoreboard.client.resource import Algorithm, Evaluation, FileUpload, UserEnvConfig
+from gym.scoreboard.client.resource import Algorithm, CommitHash, Evaluation, FileUpload, UserEnvConfig
 from gym.scoreboard.registration import registry, add_task, add_group
 
 # Discover API key from the environment. (You should never have to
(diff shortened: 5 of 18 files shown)


JKCooper2:
How did you do that so quickly? I almost fell out of my seat when I realised what it was doing. I'll try adding some new envs to a repo tomorrow or Saturday and give it a test. 

From a quick look there may need to be something added to gym.upload() as a message if the user tries to upload and env that isn't registered with the OpenAI gym. Because they both use GitHub as the usernames it will make it much simpler for custom env leaderboards to be displayed on OpenAI

gdb:
To echo @JKCooper2, I'm very impressed that you got this working so quickly!

As I said in the other thread, I don't think now is the time to add this functionality to core. It's less of an implementation question, and more of a question of getting something correct, which people will really use, which is as simple as possible, and which will still be what we want for the future. I'm pretty ok with continuing to ask people to merge their scoreboard specification into the core for now.

Fortunately, I think this functionality can easily be split out as a third-party tool, which I'd recommend trying out. That'd be a good way to move quickly on it, and see where it goes.

JKCooper2:
Set up a [few tests envs](https://github.com/JKCooper2/gym-envs) and it worked very well
```
gym.pull('jkcooper2/gym-envs/EightPuzzle', version=0)
gym.pull('jkcooper2/gym-envs/HotterColder', version=1)
gym.pull('jkcooper2/gym-envs/GuessingGame', version=1)
```

The structure is very nice and clean, having a single pull function to load/refresh the env if needed and then just running `env = gym.make("jkcooper2/EightPuzzle-v0")`


A few things that may need some adjustment:

1.  If username contained capitals the env is saved in folder with lowercase, but attempted to be loaded from upper case causing it not to be found
```
[2016-08-19 10:07:53,425] Found and registered 1 user environments.
  File "/home/jesse/git/rlagents/examples/tests/environment.py", line 34, in <module>
    main()
  File "/home/jesse/git/rlagents/examples/tests/environment.py", line 10, in main
    gym.pull('JKCooper2/gym-envs/HotterColder', version=1)
  File "/home/jesse/git/gym/gym/versioning/manager.py", line 109, in pull
    open(os.path.join(gym_abs_path, 'envs', parsed_env['username'], '__init__.py'), 'w').close()
IOError: [Errno 2] No such file or directory: '/home/jesse/git/gym/gym/envs/JKCooper2/__init__.py'
```

2. It may be better for envs to be stored in `gym/envs/custom` folder (along with .json files) which is added to .gitignore so new pulls don't overwrite

3. Partial install of envs not possible where envs in the same folder as init file requires being called raises error of modules not included. Not sure if there's a simple way around this but may just require docs that explain groups of envs need to be placed in separate folders to allow individual install (which is what I ended up doing)

4. Potentially have commit hash able to be empty and use master branch if that's the case. Currently it require two commits in order to get the right commit hash (unless I'm doing something wrong), but I like it as an option for making versioning a lot easier

5. Warning when trying to pull an environment that exists but the version doesn't could display a more descriptive warning, as currently it makes it look like the env doesn't exist, rather than just you are using a version that doesn't exist

6. Attempting to upload a custom env currently gives the error:
```
  File "/home/jesse/git/rlagents/examples/tests/environment.py", line 34, in <module>
    main()
  File "/home/jesse/git/rlagents/examples/tests/environment.py", line 31, in main
    gym.upload('examples/tests/agents/', api_key='**API_KEY_REMOVED***')
  File "/home/jesse/git/gym/gym/scoreboard/api.py", line 62, in upload
    api_key=api_key,
  File "/home/jesse/git/gym/gym/scoreboard/client/resource.py", line 324, in create
    response, api_key = requestor.request('post', url, params, headers)
  File "/home/jesse/git/gym/gym/scoreboard/client/api_requestor.py", line 41, in request
    resp = self.interpret_response(rbody, rcode, rheaders)
  File "/home/jesse/git/gym/gym/scoreboard/client/api_requestor.py", line 158, in interpret_response
    self.handle_api_error(rbody, rcode, resp, rheaders)
  File "/home/jesse/git/gym/gym/scoreboard/client/api_requestor.py", line 69, in handle_api_error
    rheaders)
gym.error.APIError: Request req_sPDW72E3RPS5dHtvfg5evw: There was an error on the server, and it was unable to complete your request. If error persists, please let us know at gym@openai.com and include your request token (req_sPDW72E3RPS5dHtvfg5evw_497652).
```
It may be worth adding a check in gym.upload for whether the env.spec.id contains a '/' (means custom env) and returning a warning that custom envs can't be uploaded to the gym

gdb:
(BTW, the place I'd like to start for core is instead have a standard best practice for defining a 3rd party pip packages of environments, and maintain the list of those manually. There are plenty of environments sitting in pull requests right now which I don't think we should merge to core and would be good places to start.)

ppaquette:
Lots of good points.

- Required field is now `commit_ref`, which can be a branch name, tag name, short hash or long hash. This is to avoid the double commit / default branch issue (Point 3)
- `commit_hash` is automatically retrieved by querying the [GitHub API](https://developer.github.com/v3/repos/commits/#get-the-sha-1-of-a-commit-reference)
- Envs are now stored in gym/envs/custom/`username`/`commit_hash`/. This is to avoid re-downloading shared files for every env using them (Point 1). 
- json are now stored under gym/envs/custom/`username`/.envs.json. This makes deleting custom envs as easy as deleting the `username` folder.
- For partial install (Point 2), all required files should be specified (which might mean all files in the repo, depending on folding structure). [YAML anchor syntax](https://en.wikipedia.org/wiki/YAML#Repeated_nodes) is very useful to avoid having to copy a long list of files (&name_of_anchor / *name_of_anchor) [Example here](https://github.com/ppaquette/gym_test/blob/master/.openai.yml)
- Added a warning when pulling an obsolete version of an env
- Added a warning when multiple versions of an env use the same `commit_ref`.
- Added a more meaningful message when env exists, but version doesn't exist (Point 4)
- Added restriction on uploading custom envs to scoreboard (Uploading to scoreboard should be quite straightforward to do if the functionality is needed in the future) (Point 5)
- For Point 0, usernames containing capitals should now work properly.
- Added ' * ' install where you can install the latest version (or a specific version if specified) of all envs in the config file, without installing them one-by-one. Syntax `gym.pull('user/repo/*')`

Also
- I didn't put `master` as the default `commit_ref`, because it might lead to unexpected results when someone bumps the version number, and there is no `commit_ref` line to change.

Will create a third-party tool if there are no more issues.

gdb:
You might also want to consider how to handle environments with C dependencies.

My instinct is that you'll ultimately be happiest with something that builds directly on Pip packages, rather than creating your own dependency + versioning solutions. But I'll be curious to see what comes of this!

gdb:
Incidentally, to make sure I've been explicit on how I'm thinking about this feature:

In core, I would like to soon solve packaging of new environments. Doing `pip install gym-myenv` by hand once is pretty simple, and I have some trouble imagining people choosing not to use an environment because of the added friction of installing it. I could *imagine* a syntax like `gym.make("gym-myenv:MyEnv-v0")`, and shell out to `pip` if needed, but I'm doubtful — when running in my cluster, I don't want unexpected HTTP requests. Fundamentally I'd like to separate dependency management from runtime.

So in core, I expect we'll ask people to package their environments as `pip` packages (and maybe eventually Docker containers), and then we'll figure out some registry solution from there. (Likely not automated, at least to start.)

I don't have the strongest sense of the target audience for auto-downloading, but there's a broad design space to explore and I'm supportive of you trying it out!

ppaquette:
Added auto-installation of pip requirements:
```
>>> import gym
>>> gym.pull('ppaquette/gym_test/*')
[2016-08-22 09:30:28,412] Downloading configuration file "ppaquette/gym_test/master/.openai.yml"
[2016-08-22 09:30:28,724] Downloading environment files for "ppaquette/Acrobot-v0"
[2016-08-22 09:30:29,031] Downloading environment files for "ppaquette/CartPole-v1"
[2016-08-22 09:30:29,032] Installing environment requirements: "gym,numpy"
Requirement already satisfied (use --upgrade to upgrade): gym in /www/ppaquette/gym
Requirement already satisfied (use --upgrade to upgrade): numpy>=1.10.4 in ./.env2.7/lib/python2.7/site-packages (from gym)
Requirement already satisfied (use --upgrade to upgrade): requests>=2.0 in ./.env2.7/lib/python2.7/site-packages (from gym)
Requirement already satisfied (use --upgrade to upgrade): six in ./.env2.7/lib/python2.7/site-packages (from gym)
Requirement already satisfied (use --upgrade to upgrade): pyglet>=1.2.0 in ./.env2.7/lib/python2.7/site-packages (from gym)
Requirement already satisfied (use --upgrade to upgrade): numpy in ./.env2.7/lib/python2.7/site-packages
--------------------------------------------------
Successfully registered the environment: "ppaquette/Acrobot-v0"
Successfully registered the environment: "ppaquette/CartPole-v1"
```

We probably see different use cases.

Lots of requests I've seen were about making quick customization to an env (e.g. changing the action space, modifying reward functions, obtaining skip frames from atari env, etc.). The easiest way to do this is just fork a repo, change a couple of lines, and then pull the customized env by just changing the username in the pull() command.

The pip / docker scenario would add probably too much complexity for this use case IMO (i.e. for env creation, and not for installation). As for the versioning, I foresee most people sticking with v0 and v1, and just updating their master branch, so I'd be surprised if it was a big issue.

ppaquette:
Published as `gym_pull` on PyPi (https://github.com/ppaquette/gym_pull)

Syntax: 
```
import gym
import gym_pull
gym.pull(...)
```

Added some more complex envs to my [gym-envs repo](https://github.com/ppaquette/gym-envs).

Closing this PR for now.

JKCooper2:
gym_pull works like a charm, Mario jumping around my screen in no time. Amazing work.

gdb:
Cool. Any chance we could keep this as `gym_pull.pull(...)`? I recognize there's a repeated word there but would prefer not to monkey-patch `gym`.

gdb:
Just checked the current implementation — I see that there's a lot of monkey-patching currently. I suspect you don't want this implementation in the medium-term, as it's likely to be pretty brittle.

PR #299, opened 2016-08-18
Added PLE env to scoreboard (from https://github.com/lusob/gym-ple)

lusob:


gym/scoreboard/__init__.py
```diff
@@ -706,6 +712,32 @@
...
+    except ValueError as e:
+        raise ValueError('{}: id={}'.format(e, id))
+    ob_desc = ram_desc if ob_type == "ram" else image_desc
+    add_task(
+        id=id,
+        group='ple',
+        contributor='lusob',
+        summary="Maximize score in the game %(game)s, with %(ob_type)s as input"%dict(game=game, ob_type="RAM" if ob_type=="ram" else "screen images"),
+        description="""\
+Maximize your score in the pygame game %(game)s.
```
gdb: You should include some instructions here so people know how to obtain your environments.

lusob:
Hi @gdb, I've already added a description to install my package,  but it's not passing travis tests due to it  scoreboard addtask function expects that the ple env is registered, and I'm doing the register of the en on my own __init__.py module. Do you know an easy way to resolve it? thxs!

gdb:
We should fix that check. @catherio, mind taking?

catherio:
I'll take a look!

ppaquette:
@lusob You are probably better off making a pullable custom env.

- rename folder 'gym_ple' to 'ple'
- 'ple/__ init __.py' should just import PLEEnv (no need to register env)
- remove 'setup.py' in root folder
- add .openai.yml to root folder (continue from [here](https://gist.github.com/ppaquette/7a39159f19e2f4d7bc086fe964a799fd))

You can have a look at the envs I did [here](https://github.com/ppaquette/gym-envs) and at the .openai.yml syntax documentation [here](https://github.com/ppaquette/gym_pull).

I also published ``ple`` as a pip package, so it can be installed with ``pip install ple``, so it would be automatically installed when your envs are pulled (because it is listed in the requirements section).

You wouldn't need to register your envs on the scoreboard anymore, but just publish a short description explaining what they do (I'm almost done with my desc for my envs), and how to download them (e.g. gym.pull('lusob/your_repo_name/MonsterKong-image')

gdb:
(For now, you will still need to register them on the scoreboard if you want them to appear on gym.openai.com.)

gdb:
@ppaquette, why rename the package to `ple` vs `gym_ple`? I think it's much better to namespace all of the environment plugins.

ppaquette:
It's not registered on PyPi. It doesn't really matter. The example .openai.yml used 'ple/__init__.py', that's why I told him to rename.

gdb:
Ok, understood. The official mechanism here is going to be to publish a PyPi package, ideally `gym_...`. We will maintain a manual registry of those for now.

That should be compatible with `gym_pull`.

catherio:
@lusob you might also want to tidy up the references to "ram" versus "image" in your code - it looks like you borrowed some code from the Atari section that's not currently needed because your code uses only image-based environments, not RAM-based environments.

Specifically:
- in ple_env.py you do not need an 'obs_type' parameter
- in your __init__.py, you do not need "for obs_type in ['image']", because that loop is not currently doing anything

PR #303, opened 2016-08-21
limit mountain car action

d1hotpep:
limit the action on mountain car so that there isn't a trivial win scenario.  otherwise, you can simply pass action=10 into the env.step() and win.  see https://gym.openai.com/evaluations/eval_4oWGpaKrRLKssArKVss0sQ for example of why this limit is needed.

also did a couple minor cleanups while I was in there

### DIFF
--- a/gym/envs/classic_control/mountain_car.py
+++ b/gym/envs/classic_control/mountain_car.py
@@ -36,15 +36,13 @@ class MountainCarEnv(gym.Env):
         return [seed]
 
     def _step(self, action):
-        # action = np.sign((self.state[0]+math.pi/2) * self.state[1])+1
+        assert self.action_space.contains(action), "%r (%s) invalid" % (action, type(action))
 
         position, velocity = self.state
         velocity += (action-1)*0.001 + math.cos(3*position)*(-0.0025)
-        if (velocity > self.max_speed): velocity = self.max_speed
-        if (velocity < -self.max_speed): velocity = -self.max_speed
+        velocity = np.clip(velocity, -self.max_speed, self.max_speed)
         position += velocity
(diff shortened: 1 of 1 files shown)


tlbtlbtlb:
The original action space, following published results, was `Discrete(3)`, meaning the possible values are `0`, `1`, and `2`. This PR turns it into a completely different continuous action space, from [-1.5 .. +1.5], while still reporting `action_space=Discrete(3)`.

The right solution is to add:
```
        assert self.action_space.contains(action), "%r (%s) invalid"%(action, type(action))
```
at the beginning of step, as is done in `cartpole.py`.

d1hotpep:
updated, per your comment...clearly didn't understand the Discrete(3), sorry.  however, with a constant action of 2, my bot still generally wins trivially.  did I somehow set up the environment wrong?

tlbtlbtlb:
I don't get that behavior. Do you still have
```
        velocity += action*0.001 + math.cos(3*position)*(-0.0025)
```
instead of the original
```
        velocity += (action-1)*0.001 + math.cos(3*position)*(-0.0025)
```

d1hotpep:
good catch...not sure how that line got updated, but yeah that fixes it (along with the input constraint you suggested)

d1hotpep:
any other feedback on this diff?

regarding the travis failure, it looks like all 739 tests passed in both python 2.7 and 3.4, and then it failed with:

`ERROR:` InvocationError: '/usr/local/gym/.tox/py34/bin/nose2'`

not sure how to interpret / fix this

tlbtlbtlb:
Merged, thanks. The tox error is nondeterministic, still under investigation.

PR #306, opened 2016-08-22
added continuous mountain car v0

osigaud:
This new environment directly derives from the gym MountainCar environment (same structure, same render function), but with continuous action (force) control. The force action is taken from the MountainCar Environment from the "FAReinforcement" library of Jose Antonio Martin H. (version 1.0), adapted by  'Tom Schaul, tom@idsia.ch' and then modified by Arnaud de Broissia.

I believe it is useful because there are very few small continuous state/continuous action problems in gym so far (apart from pendulum).

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -41,6 +41,8 @@ class Env(object):
         observation_space: The Space object corresponding to valid observations
         reward_range: A tuple corresponding to the min and max possible rewards
 
+    Note: a default reward range set to [-inf,+inf] already exists. Set it if you want a narrower range.
+
     The methods are accessed publicly as "step", "reset", etc.. The
     non-underscored versions are wrapper methods to which we may add
     functionality over time.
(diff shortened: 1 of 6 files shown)


olegklimov:
I actually came to the same idea: it's nice to have continuous version of simple environments, I have/use LunarLanderContinuous for a while, probably should do a PR...

tlbtlbtlb:
MountainCar is sort of a degenerate case of continuous action spaces, as the optimal policy consists of only actions at the + or - limit. But maybe it's a reasonable addition for testing continuous action policies.

olegklimov:
Posted `LunarLanderContinuous` I mentioned earlier, in #307. Environments are exactly the same, only control method that is different.

osigaud:
I forgot to mention that the reward function in this instance of continuous mountain is such that you should reach the goal using as few energy (sum of squared actions, in practice) as possible. That's different from reaching the goal as fast as possible, and in this case the bang-bang control is not optimal, to answer Trevor Blackwell's comment.

gdb:
It does seem like this would be useful, and this is well-implemented. Thanks!

PR #307, opened 2016-08-22
LunarLanderContinuous

olegklimov:
This is actually a very conservative change. `LunarLander-v2` works _exactly_ like it worked previously. So no version bump. Two ways to check this:

1. Diff shows exactly the same logic for `not self.continuous` branches.

2. Heuristic code works and lands exactly the same.

Additionally, heuristic now can land a continuous-control lander as well. 

Finally, heuristic now is within a function, useful for user to record demonstration rollouts, as easy as `a = heuristic(env, s)`. I have `BipedalWalker` heuristic rewritten the same way, will post later.


gdb:
Looks great, thanks!

PR #311, opened 2016-08-23
Allow envs to be registered on the scoreboard but not in gym core

catherio:
In order to enable user-contributed environments to be listed on the scoreboard, but not packaged with gym core or included in the gym.make() registry, the "finalize" test in gym.scoreboard.registration needs to become more lenient. We still ensure that all gym core envs are in the scoreboard, but we now permit envs to exist in the scoreboard registry that are not in gym core.


gdb:
LGTM!

PR #312, opened 2016-08-24
SSBM environment running in the Dolphin Emulator

vladfi1:
This PR exposes Super Smash Bros. Melee (originally for the Nintendo GameCube) as gym environment. I think this would be a fun environment in which to train RL agents!

There are a few things left to do before merging. In roughly decreasing order of importance:
- [ ] Either inform users or automatically install my [fork](github.com/vladfi1/dolphin) of the dolphin emulator.
- [ ] Python 2.7 and 3.4/5 compatibility (currently only 3.X works). This may be tricky since some libraries got renamed from 2 to 3.
- [ ] Do something about the game ISO. Currently it is assumed to live at `~/ISO/SSBM.iso`, though it can be set with a kwarg.
- [ ] Enable playing against the builtin AI. This would be useful for benchmarking.
- [ ] Headless mode and other toggles.
- [ ] Allow more stages and characters to be selected.
- [ ] Capture pixels. Currently only internal structures are read from memory.
- [ ] Actually implement reset. Current behavior is to just pause, as the game mode is set to endless.

The last few can probably be done after merging as well.


pzhokhov:
this seems to belong more into gym-retro project (https://github.com/openai/retro).. Also, currently we are not adding new environments into the gym. Putting @endrift and @christopherhesse into the loop about moving this to retro and closing the PR.

PR #313, opened 2016-08-24
Create separate descriptions for PyGame environments

catherio:
PyGame environments now each have their own descriptions, using the text from http://pygame-learning-environment.readthedocs.io/en/latest/user/games/flappybird.html, rather than generic descriptions filled into a template.


gdb:
LGTM!

PR #314, opened 2016-08-24
Fix error when running min on empty list

JKCooper2:
Running min on an empty list (which occurs on first episode from `if len(content['timestamps'])==0: continue`) causes it to throw an error

### DIFF
--- a/gym/monitoring/monitor.py
+++ b/gym/monitoring/monitor.py
@@ -360,7 +360,7 @@ def merge_stats_files(stats_files):
     timestamps = np.array(timestamps)[idxs].tolist()
     episode_lengths = np.array(episode_lengths)[idxs].tolist()
     episode_rewards = np.array(episode_rewards)[idxs].tolist()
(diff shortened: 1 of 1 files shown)


gym/monitoring/monitor.py
```diff
@@ -360,7 +360,7 @@ def merge_stats_files(stats_files):
     timestamps = np.array(timestamps)[idxs].tolist()
     episode_lengths = np.array(episode_lengths)[idxs].tolist()
     episode_rewards = np.array(episode_rewards)[idxs].tolist()
-    initial_reset_timestamp = min(initial_reset_timestamps)
+    initial_reset_timestamp = min(initial_reset_timestamps) if len(initial_reset_timestamps) > 0 else []
```
gdb: Is the else case here buggy (should it be `0`)?

I also try to avoid the Python ternary operator; mind changing to a straight if...else?

JKCooper2:
True it should be 0. Updated.

gdb:
Thx!

PR #325, opened 2016-08-29
added ffmpeg to dockerfile and fixed test

kashif:


gym/monitoring/tests/test_video_recorder.py
```diff
@@ -22,16 +22,17 @@ def render(self, mode=None):
         pass
 
 # TODO(jonas): disabled until we have ffmpeg on travis
```
jonasschneider: Great! Feel free to also remove this note.

kashif:
@jonasschneider will do... i wanted to just see the docker/ubuntu error... i will fix that first

jonasschneider:
Thanks for the patch. Instead of using the third-party `ffmpeg`, could you rather install `libav-tools` to have `avconv` (which we also support in `VideoRecorder`)? I'd rather not have the unofficial PPA here, if we have the alternative.

jonasschneider:
Awesome, thank you!

PR #326, opened 2016-08-30
Force integer conversion in blackjack env comparison function

dennybritz:
Bug in the Blackjack environment: The hand comparison function returns a boolean, which is converted into a 0/1 reward, but the rewards should be -1, 0 or +1. For example, always picking the "stick" action results in never losing a game with a -1 reward.

Numpy also gives a warning: 

```
DeprecationWarning: numpy boolean subtract, the `-` operator, is deprecated.
```

### DIFF
--- a/gym/envs/toy_text/blackjack.py
+++ b/gym/envs/toy_text/blackjack.py
@@ -3,7 +3,7 @@ from gym import spaces
 from gym.utils import seeding
 
 def cmp(a, b):
-    return (a > b) - (a < b)
+    return int((a > b)) - int((a < b))
 
 # 1 = Ace, 2-10 = Number cards, Jack/Queen/King = 10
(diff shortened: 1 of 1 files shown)


tlbtlbtlb:
Hmmm. The code in cmp works correctly for python ints, but not numpy ints.
```python
>>> (5 > 6) - (6 > 5)
-1
>>> (6 > 6) - (6 > 6)
0
>>> (6 > 7) - (7 > 6)
-1
>>> (np.int64(6) > np.int64(7)) - (np.int64(6) < np.int64(7))
True
```
so I think this was introduced with the seeding changes. Fixing it changes the scores (which is why the Travis check is failing). @jonasschneider, is the test case incorrect?

pzhokhov:
fixed upstream via conversion to floats, closing this

PR #331, opened 2016-09-06
Add universal 'dim' interface to all spaces.

gaoyuan:
Exposing the state dimension will allow more generality in designing algorithms. For example, the parameters of a linear model will share the same dimension as the state space. Without this change, algorithms have to manually configure for each different state space.

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -307,6 +307,10 @@ class Space(object):
         # By default, assume identity is JSONable
         return sample_n
 
+    @property
+    def dim(self):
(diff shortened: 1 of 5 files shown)


gdb:
Thanks for the contribution!

Unfortunately, not all spaces will have a will-defined "dim" concept. This feels more like an external helper method to me. We don't ship helper methods in Gym currently, but it might be worth starting your own repo with it (and adding other helper methods as you or others find them useful).

PR #340, opened 2016-09-09
Wrapper metadata typo

olegklimov:
Good idea, the wrappers. Started to use, found a bug, here's a fix and a sample to test:

```python
import gym
from gym.wrappers import SkipWrapper

every_two_frame = SkipWrapper(2)
env = gym.make("FrozenLake-v0")
env = every_two_frame(env)
obs = env.reset()
for _ in range(1000):
  env.render()
  obs, reward, done, info = env.step(action)
```

(doesn't work with master, works after this PR)

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -316,10 +316,9 @@ class Wrapper(Env):
 
     def __init__(self, env=None):
         self.env = env
-        # Merge with the base metadata
-        metadata = self.metadata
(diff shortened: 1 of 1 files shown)


gdb:
Good catch. It's actually intended that the metadata from the Wrapper overrides that from the underlying Env. The issue is that Wrapper shouldn't have any default metadata. Just pushed a fix!

Issue #72, opened 2016-05-08
Bugs in Taxi-v1 environment


arboo:
I think the strings for the last action of _North_ and _South_ are switched. Also sometimes there seems to be a passenger to pick up, but no destination.

Are there any plans to extend the environment in a probabilistic manner? Like in the original paper when choosing an action there is a 20 % chance to end up somewhere else.

joschu:
Thanks, should be fixed now. The problem was that sometimes the start location was the same as the destination.
I didn't notice in the paper that the transitions were random, but that's good to know.

PR #345, opened 2016-09-15
Fix Taxi (upgrade -> v2)

vadimq:
This fixes the initial state distribution of Taxi such that it no longer produces a start location equal to the goal location (issue #72). At the moment the distribution is not what's expected.

The new reward threshold was calculated using policy iteration.


vadimq:
We need them because with this fix the average reward for an episode while following the optimal policy declines from 10.2 to 8.46, and the actual threshold, as I understand it, is just this number multiplied by 0.95.

9.7 will become hardly achievable, yes, but, more importantly, it won't make sense anymore.

joschu:
Looks good

PR #364, opened 2016-10-02
swap avconv / ffmpeg. 

michalsustr:
ffmpeg is obsolete, so it shouldn't be evaled first (crashes on my pc)

### DIFF
--- a/gym/monitoring/video_recorder.py
+++ b/gym/monitoring/video_recorder.py
@@ -247,10 +247,10 @@ class ImageEncoder(object):
         self.frame_shape = frame_shape
         self.frames_per_sec = frames_per_sec
 
-        if distutils.spawn.find_executable('ffmpeg') is not None:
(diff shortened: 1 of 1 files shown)


tlbtlbtlb:
Changing this will probably break other systems. Could you simply deinstall ffmpeg if it's broken anyway?

gdb:
@tlbtlbtlb Hm, on second thought, on what systems would avconv be unhappy? (This should only change behavior if you have both ffmpeg and avconv installed, and one of them is broken but not the other.)

tlbtlbtlb:
I had such a configuration on my mac last summer. IIRC the problem was that avconv in MacPorts didn't include libx264 by default, which gym asks for with -vcodec. But that might be a very small minority of users these days.

Searching github, several projects that support both prefer avconv, so probably time to switch.

PR #372, opened 2016-10-09
Scale height reported by state to viewport height rather than width

peejaybee:
The scaling in the reported state looks off to me.  It shouldn't matter to the solution, since it's just scaling by a different constant, but seems like the vertical velocity and height should be in compatible scales.

### DIFF
--- a/gym/envs/box2d/lunar_lander.py
+++ b/gym/envs/box2d/lunar_lander.py
@@ -280,7 +280,7 @@ class LunarLander(gym.Env):
         vel = self.lander.linearVelocity
         state = [
             (pos.x - VIEWPORT_W/SCALE/2) / (VIEWPORT_W/SCALE/2),
-            (pos.y - (self.helipad_y+LEG_DOWN/SCALE)) / (VIEWPORT_W/SCALE/2),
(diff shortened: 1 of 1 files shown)


olegklimov:
You're probably right, that's a typo. 

Description says landing pad is at (0,0), so a very wrong thing would be zero out of place. But that's not the case, just scale is wrong. 

IMO, change is correct, but it's not worth environment version increase. Maybe put this in comment and apply anytime we increase version for some other reason?

peejaybee:
That makes sense, or we can just keep this PR open and I can rebase when it's time for a version bump.

pzhokhov:
has been fixed upstream; closing this

pzhokhov:
or actually version upstream has VIEWPORT_H/SCALE/2, whereas the PR has VIEWPORT_H/SCALE... @olegklimov could you comment whether that's important?

olegklimov:
I think it's correct as it is. Observation scale is the same for x and y, which is good. The modified version would be just as learnable and largely equivalent, isn't worth the change.

pzhokhov:
ok then, closing this

PR #383, opened 2016-10-20
Algorithmic refactor

colinmorris:
Refactored/documented the algorithmic environments, and added some tests. When I started working on solving some of these environments, I turned to the source code to understand what was going on, but found it a bit opaque in places. I'm hoping it will now be a bit easier to understand. (As a side benefit, it should also be a good bit faster.)

For the sake of keeping this P.R. simple, and not having to worry about bumping the version number on any envs, I've kept behaviour the same with respect to public interfaces.


gym/envs/algorithmic/copy_.py
```diff
@@ -0,0 +1,14 @@
+"""
```
colinmorris: (I renamed this file because the collision with the python standard library of the same name led to some problems. For example, running `pip` or `nose` in the gym/envs/algorithmic directory would fail with obscure error messages.)

gdb: Ah good catch.

gdb:
Thank you, this is great — Travis looks unhappy, but happy to merge after the tests are green!

colinmorris:
Hm, okay, I'm not sure what the problem is. All tests passed when I ran them locally. 

When I click through for details of the failure, it just says "The repository at openai/gym was not found.". I'm guessing it's spurious - the Travis CI status page mentions a lot of outages and build issues right now. I guess I'll wait and see if it clears up on its own.

colinmorris:
Whoops, my bad, didn't realize I'd need to run tests under python3 as well. Looks like Travis is happy now.

I had to re-generate the python 3 rollout file, to generate new hashes for `ReversedAddition-v0` and `ReversedAddition3-v0`. Which may seem like cheating, but after some digging, it looks like those envs were totally broken in python 3. You can verify by running this with python 3:

    from __future__ import print_function
    import gym
    env = gym.make('ReversedAddition-v0')
    env.seed(0)
    obs = env.reset()
    print(env.target)

Which prints:

    {0: 1, 1: 2.333333333333333, 2: 0.7777777777777777}

(Those are the 'digits' of the putative sum the agent is supposed to write)

I'm curious how this kind of thing could be caught in the future. Is there a good reason to even have separate rollout files for python2 and python3? Seems like environments should, ideally, have the same behaviour for both.

gdb:
Yeah, good question — I haven't looked at that code before. Do the files actually differ at all?

colinmorris:
It looks like the only differences are that `OneRoundNondeterministicReward-v0` and `TwoRoundNondeterministicReward-v0` differ in rewards. Wouldn't be surprised if that was just a similar bug. I'll file a ticket for making a single rollout file, and may try to tackle it shortly.

gdb:
Awesome, thanks for the investigation!

PR #389, opened 2016-10-25
rfc: allow optionally assigning names to box dimensions and discrete values

colinmorris:
Before (CartPole):

```python
>>> env.action_space
Discrete(2)
>>> env.observation_space
Box(4,)
```
After:

```python
>>> env.action_space
Categorical(['LEFT', 'RIGHT'])
>>> env.observation_space
Box({x: -4.8..4.8, x_dot: -3.4e+38..3.4e+38, theta: -0.42..0.42, theta_dot: -3.4e+38..3.4e+38})
```

I only re-did CartPole as an example. If this looks agreeable, I'm happy to go through and start changing others as well.

I think the Categorical option is a definite improvement. The box changes feel a little more kludgy. Maybe `repr` should just show the names of dimensions (if present), but not try to fit in their ranges?

Currently, this only affects `repr` output and, in the case of the `Categorical` class, adds some convenience attributes to use in place of magic numbers (e.g. `self.action_space.RIGHT` instead of `1`).

### DIFF
--- a/gym/spaces/discrete.py
+++ b/gym/spaces/discrete.py
@@ -23,6 +23,24 @@ class Discrete(gym.Space):
             return False
         return as_int >= 0 and as_int < self.n
     def __repr__(self):
-        return "Discrete(%d)" % self.n
+        return "Discrete({})".format(self.n)
     def __eq__(self, other):
         return self.n == other.n
+
+class Categorical(Discrete):
+    """A discrete space with named elements. Element indices are made available
+    as attributes (as long as they're legal Python names). e.g.
+    >>> shapes = Categorical(['circle', 'triangle', 'rectangle'])
+    >>> shapes.circle
+    0
+    """
+    def __init__(self, named_members):
+        self.n = len(named_members)
+        self.named_members = list(named_members)
+    def __repr__(self):
+        return "Categorical({!r})".format(self.named_members)
+    def __getattr__(self, attr):
+        try:
+            return self.named_members.index(attr)
+        except IndexError:
+            raise AttributeError("{} has no attribute {}".format(self, attr))
--- a/gym/spaces/discrete.py
+++ b/gym/spaces/discrete.py
@@ -37,10 +37,11 @@ class Categorical(Discrete):
     def __init__(self, named_members):
         self.n = len(named_members)
         self.named_members = list(named_members)
+        for i, named_member in enumerate(named_members):
+            if hasattr(self, named_member):
+                raise ValueError("Element name {} collides with existing attribute.")
+            else:
+                setattr(self, named_member, i)
+
     def __repr__(self):
         return "Categorical({!r})".format(self.named_members)
-    def __getattr__(self, attr):
-        try:
-            return self.named_members.index(attr)
-        except IndexError:
-            raise AttributeError("{} has no attribute {}".format(self, attr))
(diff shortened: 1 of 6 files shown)


gym/spaces/discrete.py
```diff
@@ -23,6 +23,24 @@ def contains(self, x):
...
+    >>> shapes = Categorical(['circle', 'triangle', 'rectangle'])
+    >>> shapes.circle
+    0
+    """
+    def __init__(self, named_members):
+        self.n = len(named_members)
+        self.named_members = list(named_members)
+    def __repr__(self):
+        return "Categorical({!r})".format(self.named_members)
+    def __getattr__(self, attr):
```
gdb: It's probably better to statically set the props (i.e. use `setattr` in `__init__`)

gdb:
I like it. Seems like a nice improvement with total backwards compatibility. @joschu what do you think?

gym/spaces/discrete.py
colinmorris: Done in [7edd7ea](https://github.com/openai/gym/pull/389/commits/7edd7ea9fe44fcef748f50c8801f38987ca3ce5f). Another option would be to use `__getitem__`, which would remove any concerns about namespace pollution, but require a few more keystrokes to invoke.

joschu:
I don't like that there will be two classes, Discrete and Categorical, that do basically the same thing, but algorithm code will have to deal with both of them separately.
Naming the components in Box is not worth the extra complexity, in my opinion.

gdb:
Note that Categorical descends from Discrete, so you can just ignore the fact it exists.

Could still be confusing given that it wasn't obvious...

Another option would be to have this functionality optionally present in the Categorical + Box classes.

joschu:
Sounds most reasonable to add a optional labels member to the Box and Discrete spaces.
I'm not in favor of changing the default string representation, since we might end up printing a huge amount, and it'd be complicated to get the display right.
For Box, it might make sense to have index_labels (which label the indexes of a 1d array) and axis_labels (which label the different axes of a multidimensional array).

colinmorris:
> I don't like that there will be two classes, Discrete and Categorical, that do basically the same thing, but algorithm code will have to deal with both of them separately.

FWIW, I originally tacked the functionality on to the Discrete class, but I kept branching on `self.named_members is None`, which felt like an anti-pattern. Also, I think the two classes represent a useful semantic distinction. Discrete represents (or could represent), sets where there's a meaningful ordering, or meaningful magnitudes. e.g. the sum of your hand in the Blackjack environment, or the rank of the dealer's showing card. Whereas categorical is just a bunch of things that we assign arbitrary integer labels to (e.g. 'hit' or 'stick').

> I'm not in favor of changing the default string representation, since we might end up printing a huge amount, and it'd be complicated to get the display right.

Do you mean for Box only, or Box and Discrete(/Categorical)? For Box's repr, I have a check to only use the 'rich' output if the number of indices is less than 10. Maybe it should be lower. It could also be made a lot more compact by just showing the labels, and not the high/low ranges.

> For Box, it might make sense to have index_labels (which label the indexes of a 1d array) and axis_labels (which label the different axes of a multidimensional array).

That makes sense. And you're suggesting that users should just access those attributes directly if they want to learn more, rather than them being exposed in repr? 

I'd be fine with re-doing it that way, or just cutting the Box changes entirely for now. I do feel pretty strongly that the Discrete/Categorical changes are an improvement though.

gdb:
I'm ok with the namespace pollution, but prefer avoiding dynamism when
possible!

joschu:
I vote against Categorical, people will wonder if they should use Discrete or Categorical, they'll have to look at the code to figure out that it's a subclass...benefit is not worth the extra complexity.

PR #392, opened 2016-10-26
Assert action is contained in action_space in Env.step

colinmorris:
A lot of individual environments already asserted this in their `_step` method, but not all of them.


gdb:
We actually used to do this, but unfortunately the performance overhead for some environments makes it not worth it :(. We could in theory allow for this behavior in an opt-in fashion in metadata, but I don't think that's worth it when the env can just do the check on its own!

colinmorris:
Interesting! I never would have guessed.

PR #393, opened 2016-10-26
Redirect stderr to avoid avconv version spam

colinmorris:
In addition to what it prints to stdout, `avconv -version` outputs a long-ish version string to stderr. As a result, 90% of Travis build logs are composed of that string. I've also seen it in the wild when running monitors.


gdb:
Nice catch :)

PR #406, opened 2016-11-22
Unbreak random_agent

avital:
The semantics of `env.reset()` have changed. It's no longer possible
to reset an environment until it's done.

Notably, this commit changes the behavior of random_agent. It used
to have a limit on number of actions per episode. If we wanted
that now, we'd have to close the environment and recreate it
on each episode, which may be slow.

### DIFF
--- a/examples/agents/random_agent.py
+++ b/examples/agents/random_agent.py
@@ -48,14 +48,12 @@ if __name__ == '__main__':
     agent = RandomAgent(env.action_space)
 
     episode_count = 100
-    max_steps = 200
     reward = 0
     done = False
 
     for i in range(episode_count):
         ob = env.reset()
-
-        for j in range(max_steps):
+        while True:
             action = agent.act(ob, reward, done)
             ob, reward, done, _ = env.step(action)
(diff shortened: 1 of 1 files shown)


tlbtlbtlb:
The random_agent in `universe/bin/random_agent.py` respects the new semantics.
For now, gym retains the old semantics where reset is allowed.

tlbtlbtlb:
I take that back, we should avoid early reset in gym envs too since the monitor doesn't allow it. So the PR makes sense.

avital:
Tests are failing, almost certainly not due to this change. Are there flaky tests?

```
FAIL: gym.envs.tests.test_envs_semantics.test_env_semantics:138
EnvSpec(Blackjack-v0)
  File "/usr/local/gym/gym/envs/tests/test_envs_semantics.py", line 82, in test_env_semantics
    assert rollout_dict[spec.id]['rewards'] == rewards_now, 'Rewards not equal for {}'.format(spec.id)
AssertionError: Rewards not equal for Blackjack-v0
```

PR #408, opened 2016-11-22
Implementations of bandits and tabular MDPs for rl^2

dementrock:
Implements the bandits and random tabular MDP tasks, as used in the following paper:

Duan, Y., Schulman, J., Chen, X., Bartlett, P. L., Sutskever, I., & Abbeel, P. (2016). [RL^2: Fast Reinforcement Learning via Slow Reinforcement Learning](https://arxiv.org/pdf/1611.02779v2.pdf). arXiv preprint arXiv:1611.02779.

Also registered benchmarks for them separately.

### DIFF
--- a/gym/benchmarks/__init__.py
+++ b/gym/benchmarks/__init__.py
@@ -264,3 +264,45 @@ register_benchmark(
          'max_timesteps': 1000000,
         }
     ])
+
+bandit_tasks = []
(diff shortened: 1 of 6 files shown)


gdb:
Ah sorry just saw this now! This is great; fix the conflict and then merge?

PR #416, opened 2016-12-08
Implementations of bandits and tabular MDPs for rl^2

dementrock:
Implements the bandits and random tabular MDP tasks, as used in the following paper:

Duan, Y., Schulman, J., Chen, X., Bartlett, P. L., Sutskever, I., & Abbeel, P. (2016). [RL^2: Fast Reinforcement Learning via Slow Reinforcement Learning](https://arxiv.org/pdf/1611.02779v2.pdf). arXiv preprint arXiv:1611.02779.

Also registered benchmarks for them separately.

Rebased and fixed conflicts to match the latest master.

### DIFF
--- a/gym/benchmarks/__init__.py
+++ b/gym/benchmarks/__init__.py
@@ -568,3 +568,45 @@ register_benchmark(
          'reward_ceiling': 1000.0,
         },
     ])
+
+bandit_tasks = []
+for n_arms in [5, 10, 50]:
+    for n_episodes in [10, 100, 500]:
+        bandit_tasks.append({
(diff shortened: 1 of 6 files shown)


gdb:
\o/. Ship it!

pzhokhov:
so uh this happened some time ago... And normally I would say that we should not include new environments into gym. However, @gdb gave his blessing... So I would refer to higher powers and summon @gdb and @joschu for advice on this one. We should probably remove the benchmarks portion of it, but the rest seems relatively conflict-less, so if need be, I can merge.

pzhokhov:
the higher powers were summoned and did not respond. Closing this (one last ping @joschu and @gdb - can reopen if requested)

joschu:
d'oh -- it's a shame we didn't merge.
anyway, probably doesn't make sense to do it now

PR #422, opened 2016-12-13
Monitor wrapper

colinmorris:
Extracts monitor logic from Env base class and adds it to a wrapper.

My overall impression is that this is a net loss in terms of clarity and robustness :(

Some alternative approaches:

- Make Wrapper's `__getattr__` fall through to `getattr(self.env, foo_attr)`. This would remove the need for the hacky check in `make()` to avoid breaking universe envs. The reason that check is necessary is that wrapping in gym.Wrapper hides the env's attributes (like `env.n`). This would also avoid having to copy attributes like `action_space`, `reward_range`, etc. in Wrapper's constructor.
- Contain the change to just `core.py`. Move the monitor logic to a wrapper, but export that wrapper under the name 'Env'. 

(These also seem kind of bad in their own ways.)

### DIFF
--- a/gym/envs/safety/semisuper.py
+++ b/gym/envs/safety/semisuper.py
@@ -15,6 +15,7 @@ import gym
 class SemisuperEnv(gym.Env):
     def step(self, action):
         assert self.action_space.contains(action)
+        # TODO: is this right? Should already be done in parent class. 
         self.monitor._before_step(action)
 
         observation, true_reward, done, info = self._step(action)
--- a/gym/envs/safety/semisuper.py
+++ b/gym/envs/safety/semisuper.py
@@ -15,13 +15,9 @@ import gym
 class SemisuperEnv(gym.Env):
     def step(self, action):
         assert self.action_space.contains(action)
-        # TODO: is this right? Should already be done in parent class. 
-        self.monitor._before_step(action)
 
         observation, true_reward, done, info = self._step(action)
         assert self.observation_space.contains(observation)
-
-        done = self.monitor._after_step(observation, true_reward, done, info)
 
         perceived_reward = self._distort_reward(true_reward)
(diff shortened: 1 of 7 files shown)


gym/envs/safety/semisuper.py
```diff
@@ -15,13 +15,10 @@
 class SemisuperEnv(gym.Env):
     def step(self, action):
         assert self.action_space.contains(action)
-        self.monitor._before_step(action)
```
colinmorris: I have no idea why this env was doing monitor stuff here. I'm assuming it was a mistake.

gdb: This environment looks like it's overriding `step` rather than `_step` — that's a bug.

gdb:
Ahh so I think it'd be better to have Monitor be an *optional* wrapper — so you wrap the env yourself: `env = Monitor(env, '/tmp/random-agent')`. For Universe/vectorized envs, we should then use a `VectorizedMonitor`. This allows the user to inject the monitor at various points in the wrapper stack, and try out different monitor implementations.

gym/envs/safety/semisuper.py
jietang: The safety environments want to provide one reward to the agent (observed_reward) while recording a different reward in the monitor (true_reward). 

(Aside: this would be easy to implement if monitor was a wrapper, since you could have a wrapper that distorts the reward after the monitor wrapper)

jietang:
Yeah my initial thinking was that it would be nice to maintain backward compatibility with old gym code. After reading the diff I agree that approach is less clear, and I support changing the wrapper to be optional + having the informative error message about how to use the new interface.

colinmorris:
Cool, I'll take a stab at that approach and open a fresh P.R. when it's ready

PR #423, opened 2016-12-14
Minor tweaks to random agent scripts

colinmorris:
Some more usage documentation and make benchmark_runner auto-upload

### DIFF
--- a/examples/agents/random_agent.py
+++ b/examples/agents/random_agent.py
@@ -14,8 +14,10 @@ class RandomAgent(object):
         return self.action_space.sample()
 
 if __name__ == '__main__':
-    parser = argparse.ArgumentParser(description=None)
-    parser.add_argument('env_id', nargs='?', default='CartPole-v0', help='Select the environment to run')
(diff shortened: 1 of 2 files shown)


colinmorris:
I'm confused by the `MujocoDependencyError` in CI. There's no way these changes could have caused that kind of error. I feel like I've seen this every time I've run CI in the last few days, so I assumed it was an existing issue - but [this](https://travis-ci.org/openai/gym/builds/183783869) recent build of master doesn't have it. (I tried merging very latest master and running C.I. again, and still got the error.)

gdb:
Hm, I think pull requests are supposed to skip tests requiring MuJoCo credentials: https://github.com/openai/gym/blob/master/gym/envs/tests/test_envs.py#L15-L18. Maybe we just need to ignore some extra tests?

tlbtlbtlb:
When it does try to run the mujoco tests, they all fail. For now, remove the `InvertedPendulum-v0` entry from the `rollouts.json` file.

pzhokhov:
with the benchmark / scoreboard removed, I think this is obsolete and am closing it.

PR #424, opened 2016-12-14
Optional monitor wrapper.

colinmorris:
Sequel to [#422](https://github.com/openai/gym/pull/422). 

There would need to be some non-trivial corresponding changes to universe before this is mergeable, but I thought I'd float this here first to see if it's on track.

I have some concerns:

- this is a pretty significant breaking API change (not in the sense of being a huge departure, but it affects a common operation, so it'll break a lot of people's code).
- it's not really clear how to fix the issue this introduces in SemiSuperEnv. It's true that you could recover the same behaviour by writing a separate Monitor wrapper. But how does the user know they're supposed to import `SemiSuperMonitored` rather than using the regular `Monitored` wrapper?
- without knowing how wrappers are implemented, would you guess that `Monitored(SpamWrapper(env))` enables monitoring, but `SpamWrapper(Monitored(env))` doesn't?

### DIFF
--- a/gym/monitoring/tests/test_monitor.py
+++ b/gym/monitoring/tests/test_monitor.py
@@ -4,7 +4,7 @@ import os
 import gym
 from gym import error, spaces
 from gym import monitoring
-from gym import Monitored
+from gym.wrappers import Monitored
 from gym.monitoring import monitor
 from gym.monitoring.tests import helpers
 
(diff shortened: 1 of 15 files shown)


gym/monitoring/tests/test_monitor.py
```diff
@@ -13,16 +14,18 @@ def _render(self, close=True):
...
 def test_monitor_filename():
     with helpers.tempdir() as temp:
-        env = gym.make('CartPole-v0')
-        env.monitor.start(temp)
-        env.monitor.close()
+        env = gym.make('Cartpole-v0')
+        monitor = monitoring.Monitor(env)
+        env = Monitored(env, monitor)
+        monitor.start(temp)
+        monitor.close()
```
colinmorris: @gdb is this roughly what you had in mind?

gdb: I was thinking more like this:

```env = gym.make('Cartpole-v0')
monitor = wrappers.Monitor(env, temp)
env = wrappers.Other(monitor)
env.reset()
env.step(...)
monitor.close() # if needed; generally env.close() should be enough.
 ```

tlbtlbtlb:
For the Semisuper envs, I think the right thing is to return `info={'true_reward': ...}` and have the monitor use that true_reward instead of reward if provided.

There are other wrappers that meddle with reward, like the `RewardOutlierWrapper` in vnc-agents. And I want a `SqrtRewardWrapper` that replaces reward with the delta of `sqrt(sum(rewards))`, for envs where the score goes up more than linearly with performance.

So net net, I think the Monitor wrapper should use `info.get('true_reward', reward)` and any other wrapper that meddles with reward should put the true value there.

colinmorris:
Can I make the simplifying assumption that anyone wanting to monitor an env will want to do so for its entire lifetime? i.e., no delayed starts and no premature termination of monitoring? Otherwise, this pattern might lead to awkward situations like...

    env = gym.make('Copy-v0')
    runner = MyEnvRunner(env, 12)
    for i in range(1000):
        runner.run_one_episode()
        if i == 900:
            env = Monitor(env, tmpdir) # uh oh, runner is still pointing to old env
    env.close()

In general, it seems like adding wrappers at arbitrary points in a program is going to cause trouble, because you would need to update all pointers to the previous environment. So are wrappers intended to be applied all at once right after an env is created?

gdb:
It's the right question to ask. You should definitely only wrap the env at creation time (prior to calling configure — it'll probably break if not). But, envs are pretty cheap to set up, and I don't mind making them be monitored from the start. Of course, the beauty is that if someone wants to have a monitor with start/stop functionality, they can implement it themselves!

nottombrown:
Ah, I didn't see this and resolved the issue in #434. Closing

PR #429, opened 2016-12-21
Bug in frozen lake implementation

last-g:
Because of bug terminal states do not act as terminal


hunkim:
This seems a reasonable fix, but why the test fails?

tlbtlbtlb:
The test was failing due to a different problem. Thanks for the fix

PR #434, opened 2016-12-23
Add Monitored wrapper

nottombrown:
Adds Monitored wrapper. Universe has a sibling PR here: https://github.com/openai/universe/pull/79

This contains breaking changes so I've bumped the `VERSION` to 0.6.0.

To move to the wrapper we've made the following changes:

- To add the `Monitor`, you now use `wrappers.Monitor(out_dir)(env)`
- Monitoring starts as soon as the env is wrapped
- Monitoring is closed automatically during `env.close`
- In semisupervised tasks, we now return a `true_reward` value in `info` in addition to the distorted `reward` returned by `step`


tlbtlbtlb:
Looks good to me.

Some wrappers are reasonable to use more than once in the stack, like `universe.wrappers.Recording` if you want to record both raw observations and post-rescale. Can I suggest factoring out a `_check_for_duplicate_wrappers` method out of `_update_wrapper_stack` that such special wrappers can override?

nottombrown:
Good call. I like that solution. Updated.

PeterMitrano:
Is there documentation on how to use this somewhere? [The old documentation](https://gym.openai.com/docs) is now out of date

gym/wrappers/monitoring.py
```diff
@@ -0,0 +1,41 @@
+from gym import monitoring
+from gym import Wrapper
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+def Monitor(directory, video_callable=None, force=False, resume=False,
```
gdb: I'd prefer not to do this in decorator style, but instead have the Monitor class __init__ method takes these as arguments directly.

PeterMitrano: if this is in reference to writing `Monitor(directory)(env)` I agree. It's unexpected

nottombrown: Agreed, have a fix here to move away from the decorator syntax. Will merge in the morning:
https://github.com/openai/gym/pull/442

JamesCaud:
When adding a monitor a TypeError gets raised. Could you check out this trace? Is this me or you?

  File "slither.py", line 217, in <module>
    main()
  File "slither.py", line 208, in main
    observation_n, reward_n, done_n, info = env.step(action_n)
  File "/home/james/openai/gym/gym/core.py", line 110, in step
    observation, reward, done, info = self._step(action)
  File "/home/james/openai/gym/gym/wrappers/monitoring.py", line 22, in _step
    done = self._monitor._after_step(observation, reward, done, info)
  File "/home/james/openai/gym/gym/monitoring/monitor_manager.py", line 249, in _after_step
    self.stats_recorder.after_step(observation, reward, done, info)
  File "/home/james/openai/gym/gym/monitoring/stats_recorder.py", line 53, in after_step
    self.rewards += reward
TypeError: unsupported operand type(s) for +=: 'int' and 'list'

tlbtlbtlb:
@JamesCaud Your environment is returning a list for the reward, it should be a scalar.

PR #435, opened 2016-12-25
Fixed RIGHT/LEFT/DOWN/UP for Frozen Lake

hunkim:
It seems the current macros are not correct based on the code and render.

```python
        def inc(row, col, a):
            if a==0:
                col = max(col-1,0)
            elif a==1:
                row = min(row+1,nrow-1)
            elif a==2:
                col = min(col+1,ncol-1)
            elif a==3:
                row = max(row-1,0)
            return (row, col)
```

```python
outfile.write("  ({})\n".format(["Left","Down","Right","Up"][self.lastaction]))
```


tlbtlbtlb:
Thanks!

PR #441, opened 2016-12-28
Fix video recording on Windows - skip setsid if attribute not present

bstriner:
Conditionally reference os.setsid if that attribute is present. setsid is not an attribute of os on Windows. 

With this change, it looks like video recording is working correctly on Windows. Without the change, I get an error about setsid not being an attribute of os. I'm running Anaconda python 2.7.

Shouldn't affect mac and linux users.

Cheers,
Ben


bstriner:
To clarify, the master branch is rendering video for me on Windows aside from that additional if statement. Not sure if the windows branch is being kept up to date.

tlbtlbtlb:
Thanks!

PR #443, opened 2016-12-28
Fix CEM example for Monitor wrapper change

benkant:
Fix for CEM agent example re Monitor wrapper change #434


tlbtlbtlb:
Thanks!

Issue #449, opened 2016-12-29
Why is scipy pinned to 0.17.1?


bstriner:
In `requirements.txt`, scipy is pinned to 0.17.1. Should this be a minimum or is there something that is broken in later versions that it should be pinned?

https://github.com/openai/gym/commit/645f5a0758204e3ed28a7329ce05eddb03dfc4f7

```
numpy>=1.10.4
requests>=2.0
six
pyglet>=1.2.0
scipy==0.17.1
```

The requirements in `setup.py` are `'numpy>=1.10.4', 'requests>=2.0', 'six', 'pyglet>=1.2.0',`. Depending on the answer to the above question, should update `setup.py` accordingly.

Cheers,
Ben

bstriner:
On a related note, parameter_tuning should be able to run on tensorflow or theano depending on how Keras is configured, so the dependency on theano can probably be removed.

Search gym for theano and the only reference is in setup.py.

bstriner:
Updated requirements in this PR: https://github.com/openai/gym/pull/450

tlbtlbtlb:
The gym library doesn't depend on scipy, so it's correct not to include it in setup.py
The gym scoreboard application does require scipy. Scipy 0.18 has significant API changes, which we haven't tested against. Unless this is causing a problem, I'd prefer to leave it.

Keras in PyPI depends on Theano, so I don't see the harm of including it as a dependency.

bstriner:
Thanks for the article and the explanation. The issue is really the extra scipy in requirements.txt. Sounds like that is safe to remove. Eventually, we could add a scipy dependency as an extra like 'scoreboard' in setup.py. Does that sound like the correct configuration?

Thanks for the clarification that scipy is only a requirement for the scoreboard.

Nothing urgent, I was just getting some warnings in my IDE about missing scipy, but scipy wasn't a listed dependency, so I was trying to figure out the source of the issue.

Regarding keras, the code should be able to run on a system with only keras and tf, but setup.py will require theano as well. No harm unless someone really doesn't want to install theano.

Cheers,
Ben

PR #450, opened 2016-12-29
Synchronize dependencies in requirements.txt and setup.py

bstriner:
Synchronize dependencies in requirements.txt and setup.py.

I changed scipy to a greater-than requirement instead of equality. Please let me know if there is something broken in later versions of scipy. I also added scipy to setup.py. I removed theano because it is a dependency of keras.

https://github.com/openai/gym/issues/449

Cheers,
Ben

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,4 +2,3 @@ numpy>=1.10.4
 requests>=2.0
 six
 pyglet>=1.2.0
-scipy==0.17.1


bstriner:
Sounds like a better solution might be to remove scipy from requirements.txt and add to an extra in setup.py.

pzhokhov:
currently there is a dependency on scipy (in kellycoinflip environment). Correspondingly, scipy without a version is included into setup.py and versioned into requirements.txt - which is, to my understanding, how it should be. Closing this PR.

PR #310, opened 2016-08-23
BoxWrapper 

vladfi1:
`BoxWrapper` provides a standard way to convert arbitrary (potentially nested) `Tuple`s of `Discrete` and `Box` spaces into a one-dimensional `Box`. This is especially useful for environments that expose their underlying data structures as observations.

PR #451, opened 2016-12-29
Utilities for observation spaces

bstriner:
Hello everybody,

These are generic utilities for converting observations and observation spaces into feature vectors. Using these utilities, an agent should be able to support any combination of Tuple, Box and Discrete observation spaces. The goal is to easily write generic agents without knowing the specifics of the observation space.

At the extreme, you can flatten the input entirely regardless of its structure.

Typical use case:
```
# flatten Tuple and Box spaces and one-hot encode Discrete spaces
# model input shape calculated by concatenated_input_dim
input_layer = Input(shape=(concatenated_input_dim(space),))
# 1-D numpy feature vector calculated by concatenated_input
input_features = concatenated_input(observation, space)
```

If you want more control, you can use `space_shapes` and `flatten_input` to one-hot encode input and convert a tree into a list, but not flatten the inputs.

Cheers,
Ben

### DIFF
--- a/gym/utils/space_utils.py
+++ b/gym/utils/space_utils.py
@@ -1,0 +1,90 @@
+from gym.spaces import Tuple, Box, Discrete
+import itertools
+import numpy as np
+
+
+def flatten_spaces(space):
+    """
+    Flatten a tree of Tuple spaces into a list of simple spaces.
+    :param space: A space that may be a Tuple
+    :return: a list of spaces
+    """
+    if isinstance(space, Tuple):
+        return list(itertools.chain.from_iterable(flatten_spaces(s) for s in space.spaces))
+    else:
+        return [space]
+
+
+def space_shapes(space_list):
+    """
+    Return shapes of a list of Discrete and Box spaces
+    :param space_list: list of spaces
+    :return: list of shape tuples
+    """
+    shapes = []
+    for space in space_list:
+        if isinstance(space, Discrete):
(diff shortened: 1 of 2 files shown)


vladfi1:
Seems very similar to the BoxWrapper PR https://github.com/openai/gym/pull/310 I submitted a while ago... which also never got merged :(

olegklimov:
Either API is good enough so you don't need utilities, or everyone write their own anyway... Also there is limited use of mixing RGB values with higher-level observations like joint positions -- probably it's a tuple for a reason, it should be wired into the network by a different path. Especially a generic agent (i.e. not trivial) should care about this kind of things. Another example: RGB image and depth image -- you want to concatenate them together, do you? Sorry for not being positive :sweat_smile:

nottombrown:
I tend to agree with @olegklimov here. Seems like these utilities are out of scope for `gym`. 

Thanks for taking the time to make this PR though, @bstriner!

emerrf:
+1 for having the space utilities available. My reasons are:
- it is hard to apply generic algorithms with such diversity of space's methods available, specially now that RL algos are groupped in baselines repo. I.e.: baselines.deepq.simple.learn has the env as a parameter but it assumes the obs. space has shape attribute (Box) and the action space has n attribute (Discrete).
- if everyone is writing their own, like @bstriner and @vladfi1, it is probably because there are no generic tools available.

PR #458, opened 2017-01-06
Added the Retro Learning Environment (SNES games) to gym

nadavbh12:


tlbtlbtlb:
Wow! Looks awesome!

It looks like you copied the many different versions (Deterministic, NoFrameskip, -v0, -v3) from the Atari versions. Most of these are there for backwards compatibility. Can we cut all the ones that aren't necessary and use -v0 for all of them? Specifically, I think all these might be unnecessary: https://github.com/nadavbh12/gym/blob/4918e31158772b1251921b7308d4d3f397b37270/gym/envs/__init__.py#L363-L411

Our Travis check is failing because we don't install the dependency. I'll try to add that soon.

nadavbh12:
Thanks Trevor!

I wasn't sure what the version were. I've removed them in the latest commit.

Travis is failing since rle_python_interface requires some manual installation.
I'm not very experienced with Python distutils and was having a hard time incorporating the cmake installation with it so it will install automatically.
Is being able to install with 'pip install' strictly required?

joschu:
This is great! Currently it's still too hard to install RLE and its dependencies. Could you please provide a script that downloads and installs the relevant parts of libretro to build the SNES core, along with at least one ROM (so people can get something running right away?) If you don't want to risk posting the nintendo ROMs on github, you could probably find a public-domain ROM.

nadavbh12:
Hey John,
The CMake file downloads and compiles our (now 3) supported emulators: Stella, SNES9x and Genesis-Plus-Gx.
I'm currently working on calling installing them along with RLE using distUtils, however, integration with CMake is quite horrible.
I can provide a simple shell script which simply calls CMake, setup.py and copies the core files to known locations. Will that suffice?
nice idea with the public domain ROM. We'll add support for one for ease of use.

joschu:
Ah OK, didn't notice the CMake file did all of that. I remember it was a bit of a pain to get Atari to work properly with pip. Let me consult with my colleagues (who know more about python installation) about this, and I'll also try installing RLE.

nadavbh12:
Great. Note though that my attempts to easily install RLE might have broken the "standard" installation, so it's better to checkout v1.0.2.
The latest version can be installed from Pypi using `pip install rle_python_interface`.
Though I haven't figured out yet why the cores don't install with it.

nadavbh12:
I updated RLE so you can install via pypi and added a ROM of the game Classic Kong.
Not sure what's failing in travis. Please have a look.

nadavbh12:
@joschu, while on my computer `pip install -e '.[all]'` installs RLE, it doesn't seem to do so [on travis](https://travis-ci.org/openai/gym/builds/204576253#L393).
Could you have a look and see why's this?

nadavbh12:
@joschu , @tlbtlbtlb ,
Hopefully the branch should be working now correctly using both Python 2 & 3, using pip install (classic kong rom included).
Not sure why travis won't run this. Could you give it a look?

joschu:
Hi Nadav, we probably won't be merging in other environments like this -- we're realizing that it makes sense to host them in separate repositories so their authors can more easily develop them. We'll make sure to advertise your environment prominently. Sorry if you've had to waste time trying to get your code to run in our CI.

John

tlbtlbtlb:
See https://github.com/openai/gym/tree/master/gym/envs#how-to-create-new-environments-for-gym for guidelines on structuring add-on packages for gym. So we'll hope you'll make a repo called gym-rle or gym-snes that, when imported, registers the environments.

nadavbh12:
Thanks. Well, that actually makes sense.
No harm done since your CI helped me find a few bugs and from what is seems integrating environments in both methods highly overlap.
Will open the new repository soon.

nadavbh12:
@tlbtlbtlb,
I created a [gym-rle repo](https://github.com/nadavbh12/gym-rle). 
Hopefully, once it's up and running we can add more interesting challenges :-)

Issue #471, opened 2017-01-24
Go env crashes from custom adverasry policy


partobs-mdp:
Current version of OpenAI Gym doesn't support creating Go environments with custom adversary policies. The code that illustrates this:
```python
from gym.envs.board_game import go
import random


def my_policy(curr_state, prev_state, prev_action):
    """
    My own random policy function.
    """
    avail_moves = list(map(lambda x: go._coord_to_action(curr_state.board, x), curr_state.board.get_legal_coords(curr_state.color)))
    return random.choice(avail_moves)


# Creating GoEnv, passing my_policy as opponent.
env = go.GoEnv('black', my_policy, 'image3c', 'raise', 9)
# And this instruction crashes.
env.reset()

```

This is an important issue, because it prevents (for example) training AlphaGo-like bot by self-play in OpenAI.

tlbtlbtlb:
You can work around this by overriding the `_reset_opponent` method. So
```
env = go.GoEnv('black', 'custom', 'image3c', 'raise', 9)
def my_reset_opponent(self, board):
    self.opponent_policy = my_policy
env._reset_opponent = type(env.__class__._reset_opponent)(my_reset_opponent, env, env.__class__)
```
Please don't submit results to the scoreboard against a non-standard opponent.

PR #472, opened 2017-01-24
Go environment bugfix

partobs-mdp:
- Fixed issue #471.
- Refactored `GoEnv.__init__()` (added defaults, removed one unnecessary parameter).
- Fixed bug in _play function (essentially, a syntax error).

### DIFF
--- a/gym/envs/board_game/go.py
+++ b/gym/envs/board_game/go.py
@@ -105,16 +105,15 @@ def _play(black_policy_fn, white_policy_fn, board_size=19):
     moves = []
 
     prev_state, prev_action = None, None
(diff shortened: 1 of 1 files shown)


tlbtlbtlb:
If you're removing the `observation_type`, parameter, please also remove it in gym/envs/__init__.py. Currently, `gym.make('Go9x9-v0')` fails with
```
  File "/usr/local/gym/gym/envs/tests/test_envs.py", line 47, in test_env
    env = spec.make()
  File "/usr/local/gym/gym/envs/registration.py", line 70, in make
    env = cls(**self._kwargs)
TypeError: __init__() got an unexpected keyword argument 'observation_type'
```

PR #475, opened 2017-01-30
Fix multi-discrete sampling 

alito:
The current sampling from the MultiDiscrete space is non-uniform for any sub-space larger than two values. The code rounds the values from a float ranging between the sub-space extremes, which means that the end points only get half the space of the internal points.

eg for a space between 0 and 2, all values between 0 and 0.5 will go to 0, all the ones between 0.5 and 1.5 will map to 1 and all between 1.5 and 2 will map to 2, which means 1 gets twice the "hits" of either 0 or 2.

This one-line change fixes it by grabbing the floor of the range + 1.

Testing. Before:
```
In [4]: from gym.spaces import MultiDiscrete

In [5]: space = MultiDiscrete([[0,2]])

In [6]: from collections import defaultdict

In [7]: count = defaultdict(int)

In [8]: for r in range(1000000):
   ...:     sample = space.sample()
   ...:     count[sample[0]] += 1
   ...:     

In [9]: count
Out[9]: defaultdict(<type 'int'>, {0: 249643, 1: 499561, 2: 250796})

```
After:

```
n [1]: from gym.spaces import MultiDiscrete

In [2]: from collections import defaultdict

In [3]: space = MultiDiscrete([[0,2]])

In [4]: count = defaultdict(int)

In [5]: for r in range(1000000):
   ...:     sample = space.sample()
   ...:     count[sample[0]] += 1
   ...:     

In [6]: count
Out[6]: defaultdict(<type 'int'>, {0: 332802, 1: 333425, 2: 333773})
```

### DIFF
--- a/gym/spaces/multi_discrete.py
+++ b/gym/spaces/multi_discrete.py
@@ -35,7 +35,7 @@ class MultiDiscrete(gym.Space):
         """ Returns a array with one sample from each discrete action space """
         # For each row: round(random .* (max - min) + min, 0)
         random_array = prng.np_random.rand(self.num_discrete_space)
-        return [int(x) for x in np.rint(np.multiply((self.high - self.low), random_array) + self.low)]
+        return [int(x) for x in np.floor(np.multiply((self.high - self.low + 1.), random_array) + self.low)]
     def contains(self, x):
         return len(x) == self.num_discrete_space and (np.array(x) >= self.low).all() and (np.array(x) <= self.high).all()
 


tlbtlbtlb:
Yikes! Thanks for finding this!

@ppaquette, can you double-check this?

ppaquette:
Good catch. I'm OK with the proposed fix.

tlbtlbtlb:
Committed, thanks @alito.

PR #481, opened 2017-02-01
WIP - Time limit refactor

nottombrown:
- Default to autowrapping all non-universe envs in TimeLimit wrapper
- Add `max_episode_steps`
- Include `timestep_limit` for backwards compat
- Add `play` for human debugging


gym/envs/registration.py
```diff
@@ -102,7 +114,15 @@ def __init__(self):
     def make(self, id):
         logger.info('Making new env: %s', id)
         spec = self.spec(id)
-        return spec.make()
+        env = spec.make()
+
+        # Automatically wrap envs with a TimeLimit wrapper, because the limit is part of the env semantics
+        # Omit vnc envs for now, because autowrapping breaks `env.configure`
+        # We will add TimeLimit to all envs after we remove `configure` from the env API
+        if not env.spec.tags.get('vnc'):
```
nottombrown: My biggest question in this PR is here. @joschu, would it work for you to autowrap all non-vnc envs for now, and then autowrap the VNC envs once we deprecate `env.configure`?

gym/envs/registration.py
```diff
@@ -49,7 +48,22 @@ def __init__(self, id, entry_point=None, trials=100, reward_threshold=None, loca
...
+            max_episode_steps = tags.get('wrapper_config.TimeLimit.max_episode_steps')
+            warnings.warn("tags['wrapper_config.TimeLimit.max_episode_steps'] is deprecated. Use max_episode_steps argument.")
+        tags['wrapper_config.TimeLimit.max_episode_steps'] = max_episode_steps
+        ######
+
+        # BACKWARDS COMPAT 2017/1/31
+        if timestep_limit is not None:
+            max_episode_steps = timestep_limit
+            # TODO: Add the following deprecation warning after 2017/03/01
+            # warnings.warn("register(timestep_limit={}) is deprecated. Use register(max_episode_steps={}) instead.".format(timestep_limit, timestep_limit))
```
jietang: *sigh* so many backwards compats

jietang:
lgtm pending response from joschu

joschu:
How does this branch related to unstable?

joschu:
Doesn't look like it has my monitor simplification commits.

joschu:
It should be pretty easy to fix the universe monitor so it works with those commits.

nottombrown:
I'd like to merge this in safely before doing the Monitor refactor @joschu. Is there a reason that you'd like to combine the two PRs?

joschu:
We could do that, but I've been using unstable for a while and it's working fine. 
I'm afraid these changes will just be lost if we don't put them in now.

joschu:
Another unrelated comment: DEFAULT_MAX_EPISODE_SECONDS should be infinity

joschu:
Also, why are we applying TimeLimit wrapper if there's no time limit specified in the registration?

joschu:
I'm against having the Monitor apply the TimeLImit -- that's surprising.

joschu:
Makes more sense to have `make` apply the TimeLimit if it's needed, as is done currently in `unstable`.

joschu:
Can we just apply your renaming changes (timestep_limit->max_episode_steps) on unstable, and make master=unstable? And universe can be tied to a version of gym where it's not broken.

joschu:
We can then quickly fix gym:master so universe works.

nottombrown:
## 1
I'll cherry-pick out your refactor commits and make them into a separate PR so that we can merge them in after we get them working with `universe`

## 2
I'm not sure I follow this, it seems like `DEFAULT_MAX_EPISODE_SECONDS` of infinity is what we currently have here in `gym`?

## 3
The `TimeLimit` wrapper already checks whether or not there are relevant limits that it should enforce, so it's a bit redundant to also check in the `registration` call. Would prefer not to double check, but don't have a very strong opinion on that.

gym/wrappers/monitoring.py
```diff
@@ -41,11 +40,20 @@ def set_monitor_mode(self, mode):
...
+
 def Monitor(env=None, directory=None, video_callable=None, force=False, resume=False,
             write_upon_reset=False, uid=None, mode=None):
     if not isinstance(env, gym.Env):
         raise error.Error("Monitor decorator syntax is deprecated as of 12/28/2016. Replace your call to `env = gym.wrappers.Monitor(directory)(env)` with `env = gym.wrappers.Monitor(env, directory)`")
 
     # TODO: add duration in seconds also
-    return _Monitor(TimeLimit(env, max_episode_steps=env.spec.timestep_limit), directory, video_callable, force, resume,
-                    write_upon_reset, uid, mode)
+    return _Monitor(env, directory, video_callable, force, resume, write_upon_reset, uid, mode)
```
nottombrown: @joschu I agree that having the `Monitor` apply the `TimeLimit` is surprising, so it is removed here.

joschu:
2: 
if self._max_episode_seconds is None and self._max_episode_steps is None:
            self._max_episode_seconds = DEFAULT_MAX_EPISODE_SECONDS
So it currently sets a 20 minute limit if neither of these are set. I actually think we should just have two different classes, one for time limit, and one for timestep limit.

3: In `unstable`, there's no `configure` in the TimeLimit wrapper, there are just arguments mx_episode_steps and max_episodes_seconds -- I think that's how it should be.

nottombrown:
@joschu I'd strongly prefer not to have `gym:master` ever be broken with respect to `universe:master` if we can avoid it. It seems like doing that will make `universe` unusable for the majority-case user.

joschu:
Then why don't we just fix universe:unstable + gym:master?

nottombrown:
## 2
Which line are you looking at @joschu when I search `gym` for `DEFAULT_MAX_EPISODE_SECONDS`, I see no results

joschu:
2: hmm, maybe it's only in unstable. I'm looking at time_limit.py in gym.

I can probably just fix the universe tests in a few minutes. Let me just do that.

joschu:
Oops, my bad, DEFAULT_MAX_EPISODE_SECONDS is actually in universe's time_limit.py, which is fine.

nottombrown:
Ok, my intuition is that it will take quite a bit of effort to fix the vectorized universe `Monitor` to work with the refactored stuff on `gym:unstable`. That's why I pulled them apart into two PRs. If it turns out to be easy, then we can ditch this PR and just merge in `unstable` directly.

It's past my bedtime though, so happy to talk more in the morning.

Thanks for doing the work to get the API looking nice and clean btw @joschu !

joschu:
Got tests passing on my mac
https://github.com/openai/universe/compare/master...fix-monitor-to-work-with-gym-unstable

joschu:
Totally understand the concern about breaking gym:master+universe:master. I should've fixed it before pushing to unstable. 

Apologies for causing delays on your PR and requiring more commit reshuffling, I'm a bit neurotic about making sure gym is nice and clean :P

nottombrown:
Nice work @joschu, I like your version better. I rebased it and have a PR up at #482.

Closing this now.

PR #486, opened 2017-02-03
allow partial writes in example script

MichaelBlume:


pzhokhov:
both scoreboard and benchmark_runner are removed upstream; closing this

PR #498, opened 2017-02-15
support for mujoco pixel

damodei:


Breakend:
Any reason not to merge this into the master branch?

danijar:
I think these environments are a good addition to Gym. There are already papers reporting results on performing MuJoCo tasks from pixels.

tlbtlbtlb:
@damodei, I think this is useful. Let's merge unless someone objects.

damodei:
Initially, we were going to wait to merge until someone (it was going to be
me) actually used it a few times.  I didn't end up doing so.  I believe
it's correct, so no objections to merging, but it hasn't been very well
tested.

pzhokhov:
I am inclined to close this based on the discussion in a similar PR https://github.com/openai/gym/pull/740
(current PR is nicer and has less duplicated code, but I am still in favor of a pixel observation wrapper that can be applied to other environments as well) @damodei feel free to reopen / argue back.

PR #506, opened 2017-02-21
Suit ConvergenceControl-v0 to tensorflow backend.

XericZephyr:
At present, the `ConvergenceControl-v0` environment does not support tensorflow backend. 

Mostly because the image dimension orders are different between tensorflow and theano.

This PR try to fix that with tensorflow backend.

### DIFF
--- a/gym/envs/parameter_tuning/convergence.py
+++ b/gym/envs/parameter_tuning/convergence.py
@@ -3,6 +3,7 @@ import gym
 import random
 from gym import spaces
 import numpy as np
+import keras
 from keras.datasets import cifar10, mnist, cifar100
 from keras.models import Sequential
 from keras.layers import Dense, Dropout, Activation, Flatten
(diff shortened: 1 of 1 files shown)


tlbtlbtlb:
Does this break using Theano as a backend? Do we want to support Theano here?

XericZephyr:
@tlbtlbtlb  Do we have any unit test for it? In principle, it should work with Theano.

tlbtlbtlb:
@iaroslav-ai This is your code, do you know what you want to do about Theano/Tensorflow here?

XericZephyr:
@iaroslav-ai @tlbtlbtlb Also let me know if there are additional unit/lint tests required.

iaroslav-ai:
@XericZephyr Thanks for the PR! Can confirm that that it runs fine using both theano and tensorflow.
As for a test, I could imagine something like this could work:
Make a file "test.py":
```
import gym

env = gym.make('ConvergenceControl-v0')

for episode in range(3):
    o = env.reset()
    for step in range(7):
        env.render()
```
Make a bash file "run.sh":
```
THEANO_FLAGS=device=gpu KERAS_BACKEND=tensorflow python test.py
THEANO_FLAGS=device=gpu KERAS_BACKEND=theano python test.py
```
If run.sh completes successfully, then the code works fine. This assumes that the code is tested on Ubuntu. 
It would probably be good to have such test around. But one could also just say that only tensorflow is (officially) supported and target tensorflow only. 
@tlbtlbtlb the code is compatible with both Theano and Tensorflow I propose to keep it as it is.

tlbtlbtlb:
Ideally, the test should determine that they get similar answers from tf and Theano. We do this for most environments: see gym/gym/envs/tests/test_envs_semantics.py.

XericZephyr:
@iaroslav-ai  Thanks for the confirmation. Your original env looks interesting and charming. I am looking forward to testing it on my end.

@tlbtlbtlb After looking into the code, I am still bit confused on how to use the pytest. (I used to be heavily leaned on `nosetests`. ) Do you have any example on how to produce those tests for specific env? Thanks.

XericZephyr:
Any update comment on this PR? @tlbtlbtlb Thanks.

XericZephyr:
Closed for now due to inactivity.

PR #510, opened 2017-02-24
Pyglet performance improvements + display tweaks [WIP]

louiehelm:
1) The gym display process currently consumes ~40% CPU when visualising a PongDeterministic-v3 agent via gym's pyglet interface on a modern desktop.

Gym is currently blit'ing its own frames to pyglet but not doing any vsync / framebuffer limiting. In the long run, gym should probably pass frames asyncronously to a fully managed pyglet window. Or use a different gl lib.

For now, I got a solid 85 --> 120 fps training rate boost (per thread) simply by patching the pyglet code. I've seen other pyglet developers use similar approaches when they're buffering their own frames.

This corresponds to a more manageable 14% performance penalty to display agents (vs ~40%). I'm sure the ratio is different on different machines. Highest boosts will come on fastests machines since that's where the current code is burning up the most CPU by rendering 120+ fps, even when no monitors allow that and no ones eyes can see it.

2) Double Zoom - Atari windows are very small on 4k / 5k / Retina montors. Uses scipy to quickly double the output size with minimal overhead.  ~10% performance hit vs no resizing.

3) Window Positioning - Spawn display windows at known screen coordinates. **

4) File-only framebuffer - Dumps framebuffer to ramdisk so it can be drawn with something besides pyglet. Linux only right now but I could likely make it work with Mac too if there's interest. The main reason I use this is higher performance. But also so I can peak at video of my agents directly in the terminal.

atari_terminal

This code is a WIP mock up for demonstrating what I'm actually using. But if you want these performance boosts + features integrated, let me know how best to harmonize the style to OpenAI's standards.

[*] I'm using 24 frames per second since that's the 22mm film standard.

Other principaled choices include 25 (PAL), 30 (NTSC) or 60 (VSYNC) FPS. Almost all monitors or video card drivers (or both) are vsync limited at 60Hz. Displaying beyond that is torching CPU.

I know that film is motion blurred and double-buffered so it doesn't suffer from lower frame rates as much as crisp video game scenes do. But I still settled on 24 since the visual difference from 24 -> 60, is pretty negligible IMO compared to a 20% CPU performance penalty. I would guess that most people would want higher training rates. Anyone who needs to do a CES demo or something can come tune this parameter.

[**] Note about window positioning: I have uncommitted code that increments window positioning when drawing multiple displays. It puts them at fixed, evenly spaced intervals. However, it's designed to conform to my spicific i3-gaps workspaces and monitor layout. I could try to make it more general if you want this sort of feature? This s actually the reason I was using dirty global variables to begin with. So the style could instanly be improved without even creating multiple variable blocks if this feature isn't desired.

### DIFF
--- a/gym/envs/classic_control/rendering.py
+++ b/gym/envs/classic_control/rendering.py
@@ -5,6 +5,8 @@ from __future__ import division
 import os
 import six
 import sys
+import scipy.ndimage
+from time import time
 
 if "Apple" in sys.version:
     if 'DYLD_FALLBACK_LIBRARY_PATH' in os.environ:
@@ -305,25 +307,55 @@ class Image(Geom):
 
 # ================================================================
 
+
+
+
+MAX_FPS = 24
+RATELIMIT_FPS = 1 / MAX_FPS
+DOUBLE_SIZE = False  # Good for Retina / HiDPI monitors
+FILE_ONLY = False
+SET_POSITION = False
+POSITION_X = 100
+POSITION_Y = 200
+
 class SimpleImageViewer(object):
     def __init__(self, display=None):
         self.window = None
         self.isopen = False
         self.display = display
+        self.last = 0
+
     def imshow(self, arr):
-        if self.window is None:
+        if not FILE_ONLY and self.window is None:
             height, width, channels = arr.shape
-            self.window = pyglet.window.Window(width=width, height=height, display=self.display)
+            if DOUBLE_SIZE:
+                self.window = pyglet.window.Window(width=2*width, height=2*height, display=self.display)
+            else:
+                self.window = pyglet.window.Window(width=width, height=height, display=self.display)
+            if SET_POSITION:
+                self.window.set_location(POSITION_X,POSITION_Y)
             self.width = width
             self.height = height
             self.isopen = True
-        assert arr.shape == (self.height, self.width, 3), "You passed in an image with the wrong number shape"
-        image = pyglet.image.ImageData(self.width, self.height, 'RGB', arr.tobytes(), pitch=self.width * -3)
-        self.window.clear()
-        self.window.switch_to()
-        self.window.dispatch_events()
-        image.blit(0,0)
-        self.window.flip()
+
+        if not FILE_ONLY and time() - self.last > RATELIMIT_FPS:
+            assert arr.shape == (self.height, self.width, 3), "You passed in an image with the wrong number shape"
+            if DOUBLE_SIZE:
+                image = pyglet.image.ImageData(2*self.width, 2*self.height, 'RGB', scipy.ndimage.interpolation.zoom(arr, (2,2,1), order=0).tobytes(), pitch=self.width * -6)
+            else:
+                image = pyglet.image.ImageData(self.width, self.height, 'RGB', arr.tobytes(), pitch=self.width * -3)
+            scipy.misc.imsave('/tmp/gym.png', arr)
+            self.window.clear()
+            self.window.switch_to()
+            self.window.dispatch_events()
+            image.blit(0,0)
+            self.window.flip()
+            self.last = time()
+
+        if FILE_ONLY and time() - self.last > RATELIMIT_FPS:
+            scipy.misc.imsave('/tmp/gym.png', arr)
+            self.last = time()
+
     def close(self):
         if self.isopen:
             self.window.close()


joschu:
Closing, because I don't think rendering speed really matters, and people can easily resize the images for their own favorite visualizer, e.g. cv2.imshow.

louiehelm:
I completely understand if you don't want my resizing or exporting functions. But the main feature here improves training speed by 10-40%, not rendering speed. I may not have made that clear since both can be expressed in FPS. I don't care about better rendering speed either -- I want better training rates.

Would you consider accepting a 2 line patch that only does the 40% training speedup?

olegklimov:
> simply by patching the pyglet code

Is it inside Pyglet code?

louiehelm:
I've pared down my branch to only include the performance boosting part I would like to see pulled:
https://github.com/openai/gym/compare/master...louiehelm:master

This isn't modifying anything inside pyglet. But it is mimicking the most important part of what pyglet would do if you let it manage your framebuffer instead of updating the window contents yourself.

I'm seeing per thread training rate increases of 125% on the most basic experiments I can do. The high performance increase comes from being on as especially fast system. But that's why you should want a patch like this. Visualisation currently costs O(n log n). This patch makes visualisation O(1).

olegklimov:
This code:

```python
        if time() - self.last > 0.04:
```

looks to me as you skip frames, force less than 25 fps for SimpleImageViewer. Why would you want that?

To reiterate, this looks to me like frame skipping, not improved render speed.

louiehelm:
This patch provides a framebuffer for the gym code -- something that's currently missing. It prevents pyglet from it's current behavior of using unbounded CPU trying to draw non-displayable frames.

It "skips frames" the same way an mp3 encoder "drops audio".

Most of this is a hardware issue. Monitors can't display more than 60fps. When they claim to do more, they're essentially lying. They simply have modes that allow the entire monitor to flicker and "tear" by re-drawing parts of scheduled frames on top of other undrawn or partially drawn frames. People hate this visually so monitor companies and developers who write video card drivers and even OS developers all take measures to prevent this from actually happening.

People who think they're watching >60fps on their displays are almost always mistaken. Yes, the FPS in counterstrike may say 150fps. And yes, those 150 frames are all rendered and push around memory. But most of those frames are dropped and go unseen because several hardware and software layers apply frame limiting before they ever reach the display.


Consider the random agent from https://gym.openai.com/ playing a gym atari environment (the example code is technically broken BTW since Taxi-v1 doesn't exist anymore):


Current code w/ vsync: Agent can only perform 60 actions / sec [!]. The OS forcefully hangs the entire gym process after it tries to draw it's 61st frame in a second. Agent+GYM: <1% CPU, Vis: 30% CPU, 70% CPU unused.

Current code with no vsync ( less common driver / OS setup ): Agent performs 200 actions / sec. Agent+GYM: 5% CPU, Vis: 95% CPU.

Framebuffering patch (reguardless of vsync settings): Agent can perform 1500 actions per second. Agent+GYM: 85% Vis: 15% CPU.



And just to be clear, this isn't "skipping frames" in the way Deep Mind might refer to an Atari "frame skip" where the agent is waiting between frames of the game to submit actions. This patch does nothing to impact the internal workings of the gym environments themselves. It just's only "skipping" the processing of frames that can be known in advance to be undisplayable (or in the case of the frames between 25-60 fps, indistinguishable visually).

olegklimov:
Can you reach the same effect in your code by calling `render('human')` less often?

louiehelm:
Yes. But then we're passing the buck to every gym / universe developer to write their own framebuffers.

I've never seen another developer on github implement any framebuffering in their code that calls gym or universe. So there's evidence that most, if not all, developers aren't fixing this issue in their own code (even though they technically could). I don't think we should simply lament that nearly every developer uses render() inefficiently when we could easily add basic sanity checks to gym to prevent folks from making these non-sensical tradeoffs.

Developers aren't benefitting from their current "freedom" they have to waste 30-95% of their CPU pushing extra frames around memory that are never displayed. They just don't realize this is what's happening.

olegklimov:
Umm... I think you have a misconception of what gym users do. They learn agents using RL. That doesn't involve calling `render('human')` more often or less often, it is done without calling it **at all**.

The only scenario it is useful: learned agent demo. Sure you may think of other uses, but it is far from "every developer".

louiehelm:
I agree that most researchers aren't using render() at all. And only the developers who actually call render() suffer from its current behaviour. But those aren't reasons to not fix it for those who do call it.

Probably every researcher who uses gym at least experiments with render() for a while while learning how gym works. There's a reason OpenAI calls render() in all their example code. And even experienced researchers will want to glance at certain runs while working on new algorithms.

I'm currently tuning a new Nadam optimizer I added to TF. I visualise 1 of the 8 agents I'm running on about 25% my test runs. So there are in-between cases of researchers and other developers who will want to watch their agents while training sometimes.

Newcomers who are just getting started are probably even more likely to use render() than people like me who are sporadically rendering agents or displaying saved models. OpenAI's own example code for gym calls render 1-for-1 with each action. Everyone learning how gym works will call it this exact way while getting started. Same thing for universe and universe starter agent since OpenAI's example code does the same kind of render() calling in those too. So why not make the most straightforward way to call render provide good, stable performance?

New developers are destined to call render() this way. And they are also the least likely to realize they are sacrificing 40% of their CPUs by doing so. Why not handle this internally for them since it costs nothing, requires no new dependencies, avoids needless computation that goes completly unused, and improves performance for everyone using render(), without breaking any interfaces or causing downsides for anyone?

olegklimov:
OK so why new developers would need faster than realtime? I think they need normal speed with vsync, to see the game works as expected.

What if someone wants normal behavior: call render(), have render? Why introduce undefined behavior?

Primary use case is running a trained agent. How code will look like, `render(); sleep()`? Rings the bell for me... Sleep is already there, it's vsync.

louiehelm:
I agree that beginners prob want their agents to run in realtime. And a side-effect of providing the sort of interface protection that I'm suggesting is that default performance of OpenAI reference code would be much faster than realtime. But I don't think that slowness should be enforced by flooding a physical interface (the display) with more data than it can handle and counting on 2 different side-effects to slow it down, either (1) 60fps via forced thread locking by the system video driver (vsync) or (2) 80-200fps via heavy CPU/mem spinning from choaking the display buffer with undisplayable frames that it will drop anyway (no vsync). 

Providing basic interface protection and adding an explicit `sleep()` call to the gym reference example would be a much more principled and reliable way to get the behavior you're describing.

That setup would more consistently give beginners real-time visualisation since it doesn't depend on them having vsync enabled. Vsync is usually on, but it varies a lot by video card / OS. Do we really want to explain vsync settings to newbies on github or IRC? Or just have the reference code work for everyone and not depend on them having drivers that slow it down?

And when new researchers + developers inevitably want to train faster than 60 actions per second, removing the sleep call will be the obvious way to do that. That's a great training wheels --> intermediate performance transition that allows them to keep using render() a bit longer while still becoming familiar with gym.

olegklimov:
> Vsync is usually on, but it varies a lot by video card

I believe it's not. Vsync option is available in Pyglet window constructor, and it's on by default. Please try vsync=False in render.py.

To switch to high speed, it is sufficient to call render('human') less, or not call it at all. Pretty straightforward, I can't see the problem.

louiehelm:
OK. I'll work on something else.

PR #514, opened 2017-02-26
remove configure, simplify core.py

joschu:
Made some simplifications to core.py.
I'd like to move all of the configure() logic into the Universe Env, but I can't get any of the Universe tests to pass on my machine.


joschu:
@nottombrown could you please review?

PR #531, opened 2017-03-14
Rename scoring methods

jietang:
WIP, goes with scoreboard-gym:tom-handoff-rename


jietang:
@nottombrown work remaining here is to remove all the scoring rules and the benchmark.scorer attribute in the spec (no backwards compatibility issues since I don't think external folks are using it) 

Most of the scoring tests will be irrelevant and can be removed - it should be straightforward but I'm happy to pair w/ you later in the week.

pzhokhov:
benchmarks / scoreboard got deleted upstream; closing this

PR #533, opened 2017-03-14
Pass along kwargs in reset

strin:
TL;DR extend reset to be `reset(**kwargs)` so we can have reset the environment to a given configuration.

See PR: https://github.com/openai/universe/pull/122

@joschu

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,8 @@ tutorial/*.html
 # PyCharm project files
 .idea
 vizdoom.ini
+
+# vim
+.ropeproject
+tags
+*.swo
--- a/gym/core.py
+++ b/gym/core.py
@@ -110,9 +110,12 @@ class Env(object):
         observation, reward, done, info = self._step(action)
         return observation, reward, done, info
 
-    def reset(self):
+    def reset(self, **kwargs):
         """Resets the state of the environment and returns an initial
         observation. Will call 'configure()' if not already called.
+
+        Args:
+            kwargs: forward episode-specific config to the environment.
 
         Returns: observation (object): the initial observation of the
             space. (Initial reward is assumed to be 0.)
@@ -120,7 +123,7 @@ class Env(object):
         if self.metadata.get('configure.required') and not self._configured:
             logger.warning("Called reset on %s before configuring. Configuring automatically with default arguments", self)
             self.configure()
-        observation = self._reset()
+        observation = self._reset(**kwargs)
         return observation
 
     def render(self, mode='human', close=False):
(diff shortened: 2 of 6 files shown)


gdb:
Ah, so this won't work (everyone overrides `_reset()`) themselves, and we don't want `**kwargs` in the official `reset()` API — but fortunately this isn't needed. Instead, the idea is to just have people start overriding `reset()` directly rather than `_reset()`, and then they can choose to support `reset(**kwargs)` themselves.

strin:
True. Having `reset(**kwargs)` is still backward compatible though. I am worried about inconsistency in style, i.e. overwriting `_reset` vs `reset`.

gdb:
Surprisingly, there's no actual backwards compatibility issue — we should still *support* overriding `_reset()` but *recommend* overriding `reset()`.

The Gym repo itself shouldn't recommend the `**kwargs` feature; only envs which need it should implement.

strin:
We may want to recommend `**kwargs` feature in wrappers, right? Since we want those args to be passed along, we need `**kwargs` to be present in all wrappers.

tlbtlbtlb:
Yes, all wrapper reset functions will need to pass kwargs, such as https://github.com/openai/universe/blob/a7944e65446a5484e5011f048558f312e907da61/universe/wrappers/timer.py#L13.

gdb:
This will be reasonable for some envs, but it's technically a violation of the core semantics (that an env is specified by only its name). So I'd rather not have this appear or be encouraged in the core if we can.

You're right that wrappers which aren't aware of kwargs won't nicely wrap those which do. But users can still get by, such as by maintaining a pointer to the section of the env which *does* support kwargs. They can also fork whatever wrappers they're using, or potentially get by with subclassing them — neither is particularly fun but in practice I think it's ok for an extension feature like this.

strin:
Let's discuss the core semantic. I think env should specified only by its name, but it's useful to reset with different configurations. For example, imagine a maze environment, and we want to reset to a specific starting point, i.e. `reset(start=(x,y))`. Or a game that relies on random seeds, so we can do for example `reset(seed=42)`. 

It will be cleaner in code if we can support `reset(**kwargs)` as first class citizens. So we don't have to fork different versions of wrappers, or use different styles such as `reset` vs `_reset`.

tlbtlbtlb:
The argument for reset not taking args in nearly all cases is that we want to be able to compare agents based on their score on a named env. If agents can choose different versions of the env with reset arguments, they're not comparable any more. In the maze example, performance with a specified starting point isn't the same as a random one. Allowing agents to set random seeds seems like it would enable the worst kind of cheating.

strin:
Agree. We should not change the semantic of reset to take args, due to the fact that scoreboard uses `env_id` only for fair comparison. Per our discussion, the solution would be to keep the same `reset` semantic and recommend to overwrite `reset` (not `_reset`) in the case of `kwargs`. 

Side note: for environments that overwrite `reset(**kwargs)`, if we want to evaluate the agent's average score on a particular set of `kwargs`, the correct thing to do is to create a wrapper that specifies the schedule of `kwargs`. Then use that wrapper on scoreboard.

PR #536, opened 2017-03-18
Expand the Kelly Coinflip game

gwern:


gym/envs/toy_text/kellycoinflip.py
```diff
@@ -93,7 +93,7 @@ def __init__(self, initialWealth=25, edgePriorAlpha=7, edgePriorBeta=3, maxWealt
         self.maxRounds = maxRounds
         self.rounds = self.maxRounds
         self.maxWealth = maxWealth
-        self._seed()
+        if reseed: self._seed()
```
tlbtlbtlb: Won't this crash in `step` if `_seed` isn't called in `__init__`? Otherwise, `self.np_random` is undefined.

gwern: Yes, but why would anyone call `__init__(reseed=False)` if they're not `_reset()`? I guess it could try to check whether `np_random` exists & reseed if it doesn't, just in case someone does that for some reason.

gwern: So to do a check, you could change the `__init__` line to 

```.python
if reseed or not hasattr(self, 'np_random') : self._seed()
```

I think that would work, if I've understood what `hasattr` does right.

gwern:
(I don't know why the build failed, it looks like some sort of dependency download problem caused it to error out, but the code should still pass since it was a one-liner and works locally.)

gym/envs/toy_text/kellycoinflip.py
gwern: Checking, I've realized there's another way users could be confused/misused it: if you set the hyperparameters in init and then do resets, you'll get a new game with parameters drawn according to the default hyperparameters, but you probably expected the parameters to be drawn according to your original hyperparameter options. I don't know any better way to program this in Python but to store all the overriden options in the env object and just pass them into init on reset.

gwern:
Any further comments on the generalized version of the game?

gwern:
Thanks!

PR #552, opened 2017-04-05
Add render-only wrapper

machinaut:


tlbtlbtlb:
I'd rather leave all this `.render` calling `._render` bureaucracy behind, and just have
```
class MyWrapper(gym.Wrapper)
    def render(...)
        ...
```

pzhokhov:
should be obsolete now that there is only `render` method (no `_render`) in the API, closing this

PR #561, opened 2017-04-17
Allow Discrete spaces to be iterable

allanbreyes:
Allows discrete spaces to be iterable, so the following can be accomplished:

```python
for action in env.action_space:
    print action
```

### DIFF
--- a/gym/spaces/discrete.py
+++ b/gym/spaces/discrete.py
@@ -22,6 +22,13 @@ class Discrete(gym.Space):
         else:
             return False
         return as_int >= 0 and as_int < self.n
+    def __len__(self):
+        return self.n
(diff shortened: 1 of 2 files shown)


allanbreyes:
@andrewschreiber are you accepting contributions? Should I close this? I'd _love_ to help out, but not really sure where to... the ~3 `help wanted` tags seem a tad outdated. Help me help you?

tlbtlbtlb:
Thanks for the contribution, but I'd rather keep the Spaces as simple as possible and just require people to do:
```
    for action in range(env.action_space.n):
```
if they need to enumerate for some reason.

allanbreyes:
Makes sense. Thanks!

PR #563, opened 2017-04-20
Add Dict space for access by keys

longjie:
This Dict space defines a dict-like Space. It should be useful to bridge gym to other system. I need to modify the other space a bit to create JSON representation recursively.

Example:
```
> d=Dict(('button_a', Discrete(2)), ('axes', Box(0,1,2)))
> s=d.sample()
> s
OrderedDict([('button_a', 1), ('axes', array([ 0.38438171,  0.29753461]))])
> d.contains(s)
True			  
> s['button_a']
1
> s['axes']
array([ 0.38438171,  0.29753461])
> np.concatenate([x if isinstance(x,np.ndarray) else np.array([x]) for x in s.values()])
array([ 1.        ,  0.38438171,  0.29753461])
```
I use OrderedDict to preserve the order of elements. So we need to add 'object_pairs_hook=OrderedDict' argments for json.load function when use this Dict space.

- add dict_space.py
- add to_one_jsonable() and from_one_jsonable() for other space,
  need for recursive representation
- add test case in test_spaces.py
- add test for Dict in test_dict.py

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -241,6 +241,16 @@ class Space(object):
         # By default, assume identity is JSONable
         return sample_n
 
+    def to_one_jsonable(self, sample):
+        """Convert a sample from this space to a JSONable data type."""
+        # By default, assume identity is JSONable
+        return sample
+
+    def from_one_jsonable(self, sample):
+        """Convert a JSONable data type to a sample from this space."""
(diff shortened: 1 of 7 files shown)


pzhokhov:
added upstream, closing this

Issue #402, opened 2016-11-13
Save and Restore Environment State


ricklove:
Is it possible to save the environment state (for example every 10 steps) so that the environment state could be restored to the state.

This could be useful for retrying alternative actions without having to replay the preceding environment states that lead to that state. 

This would be especially useful for learning from the state space that preceded failure to focus on learning actions that avoid failure in risky states.

It seems most reinforcement learning systems always restart the environment state upon failure, but it would be much better to restore to a state n steps before failure to learn from important state space.

ricklove:
To be more clear:

Let's say I am training on a simulation of a game that has a failure state where the character dies and the game resets at the beginning of the level.

The additional training needs to focus on the situation that caused the failure state. Because the machine is forced to waste time going through the entire simulation, it is not learning anything.

Even worse, if the simulation uses randomness, it may take a very long time for the machine to ever encounter a similar state and have a chance to try alternative actions to discover the results.

Ideally, the environment could be saved and restored at any point.

I want to be build a training algorithm that explores backward from failures to focus on exploration around failure states. However, it would require the ability to save and restore the environment at any point.

All that is needed is the ability to serialize and deserialize the environment state:

- envData = SaveEnvironment(env)
- env = RestoreEnvironemnt(envData)

tlbtlbtlb:
Not in general. Some kinds of environments allow this, but it's environment-specific. For Mujoco environments, the state can be gotten at with env.data. For Atari environments, something like
```
   snapshot = env.ale.cloneState()
   ...
   env.ale.restoreState(snapshot)
```
but you may have to dig into the Atari / ALE code to get it right.

shelhamer:
Note that for Atari/ALE the `cloneSystemState/restoreSystemState` methods are necessary to serialize the complete state (including pseudorandomness) if you want a fully deterministic return to the saved state.

I'm trying to reconcile `atari-py` and the gym with the version of ALE required to do this right, and will make a PR soon.

colinmorris:
[This MCTS implementation](https://gist.github.com/blole/dfebbec182e6b72ec16b66cc7e331110#file-openai-gym-mcts-py) is an example of a solution that uses this kind of technique. They just use python's `copy` library to make copies of the env.

ricklove:
@tlbtlbtlb, @shelhamer , @colinmorris Thanks for your input!

It's too bad there isn't a general solution for this. I would think it would be useful to the whole community to try to make a standard method to accomplish this for each major environment.

[...]

PR #575, opened 2017-05-02
Save and Restore Atari Environment State

shelhamer:
Save (clone) and load (restore) the Atari emulator state for analysis, planning, or other experiments that require returning to identical states of the environment. This addresses #402 for Atari.

Note: the ALE distinguishes two kinds of environment states. The "state" captures memory but not further details of the "system state" that are necessary to determinism. `clone_full_state()` and `restore_full_state()` are necessary to save and load the state and system state (including pseudorandomness) for fully identical restoration of the environment.

Further note: a limitation of ALE is that on restore the screen is blank, and it's necessary to take a step in the emulator to draw to it. To properly restore a a time step t, one needs to both clone the system state at t-1 and save the action at t-1, and then restore the system state and take the action to arrive at time t.

See openai/atari-py#15 for the prerequisite changes to expose these methods through gym. This can be merged once atari-py has been updated, the new version is propagated to gym, and the environment version numbers are bumped to v4 since a few ROMs have slightly altered semantics in the latest ALE.

### DIFF
--- a/gym/envs/atari/atari_env.py
+++ b/gym/envs/atari/atari_env.py
@@ -145,18 +145,34 @@ class AtariEnv(gym.Env, utils.EzPickle):
 
         return keys_to_action
 
-    # def save_state(self):
-    #     return self.ale.saveState()
+    def clone_state(self):
+        """Clone emulator state w/o system state. Restoring this state will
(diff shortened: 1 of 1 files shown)


shelhamer:
Depends on #584

PR #581, opened 2017-05-05
Add new classic control environment

PeterMitrano:
This pull request adds a the control environment as described in Uther and Veloso 1998, where a robot travels along a corridor with various temperature, and just adjust turn heating or cooling on or off, as well as move forward or backward. The actions are not deterministic, so the hotter or colder the robot gets the less likely your actions are to succeed. Reward is 1 at the finish and -1 for each time step not at the finish.

If this is accepted, I will document the new environment on the [gym wiki](https://github.com/openai/gym/wiki), as well as anywhere else it should be documented.

### DIFF
--- a/gym/envs/__init__.py
+++ b/gym/envs/__init__.py
@@ -90,6 +90,12 @@ register(
     max_episode_steps=500,
 )
 
+register(
+    id='RobotTemperature-v0',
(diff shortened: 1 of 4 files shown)


tlbtlbtlb:
This causes several test failures. You should be able to see the errors by running `py.test` in the gym directory.

PR #583, opened 2017-05-09
replace self.spec by getattr(self, 'spec', None) in core Env object

quanvuong:
Accessing missing object attribute raises an exception that's not catched. For example

```
  File "policy_evaluation.py", line 5, in <module>
    print(env)
  File "/Users/QV/code-project/learn/sutton/venv/lib/python3.5/site-packages/gym/core.py", line 205, in __str__
    if self.spec is not None:
AttributeError: 'BlackjackEnv' object has no attribute 'spec'
```

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -202,7 +202,7 @@ class Env(object):
         self.close()
 
     def __str__(self):
-        if self.spec is not None:
+        if getattr(self, 'spec', None) is not None:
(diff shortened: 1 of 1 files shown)


tlbtlbtlb:
Every Env will have a `.spec` field when created by `gym.make(name)`. If you're creating an env some other way, you should fill in the `.spec` field.

PR #587, opened 2017-05-12
add dimensionality and density to description for box space

4d55397500:
Add to description for the box space and add an example. 

Would also be useful to add a dimensionality attribute? (self.dim)


pzhokhov:
gym.Space is not meant to induce any specific / reliable distributions over the underlying space. While currently your description of point density is correct, samples being uniformly distributed is something I'd prefer users did not rely on - an example of possible change would be to change distribution to gaussian when boundaries are infinite, and to beta when boundaries are finite. Closing this.

PR #609, opened 2017-06-04
Improve compatibility handling in reraise module

daftshady:
`reraise.py` imports `reraise_impl` from two different files according to python version.
I think there is a better way to deal with the compatibility issue without using separated module.
This project already uses `six`.
`six.reraise` will handle a syntax difference between python 2 and 3 when we want to inject traceback to exception.
Also, setting `__cause__` to `None` in runtime will disable exception chaining in python 3.
There seems to be no existing testcases to test `reraise.py`.
Please review this change. If you guys think this change makes sense, I will add testcases.

### DIFF
--- a/gym/utils/reraise.py
+++ b/gym/utils/reraise.py
@@ -1,5 +1,5 @@
 import sys
-from six import reraise as reraise_impl
+from six import reraise as reraise_impl, PY3
 
 def reraise(prefix=None, suffix=None):
     old_exc_type, old_exc_value, traceback = sys.exc_info()
@@ -7,8 +7,7 @@ def reraise(prefix=None, suffix=None):
         old_exc_value = old_exc_type()
 
     e = ReraisedException(old_exc_value, prefix, suffix)
-    is_py3 = sys.version_info[0] == 3
-    if is_py3:
+    if PY3:
         # Python 3 has exception chaining, which we don't want in this case.
         # Setting `__cause__` to None is equivalent to `from None` syntax
         # which will disable the chaining.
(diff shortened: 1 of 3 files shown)


gym/utils/reraise.py
```diff
@@ -1,20 +1,20 @@
...
 
 def reraise(prefix=None, suffix=None):
     old_exc_type, old_exc_value, traceback = sys.exc_info()
     if old_exc_value is None:
         old_exc_value = old_exc_type()
 
     e = ReraisedException(old_exc_value, prefix, suffix)
-
-    reraise_impl(e, traceback)
+    is_py3 = sys.version_info[0] == 3
```
gdb: I think best practice is to do `if six.PY2: ...; else: ....`

gym/utils/reraise.py
```diff
@@ -1,20 +1,20 @@
...
+    if is_py3:
+        # Python 3 has exception chaining, which we don't want in this case.
+        # Setting `__cause__` to None is equivalent to `from None` syntax
+        # which will disable the chaining.
+        # See https://www.python.org/dev/peps/pep-0415/
+        e.__cause__ = None
```
gdb: Good idea. Is it possible to propagate `old_exc_value`'s cause, in case `old_exc_value` itself is a chained exception?

daftshady: What you mean is allowing exception chaining rather than suppressing it only if `old_exc_value` itself is a chained exception, I think it's possible. But it's different from original implementation because the original one disables every exception chaining. Please correct me if i misunderstood.

pzhokhov:
reraise logic is now removed completely (https://github.com/openai/gym/pull/1342), closing this

PR #621, opened 2017-06-13
Make sure env.spec always exists and is valid.

joschu:
Previously there was an error when you tried to print out an unregistered env.

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -51,6 +51,7 @@ class Env(object):
         env = super(Env, cls).__new__(cls)
         env._env_closer_id = env_closer.register(env)
         env._closed = False
+        env._spec = None
 
         # Will be automatically set when creating an environment via 'make'
         return env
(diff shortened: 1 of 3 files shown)


gym/core.py
```diff
@@ -51,6 +51,7 @@ def __new__(cls, *args, **kwargs):
         env = super(Env, cls).__new__(cls)
         env._env_closer_id = env_closer.register(env)
         env._closed = False
+        env._spec = None
```
gdb: Why not do `env.spec = None`?

joschu: spec is a @property of Wrapper, and I think I got an error when I tried to set it.

gdb: Ah good point. I'd rather not change it to being a property on env too unless we have to. A bit of a hack but you could do: `if not hasattr(env, "spec"): env.spec = None`.

joschu: Seems a bit hacky. Isn't it less likely to cause confusion if it's a property in both places?

gdb: I'm fine either way really, just think this might be a slight breaking change for people (I've run into code which sets .spec before, though it's usually a bug) without much benefit. I'd also have preferred that spec on wrapper is an attribute rather than property, but it's hard to do it due to when we actually set the spec.

So, if you'd like to make the change, just add something to the readme about this as a breaking change?

joschu: Well there's currently a bug, that comes up pretty fast when you're not using registration. (And usually it doesn't make sense to use registration in the early stages of research, or when you're doing something outside of the standard single-task setting.) I'll add to the changelist and merge.

PR #623, opened 2017-06-14
Action wrappers

ngc92:
Thre wrappers that change the apparent action space used by an env. With this it is possible to train an agent that is designed for single discrete actions on multi-discrete or even continuous envs. For the latter it is also possible to rescale the actions, e.g. to always assume actions in the range [0, 1] for the agent.

What is currently not implemented is interpreting a discrete space as continuous, any transformations of TupleSpace and the reverse_action method. If there is interest in any of that I can implement it, at least to a degree to which it is sensible (e.g. flattening a tuple space with continuous and discrete subspaces does not make sense).

### DIFF
--- a/gym/wrappers/__init__.py
+++ b/gym/wrappers/__init__.py
@@ -2,3 +2,4 @@ from gym import error
 from gym.wrappers.frame_skipping import SkipWrapper
 from gym.wrappers.monitoring import Monitor
 from gym.wrappers.time_limit import TimeLimit
+from gym.wrappers.action_wrappers import FlattenedActionWrapper, DiscretizedActionWrapper, RescaledActionWrapper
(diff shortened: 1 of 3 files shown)


gdb:
Thanks for the wrappers! We're actually trying to keep the core lightweight and stable, so I would recommend instead making a separate repository for these wrappers, which people can then import. (I've updated the README to no longer suggest that people submit wrappers for us to add.)

ngc92:
That makes sense if you strive for maximum stability in the core code. It would probably still be useful to have some kind of central registry for these kinds of utilities (downsampling, conversion to grayscale also come to mind here). Either another repo, or mazbe simply a list or wiki page with links towards the respective repos.

gdb:
Yep, please feel free to start a wrappers.md here: https://github.com/openai/gym/tree/master/docs!

pzhokhov:
per @gdb's point closing this (I have merged #631 )

PR #630, opened 2017-06-21
deleted double comment

FirefoxMetzger:
A comment in scoring.py was there twice. I removed it.


FirefoxMetzger:
@wojzaremba @gdb Since the main page is down, is the scoring system still relevant? If not I'll close this PR.

gdb:
Scoring system no longer relevant, but let's clean up the comment :)!

PR #634, opened 2017-06-26
MAME support

alito:
This adds support for mamele, which is a wrapper around the MAME emulator. At the moment only 5 games are supported, but more should come online soonish.

The name of the currently implemented environments are:

MameGalaxian-v0
MameCircus-v0
MameSidetrac-v0
MameRipcord-v0
MameCrash-v0

and their corresponding -watch- versions (eg MameGalaxian-watch-v0), which leave the screen and sound on and run at normal speed.

The last four ROMs can be legally downloaded from http://mamedev.org/roms/

There are some gross inefficiencies in the way that the frame data is transferred between MAME and gym, and some minor issues when running (eg the MAME windows remains onscreen, even though it doesn't get updated). It is also Python2 only for now (the code has very old roots). The installation is slow, since it compiles MAME, but shouldn't be too fragile.

The code from https://github.com/alito/deep_q_rl successfully trains on (at least) the circus environment, although not successfully on Galaxian for now (Arcade Galaxian is a tricky environment)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -209,6 +209,21 @@ These are a variety of classic control tasks, which would appear in a typical re
 	  env.reset()
 	  env.render()
 
+MAME
+-----
+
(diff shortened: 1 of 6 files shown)


alito:
Python 3 support now added. No changes to this patch, just to the underlying mamele package

pzhokhov:
This looks pretty cool! Unfortunately, at the moment our policy is not to add new environments to the gym. I would recommend keeping it as a separate project / package; and will be happy to add reference to it to docs/environments.md

PR #643, opened 2017-07-05
Add grayscale support to AtariEnv

qbx2:
ALE has getScreenGrayscale() interface for grayscale rendering, but many DQN based algorithms extracts Y' channel from RGB data since gym only supports RGB rendering output. I added grayscale rendering support for the algorithms.

Try this:

```python
import gym
env = gym.make('Breakout-v0')
x = env.render('grayscale_array')
plt.imshow(x[...,0], cmap='gray')
plt.show()
```

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -115,6 +115,8 @@ class Env(object):
         - rgb_array: Return an numpy.ndarray with shape (x, y, 3),
           representing RGB values for an x-by-y pixel image, suitable
           for turning into a video.
+        - grayscale_array: Return an numpy.ndarray with shape (x, y, 1),
+          representing luminance values for an x-by-y pixel image. (AtariEnv only)
         - ansi: Return a string (str) or StringIO.StringIO containing a
           terminal-style text representation. The text can include newlines
           and ANSI escape sequences (e.g. for colors).
--- a/gym/envs/__init__.py
+++ b/gym/envs/__init__.py
@@ -304,11 +304,15 @@ for game in ['air_raid', 'alien', 'amidar', 'assault', 'asterix', 'asteroids', '
     'private_eye', 'qbert', 'riverraid', 'road_runner', 'robotank', 'seaquest', 'skiing',
     'solaris', 'space_invaders', 'star_gunner', 'tennis', 'time_pilot', 'tutankham', 'up_n_down',
     'venture', 'video_pinball', 'wizard_of_wor', 'yars_revenge', 'zaxxon']:
-    for obs_type in ['image', 'ram']:
+    for obs_type in ['image', 'ram', 'grayscale_image']:
         # space_invaders should yield SpaceInvaders-v0 and SpaceInvaders-ram-v0
         name = ''.join([g.capitalize() for g in game.split('_')])
         if obs_type == 'ram':
             name = '{}-ram'.format(name)
+        elif obs_type == 'grayscale_image':
+            # register grayscale ?
+            continue
+            # name = '{}-grayscale'.format(name)
 
         nondeterministic = False
         if game == 'elevator_action' and obs_type == 'ram':
--- a/gym/envs/atari/atari_env.py
+++ b/gym/envs/atari/atari_env.py
@@ -20,14 +20,14 @@ def to_ram(ale):
     return ram
 
 class AtariEnv(gym.Env, utils.EzPickle):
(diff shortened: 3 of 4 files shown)


pzhokhov:
I am not convinced this is necessary... Is grayscale rendering much faster than rgb rendering / does it offer something that cannot be achieved via taking mean value of all rgb channels? If not, than I would rather have something like a GrayscaleWrapper that converts rgb images into grayscale - this way, it can be applied to other envs with pixel observations, not just atari

qbx2:
Yes, there is a lot of performance gain..

pzhokhov:
do you have specific numbers? Thanks!

qbx2:
Sorry, I do not know. Even I’m not sure if it’s much faster. It’s too old to remember.

pzhokhov:
fair point. I'll experiment with it a little bit and if confirmed will merge.

christopherhesse:
We have an OpenCV2 based wrapper called WarpFrame that does grayscale conversion, I think that should be the reference point for performance comparisons.

zuoxingdong:
Hi all @pzhokhov @christopherhesse @qbx2 , I've made a simple heuristic test how much speed up we can gain from ALE grayscale method. 

It's around ~5-10% speed up, tested on Breakout for 100 random seeds

To reproduce the result, the code is provided as below

```python
import numpy as np
import gym
from time import time

env_id = 'BreakoutNoFrameskip-v0'

T1 = []
T2 = []
for seed in range(100):
    env = gym.make(env_id)
    env.seed(seed)
    env.reset()

    t = time()
    img = env.unwrapped.ale.getScreenGrayscale()
    T1.append(time() - t)

    env.close()
    del env


    env = gym.make(env_id)
    env.seed(seed)
    obs = env.reset()

    t = time()
    obs = cv2.cvtColor(obs, cv2.COLOR_RGB2GRAY)
    T2.append(time() - t)

    env.close()
    del env

T1 = np.mean(T1)
T2 = np.mean(T2)
print(f'Average time of ALE grayscale: {T1}')
print(f'Average time of cv2 grayscale: {T2}')
print(f'ALE is {round(100*(T2/T1)-100, 1)}% faster than cv2')
```

pzhokhov:
@zuoxingdong thanks for putting the time in benchmarking this! I don't think the comparison is quite fair though in your script - the time to get the grayscale image is compared to time to convert rgb image to grayscale, and I think it should be compared to time to get rgb image, then convert it to grayscale. Here's a modified version of your script that I ran: 

```
import numpy as np
import gym 
from time import time
import cv2 

env_id = 'BreakoutNoFrameskip-v0'

T1 = []
T2 = []
env1 = gym.make(env_id)
env2 = gym.make(env_id)
for seed in range(10):
    env1.seed(seed)
    env1.reset()

    env2.seed(seed)
    env2.reset()

    for _ in range(100):
        t = time()
        img = env1.unwrapped.ale.getScreenGrayscale()
        T1.append(time() - t)

        t = time()
        obs = cv2.cvtColor(env2.unwrapped.ale.getScreenRGB2(), cv2.COLOR_RGB2GRAY)
        T2.append(time() - t)

env1.close()
env2.close()

T1 = np.mean(T1)
T2 = np.mean(T2)
print(f'Average time of ALE grayscale: {T1}')
print(f'Average time of cv2 grayscale: {T2}')
print(f'ALE is {round(100*(T2/T1)-100, 1)}% faster than cv2')
```
on osx 10.14 and ubuntu 16.04 it ALE shows speedup of 160 - 180%. So I would say that it s likely worth the change. @christopherhesse what do you think?

christopherhesse:
It does look faster, though the environment reset time may be so slow that it dominates the runtime of the environment.  I don't think we should register even more variants of atari environments, but you should be able to specify this flag either directly to `AtariEnv` or to `gym.make('Breakout-v4', obs='grayscale_image')` or something like that.

zuoxingdong:
Or another option could be merging current Atari-related wrappers in baselines as a single wrapper (e.g. `AtariPreprocessing`) which internally call ALE gray scale like `obs = env.unwrapped.ale.get_grayscale`, while keeping generic wrappers separately such as `ResizeFrame`, `GrayFrame` implemented via opencv which could be used for non-atari environments.

pzhokhov:
closing this PR; but created a new issue to implement grayscale atari envs either via a wrapper or via a gym.make kwarg.

PR #644, opened 2017-07-05
Support for Env.__init__() parameters overridding

qbx2:
```python
env = gym.make('Pong-v0', frameskip=0)
```

Above code will generate Pong-v0 with frameskip=0. Sometimes I wanted to change the spec of the environment for my study, so added support for the parameter overriding.

### DIFF
--- a/gym/envs/registration.py
+++ b/gym/envs/registration.py
@@ -77,13 +77,15 @@ class EnvSpec(object):
         self._local_only = local_only
         self._kwargs = {} if kwargs is None else kwargs
 
-    def make(self):
+    def make(self, **kwargs):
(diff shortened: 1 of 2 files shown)


qbx2:
Hmm, I made this change to use gym-breakout (where obs_type='image') with my RL project. How is that possible?

kouroshHakha:
I made those changes manually, but can these be globally resolved please? It is kind of an important issue.

pzhokhov:
kwargs for gym.make are now supported, closing this PR

tsunhenry:
Hi, I don't think kwargs for gym.make is supported by ExperimentGrid?

PR #652, opened 2017-07-12
fixed and added missing __eq__ for spaces

ngc92:
This PR adds `__eq__` to the spaces that do not have this operator yet, and fixes the old equality operator that could not distinguish some spaces (i.e. Discrete and MultiBinary with the same n).


pzhokhov:
similar changes are made upstream (and some in this PR https://github.com/openai/gym/pull/1375); closing

PR #653, opened 2017-07-12
Adding TicTacToe Environment in board_games 

aravindreddypv:
This pull request was to add a new environment called Tic-Tac-Toe . It is a two player game played on 3x3 board and rendered using pygame module . Observation was a list of 0's , 1's and 2's , where 0 is blank and 1  for 1st player's move(i.e X) and 2  for 2nd player's move(i.e O) . A reward of +1,0,-1 will be given at the end of the episode for win,draw,lose respectively and 0 for all other intermediate steps.Documentation will be done if it was accepted.


pzhokhov:
While the tictactoe env looks interesting, our current policy is to not include new environments into the gym. I would recommend creating a separate project / package; we'll gladly add a reference to it in docs/envrionments.md file.

PR #656, opened 2017-07-14
Fixed typo

kokimame:


pzhokhov:
while this could have meant if and only if, looking at the rest of the sentence regular if makes just as much sense. Merging.

PR #671, opened 2017-07-29
New text games group and two text-based games

MikulasZelinka:
Hello, I added a group for text-based games (interactive fiction) and two sample text games, Saving John and Machine of Death.

These two sample games are available in the [pyfiction](https://github.com/MikulasZelinka/pyfiction) library that supports different games which I plan to add to gym once some basic problems are resolved.

My questions are:

1. How to handle variable action space and unbounded observation space? See the pull request code for my approach (always checking for the dynamic environment action space). Is this approach valid?

2. Is it necessary that all environments support Python 2.7? The pyfiction games, for example, are only tested in Python 3.5 and 3.6.

Thank you :)

### DIFF
--- a/examples/agents/random_agent.py
+++ b/examples/agents/random_agent.py
@@ -51,6 +51,11 @@ if __name__ == '__main__':
     for i in range(episode_count):
         ob = env.reset()
         while True:
+
+            # handle dynamic action spaces (has no effect if the action space is static):
+            if agent.action_space != env.env.action_space:
+                agent.action_space = env.env.action_space
+
             action = agent.act(ob, reward, done)
             ob, reward, done, _ = env.step(action)
             if done:
--- a/gym/envs/__init__.py
+++ b/gym/envs/__init__.py
@@ -128,6 +128,19 @@ register(
     reward_threshold=900,
 )
 
+# Text games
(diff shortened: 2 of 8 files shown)


MikulasZelinka:
I'm currently working on getting pyfiction published on PyPI which should fix the CI test.

pzhokhov:
Hi @MikulasZelinka ! Currently, our policy is not to accept new environments to gym; if still interested, I would recommend creating a separate/project. From our side, we can put a reference to it into the environments.md file. As such, closing this PR. 
To your questions - the problem with handling variable / unbounded action and observation spaces is really less about how to handle them from the environment side; but how to make standard RL agents know about / work with the variable spaces. The reasonable and simple enough approach for action space is to simply  set number of actions to maximum possible number of actions in the game; and if the agent outputs action that is currently invalid, terminate the episode with large negative reward. Unbounded observation space - that depends; from the PR I just see that the observation is the simulator state. How does that state look like? In principle, Box observation space supports infinite limits; moreover, an observation space can be multidiscrete with number of subspaces equal to number of bytes in simulator state, and number of choices for each 256 ( this way any state of finite-size memory can be represented).

PR #683, opened 2017-08-11
For mujoco_py 1.5

jaesik817:
Hi All,

Currently I researched my own project with mujoco environment with mujoco_py and gym.

I hoped to use a great structure of gym, however I found some not proper linked lines between mujoco_py 1.5 and gym.

Thus, I checked those things and changed gym source code.

Please review that, and if that is correct, then sharing that to other users!

Thanks. Jaesik.


matthiasplappert:
This is addressed in https://github.com/openai/gym/pull/834.

PR #692, opened 2017-08-15
Fixed import error message for pachi-py.

tiberiu:
The import error message for pachi-py suggested a fix which was not correct.

It should be pip install gym[board_game] instead.


pzhokhov:
fix is obsolete as board_game environments got removed; closing this

PR #703, opened 2017-08-27
Info about the osim-rl environment added to external environments

kidzik:
We built an RL environment within and motivated by OpenAI gym framework and we would be happy to share the link in gym docs.


olegklimov:
Hmm... http://opensimulator.org/ doesn't appear to be an alive project. Do you have a screenshot, or video?

kidzik:
It's actually not the opensimulator but OpenSim from Stanford http://opensim.stanford.edu/ / https://github.com/opensim-org/opensim-core 
This is a biomechanics simulator built and used in biomechanics community. Recently we created a reinforcement learning environment using this engine ( https://github.com/stanfordnmbl/osim-rl ).
Best examples of how it works are in our NIPS competition itself
https://www.crowdai.org/challenges/nips-2017-learning-to-run
especially videos in the leaderboard
https://www.crowdai.org/challenges/nips-2017-learning-to-run/leaderboards

olegklimov:
Oh, I've seen this. Didn't look serious back then, we skipped. So you just want to have link posted?

kidzik:
Sure, I posted the link in the commit, should I add more info?
Indeed, our RL is quite fresh but in biomechanics it's a well-established simulation engine.

PR #712, opened 2017-09-06
Add automatic discovery of envs defined in custom packages (plugins)

Phylliade:
This PR adds plugin support to gym envs.

Currently, packages that add additional plugins require to be manually imported in order to be registered. 
For example, if we have a `gym_custom` package, providing the `custom_env`, we should do:
```
import gym 
import gym_custom

env = gym.make("Custom-v0")
```

With this PR, discovery of plugin packages is automatic during the `gym import`:
```
import gym

env = gym.make("Custom-v0")
```
This is working using a naming convention (as Flask does): every package whose toplevel dir begins with (`gym_`) is considered a plugin package and will be imported.

N.B.: While there are other possible solutions, this one is the most lightweight and won't break current custom envs.

### DIFF
--- a/gym/envs/README.md
+++ b/gym/envs/README.md
@@ -37,7 +37,7 @@ ant.AntEnv
         foo_env.py
         foo_extrahard_env.py
   ```
-  Note: Ensure that your top level package name is of the format `gym_<yourname>`, otherwise **it won't automatically be imported by gym**.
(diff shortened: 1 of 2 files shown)


pzhokhov:
analogous solution is in the works: https://github.com/openai/gym/pull/1344, closing this

PR #728, opened 2017-09-21
Changed MujocoEnv and inheriting environment classes to use mujoco_py 1.5

aravind0706:
Changed MujocoEnv and inheriting task environment classes to use the new mujoco_py 1.5 API.
Issues to be figured out:

1. mjlib.mj_resetData(self.model.ptr, self.data.ptr) - the new API's equivalent of it is not clear
2. viewer.autoscale() - no equivalent in new API. Not sure if it's needed. 

Changed setup.py to remove @jonasschneider's mujoco pin to <1.0.0


matthiasplappert:
This is addressed by https://github.com/openai/gym/pull/834

PR #733, opened 2017-09-27
Mujoco 1.0

adamlerer:
Some people have been asking for mujoco-1.0 (1.5.0.x) integration with gym (e.g. for headless rendering). This got it working for us.

### DIFF
--- a/gym/envs/mujoco/mujoco_env.py
+++ b/gym/envs/mujoco/mujoco_env.py
@@ -29,7 +29,7 @@ class MujocoEnv(gym.Env):
         self.model = load_model_from_path(fullpath)
         self.sim = MjSim(self.model)
         self.data = self.sim.data
-        self.viewer = None
+        self.viewer = MjViewer(self.sim)  # None
 
         self.metadata = {
             'render.modes': ['human', 'rgb_array'],
--- a/gym/envs/mujoco/mujoco_env.py
+++ b/gym/envs/mujoco/mujoco_env.py
@@ -9,9 +9,48 @@ import six
 
 try:
     import mujoco_py
-    from mujoco_py.mjlib import mjlib
+    from mujoco_py import load_model_from_path, MjSim, MjViewer
 except ImportError as e:
     raise error.DependencyNotInstalled("{}. (HINT: you need to install mujoco_py, and also perform the setup instructions here: https://github.com/openai/mujoco-py/.)".format(e))
+
+def _get_image_hack(self):
+    # modified version of _read_pixels_as_in_window
+    # Reads pixels with markers and overlay from the same camera as screen.
+    resolution = glfw.get_framebuffer_size(
+        self.sim._render_context_window.window)
+ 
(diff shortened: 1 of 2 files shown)


rhaps0dy:
Thank you, very useful! I'm using your code with the latest MuJoCo.

gym/envs/mujoco/mujoco_env.py
```diff
@@ -9,41 +9,49 @@
...
         observation, _reward, done, _info = self._step(np.zeros(self.model.nu))
         assert not done
         self.obs_dim = observation.size
 
-        bounds = self.model.actuator_ctrlrange.copy()
-        low = bounds[:, 0]
-        high = bounds[:, 1]
-        self.action_space = spaces.Box(low, high)
+        bounds = self.model.actuator_ctrlrange
+        # I'm not sure why bounds is at least sometimes None ... bug?
```
machinaut: @adamlerer / @yayitsamyzhang could you share examples of when this happens?  Possibly a bug.

gym/envs/mujoco/mujoco_env.py
```diff
@@ -76,65 +84,53 @@ def viewer_setup(self):
...
 
     def set_state(self, qpos, qvel):
         assert qpos.shape == (self.model.nq,) and qvel.shape == (self.model.nv,)
-        self.model.data.qpos = qpos
-        self.model.data.qvel = qvel
-        self.model._compute_subtree()  # pylint: disable=W0212
-        self.model.forward()
+        sim_state = self.sim.get_state()
+        sim_state.qpos[:] = qpos
+        sim_state.qvel[:] = qvel
```
machinaut: I'm pretty sure what we want to do here is
```
self.data.qpos[:] = qpos
self.data.qvel[:] = qvel
```

The state is a separate thing, and if you set those parameters on it they'll just get thrown away.

gym/envs/mujoco/mujoco_env.py
adamlerer: Unfortunately I've lost track of when this happened. You could try setting it to None and then run an environment and see what happens!

machinaut: It's constructed by generated code (here in the most recent version: https://github.com/openai/mujoco-py/blob/master/mujoco_py/generated/wrappers.pxi#L1285).

Wrapper is defined here: https://github.com/openai/mujoco-py/blob/master/mujoco_py/generated/wrappers.pxi#L3824

So it looks like we set it to `None` instead of a numpy array with 0 as one of the dimensions (which would just be empty anyways).

This would be true for models without any actuators (so all the `ctrl` related arrays are size 0).

I'm alright with this handling -- maybe update the comment to say that it's None when there are no actuators.

If you wanted to, you could check `sim.model.nu == 0`, which is the same thing (no actuators in the model).

matthiasplappert:
This is addressed in https://github.com/openai/gym/pull/834.

PR #740, opened 2017-10-04
Get raw pixels as observations from Mujoco env

sumethy:
Hi, recently I had to get raw pixels out from Mujoco env in order to train rl agent using images, so I'd like to share what I did. The details of the changes made can be found at  gym/gym/envs/GET_PIXELS_AS_OBSERVATION_MUJOCO.md


pzhokhov:
I am not a big fan of this approach - it leads to code duplication; also hard to effort scales linearly with number of environments. Given that environments support `render(mode='rgb_array')`, to my taste a better approach would be to make a wrapper that calls `render` and returns that as an observation, something like (roughly)

```
from gym.core import ObservationWrapper
class PixelObservationWrapper(ObservationWrapper)
     def __init__(env, img_shape=None):
           super().__init___(env)
           self.img_shape = img_shape

          img = self.env.render(mode='rgb_array')
          return img if self.img_shape is None else img.image_resize(self.img_shape)
```
As such (and due to age) closing this, also creating a respective issue to implement PixelObservationWrapper

https://github.com/openai/gym/issues/1374

biggzlar:
This has been implemented in the form of the PixelObservationWrapper since, but is there actually a way **not to render** the environment using that wrapper?

PR #763, opened 2017-11-03
Shape attribute to Dict space. 

Kismuz:
Added .shape attribute to Dict space. Fixed from_jsonable() method for Tuple space as it has been returning `zip` object instead of list of tuples. 
Test with nested Dict space passed.


Kismuz:
@wojzaremba, 
@bmcgrew,
is there any chance for this commit to get reviewed/merged?

wojzaremba:
LGTM

PR #767, opened 2017-11-08
Support MuJoCo 1.5

lobachevzky:
To test these changes, I checked that gym ran on all mujoco environments and that rendering works. Also, I have been working on a custom environment using these changes for about a month with no problems. That said, I cannot swear that these changes don't break anything so I certainly welcome any feedback.

### DIFF
--- a/gym/envs/mujoco/half_cheetah.py
+++ b/gym/envs/mujoco/half_cheetah.py
@@ -8,9 +8,9 @@ class HalfCheetahEnv(mujoco_env.MujocoEnv, utils.EzPickle):
         utils.EzPickle.__init__(self)
 
     def _step(self, action):
-        xposbefore = self.model.data.qpos[0, 0]
+        xposbefore = self.sim.data.qpos[0]
         self.do_simulation(action, self.frame_skip)
-        xposafter = self.model.data.qpos[0, 0]
+        xposafter = self.sim.data.qpos[0]
         ob = self._get_obs()
         reward_ctrl = - 0.1 * np.square(action).sum()
         reward_run = (xposafter - xposbefore)/self.dt
@@ -20,8 +20,8 @@ class HalfCheetahEnv(mujoco_env.MujocoEnv, utils.EzPickle):
 
     def _get_obs(self):
         return np.concatenate([
-            self.model.data.qpos.flat[1:],
-            self.model.data.qvel.flat,
+            self.sim.data.qpos.flat[1:],
+            self.sim.data.qvel.flat,
         ])
 
     def reset_model(self):
--- a/gym/envs/mujoco/inverted_double_pendulum.py
+++ b/gym/envs/mujoco/inverted_double_pendulum.py
@@ -11,22 +11,22 @@ class InvertedDoublePendulumEnv(mujoco_env.MujocoEnv, utils.EzPickle):
     def _step(self, action):
         self.do_simulation(action, self.frame_skip)
         ob = self._get_obs()
-        x, _, y = self.model.data.site_xpos[0]
+        x, _, y = self.sim.data.site_xpos[0]
         dist_penalty = 0.01 * x ** 2 + (y - 2) ** 2
-        v1, v2 = self.model.data.qvel[1:3]
+        v1, v2 = self.sim.data.qvel[1:3]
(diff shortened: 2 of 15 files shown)


gym/envs/mujoco/half_cheetah.py
```diff
@@ -8,9 +8,9 @@ def __init__(self):
         utils.EzPickle.__init__(self)
 
     def _step(self, action):
-        xposbefore = self.model.data.qpos[0, 0]
+        xposbefore = self.sim.data.qpos[0]
```
lobachevzky: My understanding is that `qpos` lists joint angles. In `half_cheetah.xml`, it looks like the first joint is "rootx", a slide joint that should give displacement in the x direction. So `qpos[0]` should give the desired value.

gym/envs/mujoco/inverted_double_pendulum.py
```diff
@@ -11,22 +11,22 @@ def __init__(self):
...
         ob = self._get_obs()
-        x, _, y = self.model.data.site_xpos[0]
+        x, _, y = self.sim.data.site_xpos[0]
         dist_penalty = 0.01 * x ** 2 + (y - 2) ** 2
-        v1, v2 = self.model.data.qvel[1:3]
+        v1, v2 = self.sim.data.qvel[1:3]
         vel_penalty = 1e-3 * v1**2 + 5e-3 * v2**2
         alive_bonus = 10
-        r = (alive_bonus - dist_penalty - vel_penalty)[0]
+        r = alive_bonus - dist_penalty - vel_penalty
```
lobachevzky: Not sure why `alive_bonus - dist_penalty - vel_penalty` used to be an array. Please double check if I broke something here.

gym/envs/mujoco/mujoco_env.py
```diff
@@ -111,30 +111,19 @@ def _render(self, mode='human', close=False):
...
-        idx = self.model.body_names.index(six.b(body_name))
-        return self.model.data.com_subtree[idx]
-
-    def get_body_comvel(self, body_name):
-        idx = self.model.body_names.index(six.b(body_name))
-        return self.model.body_comvels[idx]
-
-    def get_body_xmat(self, body_name):
-        idx = self.model.body_names.index(six.b(body_name))
-        return self.model.data.xmat[idx].reshape((3, 3))
```
lobachevzky: These methods don't have obvious implementations in mujoco 1.5 and they aren't used anywhere so I deleted them.

episodeyang:
@gdb would someone review this change? this one adds support for mujuco_py@1.5. 
Thanks!

aravind0706:
@lobachevzky Does your GLFW in MuJoCo 1.5 close on calling `env.close()` ? I haven't tried using this branch but independently, the version I have has an issue with closing the renderer automatically from `env.close()` ....  Is there some change to `mujoco-py` source that needs to be made?

lobachevzky:
@aravind0706 My pull request does not support closing on `env.close()`. The original version of `mujoco-py` calls the [`finish()` method](https://github.com/openai/mujoco-py/blob/0.5/mujoco_py/mjviewer.py#L323) of `MjViewer` in `mujoco-py` version .5. This method no longer exists. 

What is your use case? I have never had to close a window programmatically (usually I use my good old mouse).

matthiasplappert:
I'm currently looking into getting this PR merged. I have benchmarked your changes with PPO to ensure that they are not broken and not much harder or easier to learn than before.

Here are these results:
fig_ant
fig_halfcheetah
fig_hopper
fig_humanoid
fig_humanoidstandup
fig_inverteddoublependulum
fig_invertedpendulum
fig_reacher
fig_swimmer
fig_walker2d

This looks very good overall. I'm just slightly concerned about `Walker2d`, but it seems to eventually converge to the same performance.

matthiasplappert:
Could you please bump the versions of all MuJoCo environments (e.g. `HalfCheetah-v1` should become `HalfCheetah-v2` and so on)? After this, I think we can get this merged quickly.

matthiasplappert:
I've moved your commits over to https://github.com/openai/gym/pull/834 and made the necessary version bumps myself. Closing this since further progress will be tracked in the new PR. Thank you so much for this contribution!

PR #773, opened 2017-11-12
Modernize Python 2 code to get ready for Python 3

cclauss:
Make the minimal, safe changes required to convert the repo's code to be syntax compatible with both Python 2 and Python 3. More work _might_ be required to complete the port to Python 3 but this is a minimal, safe first step.

* [futurize stage 1](http://python-future.org/futurize_cheatsheet.html#step-1-modern-py2-code)
* `from __future__ import print_function`


cclauss:
Closing because there are too many conflicts to cleanly resolve.

PR #781, opened 2017-11-21
_reset function and _step function not consistency

liaolushen:
In the `_reset` function in line 312, the function use `_observation()`, however, in the `_step` function in line 316, the function use `observation()`. It's not a big deal, but I think we should keep the consistency.


pzhokhov:
fixed upstream; closing this

PR #785, opened 2017-11-23
Full Keras 2 Compatibility w/TF or Theano Backend

dexterfichuk:
Updated keras layers to use the v2+, reshaped MNIST to be the same order as Cifar, and fixed so input shape will adjust to both Theano and Tesnorflow Backends.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,6 +1,6 @@
 OpenAI Gym
 **********
-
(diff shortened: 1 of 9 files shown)


pzhokhov:
closing this for the following reasons: a) keras code in parameter tuning is now not part of the gym
2) generally speaking, our policy is not to accept new environments and rather to advise to move them in separate projects

Issue #784, opened 2017-11-22
explainations for environment state parameters?


evanthebouncy:
I am using the BipedalWalker-v2 environment at the moment, and the state space is this monster of array of numbers:

```
array([  2.74756690e-03,   1.39075128e-07,  -1.81091944e-05,
        -1.60000002e-02,   9.21555310e-02,   4.20305405e-05,
         8.60153288e-01,   1.36994788e-03,   1.00000000e+00,
         3.25597860e-02,   4.20435208e-05,   8.53705347e-01,
        -2.86064533e-05,   1.00000000e+00,   4.40813929e-01,
         4.45820034e-01,   4.61422682e-01,   4.89550084e-01,
         5.34102678e-01,   6.02460921e-01,   7.09148765e-01,
         8.85931671e-01,   1.00000000e+00,   1.00000000e+00])

```

So what do each number mean? I have been searching online for a documentation for a few hours and I have found nothing. I understand I can probably just train it with a black-box but it will be very useful to know what these numbers even represent. 

If someone can point me to a definitive documentation page that will be good. I have _not_ been able to find any. There is a wiki page for cart-pole explaining what the parameters mean in that environment:

https://github.com/openai/gym/wiki/CartPole-v0

But I have not found an explaination for the walker environment.

Thanks

olegklimov:
1. Just look at the code, it's nice and clean.
2. You can't hope to learn walker to walk using your own insight. It's too hard. You need to use a learning algorithm. And if you use that, you no longer need to know what each observation means.

PR #787, opened 2017-11-24
Create ReadMe.md

ParthaEth:
Describing the input out-put meaning by reading the source code

### DIFF
--- a/gym/envs/box2d/ReadMe.md
+++ b/gym/envs/box2d/ReadMe.md
@@ -1,0 +1,64 @@
+## Overview
+
+
+
+### Details
+
+* Name: BipedalWalker-v2  
+* Category: Box2D  
+* [Leaderboard Page](https://gym.openai.com/envs/BipedalWalker-v2/)
+
+### Description
+
+Get a 2D biped walker to walk trough rough terrain.
+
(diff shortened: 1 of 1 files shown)


ParthaEth:
@olegklimov could you please review this and merge?

ImpulseAdventure:
Hi @ParthaEth -- thanks for creating the summary.
I have integrated your content into the wiki: [BipedalWalker-v2](https://github.com/openai/gym/wiki/BipedalWalker-v2)

olegklimov:
Hi @ParthaEth @ImpulseAdventure . I've written BipedalWalker two years from now. I've also learned a lot since then. I'd recommend two options if you want to learn continuous control RL:

1. Continue with roboschool, all environments are well tested, have zoo policies.

2. Modify BipedalWalker so it is suitable for RL, not only evolution strategies. First thing I'd do, remove -100 for crashing, because it discourages agent from doing anything dangerous, and it has to. Actially not receiving alive bonus is already bad enough. I don't have time to do this (modify and test), but for someone else this can be a very good practice to understand what env should be to be learnable by RL.

@ImpulseAdventure nice wiki.

Also dup #784

PR #788, opened 2017-11-30
Video Recorder: skip current frame if env._recorder_skip_frame is True

Sohojoe:
I need a way to skip recording some frame. 

I add a lag between the agent and the enviroment so the first n frames are blank which creates a flicker in the video

This allows me to offset ping and gives a significant speed boost to training in a distributed enviroment.

### DIFF
--- a/gym/monitoring/video_recorder.py
+++ b/gym/monitoring/video_recorder.py
@@ -104,6 +104,10 @@ class VideoRecorder(object):
 
         render_mode = 'ansi' if self.ansi_mode else 'rgb_array'
         frame = self.env.render(mode=render_mode)
+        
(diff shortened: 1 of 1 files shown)


pzhokhov:
VideoRecorder is not an environment Wrapper - so not every state/observation from every step is automatically passed to the VideoRecorder, only the states in which capture_frame is called. Thus, I am in rather in favor of keeping this logic where capture_frame method is being called (instead of making VideoRecorder make a decision whether to store the frame or not). Closing this; but if the change is still relevant / necessary, feel free to reopen a new PR.

Sohojoe:
@pzhokhov - no problem - i moved to a ml-agents so no longer use this approach

PR #806, opened 2017-12-18
Fix sim_env script for python3

goriccardo:
```raw_input``` has become ```input``` in python3. Using ```six``` to support both python2 and python3.


goriccardo:
Closing as #1246 solved it

PR #807, opened 2017-12-18
Fix sim_env script

goriccardo:
I kept having failures due to calling env.step after done was set to True.

I think the last ```_, _, done, _ = env.step(a)``` belongs to the "human" mode elif.

I also cleaned up a bit the indentation following pep8 rules.


goriccardo:
Hi there, sim_env is still broken, should I rebase the request, close it or make some changes to it?
thanks!

pzhokhov:
similar fixes to sim_env made upstream, closing this

PR #834, opened 2018-01-24
Migrate to mujoco-py 1.5

matthiasplappert:
Based on #767 but with bumped versions. For completeness, here are the before and after benchmark results with PPO over 10 seeds:

fig_ant
fig_halfcheetah
fig_hopper
fig_humanoid
fig_humanoidstandup
fig_inverteddoublependulum
fig_invertedpendulum
fig_reacher
fig_swimmer
fig_walker2d

### DIFF
--- a/gym/envs/__init__.py
+++ b/gym/envs/__init__.py
@@ -204,89 +204,89 @@ register(
 # 2D
 
 register(
-    id='Reacher-v1',
+    id='Reacher-v2',
     entry_point='gym.envs.mujoco:ReacherEnv',
     max_episode_steps=50,
     reward_threshold=-3.75,
 )
 
 register(
-    id='Pusher-v0',
+    id='Pusher-v1',
     entry_point='gym.envs.mujoco:PusherEnv',
     max_episode_steps=100,
     reward_threshold=0.0,
 )
 
 register(
-    id='Thrower-v0',
+    id='Thrower-v1',
     entry_point='gym.envs.mujoco:ThrowerEnv',
     max_episode_steps=100,
     reward_threshold=0.0,
 )
 
 register(
-    id='Striker-v0',
+    id='Striker-v1',
     entry_point='gym.envs.mujoco:StrikerEnv',
     max_episode_steps=100,
     reward_threshold=0.0,
 )
 
 register(
-    id='InvertedPendulum-v1',
+    id='InvertedPendulum-v2',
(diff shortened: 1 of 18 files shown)


wojzaremba:
Could you bump version in gym/version.py
Please push to pypi after merge.

gym/envs/__init__.py
```diff
@@ -204,89 +204,89 @@
...
-    id='Reacher-v1',
+    id='Reacher-v2',
     entry_point='gym.envs.mujoco:ReacherEnv',
     max_episode_steps=50,
     reward_threshold=-3.75,
 )
 
 register(
-    id='Pusher-v0',
+    id='Pusher-v1',
```
machinaut: I think these should all be the same version (even if that means some environments skip versions).

This makes it obvious that "all the v2+ mujoco environments use version 1.50" or something similar.

setup.py
```diff
@@ -11,7 +11,7 @@
   'board_game' : ['pachi-py>=0.0.19'],
   'box2d': ['Box2D-kengz'],
   'classic_control': ['PyOpenGL'],
-  'mujoco': ['mujoco_py<1.0.0,>=0.4.3', 'imageio'],
+  'mujoco': ['mujoco_py>=1.5', 'imageio'],
```
machinaut: While this is gonna match the right thing anyways, I think this should be `mujoco-py>=1.50` since the version we jumped from 0.5.x -> 1.50.x.y.

machinaut:
+cc @lobachevzky

matthiasplappert:
Remaining feedback has been addressed (version bump, unified version number for all mujoco envs).

@machinaut I have not tested if existing policies do transfer. I think this is a non-blocker though since anyone can install `gym 0.9.3` and use the old environments with old the `mujoco-py`. Thoughts?

machinaut:
Oh yeah no need for that.  I was mostly curious.

quanvuong:
@matthiasplappert Is it possible to release the raw numbers used to draw the benchmark graph above? We're working on extending PPO and would love to have these numbers to compare our results against PPO without having to run the benchmarking again.

Thank you!

matthiasplappert:
@quanvuong Here are the raw numbers from the above plots: https://www.dropbox.com/s/2chw5biyvmd6xvf/ppo-gym-v2.zip?dl=0

PR #851, opened 2018-01-31
Try to fix the recursive import error on Python 2

pcmoritz:
Be careful with this, it might break other things.

There should also be a proper way of getting all the dependencies sorted out nicely.

### DIFF
--- a/gym/__init__.py
+++ b/gym/__init__.py
@@ -9,7 +9,11 @@ from gym.version import VERSION as __version__
 
 from gym.core import Env, Space, Wrapper, ObservationWrapper, ActionWrapper, RewardWrapper
 from gym.envs import make, spec
-from gym import wrappers, spaces, logger
+from gym import wrappers, logger
+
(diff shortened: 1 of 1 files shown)


simontudo:
Solved the error for me! Thanks a lot!

Astronirav:
It kinda solved the issue for me. But when I tried running this code: [https://gist.github.com/awjuliani/4d69edad4d0ed9a5884f3cdcf0ea0874#file-q-net-learning-clean-ipynb](url)
I got following error:
screenshot from 2018-02-01 18 53 31

frostburn:
This solution feels like a hack.
Usually the proper way to resolve circular imports is to move the offending entity into a separate module that has no further imports.
You can also try moving one of the imports out of module level inside a method so that the module loading can complete without actually running the circular part.

pcmoritz:
The proper fix is being developed here: https://github.com/openai/gym/pull/858

PR #862, opened 2018-02-05
Add tracking cameras to mujoco environments

machinaut:
This adds a camera which tracks the robot from a fixed pose relative to
the center of mass.


matthiasplappert:
LGTM.

aravind0706:
@machinaut @matthiasplappert Is there a way to use the GUI viewer directly with the manually added camera instead of having to toggle camera with TAB? Basically, the kind of functionality I am looking for is something like `sim.render(camera_name='track')` for `viewer.render`. It looks like the mjviewer class doesn't support this functionality at the moment.

PR #869, opened 2018-02-08
Removed Scoreboard reference & Q-Learning agent

FirefoxMetzger:
Both the agent and the scoreboard have been depricated, thus were removed from this documentation.


FirefoxMetzger:
@gdb Small doc update removing references to deprecated content. Does it look useful?

gdb:
Thanks!

PR #870, opened 2018-02-08
Updated comments in cliff walking environment

hidehiro98:
I updated the URL of the textbook and page number.
And also I added another source code of cliff walking environment.


pzhokhov:
the github repo https://github.com/ShangtongZhang/reinforcement-learning-an-introduction/blob/master/chapter06/CliffWalking.py
either does not exist or is not public; as such I'll remove reference to it. Otherwise, LGTM

PR #873, opened 2018-02-08
Generic interface to propagate custom commands through wrapper/env stack

glennpow:


gym/core.py
```diff
@@ -320,3 +354,11 @@ def action(self, action):
     def reverse_action(self, action):
         deprecated_warn_once("%s doesn't implement 'reverse_action' method. Maybe it implements deprecated '_reverse_action' method." % type(self))
         return self._reverse_action(action)
+
+
+class CommandWrapper(Wrapper):
```
welinder: what does this wrapper do, other than remove deprecation warnings from Wrapper?

glennpow: Nothing.  I believe that is the point of these "helper" wrappers.  Someone can inherit this class, and only override the "command" method, and ignore the rest.

PR #903, opened 2018-02-22
Car Racing env fixed

qsheeeeen:
After this small change it work. I test it by running this file. Car can be controlled by keyboard.


hortune:
I have just face the same bug...

Issue #859, opened 2018-02-03
Could reraise logic be simplified with six.raise_from() or six.reraise()?


cclauss:
Would it be possible to leverage [six.raise_from() or six.reraise()](https://pythonhosted.org/six/#six.raise_from) to simplify
* https://github.com/openai/gym/blob/master/gym/utils/reraise.py +
* https://github.com/openai/gym/blob/master/gym/utils/reraise_impl_py2.py +
* https://github.com/openai/gym/blob/master/gym/utils/reraise_impl_py3.py ?

pzhokhov:
closing per discussion in https://github.com/openai/gym/pull/906

PR #906, opened 2018-02-25
Use six.raise_from() and six.reraise() to unify implementation

cclauss:
http://six.readthedocs.io/#six.raise_from should allow us to get to a single implementation for both Python 2 and Python 3.  Related to #859


pzhokhov:
re-raise logic is being removed: https://github.com/openai/gym/pull/1342. As such, closing this PR

PR #909, opened 2018-02-25
Fix rgb_array mode for MujocoEnv

welinder:
`get_image()` is not a valid instance method on `MjViewer` anymore.


matthiasplappert:
LGTM.

PR #910, opened 2018-02-26
Update requirements.txt

wagonhelm:
box2d not in requirements


machinaut:
`requirements.txt` is requirements for all of gym.

For specific environments, different requirements are defined in `setup.py`

See: https://github.com/openai/gym/blob/master/setup.py#L11 for the Box2D environment requirements.

The readme documentation: https://github.com/openai/gym#box2d
Shows this as the way to get box2d requirements `pip install -e '.[box2d]'`

PR #911, opened 2018-02-26
Add `reward_range` to FrozenLake-v0

stevenschmatz:


machinaut:
@stevenschmatz are you using reward range in practice?

In any case LGTM.

PR #916, opened 2018-02-28
Brings back compatible function.

wojzaremba:


machinaut:
It'd be nice to get some tests with this.

Like some examples of compatible and incompatible spaces.

joschu:
I don't think this belongs in gym. Why not just write a free function with some if statements?

joschu:
My objection is that it's not clear what "compatible" means, or what it's used for.

pzhokhov:
@wojzaremba @lilianweng is this PR still relevant / useful? If so, I'd be happy to talk about including it / resolving current merge conflicts

PR #836, opened 2018-01-26
Cleanup, removal of unmaintained code

joschu:
Made some aesthetic improvements and removed unmaintained parts of gym. This may seem like a downgrade in functionality, but it is actually a long-needed cleanup in preparation for some great new things that will be released in the next month.

    + Now your `Env` and `Wrapper` subclasses should define `step`, `reset`, `render`, `close`, `seed` rather than underscored method names.
    + Removed the `board_game`, `debugging`, `safety`, `parameter_tuning` environments since they're not being maintained by us at OpenAI. We encourage authors and users to create new repositories for these environments.
    + Changed `MultiDiscrete` action space to range from `[0, ..., n-1]` rather than `[a, ..., b-1]`.
    + No more `render(close=True)`, use env-specific methods to close the rendering.
    + Removed `scoreboard` directory, since site doesn't exist anymore.
    + Moved `gym/monitoring` to `gym/wrappers/monitoring`
    + Add `dtype` to `Space`.
    + Not using python's built-in module anymore, using `gym.logger`

PR #923, opened 2018-03-03
update usage of `MultDiscrete` after #836 commits

chuchuhao:
#836 modify the usage of 'MultiDiscrete`
+ Changed `MultiDiscrete` action space to range from `[0, ..., n-1]` rather than `[a, ..., b-1]`.

### DIFF
--- a/gym/spaces/dict_space.py
+++ b/gym/spaces/dict_space.py
@@ -19,7 +19,7 @@ class Dict(gym.Space):
             )),
             'rear_cam': spaces.Box(low=0, high=1, shape=(10, 10, 3)),
         }),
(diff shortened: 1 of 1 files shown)


gym/spaces/dict_space.py
```diff
@@ -19,7 +19,7 @@ class Dict(gym.Space):
             )),
             'rear_cam': spaces.Box(low=0, high=1, shape=(10, 10, 3)),
         }),
-        'ext_controller': spaces.MultiDiscrete([ [0,4], [0,1], [0,1] ]),
+        'ext_controller': spaces.MultiDiscrete([ 4, 1, 1 ]),
```
pzhokhov: should be [5, 2, 2], as @bzier pointed out

pzhokhov:
fixed in b1aa912, closing this

PR #924, opened 2018-03-03
Fixed boxcar 2d rendering and reset function

LukeWood:
This pull request makes a few changes that fix box2d Car Racing.  These include:

- making the render function names fit the new spec for function naming
- make the render function for humans actually render to the proper location (previously the translation was incorrect and the course rendered only to the bottom left corner of the window)
- also made the window size smaller as 1200/1000 is too big even for the 2017 15 inch macbook if you have your dock enabled

Let me know if I need to change anything else


pzhokhov:
similar fixes are in, closing this PR

PR #932, opened 2018-03-06
Updated Humanoid left_hip_y range to match right_hip_y

stephaniexd11:
right_hip_y has range -110 to 20. left_hip_y should have the same range.


scostrell:
This still seems to be an issue? Can this thread be bumped up?

pzhokhov:
very valid point. Merging.

PR #933, opened 2018-03-06
Box2d carracing: fix render typo; fix 2x display problem; add key to quit the game

Kelym:
1) fix typo in the render function name (similar to other pull requests: [#924](https://github.com/openai/gym/pull/924) [#903](https://github.com/openai/gym/pull/903)). Otherwise cannot run the carracing env. 
2) fix display problem. Previously the game display only occupied the bottom left corner of the window. This is done by specifying viewport. 
3) add a key "esc" to quit the game when human play.

### DIFF
--- a/gym/envs/box2d/car_racing.py
+++ b/gym/envs/box2d/car_racing.py
@@ -43,8 +43,8 @@ STATE_W = 96   # less than Atari 160x192
 STATE_H = 96
 VIDEO_W = 600
 VIDEO_H = 400
-WINDOW_W = 1200
-WINDOW_H = 1000
+WINDOW_W = 600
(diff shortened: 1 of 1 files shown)


Kelym:
I tested it on my machine. ESC key is working. The rendering seems to work also.
(But I don't understand why scaling 2x in glViewPort could fix the rendering bug).

pzhokhov:
the names with underscores vs normal names fixed upstream; in this PR (https://github.com/openai/gym/pull/1339/files) I added quit on close button (X), fixed video recording, and added similar viewport fix. Closing this PR, feel free to reopen (but then you'll need to convince @olegklimov that quit on ESC is really necessary :))

PR #942, opened 2018-03-09
Replace deprecated load(False) with resolve()

Randl:
As per https://github.com/pypa/setuptools/blob/v38.5.2/pkg_resources/__init__.py#L2423 `entry_point.load(False)` is deprecated


AdamGleave:
This change seems unobjectionable, and will get rid of an annoying warning every time an environment is loaded. Can this be merged? @jonasschneider Closes #1219

zuoxingdong:
@pzhokhov It would be nice to merge this fix, because it resolves the annoying warnings for Python 3.7.

Pibastte:
I have removed the env.load(False) by env.resolve() and it s now working fine.

muGithub:
I've tried this,  but it's just get rid of the warning, the gift is still not appeared, but the code runs without errors and warnings.

PR #951, opened 2018-03-15
Add new item to list in environments.md

mpSchrader:
I added a reference to my recently published Sokoban environment in the docs/environments.md. 
I am not 100% sure if this is the right place to add the environment. In case it is not the right place please let me know which is better. ;-)


gdb:
Looks good, thanks!

PR #955, opened 2018-03-16
fix Box dtype in pendulum env

inoryy:
Without this fix you get spammed with annoying logger warnings:
> WARN: gym.spaces.Box autodetected dtype as <class 'numpy.float32'>. Please provide explicit dtype.


pzhokhov:
fixed already

PR #957, opened 2018-03-19
Add _gym_disable_underscore_compat to force-disable compatibility mode

gdb:
Advanced users may have their own _step/_reset methods


joschu:
👍

PR #959, opened 2018-03-20
Headless versions for CartPole, Pendulum and Acrobot

adamtiger:
I have implemented a modified version for each of the mentioned environments. All of them returns the image of the controlled system as the observation instead of the low level states. The motivation behind this:
1. Faster and easier to use than using the render() method with rgb_array mode. 
2. In order to access the image it is not necessary to call the render() method therefore no window will be opened. This has advantage if you run your code on a server and the need for opening a window causes errors.
The implementation draws into a numpy array directly without pyglet. The design is simpler but expressive.

### DIFF
--- a/gym/envs/__init__.py
+++ b/gym/envs/__init__.py
@@ -103,6 +103,12 @@ register(
     max_episode_steps=500,
 )
 
+register(
+    id='AcrobotRawImg-v1',
+    entry_point='gym.envs.classic_control:AcrobotRawImgEnv',
+    max_episode_steps=500,
+)
+
 # Box2d
 # ----------------------------------------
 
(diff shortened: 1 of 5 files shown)


pzhokhov:
Hi @adamtiger ! Apologies for the huge delay. This does look neat; however, as a general policy we are not accepting new environments to the gym. There are two ways out, however - you can create your own project with this environment, and we'll put a reference to it in the README; or you can modify this PR such that instead of new environments current CartPole, Pendulum and Acrobot accept `gym.make` kwarg that tells the env to return pixels instead of simulator state (gym.make supports env kwargs now). I don't want to have two versions of CartPole, Pendulum and Acrobot physics out of general DRY software engineering mantra; also, right now there is no unit tests that guarantee same behavior between your version and existing envs. Thanks!

PR #962, opened 2018-03-22
Remove joints from table to avoid that Fetch can slide it to cheat

### LINKED ISSUE #920: Make the table/big block in Fetch environments fixed.
When I am training HER on FetchPush-v0 the agent sometimes learns to push the big block to achieve it's goal. If you could make it unmovable then the agent would at least not learn such behaviours to achieve the task.

matthiasplappert:
Especially in the FetchSlide task, Fetch sometimes learns to move the table in order to achieve the desired puck position. While entertaining, this is clearly not what Fetch is supposed to do.

Closes #920.


gym/envs/__init__.py
```diff
@@ -306,116 +306,116 @@ def _merge(a, b):
...
-        id='HandReach{}-v0'.format(suffix),
+        id='HandReach{}-v1'.format(suffix),
         entry_point='gym.envs.robotics:HandReachEnv',
         kwargs=kwargs,
         max_episode_steps=50,
     )
 
     register(
-        id='HandManipulateBlockRotateZ{}-v0'.format(suffix),
+        id='HandManipulateBlockRotateZ{}-v1'.format(suffix),
```
wojzaremba: why you bump version of hand environments ?

PR #965, opened 2018-03-25
This Pull Request Creates a New Environment for Car Racing with only 2 Inputs

LukeWood:
For a course that I am helping develop demonstrations with that touches on reinforcement learning we needed a simple environment to demonstrate q learning and visualize the state space.  We thought a great way to do this would be to take the box2d car racing environment and simplify the state down to 2 signals - the distance from the front right of the car to the edge of the track and the fistance from the front left of the car to the track (both on 45 degree angles).

The image below shows the angles where these angles are found:
screen shot 2018-03-25 at 4 56 29 pm

I figured I would offer this up to the community in case anybody else wants this.


LukeWood:
This environment turned out to be way hard with smaller inputs

PR #975, opened 2018-04-04
Added a classic control environment to simulate the cart table model.

chrisspen:


pzhokhov:
Hi @chrisspen ! This looks neat, however, at the moment we are not accepting new environments to the gym. I would recommend creating a separate repo / project with it (assuming you are still interested - apologies for the huge delay); we'll be happy to mention your project in the README of the gym.

PR #994, opened 2018-04-12
specify explicit dtype in Box objects

### LINKED ISSUE #993: MuJoCo env Box initialization causes warnings
In gym/gym/envs/mujoco/mujoco_env.py on line 46, Box object initialization is missing dtype=np.float32 parameter which causes autodetect warning spam.

normster:
fixes #993


smilerz:
The same change needs to be made in envs/classic_control/cartpole.py/cartpole.py at line 41.
`self.observation_space = spaces.Box(-high, high)`
should be:
`self.observation_space = spaces.Box(-high, high,dtype=np.float32)`

PR #996, opened 2018-04-15
Update LunarLander description.

Tautvis:
Update description to include side engine cost.
Otherwise is it a bit misleading since main engine cost is mentioned.


pzhokhov:
good point, thanks!

Issue #980, opened 2018-04-07
Memory leak issue with Box2D environments, e.g, BipedalWalkerHardcore-v2


ruiwang2uber:
Looks like Box2D envs, such as "BipedalWalkerHardcore-v2" is leaking memory whenever we call reset().
```
import gym

env = gym.make("BipedalWalkerHardcore-v2")

while 1:
    env.reset()
```
Other games like Humanoid, run the above script, this memory will not go up, but Box2D envs, such as "BipedalWalkerHardcore-v2", the memory increases very fast.

Could you please help?

ruiwang2uber:
I copy some code out of bipedal_walker.py, did a little bit research:

Code 1: memory increase very fast!
```
#code 1
import gym
print(gym.__file__)
import Box2D
from Box2D.b2 import (edgeShape, circleShape, fixtureDef, polygonShape, revoluteJointDef, contactListener)



HULL_POLY =[
    (-30,+9), (+6,+9), (+34,+1),
    (+34,-8), (-30,-8)
    ]
SCALE = 30.0

world = Box2D.b2World()

def reset():
    hull = world.CreateDynamicBody(
        position = (0, 0),
        fixtures = fixtureDef(
        shape=polygonShape(vertices=[ (x/SCALE,y/SCALE) for x,y in HULL_POLY ]),
        density=5.0,
        friction=0.1,
        categoryBits=0x0020,
        maskBits=0x001,  # collide only with ground
        restitution=0.0)
        )
    world.DestroyBody(hull)


while 1:
    reset()
```
But if i comment out the "fixtures = fixtureDef (...)", the code above runs with no increase in memory

That leads to Code 2, whose memory still increase but in a much slower speed
```
#code 2
import gym
print(gym.__file__)
import Box2D
from Box2D.b2 import (edgeShape, circleShape, fixtureDef, polygonShape, revoluteJointDef, contactListener)



HULL_POLY =[
    (-30,+9), (+6,+9), (+34,+1),
    (+34,-8), (-30,-8)
    ]
SCALE = 30.0

world = Box2D.b2World()

fd = fixtureDef(
        shape=polygonShape(vertices=[ (x/SCALE,y/SCALE) for x,y in HULL_POLY ]),
        density=5.0,
        friction=0.1,
        categoryBits=0x0020,
        maskBits=0x001,  # collide only with ground
        restitution=0.0)

def reset():
    hull = world.CreateDynamicBody(
        position = (0, 0),
        fixtures = fd
        )
[...]

PR #997, opened 2018-04-16
fix a memory issue related to a box2D env in bipedal_walker.py

ruiwang2uber:
https://github.com/openai/gym/issues/980
@olegklimov


olegklimov:
I used this test to confirm the patch works:

```
import gym

env = gym.make("BipedalWalkerHardcore-v2")

while 1:
    import os
    import psutil
    process = psutil.Process(os.getpid())
    print(process.memory_info().rss)
    env.reset()
```

...and it works. Thanks a lot @ruiwang2uber!

PR #1009, opened 2018-04-22
Finish removing `close` as an argument from render

wdoug:
This cleans up some leftover changes after PR #836 which changed `close` to be a method on `Env` and `Wrapper` subclasses instead of being an argument to `render`. There were a few cases where that were missed as part of that pull request and this pull request aims to fix those.


Watesoyan:
Hi, @wdoug, how to close the window by directly clicking the button 'X'?

pzhokhov:
@Watesoyan for most environments you can read out the return value of render (in mode='human'), and if it is False, then close (X) button has been clicked. Then you can call env.close(), which should close the window.

pzhokhov:
@wdoug lgtm

PR #1012, opened 2018-04-22
Implement context handler protocol for Env

tdb-alcorn:
This enables writing
```
import gym

with gym.make(...) as env:
    ...
```
without the need to remember to call `env.close()` later.


pzhokhov:
lgtm

PR #1013, opened 2018-04-23
MuJoCo render (fix #1000)

### LINKED ISSUE #1000: MuJoCo render incompatible with newer versions of MuJoCo Py
MujucoEnv's render method assumes the viewer has a 500x500 display. If it is larger than this, it will select pixels from the bottom-left corner when called with `mode='rgb_array'`. There's a comment which suggests this size used to be the default for mujoco-py, but with the latest mujoco-py it defaults to the screen resolution: see opengl_context.pyx:89 in _create_window.

This issue can be fixed by adding the line `self.viewer.opengl_context.set_buffer_size(W, H)` in `_get_viewer` to manually specify the desired frame size. (Oddly, I need to set both dimensions to twice the desired size, I think due to an overzealous workaround for Mac's on opengl_context.pyx:113.) But this is pretty hacky. 

I'm a little confused why we're rendering anything on the screen for `render(mode='rgb_array')`, which is typically used for video capture in a batch setting. In particular, MjViewer has the annoyance of showing an overlay menu by default, rarely desired for video capture. It might be better to use `mujoco_py.MjRenderContextOffscreen` instead for this purpose. If we allow render to be called with different modes, however, this would require making a breaking change to the interface for `viewer_setup` (or some hacking to make `self.viewer` point to different objects depending on the last call).

I'd be happy to work on a pull request to address these issues if we can agree on a good way forward for this.

AdamGleave:
Fixes #1000 except for not cleaning up the OpenGL context (not possible with current MuJuCo-Py bindings AFAICS). 

I'm not happy with overloading self.viewer as I have done but the alternative would require an API change in MujocoEnv that would force all subclasses to be updated.

### DIFF
--- a/gym/envs/mujoco/mujoco_env.py
+++ b/gym/envs/mujoco/mujoco_env.py
@@ -12,6 +12,8 @@ try:
 except ImportError as e:
     raise error.DependencyNotInstalled("{}. (HINT: you need to install mujoco_py, and also perform the setup instructions here: https://github.com/openai/mujoco-py/.)".format(e))
 
+DEFAULT_SIZE = 500
+
 class MujocoEnv(gym.Env):
     """Superclass for all MuJoCo environments.
     """
@@ -28,6 +30,7 @@ class MujocoEnv(gym.Env):
         self.sim = mujoco_py.MjSim(self.model)
         self.data = self.sim.data
         self.viewer = None
+        self._viewers = {}
 
         self.metadata = {
             'render.modes': ['human', 'rgb_array'],
@@ -78,8 +81,11 @@ class MujocoEnv(gym.Env):
     def reset(self):
         self.sim.reset()
         ob = self.reset_model()
-        if self.viewer is not None:
+        old_viewer = self.viewer
+        for v in self._viewers.values():
+            self.viewer = v
             self.viewer_setup()
+        self.viewer = old_viewer
         return ob
 
     def set_state(self, qpos, qvel):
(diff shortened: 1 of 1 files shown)


matthiasplappert:
I'm okay with this as a reasonable workaround for the current situation but would like to get a second opinion on this if possible (cc @welinder).

jachiam:
I tried it out. It works, but currently wrappers around envs don't support passing args/kwargs through to render, so if you want to use this feature, you have to call render from the wrapped env, which is a bit kludgy. (Modify at https://github.com/openai/gym/blob/master/gym/core.py#L283)

AdamGleave:
Thanks for testing this @jachiam I'm confused how the Wrapper limitation is related to this PR. Is it because I've added additional arguments to render() specifying the size? Note that these arguments are optional, and their default value reproduces the current Gym behavior.

Having said that, it seems reasonable to update the Wrapper class to be more general -- I'm happy to make that change, although it should perhaps be in a separate PR.

jachiam:
Ah, you're right, it should be a separate PR. My mistake. 

Why I brought it up in the first place: the usual way to build gym environments is with gym.make(env_name). Calling this for the MuJoCo envs returns TimeLimit wrapper envs containing the MuJoCo env inside. So if you've built envs this way, and you want to use the additional arguments for render that you've designed, then at your render calls you would have to expose the wrapped env. But I realize now that MujuocoEnv already accepts an argument ('mode') that the wrapper can't pass to it, so this is not a problem you created.

Would be great if you could make that change! :)

AdamGleave:
To clarify, the mode argument is already defined in the Wrapper class and should be passed-through. So the width and height arguments I've introduced do create a "new" problem of inaccessible arguments, however since the width and height were previously hard-coded this is no less expressive than before.

The change is pretty simple though so I've made it in PR #1055

matthiasplappert:
LGTM, thanks for taking the time to work through this and thanks for trying it out @jachiam.

R1ckF:
Hi guys,

Just wanted to inform you of my experience in case other people have similar problems. I finally got it working, but I needed to install Nvidia Cuda Drivers. Without those drivers I would only view what I think is the top left corner of the screen. Now I can see the whole image again.

Issue #1018, opened 2018-04-24
Wrong ordering of x and theta calculations


Rampagy:
`theta _dot` and `x_dot` should be calculated before `x` and `theta` otherwise `x` and `theta` will always be one timestep behind `theta_dot` and `x_dot`.  Below is the order that they should be calculated in.

    x_dot = x_dot + self.tau * xacc
    x  = x + self.tau * x_dot
    theta_dot = theta_dot + self.tau * thetaacc
    theta = theta + self.tau * theta_dot

https://github.com/openai/gym/blob/293eea787a662f501b0e4aab512d3769e830ece2/gym/envs/classic_control/cartpole.py#L63#L66

pzhokhov:
New version is valid, but so is the old one. I have added suggestions to your PR, added a label, and assigned the issue back to you, @Rampagy :)

christopherhesse:
Looks like this was fixed in #1019

PR #1019, opened 2018-04-24
Updated Calculation Order

Rampagy:
Updated order of `x` and `theta` calculations so that they are no longer one timestamp behind `x_dot` and `theta_dot`.

See issue below:
https://github.com/openai/gym/issues/1018

### DIFF
--- a/gym/envs/classic_control/cartpole.py
+++ b/gym/envs/classic_control/cartpole.py
@@ -25,6 +25,7 @@ class CartPoleEnv(gym.Env):
         self.polemass_length = (self.masspole * self.length)
         self.force_mag = 10.0
         self.tau = 0.02  # seconds between state updates
+        self.kinematics_integrator = 'euler'
 
         # Angle at which to fail the episode
         self.theta_threshold_radians = 12 * 2 * math.pi / 360
@@ -60,10 +61,16 @@ class CartPoleEnv(gym.Env):
         temp = (force + self.polemass_length * theta_dot * theta_dot * sintheta) / self.total_mass
         thetaacc = (self.gravity * sintheta - costheta* temp) / (self.length * (4.0/3.0 - self.masspole * costheta * costheta / self.total_mass))
         xacc  = temp - self.polemass_length * thetaacc * costheta / self.total_mass
-        x_dot = x_dot + self.tau * xacc
-        x  = x + self.tau * x_dot
-        theta_dot = theta_dot + self.tau * thetaacc
-        theta = theta + self.tau * theta_dot
+        if self.kinematics_integrator == 'euler':
+            x_dot = x_dot + self.tau * xacc
+            x  = x + self.tau * x_dot
+            theta_dot = theta_dot + self.tau * thetaacc
+            theta = theta + self.tau * theta_dot
+        else: # semi-implicit euler
+            x  = x + self.tau * x_dot
+            x_dot = x_dot + self.tau * xacc
+            theta = theta + self.tau * theta_dot
+            theta_dot = theta_dot + self.tau * thetaacc
         self.state = (x,x_dot,theta,theta_dot)
         done =  x < -self.x_threshold \
                 or x > self.x_threshold \
--- a/gym/envs/classic_control/cartpole.py
+++ b/gym/envs/classic_control/cartpole.py
@@ -60,10 +60,10 @@ class CartPoleEnv(gym.Env):
         temp = (force + self.polemass_length * theta_dot * theta_dot * sintheta) / self.total_mass
         thetaacc = (self.gravity * sintheta - costheta* temp) / (self.length * (4.0/3.0 - self.masspole * costheta * costheta / self.total_mass))
         xacc  = temp - self.polemass_length * thetaacc * costheta / self.total_mass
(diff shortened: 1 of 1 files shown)


Rampagy:
@joschu @pzhokhov I can't figure out how to add reviewers so I'm just tagging you.

gym/envs/classic_control/cartpole.py
```diff
@@ -60,10 +60,10 @@ def step(self, action):
         temp = (force + self.polemass_length * theta_dot * theta_dot * sintheta) / self.total_mass
         thetaacc = (self.gravity * sintheta - costheta* temp) / (self.length * (4.0/3.0 - self.masspole * costheta * costheta / self.total_mass))
         xacc  = temp - self.polemass_length * thetaacc * costheta / self.total_mass
-        x  = x + self.tau * x_dot
         x_dot = x_dot + self.tau * xacc
-        theta = theta + self.tau * theta_dot
+        x  = x + self.tau * x_dot
         theta_dot = theta_dot + self.tau * thetaacc
+        theta = theta + self.tau * theta_dot
```
pzhokhov: why should it be in this new order? To me it would seem that Euler-method-like step should be
x_n+1 = x_n + tau * (xdot_n)
xdot_n+1 = xdot_n + (acc_n) 
etc which corresponded to the old version... Are there some stability issues?

Rampagy: I looked at the permalink and I think even the original had this bug.  It is the order of operations in calculus.  You need to integrate acceleration to get velocity, and integrate velocity to get position.  If you switch up the order and do acceleration -> position -> velocity, the position will always be one timestamp behind because it will have used the previous velocity rather than the current velocity.

So for example if I gave you a math problem where you had the acceleration given you would have to calculate the velocity first to get the position, it's the same thing here.  The reason it works is because on line 57 you are reading in the previous velocity.

https://github.com/openai/gym/blob/293eea787a662f501b0e4aab512d3769e830ece2/gym/envs/classic_control/cartpole.py#L57

joschu: The integration that the code does now is called semi-implicit / symplectic euler integration, and is much more stable than euler integration. Most game physics simulators use this method. This isn't a bug.

joschu: We should probably comment this on the code

joschu: Oops, after looking more closely, it's just normal Euler integration as Peter said, and it's correct.

Rampagy: I am pretty sure they are kinematic equations #1 and #2 with some simplifications.  If they are, the order of the lines should be switched.

https://www.khanacademy.org/science/physics/one-dimensional-motion/kinematic-formulas/a/what-are-the-kinematic-formulas

Rampagy: An airplane accelerates down a runway at 3.20 m/s2 for 32.8 s until is finally lifts off the ground. Determine the distance traveled before takeoff.

If you take the above example, you would have to calculate the velocity before the position using the kinematic equations.  The process is the same as what is being used in the code.

pzhokhov: Per discussion with @joschu - both versions are correct. The old one is the vanilla Euler, and the suggested is semi-implicit Euler. While the new one is more stable, we'd rather not change the behavior of the existing environment. If you'd still like to use semi-implicit Euler, could you add a separate environment flag that turns it on (i.e. by default it is off, and a flag can turn it on)?

Rampagy:
I made the change and pushed it.  Do I have to make a new pull request or anything?

gym/envs/classic_control/cartpole.py
```diff
@@ -60,10 +61,16 @@ def step(self, action):
         temp = (force + self.polemass_length * theta_dot * theta_dot * sintheta) / self.total_mass
         thetaacc = (self.gravity * sintheta - costheta* temp) / (self.length * (4.0/3.0 - self.masspole * costheta * costheta / self.total_mass))
         xacc  = temp - self.polemass_length * thetaacc * costheta / self.total_mass
-        x  = x + self.tau * x_dot
-        x_dot = x_dot + self.tau * xacc
-        theta = theta + self.tau * theta_dot
-        theta_dot = theta_dot + self.tau * thetaacc
+        if self.kinematics_integrator == 'euler':
```
pzhokhov: looks good to me, but I think the if and else blocks need to be swapped. In this context euler is the method when we update position first (using old velocity), then velocity, and semi-implicit euler - when we first update velocity, then position using new velocity

pzhokhov:
No, no need for a new PR, this one is getting updated when you push new commits. I like it save for the comment above. @joshu ?

Rampagy:
Do you want me to change the default setting to semi-implicit then as well?  That way the behaviour is only changed from the previous if they change the option?

pzhokhov:
No, I'd like default to be euler,.so that it behaves like it used to by default (first coordinate is updated , then velocity)

Rampagy:
Okay yea, you're right.  I was getting everything mixed up in my head.  You're 100% correct.

Rampagy:
Do you get notified when I commit?

PR #1020, opened 2018-04-26
Speed up time for `import gym`

gdb:
This change avoids loading Numpy and pkg_resources by default, which collectively eat up about 600-700ms worth of load time on my laptop. While you need both of these loaded to use Gym, often I'm importing gym due to a dependency which imports gym, even though I'm not doing anything with Gym directly.

### DIFF
--- a/gym/__init__.py
+++ b/gym/__init__.py
@@ -9,7 +9,7 @@ from gym.version import VERSION as __version__
 
 from gym.core import Env, GoalEnv, Space, Wrapper, ObservationWrapper, ActionWrapper, RewardWrapper
 from gym.envs import make, spec
-from gym import wrappers, spaces, logger
+from gym import logger
 
 def undo_logger_setup():
     warnings.warn("gym.undo_logger_setup is deprecated. gym no longer modifies the global logging configuration")
--- a/gym/core.py
+++ b/gym/core.py
@@ -1,5 +1,4 @@
 from gym import logger
-import numpy as np
 
 import gym
 from gym import error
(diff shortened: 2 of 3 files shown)


joschu:
The only thing I'm worried about here is that it'll break some code like
`import gym; gym.spaces.Discrete(3)`
which assumes that `spaces` is already in the gym namespace.

On my laptop, 
In [1]: %time import gym
CPU times: user 57.3 ms, sys: 19.7 ms, total: 77 ms
Wall time: 92.6 ms

gdb:
(Are your laptop times with this commit, or without it?)

joschu:
Without it. Though I'm also on an old commit. Lemme update and see if it's still the same time.

joschu:
Yup, still the same with HEAD

gdb:
Without this commit and with a warm cache:

```$ python
>>> import time; start = time.time(); import gym; time.time()-start
0.49982690811157227```

```$ ipython
In [1]: %time import gym
CPU times: user 64.6 ms, sys: 23.8 ms, total: 88.4 ms
Wall time: 120 ms```

Dumping `sys.modules`, it looks like `ipython` imports `pkg_resources` on its own.

td2014:
Hi @gdb @joschu 
I'm seeing that the latest gym build is showing as failed by travis with this PR.  Are there plans to fix the code and recommit, or will the commit be abandoned for now.  Just wondering if I should wait a bit before cloning the repo if a fix is in the works.  Thanks.

pzhokhov:
@td2014 I think the build is broken due to unrelated problem with the build itself. We are working on the fix (I have created an issue about that, https://github.com/openai/gym/issues/1024), in the mean time,  it should be ok to use the latest commit.

gdb:
@pzhokhov Talked to @joschu — let's merge!

Androbin:
@gdb May I ask, what's the dependency referencing gym?

PR #1021, opened 2018-04-26
Allow discrete space to be iterable

EugenHotaj:
It seems intuitive that spaces should be iterable (particularly discrete spaces). For example, if this represents an action space one could iterate as follows:

for action in actions:
  ...

instead of:

for action in range(actions.n):

### DIFF
--- a/gym/spaces/discrete.py
+++ b/gym/spaces/discrete.py
@@ -10,6 +10,7 @@ class Discrete(gym.Space):
     """
     def __init__(self, n):
         self.n = n
+        self._index = -1
         gym.Space.__init__(self, (), np.int64)
     def sample(self):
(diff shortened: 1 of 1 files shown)


episodeyang:
Good idea, non-ideal stateful implementation.

pzhokhov:
this feels vaguely unnecessary... I am not sure statement `for action in action_space:` is more lucid than `for action in range(action_space.n)`. In order for this to make sense, I'd like two things:
1) examples of use cases for box (the iterator can return a random walk)
2) as @episodeyang pointed out, stateless implementation (for instance iterable pattern instead of an iterator, i.e. an iter() method that returns a stateful iterator). Closing this, feel free to reopen with respective changes.

PR #1025, opened 2018-05-01
Allow Car to start at random point when env.reset() is called

dariocazzani:
When training with different policies, I found it useful to be able to let the car start from different parts in the track.


pzhokhov:
I agree with @23pointsNorth 's argument; also, this can now be implemented in a neater way by using `gym.make` kwargs.

pzhokhov:
in fact, let me close this PR; but feel free to reopen one with kwargs for gym.make that would choose if racing car starts at random position or not

PR #1026, opened 2018-05-02
Update play.py

Kautenja:
resolve KeyError from missing key in action dict. use `(dict).get` with a default value of NOP (0).

### DIFF
--- a/gym/utils/play.py
+++ b/gym/utils/play.py
@@ -115,7 +115,7 @@ def play(env, transpose=True, fps=30, zoom=None, callback=None, keys_to_action=N
             env_done = False
             obs = env.reset()
         else:
-            action = keys_to_action[tuple(sorted(pressed_keys))]
+            action = keys_to_action.get(tuple(sorted(pressed_keys)), 0)
             prev_obs = obs
             obs, rew, env_done, info = env.step(action)
             if callback is not None:


Kautenja:
review-requested:@pzhokhov

pzhokhov:
The change seems reasonable, although I am afraid that it is masking a bigger problem... Could you provide more details on the failure? From what I understand of the gym's code, the situation when provided action tuple is not in the action set is rather extraordinary and should not happen

Kautenja:
Yes absolutely. I had originally noticed an issue with an external environment I am developing, but can reproduce the bug with any env. Take pong for example below:

```python
import gym
from gym.utils.play import play
play(gym.make('Pong-v0'))
```

Pong defines the `keys_to_action` mapping `{(): 0, (32,): 1, (100,): 2, (97,): 3, (32, 100): 4, (32, 97): 5}`. When pressing miscellaneous keys on the keyboard (e.g. `K` or `P` or whatever random key _not_ defined in the mapping), the environment behaves predictably with no movement or system crash. However, when pressing `A` and `D` at the same time (e.g. the tuple `(97, 100)`), the script crashes bearing the following error message

```
  File "play.py", line 7, in <module>
    play(env)
  File "/usr/local/lib/python3.6/site-packages/gym/utils/play.py", line 118, in play
    action = keys_to_action[tuple(sorted(pressed_keys))]
KeyError: (97, 100)
```

The proposed fix resolves this error by defaulting to 0 (NOP) when `tuple(sorted(pressed_keys))` is not defined in the `keys_to_action` mapping.

pzhokhov:
I see. Good point, let's merge it then.

Kautenja:
Awesome. who has write access that I can mention to get this merged in?

PR #1039, opened 2018-05-15
Fix debug level in logger

bluddy:
Seems like this is what was intended?


pzhokhov:
fixed here: https://github.com/openai/gym/pull/1112

PR #1042, opened 2018-05-20
 fix sample() function never return lowest case when dtype is not 'f'

rk37:


pzhokhov:
cannot reproduce: `Box(low=0, high=1, shape=(), dtype='int32')`  returns both 0's and 1's to me. Maybe got fixed accidentally in some other PR, closing this.

PR #1046, opened 2018-05-21
fixed docs, if no comma then 'int is not iterable' error

pascalwhoop:


gym/spaces/dict_space.py
```diff
@@ -25,7 +25,7 @@ class Dict(gym.Space):
             'system_checks': spaces.MultiBinary(10),
             'job_status': spaces.Dict({
                 'task': spaces.Discrete(5),
-                'progress': spaces.Box(low=0, high=100, shape=()),
+                'progress': spaces.Box(low=0, high=100, shape=(1,)),
```
pzhokhov: there is technically nothing wrong with shape=(), why changing it?

pzhokhov:
the (3) -> (3,) shape for position and velocity is has been fixed upstream; the () -> (1,) change for progress seems unnecessary. Closing this.

PR #1048, opened 2018-05-23
finish mujoco and robotics envs that can be finished on close()

pzhokhov:


aravind0706:
`hasattr(env.viewer, 'finish')` is `False` for these environments I suppose? 
setting `env.viewer = None` doesn't stop the rendering screen.

pzhokhov:
@aravind0706 if I understand correctly, the suggested fix does not close the environments in question correctly, because their viewer does not have .finish() method? 
If that's the case, I'll need to dig deeper to understand why finish method was removed in the first place (that part was not me :) ) , and how its functionality can be replicated

pzhokhov:
In fact, @aravind0706 , could you provide an example of the code that misbehaves?

pzhokhov:
closing this for inactivity; will reopen if the issue is still being faced.

PR #1060, opened 2018-06-04
LunarLander: use dtype for Boxes, replace assert with clip

olegklimov:
subj


zuoxingdong:
@olegklimov It seems all other classic control environments like CartPole, Acrobot, Pendulum etc. still does not have `dtype=np.float32` explicitly with warning in runtime. Is it better to fix them at once ?

PR #1061, opened 2018-06-04
remove double-wrap warning

unixpickle:
Double wrapping is useful in some cases. For example, if I have a wrapper that modifies rewards, I usually want to put a Monitor before and after the wrapper.


joschu:
👍

PR #1075, opened 2018-06-14
[Update README.rst] New `apt` command. 

zuoxingdong:


pzhokhov:
updated here: https://github.com/openai/gym/pull/1168

PR #1084, opened 2018-06-22
Added explicit dtype to observation_space

madvn:
To suppress warning coming from line 31 of spaces.Box


episodeyang:
@jonasschneider @jietang This one needs your attention as well

PR #1097, opened 2018-07-11
box2d kengz vs py

pzhokhov:
use box2d-py instead of box2d-kengz; and enable tests of environments that use box2d

### DIFF
--- a/gym/envs/tests/spec_list.py
+++ b/gym/envs/tests/spec_list.py
@@ -11,8 +11,6 @@ def should_skip_env_spec_for_tests(spec):
         return True
     if (    'GoEnv' in ep or
             'HexEnv' in ep or
-            ep.startswith('gym.envs.box2d:') or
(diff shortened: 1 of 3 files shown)


christopherhesse:
Looks good, but https://github.com/openai/gym/blob/74229c8dc289e07a515740e3652573bab060ea52/tox.ini references box2d-py directly, but it seems like a test script should use the dependencies specified in setup.py, right?

pzhokhov:
yeah, that's also a good point... let me try to rework tox.ini a bit

christopherhesse:
Nice!  You left pachi-py and doom_py in there, are those unused at this point?

pzhokhov:
more like I did not know what will happen if I remove them. but as we are cleaning up, might as well try and see :)

DuaneNielsen:
Just so you know, I had to go back to box2d-kengz to get the gym environment working.. with the default box2d I get the missing LIMIT_RAND error on mac.

PR #1101, opened 2018-07-16
Rendering fix for envs with outputs with odd-dimensions

EndingCredits:
Due to YUV h264 encoding, image size has to be even. This just forces the input size to be even (by cropping the offending dimension).


pzhokhov:
LGTM

PR #1111, opened 2018-07-24
Add __contains__ to spaces

alok:
Allows us to write expressions like `if x in env.action_space`.


pzhokhov:
please resolve the conflict, otherwise, looks good to me

alok:
Done. Also simplified the contains check to use chained comparison, but feel free to revert that

PR #1114, opened 2018-07-30
Add depth rendering from camera

23pointsNorth:
Allow the mujoco renderer to return the depth image using the `depth_array` mode. 

It follows the `rgb_array` extraction method.

### DIFF
--- a/gym/envs/mujoco/mujoco_env.py
+++ b/gym/envs/mujoco/mujoco_env.py
@@ -33,7 +33,7 @@ class MujocoEnv(gym.Env):
         self._viewers = {}
 
         self.metadata = {
-            'render.modes': ['human', 'rgb_array'],
+            'render.modes': ['human', 'rgb_array', 'depth_array'],
             'video.frames_per_second': int(np.round(1.0 / self.dt))
         }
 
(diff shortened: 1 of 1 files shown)


pzhokhov:
looks good but, could you comment more on why is that needed?

23pointsNorth:
As part of my research I am working on adding visual env for the mujoco [robotics](https://gym.openai.com/envs/#robotics) env. 
Variations of those would include rgb/d/rgbd data as input. 

Given the trend of applying RL for robotics, we should add more of the tools used in that community.

gym/envs/mujoco/mujoco_env.py
```diff
@@ -112,6 +112,13 @@ def render(self, mode='human', width=DEFAULT_SIZE, height=DEFAULT_SIZE):
             data = self._get_viewer(mode).read_pixels(width, height, depth=False)
             # original image is upside-down, so flip it
             return data[::-1, :, :]
+        elif mode == 'depth_array':
+            self._get_viewer(mode).render(width, height)
+            # window size used for old mujoco-py:
+            # Extract depth part of the read_pixels() tuple
+            data = self._get_viewer(mode).read_pixels(width, height, depth=True)[1]
+            # original image is upside-down, so flip it
+            return data[::-1, :]
```
pzhokhov: actually, you lost me now... Why is depth data 2D? Do we still need to flip it upside down?

23pointsNorth: RGB is [h, w, 3] where 3 is for the r, g, b.
Depth I initially assumed is [h, w, 1] , after local testing, it turned out to be [h, w].

pzhokhov: oh got it. Ok.

PR #1137, opened 2018-08-22
Fix: LunarLander env state variable pos.y

RedLeader962:
Am I mistaken or it's a typo? The pos.y state should be over VIEWPORT_H instead of VIEWPORT_W

### DIFF
--- a/gym/envs/box2d/lunar_lander.py
+++ b/gym/envs/box2d/lunar_lander.py
@@ -280,7 +280,7 @@ class LunarLander(gym.Env):
         vel = self.lander.linearVelocity
         state = [
             (pos.x - VIEWPORT_W/SCALE/2) / (VIEWPORT_W/SCALE/2),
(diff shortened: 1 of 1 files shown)


pzhokhov:
seems reasonable to me. @olegklimov, any objections?

olegklimov:
I'm not 100% sure what I meant at the time (maybe scale should be the same for x and y) but there is definitely an inconsistency. On the next line there is velocity scaling and it has _H. So @pzhokhov this can be merged, but please test that heuristic still works. Version increment I think is not really necessary, because 400 or 600 is almost the same.

foxik:
BTW, I think it was an unwise think to do, changing the state observation without bumping version number:
- we had trained agents, which do not work now, given that state observations are different even for the same environment version
- the rewards are different too, because they use reward shaping, which depends on the range of `pos.y` (and so now the final return is ~+50)
- the heuristic was not modified, so it now gives suboptimal results

PR #1143, opened 2018-08-27
tests for lunarlander

pzhokhov:


olegklimov:
You don't have to render when you test.

olegklimov:
Also reward sometimes will be low or negative, because heuristic is not perfect. So you can test for average reward among ~5 episodes (or fix random seed).

PR #1146, opened 2018-08-27
lunarlander add rendering back to test_lunar_lander.py

pzhokhov:


pzhokhov:
oh ok so you want it it lunar_lander.py :) no worries

pzhokhov:
@olegklimov is that good ? Anything else that is not quite right?

olegklimov:
Perfect!

No, not perfect. One thing I still see as a problematic, 'enjoy' usually means trained policy. In this case, it is heuristic, not a trained policy.

pzhokhov:
picky you are, sir :) renamed enjoy_lander into demo_heuristic_lander

Issue #1144, opened 2018-08-27
external PR builds are failing


pzhokhov:
possibly due to missing mujoco keys?

PR #1147, opened 2018-08-28
rm unnecessary `__contains__` duplicate code

alok:
`contains` really should not exist when it does exactly what the builtin
magic method `__contains__` was meant for, but that would break backward
compatibility.

@joschu This implements the change you mentioned.

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -215,6 +215,8 @@ class Space(object):
         """
         raise NotImplementedError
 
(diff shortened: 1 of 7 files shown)


pzhokhov:
lgtm, test is failing due to lack of mujoco key in pr builds (https://github.com/openai/gym/issues/1144)

joschu:
LGTM

PR #1149, opened 2018-08-28
allow nd array for multidiscrete

### LINKED ISSUE #1148: Why is there a restiction on shape in multi discrete?
https://github.com/openai/gym/blob/422c9c7bb3c3c5a756c3b12dfe99733bfbfe3920/gym/spaces/multi_discrete.py#L10

Why is this imposed?
Say one may need a high dimensional space like a (3,3,3) grid with discretised values for each element in the grid. 
It can be very easily circumvented by using np.random.random_sample and passing the shape
Is there some specific reason for doing this?

Also note this example provided in dict_space doesn't work currently.
https://github.com/openai/gym/blob/422c9c7bb3c3c5a756c3b12dfe99733bfbfe3920/gym/spaces/dict_space.py#L22

JoyChopra1298:
Fixes #1148
Currently multidiscrete only supports 1 dimensional array which seems like an unnessecary restriction.

### DIFF
--- a/gym/spaces/multi_discrete.py
+++ b/gym/spaces/multi_discrete.py
@@ -7,10 +7,9 @@ class MultiDiscrete(gym.Space):
         nvec: vector of counts of each categorical variable
         """
         self.nvec = np.asarray(nvec, dtype=np.int32)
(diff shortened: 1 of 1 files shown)


joschu:
LGTM

PR #1151, opened 2018-08-29
add maxwidth parameter to viewer; resize image if it’s larger

joschu:
Some envs produce a very large image when you try to render them.
I modified SimpleImageViewer to resize the image if you try to render too large of an image.

### DIFF
--- a/gym/envs/classic_control/rendering.py
+++ b/gym/envs/classic_control/rendering.py
@@ -310,14 +310,20 @@ class Image(Geom):
 # ================================================================
 
 class SimpleImageViewer(object):
-    def __init__(self, display=None):
+    def __init__(self, display=None, maxwidth=500):
         self.window = None
         self.isopen = False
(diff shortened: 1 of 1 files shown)


pzhokhov:
I like the idea, but I think default value should be None (or something that does not change the behaviour as it is right now). Rationale is as follows - by changing it here the users that were using it through baselines and, for instance, were recording a full-screen video with it or something like that, will be cropped to 500 pixels of width (and will be fairly hard to trace why). I am ok with setting a default maxwidth to 500 where we actually call SimpleImageViewer (VecEnvWrapper most likely).

joschu:
AFAIK, the setting of 500 doesn't change the rendering of any of the existing envs.
If we provide a keyword argument from baselines, then we'll have to bump the required gym version.

joschu:
Anyway, your call Peter.

PR #1167, opened 2018-09-14
pass device_id=-1 when rendering mujoco in rgb_array mode

pzhokhov:


joschu:
I don't know anything about this parameter, but if it works, please go ahead and merge.

colinxs:
It is supposed to correspond to the device used for rendering, where 0 is cpu I believe. Either way, no window should open when using MjRenderContextOffscreen? Perhaps there's a bug internal to MuJoCo.

PR #1174, opened 2018-09-21
remove vertical camera motion on env.reset()

pzhokhov:


jachiam:
LGTM! I added one other tweak for improved user experience. Frequently, when I open up the viewer to watch rollouts, I like to move the camera around to see things from different angles. The old behavior would reset the camera every time ``env.reset()`` was called (whenever the rollout terminated), which was quite frustrating. The new behavior removes viewer reset calls on ``env.reset()``. The viewer reset is now only called when the viewer is initialized.

pzhokhov:
Huh... @jachiam with your change in mujoco_env the other changes are actually not needed (if we only call setup_viewer in env constructor, then camera will not do that weird motion every time env.reset() is called). That's fine with me, but let's also change the docstring here then: 

https://github.com/openai/gym/blob/42f9e14c00c1016250d932367cc96b12f23fbe46/gym/envs/mujoco/mujoco_env.py#L71

jachiam:
I still think the original changes are needed. If a user finds it valuable to reset the camera periodically, it should consistently reset to the same starting position (as caused by your original changes), because this is the behavior that they would expect. I think predictability, even in off-standard use cases, is good. 

Agreed on the recommended docstring change, I'll go make that now.

PR #1177, opened 2018-09-23
Makes Box2D environments pickleable.

jachiam:
Used EzPickle to make Box2D environments pickleable.

### DIFF
--- a/gym/envs/box2d/bipedal_walker.py
+++ b/gym/envs/box2d/bipedal_walker.py
@@ -6,7 +6,7 @@ from Box2D.b2 import (edgeShape, circleShape, fixtureDef, polygonShape, revolute
 
 import gym
 from gym import spaces
-from gym.utils import colorize, seeding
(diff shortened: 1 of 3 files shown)


olegklimov:
Hi Josh. Does it restore state back to a pickled one?

jachiam:
It doesn't - if I understand correctly, EzPickle unpickles by creating a new instance of the Box2D env via the constructor. None of the constructors take any arguments to restore state, so EzPickle couldn't be used to do that. 

But I think this is reasonable behavior, considering that the other main gym environments (MuJoCo and Atari) do the same thing. The only reason it would be bad is if the Box2D envs do some kind of implicit curriculum, where the distribution over random environment parameters changes over training. Are any of them like this?

olegklimov:
> parameters changes over training

No, they are simple.

Well I'm OK with this, let's see if Peter will merge it.

pzhokhov:
LGTM

PR #1182, opened 2018-09-28
Fix RobotEnv rgb_array rendering

mrahtz:
Issue #1000 also applies to `RobotEnv`: when `env.render` is called with `mode='rgb_array'`, currently only a 500x500 square from the bottom-left corner of the image is drawn. We fix it here using the same patch.

(As Issue #1000 notes, this code is shared between `RobotEnv` and `MujocoEnv`. Ideally we would move this to a more abstract class. I'm not sure what the best way of doing this is; for the time being I propose we just put up with a kludgy fix in two separate places.)

### DIFF
--- a/gym/envs/robotics/fetch_env.py
+++ b/gym/envs/robotics/fetch_env.py
@@ -185,3 +185,6 @@ class FetchEnv(robot_env.RobotEnv):
         self.initial_gripper_xpos = self.sim.data.get_site_xpos('robot0:grip').copy()
         if self.has_object:
             self.height_offset = self.sim.data.get_site_xpos('object0')[2]
+
+    def render(self, mode='human', width=500, height=500):
+        return super(FetchEnv, self).render(mode, width, height)
--- a/gym/envs/robotics/hand_env.py
+++ b/gym/envs/robotics/hand_env.py
@@ -47,3 +47,6 @@ class HandEnv(robot_env.RobotEnv):
         self.viewer.cam.distance = 0.5
         self.viewer.cam.azimuth = 55.
         self.viewer.cam.elevation = -25.
+
+    def render(self, mode='human', width=500, height=500):
+        return super(HandEnv, self).render(mode, width, height)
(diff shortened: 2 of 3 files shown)


pzhokhov:
I like this, but ideally the environments for which this is a problem should also be changed to call render with proper sizes (i.e. I don't think it is a responsibility of a user to know what's the proper screen size for an environment). Do you have an example of an environment for which this is a problem? If this is only for external environments, we can merge it as is.

mrahtz:
The environments where I was seeing issues were the Fetch ones.

> ideally the environments for which this is a problem should also be changed to call render with proper sizes

Not sure I got you here - wouldn't it be better to fix it in the abstract class rather than in individual subclasses?

pzhokhov:
What I meant is that the fix should be (as it is in your PR) in the abstract class, however, subclasses of specific environments should, if possible, know what size of the viewer is needed for them (so that the end user does not have to experiment with that), and modify their defaults accordingly

mrahtz:
Oh, I'm with you. The latest commit explicitly specifies the viewer size for the only two subclasses of RobotEnv I can see, FetchEnv and HandEnv. Did a quick test on the two environments to confirm with the following:
```
import gym
import argparse
from pylab import *

ion()

parser = argparse.ArgumentParser()
parser.add_argument('env_id')
args = parser.parse_args()

env = gym.make(args.env_id)
while True:
    env.reset()
    done = False
    while not done:
        obs, reward, done, info = env.step(env.action_space.sample())
        imshow(env.render(mode='rgb_array'))
        pause(1/30)
```

pzhokhov:
Oh now I get what's going on. My bad, I did not realize there is already a fix like that for mujoco environments. Ok, this is good enough - merging.

alirezakazemipour:
> Oh, I'm with you. The latest commit explicitly specifies the viewer size for the only two subclasses of RobotEnv I can see, FetchEnv and HandEnv. Did a quick test on the two environments to confirm with the following:
> 
> ```
> import gym
> import argparse
> from pylab import *
> 
> ion()
> 
> parser = argparse.ArgumentParser()
> parser.add_argument('env_id')
> args = parser.parse_args()
> 
> env = gym.make(args.env_id)
> while True:
>     env.reset()
>     done = False
>     while not done:
>         obs, reward, done, info = env.step(env.action_space.sample())
>         imshow(env.render(mode='rgb_array'))
>         pause(1/30)
> ```
Thank you a billion. It works :heart_eyes:. Helped me accurately.

PR #1188, opened 2018-10-05
Add install instructions for Fedora 28

elguitar:
There were no install instructions other than OSX and Ubuntu. With this PR I would like to add instructions for Fedora 28 as well.


pzhokhov:
actually, now most of the dependencies that require compilation have manylinux-compatible binary wheels, so in principle, explicit installation of system packages should not be needed.

pzhokhov:
closing this, feel free to argue back

PR #1196, opened 2018-10-12
Delete prng.py

zuoxingdong:
Because it seems like this seeding function is rarely used.


pzhokhov:
while I am in principle in favor of this change (I'd rather have prng state in the env than a separate global one) looks like there are more strings attached (judging by the test failures). Also putting @gdb in the loop - is there a reason to have a gym-global prng (vs local to environment)?

gdb:
I think @joschu chose to do it this particular way; I'd originally had the prng always in the environment. I don't recall the exact reasoning.

pzhokhov:
okay then... @zuoxingdong, please fix the test failures, and we can merge.

zuoxingdong:
@pzhokhov Previous failures are fixed. Could you help to check what is the current failure  ?

```bash
gym/envs/algorithmic/tests/test_algorithmic.py .................         [  4%]
gym/envs/box2d/test_lunar_lander.py ..                                   [  5%]
gym/envs/tests/test_determinism.py FERROR: InvocationError for command '/usr/local/gym/.tox/py3/bin/pytest' (exited with code 1)
___________________________________ summary ____________________________________
ERROR:   py3: commands failed
```

pzhokhov:
I think the problem is lack of mujoco license key in the build... Which means tests on all external PR's are failing at the moment. Will fix shortly.

pzhokhov:
should be fixed now, please pull upstream master into your branch to update the PR

zuoxingdong:
@pzhokhov Now it seems okay

zuoxingdong:
@pzhokhov oops, the current error seems from the random seed

zuoxingdong:
What I do now is to create a numpy random state as a class member within `Space`, do you think it can be a good idea to support an optional flag to seed the space object independently ?

Or provide a `seed()` API to the space, so that the user can seed the space object independently. 

In specific either

```python
    def __init__(self, shape=None, dtype=None, seed=0):
        import numpy as np # takes about 300-400ms to import, so we load lazily
        self.shape = None if shape is None else tuple(shape)
        self.dtype = None if dtype is None else np.dtype(dtype)
        
        self.np_random = np.random.RandomState(seed)
```

or 

```python
class Space(object):
...
    def seed(self, seed=0):
        self.np_random = np.random.RandomState(seed)
```

pzhokhov:
I am not sure I understand a use case for the seed() method in the API, could you explain?

gym/core.py
```diff
@@ -1,3 +1,5 @@
+import numpy as np
```
pzhokhov: do we need to import it here? looks like Space.__init__ intentionally uses lazy load of numpy

zuoxingdong:
@pzhokhov The motivation here is the user can set the environment and its space objects with different seeds.

zuoxingdong:
@pzhokhov Now it seems works fine.

zuoxingdong:
@pzhokhov Shall we merge this ?

zuoxingdong:
@pzhokhov Shall we merge this ?

zuoxingdong:
@pzhokhov Conflicts resolved

PR #1203, opened 2018-10-22
fix degree character encoding problem

alcharkov:
SyntaxError: Non-ASCII character '\xc2' in file /home/asc/gym/gym/envs/classic_control/cartpole.py on line 27, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details

while running Atari game on Linux.


pzhokhov:
obsolete now - following discussion in https://github.com/openai/gym/pull/1314

Issue #1191, opened 2018-10-07
Multidiscrete space shape has changed 


araffin:
Since https://github.com/openai/gym/commit/750063055f919d506b3ffa94cadde863a17b7469, the shape returned by Multidiscrete has changed, possibly breaking code (e.g. DummyVecEnv of Baselines).

Before the commit:
```
>> from gym import spaces
>> spaces.MultiDiscrete([10, 10]).shape
 Out: (2,)
```

After the commit:
```
>> from gym import spaces
>> spaces.MultiDiscrete([10, 10]).shape
 Out: ((2,),)
```

PR #1209, opened 2018-11-05
remove extra wrapping of multi_discrete space shape into a tuple

pzhokhov:
solves this: https://github.com/openai/gym/issues/1191
and also failures in baselines: https://github.com/openai/baselines/pull/694


pzhokhov:
I don't think so... In fact, the version of gym we are using internally predates https://github.com/openai/gym/commit/750063055f919d506b3ffa94cadde863a17b7469 (which introduced the bug)

PR #972, opened 2018-03-31
Fix spaces.Box autodetection warning for classic control environments

rlouyang:
Fixes the following warning when making an environment:

> WARN: gym.spaces.Box autodetected dtype as <class 'numpy.float32'>. Please provide explicit dtype.

PR #1234, opened 2018-11-24
Fix autodetect dtype warnings

araffin:
- Remove all remaining warnings due to dtype autodetection
- Add a test to avoid future regression

Fixed:
- Acrobot
- ContinuousMountainCar
- MountainCar
- BidepalWalker / BipedalWalkerHardcore
- GuessingGame
- HotterColder
- KellyCoinFlip

Similar to #972 but with all remaining gym environments.

### DIFF
--- a/gym/spaces/box.py
+++ b/gym/spaces/box.py
@@ -1,4 +1,7 @@
+import warnings
+
 import numpy as np
+
 import gym
 from gym import logger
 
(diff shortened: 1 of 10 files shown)


gym/spaces/box.py
```diff
@@ -28,7 +31,7 @@ def __init__(self, low=None, high=None, shape=None, dtype=None):
                 dtype = np.uint8
             else:
                 dtype = np.float32
-            logger.warn("gym.spaces.Box autodetected dtype as %s. Please provide explicit dtype." % dtype)
+            warnings.warn("gym.spaces.Box autodetected dtype as {}. Please provide explicit dtype.".format(dtype))
```
pzhokhov: I think it would be better to modify logger.warn method to use warnings

pzhokhov:
I like this; other than a single comment - lgtm

gym/spaces/box.py
araffin: You mean replacing the print in https://github.com/openai/gym/blob/master/gym/logger.py#L28 by `warnings.warn` ?
(as you have certainly noticed, I did that edit in order to create a test)

araffin:
@pzhokhov I updated the logger so it should be ok now ;)

PR #1023, opened 2018-04-29
Hotfix/close monitor

hyzhak:
should fix issue https://github.com/openai/gym/issues/1002

PR #1240, opened 2018-11-29
Temporary fix to TypeError thrown when calling env.close() in examples

daniel-s-ingram:


daniel-s-ingram:
I just saw that @hyzhak implemented a much better fix than mine in #1023. Is there a reason his fix has not been merged?

PR #1242, opened 2018-12-02
[Update monitor.py] Fix close() for Monitor wrapper

zuoxingdong:


zuoxingdong:
@pzhokhov @christopherhesse Propose to close this PR since it is fixed in #1321

PR #1246, opened 2018-12-03
Fix input in sim_env

daniel-s-ingram:


goriccardo:
See also #806 for the same fix and #807 for a logic fix of sim_env. Probably fixing sim_env is not a priority.

PR #1247, opened 2018-12-04
Mountain Car Environment No longer always a reward of -1.0

EliCDavis:
Now if the environment reaches a terminal state the reward is no longer -1.0

### DIFF
--- a/gym/envs/classic_control/mountain_car.py
+++ b/gym/envs/classic_control/mountain_car.py
@@ -49,7 +49,7 @@ class MountainCarEnv(gym.Env):
         if (position==self.min_position and velocity<0): velocity = 0
 
         done = bool(position >= self.goal_position)
-        reward = -1.0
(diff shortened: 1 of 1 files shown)


pzhokhov:
I think this is will not really help learning (given that you'll need to reset the environment after it reaches the terminal state), but may change the results of existing benchmarks -- imagine the task, once a good policy is learnt, can be solved in 10 steps. In that case the reward was -10, and now will be -9.  Also, technically, description of classical mountain car (for instance, https://en.wikipedia.org/wiki/Mountain_car_problem) suggests that reward at every time step (not only non-terminal) is -1. Bottom line - in order to introduce this change, a) a strong argument to do so is needed, and 2) it will probably require a new version of MountainCar (so that one could still run learning with old version). Closing for now, but feel free to argue / update the PR to convince me otherwise :)

PR #1249, opened 2018-12-05
small fix to multidiscrete

pzhokhov:


gym/spaces/multi_discrete.py
```diff
@@ -7,13 +7,15 @@ def __init__(self, nvec):
...
         self.nvec = np.asarray(nvec, dtype=np.int32)
-        gym.Space.__init__(self, self.nvec.shape, np.int8)
+        gym.Space.__init__(self, self.nvec.shape, np.uint32)
 
     def sample(self):
         return (gym.spaces.np_random.random_sample(self.nvec.shape) * self.nvec).astype(self.dtype)
 
     def contains(self, x):
-        return (0 <= x).all() and (x < self.nvec).all() and x.dtype.kind in 'ui'
+        # if nvec is int32 and space dtype is unit32, then 0 <= x < self.nvec guarantees that x
```
christopherhesse: There's a typo there for uint32.  Why isn't nvec also uint32?

pzhokhov: good point, we can make nvec uint32 as well

pzhokhov: ah but np.array([-1], dtype=np.unit32) converts signed to unsigned without error, which is probably not the behaviour we want. I'll add an extra check for that.

PR #1260, opened 2018-12-16
Create AcrobotEnv_modified_Reward_1

Roh1t129:
Variable reward depending upon Theta1 and Theta2 instead of a constant -1 reward in non terminal state.
reward = -np.cos(s[0]) - np.cos(s[1] + s[0])  if not terminal else 0.


pzhokhov:
While this looks pretty useful, our general policy is to not add new environments to the gym at the moment. I would recommend creating a separate python package for this (we can add a reference to it in the README page)

PR #1300, opened 2019-01-29
Add rocket lander environment

EmbersArc:
I made this environment a while back and thought it could be added to gym. The aim is to land a SpaceX-style rocket on a platform. More information about its inner workings can be found here: https://github.com/EmbersArc/gym_rocketLander

I've successfully tested and trained it with a variant of PPO. It's a more challenging task for an algorithm, but definitely doable.

Let me know if you want me to make any changes to it first.


pzhokhov:
Hi @EmbersArc! This looks pretty cool; however, we are not expanding the set of the environments in the gym. I would recommend creating and releasing it as a separate package (we'd be happy to add a reference to it to our README page).

PR #1304, opened 2019-02-02
Refactor mujoco envs to support dynamic arguments

hartikainen:
I'm glad to see that gym finally supports dynamic parameterization of the environments (#1301).

We've been using custom parameterizable versions of the basic mujoco environments in our [softlearning codebase](https://github.com/rail-berkeley/softlearning/tree/master/softlearning/environments/gym/mujoco) for a while now, and I thought I would share those environments here as well. The implementations are imo a bit more readable and they support parameterization of the environments via init args. For example, certain environments can be made to not terminate and all the rewards and costs can be tuned via init arguments. With the default parameters, these environments are exact copies of the original environments (I've verified the step and reset functions by testing them against the old implementations).

Currently, the only refactored environments are Ant, Hopper, HalfCheetah, Humanoid, Swimmer, and Walker2d. I'm happy to refactor the rest of the mujoco environments if these seem useful.

### DIFF
--- a/gym/envs/__init__.py
+++ b/gym/envs/__init__.py
@@ -253,8 +253,22 @@ register(
 )
 
 register(
+    id='HalfCheetah-v3',
+    entry_point='gym.envs.mujoco.half_cheetah_v3:HalfCheetahEnv',
+    max_episode_steps=1000,
+    reward_threshold=4800.0,
+)
+
+register(
     id='Hopper-v2',
     entry_point='gym.envs.mujoco:HopperEnv',
+    max_episode_steps=1000,
+    reward_threshold=3800.0,
+)
+
+register(
+    id='Hopper-v3',
+    entry_point='gym.envs.mujoco.hopper_v3:HopperEnv',
     max_episode_steps=1000,
     reward_threshold=3800.0,
 )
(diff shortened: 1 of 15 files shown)


pzhokhov:
this looks neat to me! I am a little hesitant about the exact reproduction of the original mujoco envs - could you add some unit tests to that extent? For instance, instead of replacing the old mujoco envs you could create the -v3 version of them, and add unit tests that compare v2 and v3. If all of them pass, we can create a separate PR removing the non-configurable versions.

hartikainen:
@pzhokhov thanks for the comment. I moved the new environments under v3, and added tests cases to make sure that all the new environments match the corresponding old ones.

pzhokhov:
there were some issues with the tests requiring mujoco; I have added respective skip modifiers and made a new PR: https://github.com/openai/gym/pull/1328. Unfortunately, I squash-merged your branch instead of merging it; which removed commit history. Feel free to either update your PR with the changes (the only changes are in `gym/env/tests/spec_list.py` and `gym/envs/tests/test_mujoco_v2_to_v2_conversion.py`

hartikainen:
@pzhokhov I merged your changed into this branch and at the same time rebased this one from master. Let me know if there's anything else that needs to change.

pzhokhov:
Awesome, thanks! Merging

hartikainen:
Nice, thanks @pzhokhov!

PR #1312, opened 2019-02-07
Make super() calls Python 2 compatible.

### LINKED ISSUE #1311: TypeError creating gym.spaces.Box in Python 2 (gym==0.11.0).
With Python 2 and `gym==0.11.0`, I see an error when running the following. It works with `gym==0.10.11`.

```
$ python --version
Python 2.7.14 :: Anaconda, Inc.

$ python -c "import gym; print(gym.__version__)"
0.11.0
```

```python
from gym.spaces import Box
import numpy as np

Box(0, 1, shape=(3, ), dtype=np.float32)
```

The last line fails with

```
TypeError                                 Traceback (most recent call last)
<ipython-input-3-a5c4264e2cd6> in <module>()
----> 1 Box(0, 1, shape=(3, ), dtype=np.float32)

/Users/rnishihara/anaconda2/lib/python2.7/site-packages/gym/spaces/box.pyc in __init__(self, low, high, shape, dtype)
     35         self.low = low.astype(dtype)
     36         self.high = high.astype(dtype)
---> 37         super().__init__(shape, dtype)
     38         self.np_random = np.random.RandomState()
     39 

TypeError: super() takes at least 1 argument (0 given)
```

robertnishihara:
This should fix #1311. The issue was introduced in #1196. cc @pzhokhov @zuoxingdong 

However, without testing Python 2 in Travis, it could break again in the future.

### DIFF
--- a/gym/spaces/box.py
+++ b/gym/spaces/box.py
@@ -34,7 +34,7 @@ class Box(Space):
             logger.warn("gym.spaces.Box autodetected dtype as {}. Please provide explicit dtype.".format(dtype))
         self.low = low.astype(dtype)
         self.high = high.astype(dtype)
-        super().__init__(shape, dtype)
+        super(Box, self).__init__(shape, dtype)
         self.np_random = np.random.RandomState()
 
     def seed(self, seed):
--- a/gym/spaces/dict_space.py
+++ b/gym/spaces/dict_space.py
@@ -41,7 +41,7 @@ class Dict(Space):
         if isinstance(spaces, list):
             spaces = OrderedDict(spaces)
         self.spaces = spaces
-        super().__init__(None, None) # None for shape and dtype, since it'll require special handling
+        super(Dict, self).__init__(None, None) # None for shape and dtype, since it'll require special handling
 
     def seed(self, seed):
         [space.seed(seed) for space in self.spaces.values()]
--- a/gym/spaces/discrete.py
+++ b/gym/spaces/discrete.py
@@ -12,7 +12,7 @@ class Discrete(Space):
     """
     def __init__(self, n):
         self.n = n
-        super().__init__((), np.int64)
+        super(Discrete, self).__init__((), np.int64)
         self.np_random = np.random.RandomState()
 
     def seed(self, seed):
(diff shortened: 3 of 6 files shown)


pzhokhov:
@robertnishihara I tried setting up python 2.7 CI tests of gym at some point, and they were broken in many ways (other than this particular one). If there is an actual need to use gym with python 2.7, I'd be happy to look more into those.

zuoxingdong:
@pzhokhov sounds great ! IMHO, because of the official status of Python 2 (maintenance until 2020), it might be nice to keep Python 2 specific fixes to e.g. another branch. What do you think ?

pzhokhov:
Oh man I was really hoping you'd say "No, who needs python 2 these days anyways" :) Okay, I'll look into that, in the mean time, let's merge this.

zuoxingdong:
@pzhokhov Haha, it's exactly the words in my mind ! tooo polite man :)

robertnishihara:
@pzhokhov If gym stops supporting Python 2, then the wheels should be tagged as just supporting Python 3, so that `pip2 install gym` pulls the latest version of gym that does support Python 2. E.g., see https://pypi.org/project/ipython/#files for an example.

FWIW, I haven't had any issues with gym in Python 2 up until this one.

I submitted https://github.com/openai/gym/pull/1313 just to see what goes wrong..

pzhokhov:
so as it turns out the gym plays nicely with python 2.7; but the extensions (box2d-py, atari-py, mujoco-py) do not. Almost done with a PR with a bare version of gym being tested with python 2.

pzhokhov:
yup, all good, tests pass and all :) https://github.com/openai/gym/pull/1314

pzhokhov:
also, @robertnishihara, on the subject of wheels - can you mark source distributions (as opposed to binary wheels) as python3-only compatible? We do not release binaries for gym on pypi, only the source code

robertnishihara:
@pzhokhov I'm not sure, but it might be enough to add something like this line https://github.com/numpy/numpy/blob/de1ca61e038548c73ede1e8bd11a9c7dfa02794d/setup.py#L372 to the `setup.py`.

PR #1313, opened 2019-02-07
[DO NOT MERGE] Try adding Python 2 to Travis.

robertnishihara:
Just seeing what goes wrong.. Related to conversation in  #1312.


robertnishihara:
Unfortunately this doesn't seemed to have used Python 2.

pzhokhov:
nope; because the environment in which tests are run is configured in tox.ini. (In fact, you don't need a separate docker image for that). PR is coming up in a few minutes.

pzhokhov:
here: https://github.com/openai/gym/pull/1314

PR #1320, opened 2019-02-11
[Update core.py] Remove underscore methods and its deprecation warning

zuoxingdong:
@pzhokhov This PR removes the deprecation warnings for underscore methods (e.g. _step, _reset), since this change was introduced already quite a while and cleaning up can keep the code cleaner and lightweight.


pzhokhov:
looks good to me, but a bunch of tests are failing with the NotImplementedError

zuoxingdong:
Thanks @pzhokhov , now fixed

christopherhesse:
This changes `seed()` to be an abstract method on env instead of an empty one, which might be fine, but is that going to break a bunch of code?

zuoxingdong:
@christopherhesse this is since as original version will simply return None. However, in general, it might be a good practice to enforce the `seed()` to be explicitly defined. 

The code where `seed()` is undefined will be broken by this change, although it is not the case for the environments in this repo. And perhaps it is also unlikely in the community that other environments are implemented without being able to control the randomness.

gym/core.py
```diff
@@ -262,44 +243,45 @@ def spec(self):
...
-    def step(self, action):
-        observation, reward, done, info = self.env.step(action)
-        return observation, self.reward(reward), done, info
+    def process_observation(self, observation):
+        raise NotImplementedError
 
-    def reward(self, reward):
-        deprecated_warn_once("%s doesn't implement 'reward' method. Maybe it implements deprecated '_reward' method." % type(self))
-        return self._reward(reward)
+    def update_observation_space(self, observation_space):
```
pzhokhov: why is this needed?

gym/core.py
```diff
@@ -262,44 +243,45 @@ def spec(self):
...
-        action = self.action(action)
+        action = self.process_action(action)
         return self.env.step(action)
 
-    def reset(self):
-        return self.env.reset()
+    def process_action(self, action):
+        raise NotImplementedError
+
+    def update_action_space(self, action_space):
```
pzhokhov: why is this function needed?

pzhokhov:
I have two more comments about new functions in ObservationWrapper and RewardWrapper. As for the point that @christopherhesse made - I would say let's keep it working for now and add a deprecation warning (at least for one release). After that, we can delete it. I agree it is @zuoxingdong's way is cleaner; however, we have to respect the users and give them at least some heads-up.

christopherhesse:
@zuoxingdong thanks for making this PR!  I really want the cleanup in here, but breaking `seed` and all existing wrappers is probably too much for this single PR, would you be fine removing the seed change and the change to the wrappers so we can merge in the rest?

zuoxingdong:
Thanks a lot @christopherhesse ! Now I've made the change without breaking `seed`

zuoxingdong:
btw, for the wrappers, do you think it's better to remain the deprecation warning for underscored method ?

pzhokhov:
I think it is ok to remove those and not bother with warnings, but this will merit a new gym version (as technically this breaks backwards compatibility)

PR #1333, opened 2019-02-20
Switch to np.frombuffer in classic_control/rendering.py

abdel:
This fixes the deprecation warning when rendering classic control or Box2D environments:

```
/path/to/gym/envs/classic_control/rendering.py:96: DeprecationWarning: The binary mode of fromstring is deprecated, as it behaves surprisingly on unicode inputs. Use frombuffer instead
  arr = np.fromstring(image_data.data, dtype=np.uint8, sep='')
```

### DIFF
--- a/gym/envs/classic_control/rendering.py
+++ b/gym/envs/classic_control/rendering.py
@@ -93,7 +93,7 @@ class Viewer(object):
         if return_rgb_array:
             buffer = pyglet.image.get_buffer_manager().get_color_buffer()
             image_data = buffer.get_image_data()
-            arr = np.fromstring(image_data.data, dtype=np.uint8, sep='')
+            arr = np.frombuffer(image_data.data, dtype=np.uint8)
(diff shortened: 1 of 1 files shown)


pzhokhov:
while this change seems innocuous and reasonable enough, I am having hard time reproducing this warning. Could you provide more details (numpy version, python version, os version) please? Thanks!

abdel:
> while this change seems innocuous and reasonable enough, I am having hard time reproducing this warning. Could you provide more details (numpy version, python version, os version) please? Thanks!

Python version: 3.6.4
NumPy version: 1.16.1
OS version: macOS 10.14.3

Let me know if you need more details

pzhokhov:
hm... that is very close (matches everything except for python patch number, and I doubt that matters) to my system; but still no warning. @abdel could you also post here the code that you are running? 
Thanks!

pzhokhov:
eh, the reproduction discussion becomes purely academic at this point; let's just merge it :)

PR #1339, opened 2019-02-26
Fixes to car_racing.py (rendering in mode=human and main method)

pzhokhov:


gym/envs/box2d/car_racing.py
```diff
@@ -492,7 +495,7 @@ def key_release(k, mod):
                 #plt.imshow(s)
                 #plt.savefig("test.jpeg")
             steps += 1
-            if not record_video: # Faster, but you can as well call env.render() every time to play full window.
-                env.render()
```
pzhokhov: if statement is removed, as it is hard to record a sensible video if you don't see the car being controlled from the keyboard

olegklimov: IIRC if video is being recorded it already calls render inside, and all environments based on classic.Viewer cannot render without a window.

pzhokhov: hmm.. on my machine that does not actually work (i.e. if I remove env.render() call, nothing is being shown). At the same time, even if it did, we won't be able to close the game with the close (X) button.

olegklimov: re: render() -- ah ok maybe my memories are outdated.

isopen is handled elsewhere as `if still_open==False: break` because if an environment returns nothing the loop should continue to work -- a bit of backward compatibility. It is here only working with CarRacing so no backwards compatibility is necessary, but maybe it's better to be consistent with other places.

gym/envs/box2d/car_racing.py
```diff
@@ -383,14 +383,15 @@ def render(self, mode='human'):
             self.human_render = True
             win.clear()
             t = self.transform
-            gl.glViewport(0, 0, WINDOW_W, WINDOW_H)
+            gl.glViewport(0, 0, 2 * WINDOW_W, 2 * WINDOW_H)
```
Qinusty: This appears to cause issues with the rendering of the environment, the camera is locked to follow the car on the far right of the screen rather than the center. What was the intended purpose of this exact change?

olegklimov: If true, thanks for tracking down changes to a PR!

pzhokhov: this code seems to work somewhat differently on osx and in linux. In osx, without this change, the viewport only occupies lower left corner of the window, whereas in linux I was able to reproduce behaviour described by @Qinusty. Let's revert for now, and I'll investigate what the problem is

pzhokhov: seems to be related to this: https://github.com/vispy/vispy/issues/99

pzhokhov: https://github.com/openai/gym/pull/1372/files

PR #1340, opened 2019-02-28
kwargs on reset for Reward and Action Wrapper

sidmysore:
Was having trouble passing reset arguments through some wrappers and traced it to the ActionWrapper and RewardWrapper not using reset(**kwargs) -- unlike ObservationWrapper for example. The simple fix of just adding that in seems to work for anything I've been doing

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -283,8 +283,8 @@ class ObservationWrapper(Wrapper):
 
 
 class RewardWrapper(Wrapper):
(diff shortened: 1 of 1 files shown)


pzhokhov:
I am going to merge this; however, please keep in mind that there are primitives that still don't support kwargs for reset, for instance, VecEnvs (vectorized environments) in baselines.

PR #1341, opened 2019-03-01
Add verbosity to CarRacing-v0

Qinusty:
While using the CarRacing-v0 environment, stdout is drowned with messages about track generation. This PR allows this to be disabled as follows:

```python
import gym
env = gym.make("CarRacing-v0", verbose=0)
```

The default value for verbose is currently 1 to continue previous behaviour by default.

**EDIT**: Edited for use of kwargs as suggested by @pzhokhov

### DIFF
--- a/gym/envs/box2d/car_racing.py
+++ b/gym/envs/box2d/car_racing.py
@@ -386,7 +386,7 @@ class CarRacing(gym.Env, EzPickle):
             self.human_render = True
             win.clear()
             t = self.transform
-            gl.glViewport(0, 0, 2 * WINDOW_W, 2 * WINDOW_H)
+            gl.glViewport(0, 0, WINDOW_W, WINDOW_H)
             t.enable()
             self.render_road()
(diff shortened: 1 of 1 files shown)


pzhokhov:
now that gym.make supports kwargs, this idea can be more concisely implemented as 
```
import gym
gym.make('CarRacing-v0', verbose=False)
```
@Qinusty would you like to make a respective change of the PR? (the only thing it requires is to add verbose argument to CarRacing class constructor

Qinusty:
> now that gym.make supports kwargs, this idea can be more concisely implemented as
> 
> ```
> import gym
> gym.make('CarRacing-v0', verbose=False)
> ```
> @Qinusty would you like to make a respective change of the PR? (the only thing it requires is to add verbose argument to CarRacing class constructor

Done, hadn't noticed the support for kwargs there.

I've also noticed an issue with the rendering of this environment as of a previous commit. I've detailed this here https://github.com/openai/gym/pull/1339/files#r261850847

PR #1342, opened 2019-03-01
remove reraise logic

pzhokhov:
- replace with re-raising import errors in the only place it is used


olegklimov:
Er... This way more straightforward to me. :+1:

PR #1343, opened 2019-03-01
change car racing render to fix #976

### LINKED ISSUE #976: CarRacing corrupt frames when not rendering visually
When using CarRacing-v0 and not using render() explicitly, the frames from step() are corrupt. Mainly it's always 4 quadrants with a green or gray L shape regardless of what steps I perform (Attached 2 examples).
I've tracked it down to the call to 'win.dispatch_events()' in the render() function of the CarRacing() class, which isn't called in case of mode="state_pixels".

Calling the dispatch_events() even once after creating the window seems to solve it.
Attached also notebook which shows the issue and workaround.

corrupt2
corrupt1
[car_race_consistency.zip](https://github.com/openai/gym/files/1879931/car_race_consistency.zip)

christopherhesse:
The car racing environment doesn't render correctly: https://github.com/openai/gym/issues/976

This PR attempts to fix that by calling dispatch_events in all cases.

I think the better solution to this is to have a "renderer" which renders to an offscreen buffer and a "viewer" which is just what the human uses in `.render(mode='human')` or in the interactive mode you get when using the script.  That's a bit more involved though, so this version still leaves an unused rendering window sitting around.

### DIFF
--- a/gym/envs/box2d/car_racing.py
+++ b/gym/envs/box2d/car_racing.py
@@ -282,7 +282,6 @@ class CarRacing(gym.Env, EzPickle):
         self.tile_visited_count = 0
         self.t = 0.0
         self.road_poly = []
-        self.human_render = False
 
         while True:
             success = self._create_track()
@@ -323,6 +322,7 @@ class CarRacing(gym.Env, EzPickle):
         return self.state, step_reward, done, {}
 
     def render(self, mode='human'):
+        assert mode in ['human', 'state_pixels', 'rgb_array']
         if self.viewer is None:
             from gym.envs.classic_control import rendering
             self.viewer = rendering.Viewer(WINDOW_W, WINDOW_H)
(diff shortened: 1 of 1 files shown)


gym/envs/box2d/car_racing.py
```diff
@@ -355,48 +355,38 @@ def render(self, mode='human'):
...
+        win.clear()
+        t = self.transform
+        if mode=='rgb_array':
+            VP_W = VIDEO_W
+            VP_H = VIDEO_H
+        elif mode == 'state_pixels':
+            VP_W = STATE_W
+            VP_H = STATE_H
+        else:
+            VP_W = 2 * WINDOW_W
```
Qinusty: This appears to cause a rendering issue for human mode as shown here

image

pzhokhov: yeah @christopherhesse could you make the 2x factor being optional and only enabled for osx? 
similar discussion here: https://github.com/openai/gym/pull/1372/files

PR #1344, opened 2019-03-02
import modules when specified to gym.make

christopherhesse:
Right now in order to use `gym.make()` you must first register an environment.  This works fine for environments included in gym itself, but for environments in external packages you must import all packages that you need environments from.  This can look like this: https://github.com/openai/baselines/blob/master/baselines/run.py#L21 and also leaves around a number of unused imports (since they are only used for their side-effects of calling `register()`).

One alternative would be to allow the user to specify the module to import, so instead of `gym.make('gvgai-sokoban-lvl0-v0')` you could do `gym.make('gym_gvgai:gvgai-sokoban-lvl0-v0') `.  The python module being imported must still call register, but this means that it's not a requirement to import the environment package ahead of time, just gym.

### DIFF
--- a/gym/envs/README.md
+++ b/gym/envs/README.md
@@ -90,6 +90,8 @@ ant.AntEnv
       ...
   ```
 
+* After you have installed your package with `pip install -e gym-foo`, you can create an instance of the environment with `gym.make('gym_foo:foo-v0')`
+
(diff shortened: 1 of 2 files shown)


christopherhesse:
What do you think @pzhokhov ?

pzhokhov:
I like this!

PR #1354, opened 2019-03-03
mountain_cartpole.py

Likhitha-k:
Cartpole with a curvy track.


pzhokhov:
while this looks like an interesting environment, our current policy is to not add new environments into the gym. I would recommend creating a separate package / project for it; if you want, we can put a link to this project in our README file.

PR #1355, opened 2019-03-03
__init__.py

Likhitha-k:
Cartpole with a curvy track.


pzhokhov:
This seems to be a shard of this PR: https://github.com/openai/gym/pull/1354

PR #1369, opened 2019-03-06
Create MyEnv.py

nehasingh294:


pzhokhov:
this looks accidental more than an intentional PR (naming, splitting into multiple pieces etc); the code at a glance seems to the the copy of cartpole code. Closing

PR #1370, opened 2019-03-06
Update __init__.py

nehasingh294:


pzhokhov:
see https://github.com/openai/gym/pull/1369

PR #1372, opened 2019-03-06
fix rendering issue on retina

pzhokhov:


gym/envs/box2d/car_racing.py
```diff
@@ -120,6 +121,7 @@ def __init__(self, verbose=1):
 
         self.action_space = spaces.Box( np.array([-1,0,0]), np.array([+1,+1,+1]), dtype=np.float32)  # steer, gas, brake
         self.observation_space = spaces.Box(low=0, high=255, shape=(STATE_H, STATE_W, 3), dtype=np.uint8)
+        self.framebuffer_ratio = 2 if platform.system() == 'Darwin' else 1
```
pzhokhov: technically, this may have an issue on old macs

olegklimov:
I guess we just have to find a better way to detect retina (2x real pixels) or not?

pzhokhov:
fixed in another PR, closing

PR #1375, opened 2019-03-07
spaces equality fixes and tests

pzhokhov:


pzhokhov:
Good point, I'll add respective methods to Dict and Tuple (and also tests)

PR #1376, opened 2019-03-09
better errors for play.py, fixes #1350

### LINKED ISSUE #1350: Cannot use gym.utils.play.PlayPlot
```
from gym.utils.play import play, PlayPlot
def callback(obs_t, obs_tp1, action, rew, done, info):
    return [rew,]
env_plotter = PlayPlot(callback, 30 * 5, ["reward"])
env = gym.make("IceHockey-v0")
```
I got : self.fig, self.ax = plt.subplots(num_plots) NameError: name 'plt' is not defined

christopherhesse:
Better error message for play.py

### DIFF
--- a/gym/utils/play.py
+++ b/gym/utils/play.py
@@ -1,25 +1,18 @@
 import gym
 import pygame
-import sys
-import time
 import matplotlib
 import argparse
+from gym import logger
 try:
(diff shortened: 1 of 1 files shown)


pzhokhov:
@christopherhesse shall we merge this?

PR #1383, opened 2019-03-14
PEP-8 fixes in atari env

instance01:
Remove trailing whitespaces.
Make line breaks adhere to 80 character limit (not all, but quite a few).
Other miscellaneous PEP-8 fixes.

--
Let me know what you think of having a parameter on each line in the constructor and subsequent call to `utils.EzPickle.__init__`, so I know whether that's ok to do in possible future PRs.

### DIFF
--- a/gym/envs/atari/atari_env.py
+++ b/gym/envs/atari/atari_env.py
@@ -8,28 +8,43 @@ from gym.utils import seeding
 try:
     import atari_py
 except ImportError as e:
-    raise error.DependencyNotInstalled("{}. (HINT: you can install Atari dependencies by running 'pip install gym[atari]'.)".format(e))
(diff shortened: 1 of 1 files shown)


pzhokhov:
I like the trailing whitespaces fixes. The one parameter per line in the constructor call - I have no strong feelings; sometimes it does improve readability, sometimes it does not (feel free to use your best judgement :)). I don't like strict adherence to lines of 80 characters long (I feel like in many cases this actually makes code less readable). It is usually ok for test messages though.

joschu:
I like the 80 character width, except some places where it makes the code way less readable, e.g. with quoted strings.
I'm not a big fane of one parameter per line, though -- too much whitespace.

PR #1385, opened 2019-03-14
Update __init__.py

meghaSingh1:


pzhokhov:
@meghaSingh1  this seems unintentional - or a shard of a bigger set of changes. As a rule or thumb, please submit all changes in a single PR (so that the CI build has a chance to succeed). On the other note, we are currently not accepting new environments to the gym; rather, if you are interested in publicizing your environment, I would recommend creating a separate project / package for it. We can add a reference to such project into the environments.md file. Closing this PR and related ones.

PR #1388, opened 2019-03-14
Update __init__.py

meghaSingh1:


meghaSingh1:
register new ennvironment

PR #1396, opened 2019-03-20
[Discrete tasks]: automatic un-nest the scalar action

zuoxingdong:
It can be very convenient for cases that action output from neural network with a batch size one, while still keeping minimal code change.

### DIFF
--- a/gym/envs/atari/atari_env.py
+++ b/gym/envs/atari/atari_env.py
@@ -84,6 +84,8 @@ class AtariEnv(gym.Env, utils.EzPickle):
         return [seed1, seed2]
 
     def step(self, a):
+        if not np.isscalar(a):
+            a = np.asarray(a).item()
         reward = 0.0
         action = self._action_set[a]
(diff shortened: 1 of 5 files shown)


pzhokhov:
while the code change is indeed small, I am not sure I like this very much. On one hand, this can be solved with a wrapper (so no need to change code for all discrete tasks). On the other hand, a better alternative for environments that support batch observations and batch actions are vecenvs (https://github.com/openai/baselines/blob/1b092434fc51efcb25d6650e287f07634ada1e08/baselines/common/vec_env/vec_env.py#L29); if you create a vecenv of size 1, it will be compatible with array actions, but you also get compatibility with larger sizes and process-parallel env.step().  There is actually a decent chance that we will move vecenvs into the gym some time in the future.

zuoxingdong:
@pzhokhov Thanks a lot for the comment ! I agree with you that a more decent solution is to use VecEnv

PR #1398, opened 2019-03-22
[envs/registration.py] Fixed import module

iory:
In current registration.py, there is no `import warnings`. So if warnings called, a program raises an error.


pzhokhov:
nice, thanks! ModuleNotFoundError -  I'd rather just catch ImportError everywhere (if we are maintaining python 2.7 compatibility I feel like that's the cleanest solution, renaming ImportError into ModuleNotFoundError is somewhat hacky)

iory:
Thanks for your advice!
I pushed a new commit considering your comment.
Please check it.

PR #1401, opened 2019-03-22
[Update setup.py] Support latest mujoco_py 2.0

zuoxingdong:


pzhokhov:
hm... let me try this manually - trick is the external pr's don't get injected with license key (and skip the mujoco tests), so the passing travis build does not really tell me if it works :)

pzhokhov:
seems to work (even though now badges are updated to broken for some reason): https://travis-ci.org/openai/gym/builds/510444822. Let's ship it.

PR #1405, opened 2019-03-23
add optional tracking for mujoco locomotion in the rgb rendering mode (issue 1135)

pzhokhov:


pzhokhov:
I was thinking if someone had a pixel-observation policy trained on current version, such policy will stop working if we update always set the tracking flag to True. I agree that tracking rgb observations make more sense; what we can do is hold off on setting default flag to True until we are about to release next (possibly breaking) version of mujoco locomotion environments.

pzhokhov:
on the other hand, if the policy is based on pixel observations from non-tracking camera, it cannot work very well :) Okay, let's change the default version to True in master.

PR #1406, opened 2019-03-24
[Update core.py] Remove unnecessary parts

zuoxingdong:
Changing `.spec` from property to attribute allows modifications

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -6,7 +6,6 @@ from gym.utils import closer
 
 env_closer = closer.Closer()
 
-# Env-related abstractions
 
 class Env(object):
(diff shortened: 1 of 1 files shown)


gym/core.py
```diff
@@ -300,9 +295,6 @@ def step(self, action):
         action = self.action(action)
         return self.env.step(action)
 
-    def reset(self, **kwargs):
```
pzhokhov: wouldn't this cause warnings from default reset method of Wrapper class to show up?

pzhokhov:
other than removal of the reset method, lgtm

zuoxingdong:
Now it's fixed, thanks ! @pzhokhov

PR #1346, opened 2019-03-02
Refactor spaces formatting

zuoxingdong:
@christopherhesse I've extracted from #1319 for formatting-related fixes

PR #1407, opened 2019-03-24
Refactor spaces formatting

zuoxingdong:
updated PR from #1346


zuoxingdong:
Hi @pzhokhov , I was confused how to proceed with resolving conflicts, it seems the right way is to give write access to you as collaborator in my forked repo. Already added. Could you help to have a look if it is possible now ? Because in my page here, the 'Resolve conflicts' button is disabled.

PR #1422, opened 2019-04-02
Bugfix: frozen_lake.py generate_random_map can now reliably create mu…

anuzis:
Bugfix: frozen_lake.py generate_random_map can now reliably create much larger maps (300x300+) without recursion errors.

Context: analyzing RL algorithms time to convergence as state space increases and frozen_lake.py generate_random_map() had a high risk of maximum recursion errors for any grids above 30x30. Current implementation doesn't track a frontier or explored states, and can redundantly check previously explored positions thousands (or millions) of times for sufficiently large grids.

Solution: swapped in proper BFS and DFS implementations to compare performance. Both reliably generate maps of any size, but DFS performs around 20% faster on average for larger grids. Considered A* as a better solution, but it'd require another import (heapq) and probably grow the code more than desired for this simple FrozenLake environment.

Overall, this updated DFS implementation take fewer lines of code while increasing reliability for bigger grid sizes.

### DIFF
--- a/gym/envs/tests/test_frozenlake_dfs.py
+++ b/gym/envs/tests/test_frozenlake_dfs.py
@@ -1,0 +1,33 @@
+import pytest
+import numpy as np
+
+from gym.envs.toy_text.frozen_lake import generate_random_map
+
+# Test that FrozenLake map generation creates valid maps of various sizes.
+def test_frozenlake_dfs_map_generation():
+
+    def frozenlake_dfs_path_exists(res):
+        frontier, discovered = [], set()
+        frontier.append((0,0))
+        while frontier:
+            r, c = frontier.pop()
+            if not (r,c) in discovered:
+                discovered.add((r,c))
+                directions = [(1, 0), (0, 1), (-1, 0), (0, -1)]
+                for x, y in directions:
+                    r_new = r + x
+                    c_new = c + y
+                    if r_new < 0 or r_new >= size or c_new < 0 or c_new >= size:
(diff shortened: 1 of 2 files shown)


anuzis:
Happy to add unit tests, but don't see the best place to do so consistent with the existing code base. I looked for similar examples, but didn't see anything related in:
- gym/envs/tests/
- gym/envs/unittest/
- gym/envs/toy_text/

Any suggestions on where to add them? The closest file I see is /gym/envs/tests/test_envs.py, but there are minimal existing tests for any other environments.  Maybe a new test_frozenlake_generation() test below test_env_render_result_is_immutable() ?

Re - "wouldn't an iterative DFS work as well?": Yes, that's what I implemented.

pzhokhov:
@anuzis looks good to me! I second the opinion about the unit test(s), feel free to add it at any place you see fit (the test runner should be able to find them in any of the locations you mentioned). A suggested spot could be `gym/envs/tests/test_frozenlake_dfs.py`.

anuzis:
Sounds good! Created a test in gym/envs/tests/test_frozenlake_dfs.py as suggested and confirmed it passes locally. The unit test creates maps of sizes [5, 10, 200], and confirms validity.

The unit test includes a redundant DFS method (non-DRY) since it didn't seem apt to use the target code in its own validity test code, and it didn't seem apt for the unit test to not explicitly confirm validity. Open to suggestions if anyone has other ideas.

PR #1424, opened 2019-04-04
Bugfix: wrapper should make spec optional for backward compatibility

yulunli:
`spec` was first required in https://github.com/openai/gym/pull/1406


zuoxingdong:
That sounds good, perhaps a cleaner way is to use default value

```python
getattr(self.env, 'spec', None)
```
What do you think @yulunli ?

yulunli:
I like that. Updated.

yulunli:
What's the next step to merge this change?

PR #1428, opened 2019-04-05
fix car racing pixel scale dtype

christopherhesse:


gym/envs/box2d/car_racing.py
```diff
@@ -369,7 +369,7 @@ def render(self, mode='human'):
         else:
             pixel_scale = 1
             if hasattr(win.context, '_nscontext'):
-                pixel_scale = win.context._nscontext.view().backingScaleFactor()  # pylint: disable=protected-access
+                pixel_scale = int(win.context._nscontext.view().backingScaleFactor())  # pylint: disable=protected-access
```
pzhokhov: I doublt this will ever be a problem in future retina screens, but should we rather convert VP_W and VP_H to integer in case pixel_scale is something like 1.5?

christopherhesse: yeah that's fine too

PR #1435, opened 2019-04-09
[Update dict.py]: Support `__getitem__`

zuoxingdong:


gym/spaces/dict.py
```diff
@@ -60,6 +60,9 @@ def contains(self, x):
                 return False
         return True
 
+    def __getitem__(self, key):
+        return self.spaces[key]
```
OktayGardener: What if `key` is None or does not exist in `self.spaces`? This will break.

pzhokhov: I think if key does not exist in self.spaces it should break - similarly to trying to get the non-existent element from a dictionary. As for `key == None` - why would that be a problem?

PR #1438, opened 2019-04-12
fix for classic_control rendering

johny-c:
I just tried rendering CartPole-v0 with mode 'rgb_array' and I got the error:
```
    arr = env.render(mode=mode)
  File "gym/core.py", line 249, in render
    return self.env.render(mode, **kwargs)
  File "gym/envs/classic_control/cartpole.py", line 188, in render
    return self.viewer.render(return_rgb_array = mode=='rgb_array')
  File "gym/envs/classic_control/rendering.py", line 105, in render
    arr = np.frombuffer(image_data.data, dtype=np.uint8)
AttributeError: 'ImageData' object has no attribute 'data'
```

Changing 
```
arr = np.frombuffer(image_data.data, dtype=np.uint8)
```
to
```
arr = np.frombuffer(image_data.get_data(), dtype=np.uint8)
```
seems to fix it.

### DIFF
--- a/gym/envs/classic_control/rendering.py
+++ b/gym/envs/classic_control/rendering.py
@@ -102,7 +102,7 @@ class Viewer(object):
         if return_rgb_array:
             buffer = pyglet.image.get_buffer_manager().get_color_buffer()
             image_data = buffer.get_image_data()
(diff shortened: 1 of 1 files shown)


johny-c:
Sorry, seems to be caused by new pyglet version 1.4.0b1

Issue #1450, opened 2019-04-19
Implement grayscale mode or wrapper for atari envs


pzhokhov:
along either of the lines suggested in https://github.com/openai/gym/pull/643

PR #1455, opened 2019-04-25
Create atari_preprocessing.py

zuoxingdong:
Related to #1450, rather than changing current API to optionally using ALE to generate gray scale image, it perhaps better to simply have a wrapper to do Atari preprocessings with all-in-one style.


pzhokhov:
this looks good, but for the wrapper with otherwise rather customizable functionality, it would be nice to have grayscale optional (if I understand the logic correctly, currently this wrapper will always produce grayscale images). And tests please - maybe that the generated frames are indeed of the right size, that the grayscale is equivalent to render and then convert to grayscale, etc

zuoxingdong:
@pzhokhov Now the PR is changed accordingly.

PR #1459, opened 2019-04-26
fix #1425

### LINKED ISSUE #1425: Reset in GoalEnv never called
What is the purpose of the reset in GoalEnv base class?

https://github.com/openai/gym/blob/7006c7a182a991b9409c206c56abc121eaffafb5/gym/core.py#L169

If I derive my own env from `GoalEnv`, the base reset is never called - so it never checks observation_space.

Is there an example of when this is actually called?

pzhokhov:


gym/envs/robotics/robot_env.py
```diff
@@ -76,6 +76,7 @@ def reset(self):
         # Gimbel lock) or we may not achieve an initial condition (e.g. an object is within the hand).
         # In this case, we just keep randomizing until we eventually achieve a valid initial
         # configuration.
+        super(GoalEnv, self).reset()
```
christopherhesse: Is this the same as `super().reset()`?

pzhokhov: yes, but only for python 3, so let's keep it as is for python 2.7 compatibility

PR #1460, opened 2019-04-27
Fixed g constant for better accuracy

abhinavsagar:


gym/envs/classic_control/pendulum.py
```diff
@@ -29,7 +29,7 @@ def seed(self, seed=None):
     def step(self,u):
         th, thdot = self.state # th := theta
 
-        g = 10.
+        g = 9.81.
```
pzhokhov: I am not really changing g provides better accuracy (but it will definitely make the environment behave differently). From the update equations for theta and theta dot you can get the same behavior (as with changing g) by scaling l and m, or m and dt. So while it is true that, as written, the pendulum does not simulate the pendulum of length 1m and mass 1 kg and dt = 0.05 s, it does simulate some pendulum :) I feel like having a reproducible pendulum is more important for RL purposes than a calibrated one (after all, a very simple physics integration rule is used, and it is definitely not as precise as possible).

pzhokhov: as such, I see 2 ways forward. 1 - create a new environment, Pendulum-v2, which has dynamics expressed in SI units with the correct gravity constant. 2 - pass g as an argument to the constructor, with the default value being 10 (for compatibility with previous results).

abhinavsagar: @pzhokhov Got it. I feel we should leave this as it is. I am closing this PR and will open the new PendulumV2 in Si units.

PR #1474, opened 2019-05-07
test docker images based on python images

pzhokhov:
allows for testing with various versions of python (relatively independently of os version), simplifies and parallelizes testing logic.


py.Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+    mkdir /root/.mujoco && \
+    cd /root/.mujoco  && \
+    curl -O https://www.roboti.us/download/mjpro150_linux.zip  && \
+    unzip mjpro150_linux.zip && \
+    curl -O https://www.roboti.us/download/mujoco200_linux.zip && \
+    unzip mujoco200_linux.zip && \
+    mv mujoco200_linux mujoco200
+
+ARG MUJOCO_KEY
+ARG PYTHON_VER
```
christopherhesse: This seems to be here twice.

pzhokhov: apparently, FROM overwrites build arguments, so it does have to be here twice... I know, that is a pretty strange behavior :/

PR #1483, opened 2019-05-14
[Wrappers]: add ClipAction

zuoxingdong:
This action wrapper clips the received continuous action. Although this is not always recommended, (e.g. better to use boundary-aware probability distribution to sample action or some tricks like in SAC to output action always lie between [-1, 1] and recover the original action for each dim by a wrapper, say NormalizeAction), but it sometimes quite useful for some standard policy gradient methods like VPG, PPO, and it often improve the performance in certain setups, like Mujoco environments.


gym/wrappers/test_clip_action.py
```diff
@@ -0,0 +1,18 @@
...
+
+    env.reset()
+    clipped_env.reset()
+
+    action = [10000.]
+
+    _, reward, _, _ = env.step(action)
+    _, clipped_reward, _, _ = clipped_env.step(action)
+
+    assert abs(clipped_reward) < abs(reward)
```
pzhokhov: this particular assertion seems to correlate rather weakly with action clipping (also wording clipped_reward seems confusing given that it is the action that is being clipped). If I may suggest an alternative - how about creating two environments and seeding them the same way, then passing an action at the action space boundary to the unwrapped, and outside of action space limits to wrapped, and making sure the results are identical?

pzhokhov: or maybe even create a sequence of actions, some of which can be outside action space boundary, clipping them, and passing clipped sequence to unwrapped env and unclipped sequence to a wrapped env and comparing the results?

zuoxingdong: Hi @pzhokhov , thank you for the comment, it is a great idea ! I've updated the test function accordingly to the second proposal, such that we firstly seed the environments, and generate action sequence with some of the actions out of the boundary, and passing clipped action to original environment and unclipped to the wrapped environment with their rewards compared.

pzhokhov: I am still not satisfied :) If I understand correctly, current test is actually taking the same steps twice on the same environment (because clipped_env.step() calls env.step()) - instead of taking same steps of two *copies* of the environment. The reason why rewards match is because the mountain car produces -1 reward on every step :). Ideally, the *observations* should match

zuoxingdong: Indeed, now new assertion comparing observations is added.

PR #1486, opened 2019-05-14
[Wrappers]: add GrayScaleObservation

zuoxingdong:


gym/wrappers/gray_scale_observation.py
```diff
@@ -0,0 +1,25 @@
+import numpy as np
+import cv2
```
pzhokhov: opencv is not a direct dependency, rather, it is installed when atari_py is installed; so I think we should move opencv into the class definition (otherwise, gym import / wrapper import will fail when opencv is not installed) (kinda like in this commit: https://github.com/openai/gym/commit/98917cf5d3a4808de65254bd8fdf380b0f459a53)

gym/wrappers/test_gray_scale_observation.py
```diff
@@ -0,0 +1,22 @@
+import pytest
+
+import gym
+from gym.wrappers import GrayScaleObservation
+
+
+@pytest.mark.parametrize('env_id', ['Pong-v0', 'SpaceInvaders-v0'])
```
pzhokhov: these tests should be condidtioned on atari_py present

gym/wrappers/test_gray_scale_observation.py
```diff
@@ -0,0 +1,22 @@
...
+    obs = env.reset()
+    wrapped_obs = wrapped_env.reset()
+
+    assert env.observation_space.shape[:2] == wrapped_env.observation_space.shape[:2]
+    if keep_dim:
+        assert wrapped_env.observation_space.shape[-1] == 1
+        assert len(wrapped_obs.shape) == 3
+    else:
+        assert len(wrapped_env.observation_space.shape) == 2
+        assert len(wrapped_obs.shape) == 2
```
pzhokhov: now that we have AtariPreprocessing wrapper merged in, a more detailed test is possible that actually compares the observations and their grayscale counterparts

gym/wrappers/gray_scale_observation.py
zuoxingdong: It's adjusted accordingly to 98917cf

zuoxingdong:
I've run some test, it seems ALE internal `get_Gray_observation` results in different values compared with cv2's transformation RGB->Gray. This will make the test fails. It's also related to [L22-L24 here](https://github.com/openai/gym/blob/98917cf5d3a4808de65254bd8fdf380b0f459a53/gym/wrappers/test_atari_preprocessing.py#L22-L24)

zuoxingdong:
@pzhokhov The testing error is fixed now.

gym/wrappers/test_gray_scale_observation.py
```diff
@@ -0,0 +1,39 @@
...
+    if keep_dim:
+        assert wrapped_env.observation_space.shape[-1] == 1
+        assert len(wrapped_obs.shape) == 3
+        wrapped_obs = wrapped_obs.squeeze(-1)
+    else:
+        assert len(wrapped_env.observation_space.shape) == 2
+        assert len(wrapped_obs.shape) == 2
+
+    # TODO: ALE gray scale has different result than CV2 conversion
+    #assert np.allclose(gray_obs, wrapped_obs)
```
pzhokhov: the AtariPreprocessing tests are enabled in this PR: https://github.com/openai/gym/pull/1652/files. Once that is merged, I think we can try re-enabling the assertion

PR #1487, opened 2019-05-14
[Wrappers]: ResizeObservation

zuoxingdong:


pzhokhov:
comments about opencv and depending on atari-py from https://github.com/openai/gym/pull/1486 for tests apply here as well

gym/wrappers/resize_observation.py
```diff
@@ -0,0 +1,22 @@
+import numpy as np
+import cv2
+
+from gym.spaces import Box
+from gym import ObservationWrapper
+
+
+class ResizeObservation(ObservationWrapper):
+    r"""Downsample the image observation to a square image. """
+    def __init__(self, env, size):
```
hartikainen: Should this accept `shape` instead of `size`?

```
    def __init__(self, env, shape):
	        super(ResizeObservation, self).__init__(env)
        if isinstance(shape, int):
            shape = (shape, shape)
        assert all(x > 0 for x in shape), shape
        self.shape = shape
        ...
```

zuoxingdong:
> comments about opencv and depending on atari-py from #1486 for tests apply here as well

It's updated accordingly.

gym/wrappers/resize_observation.py
zuoxingdong: Hi @hartikainen , thanks for the comment, that's a good idea ! It allows to resize to rectangular frames.

PR #1488, opened 2019-05-14
[Wrappers]: TransformReward

zuoxingdong:


gym/wrappers/scale_reward.py
```diff
@@ -0,0 +1,27 @@
...
+        >>> observation, reward, done, info = env.step(env.action_space.sample())
+        >>> reward
+        0.1
+
+    Args:
+            env (Env): environment
+            scale (float): reward scaling factor
+
+    """
+    def __init__(self, env, scale=0.01):
```
hartikainen: This could also take in an offset. Or be even more general and accept an arbitrary function?

zuoxingdong: It sounds great to me ! Do you think it can be more convenient if we have another wrapper, like `TransformReward` which could take a lambda function as consturctor argument, e.g.

```python
class TransformReward(RewardWrapper):
    def __init__(self, f):
    ...
    def reward(self, reward):
        return self.f(reward)
```

hartikainen: I think that would be convenient.

pzhokhov: I this idea too

zuoxingdong: @pzhokhov @hartikainen Shall we consider either to remain co-existence of `TransformReward` and `{Scale, Clip, Sign}Reward` or we better off having `TransformReward` only and remove `{Scale, Clip, Sign}Reward` ?

pzhokhov: I suppose we can have {Scale, Clip, Sign}Reward using TransformReward, i.e.
```python
ScaleReward = lambda env, scale: TransformReward(env, f = lambda r: r * scale))
...
```

zuoxingdong: @pzhokhov @hartikainen Thanks a lot for the great advice ! Now it is implemented unifying all current reward wrappers as mentioned above.

PR #1491, opened 2019-05-14
[Wrappers]: RescaleAction

zuoxingdong:


gym/wrappers/test_normalize_action.py
```diff
@@ -0,0 +1,17 @@
...
+    env = gym.make('CartPole-v1')
+    with pytest.raises(AssertionError):
+        env = NormalizeAction(env)
+    del env
+
+    env = gym.make('Pendulum-v0')
+    env = NormalizeAction(env)
+    env.reset()
+    with pytest.raises(AssertionError):
+        env.step(10+env.action_space.sample())
```
pzhokhov: nitpicking, but I would prefer a little more testing - say, for pendulum action limits are [-2, 2], so make sure action 1.5 causes an error after wrapping; and before wrapping produces the same observation as action 0.75 after wrapping?

gym/wrappers/normalize_action.py
```diff
@@ -0,0 +1,19 @@
+import numpy as np
+
+import gym
+from gym import spaces
+
+
+class NormalizeAction(gym.ActionWrapper):
+    r"""Rescale the continuous action space of the environment from [-1, 1]. """
```
hartikainen: Should the wrapper allow rescaling to other values than just `-1, 1`?

gym/wrappers/normalize_action.py
```diff
@@ -0,0 +1,19 @@
+import numpy as np
+
+import gym
+from gym import spaces
+
+
+class NormalizeAction(gym.ActionWrapper):
```
hartikainen: Should this be named `RescaleAction`? Normalization is a bit confusing because the rescaling is not data dependent.

zuoxingdong: @hartikainen @pzhokhov I like this idea ! That would allow more general range of the actions that user can produce prior to passing to the environment.

ngc92:
Hey, there seem to be a quite a lot of wrappers proposed now. Some time ago I wrote something similar, but at the time it was decided that the functionality of gym should remain narrow, and it was not included in gym, but just mentioned in the list at docs/wrappers.md.
It does not cover all wrappers that you proposed, but mostly the ones that transform action and observation space.
The code I wrote is currently not compatible with the newest gym version, but I can look into updating it. It might still be worth be looking into, as it provides a more general implementation of state remappings, that can be shared between action and observation space wrappers.

gym/wrappers/normalize_action.py
zuoxingdong: @hartikainen Thanks a lot for the advice, now it is supported for arbitrary range scaling.

gym/wrappers/normalize_action.py
zuoxingdong: @hartikainen @pzhokhov Done

zuoxingdong:
@ngc92 That's great ! Let's work together to iterate the improvements of related wrappers. I believe you also have some great ideas :)

zuoxingdong:
@pzhokhov @hartikainen Now it is updated according to the discussions above.

gym/wrappers/rescale_action.py
```diff
@@ -0,0 +1,28 @@
+import numpy as np
+
+import gym
+from gym import spaces
+
+
+class RescaleAction(gym.ActionWrapper):
+    r"""Rescale the continuous action space of the environment from a range of [a, b]. """
+    def __init__(self, env, a, b):
+        assert isinstance(env.action_space, spaces.Box), 'expected Box action space.'
```
hartikainen: It's generally nice to show the unexpected value for when the assert fails:
```suggestion
        assert isinstance(env.action_space, spaces.Box), (
            "expected Box action space, got {}".format(type(env.action_space)))
```

gym/wrappers/rescale_action.py
```diff
@@ -0,0 +1,28 @@
...
+        assert np.less_equal(a, b).all()
```
hartikainen:
```suggestion
        assert np.less_equal(a, b).all(), (a, b)
```

gym/wrappers/rescale_action.py
```diff
@@ -0,0 +1,28 @@
...
+            self.a = np.full(env.action_space.shape, a, dtype=env.action_space.dtype)
+        else:
+            self.a = a
+        if np.isscalar(b):
+            self.b = np.full(env.action_space.shape, b, dtype=env.action_space.dtype)
+        else:
+            self.b = b
+
+    def action(self, action):
+        assert np.greater_equal(action, self.a).all() and np.less_equal(action, self.b).all()
```
hartikainen:
```suggestion
        assert np.all(np.greater_equal(action, self.a)), (action, self.a)
        assert np.all(np.less_equal(action, self.b)), (action, self.b)
```

gym/wrappers/rescale_action.py
```diff
@@ -0,0 +1,28 @@
...
+from gym import spaces
+
+
+class RescaleAction(gym.ActionWrapper):
+    r"""Rescale the continuous action space of the environment from a range of [a, b]. """
+    def __init__(self, env, a, b):
+        assert isinstance(env.action_space, spaces.Box), 'expected Box action space.'
+        assert np.less_equal(a, b).all()
+        super(RescaleAction, self).__init__(env)
+        if np.isscalar(a):
```
hartikainen: You could squeeze these 4 lines into just 1 line and allow broadcasting with something like:
```suggestion
        self.a = np.zeros(env.action_space.shape, dtype=env.action_space.dtype) + a
```

gym/wrappers/rescale_action.py
```diff
@@ -0,0 +1,28 @@
...
+            self.a = np.full(env.action_space.shape, a, dtype=env.action_space.dtype)
+        else:
+            self.a = a
+        if np.isscalar(b):
```
hartikainen: Same here:
```suggestion
        self.b = np.zeros(env.action_space.shape, dtype=env.action_space.dtype) + b
```

gym/wrappers/rescale_action.py
zuoxingdong: Thanks a lot ! @hartikainen

gym/wrappers/rescale_action.py
```diff
@@ -0,0 +1,24 @@
...
+
+class RescaleAction(gym.ActionWrapper):
+    r"""Rescale the continuous action space of the environment from a range of [a, b]. """
+    def __init__(self, env, a, b):
+        assert isinstance(env.action_space, spaces.Box), (
+            "expected Box action space, got {}".format(type(env.action_space)))
+        assert np.less_equal(a, b).all(), (a, b)
+        super(RescaleAction, self).__init__(env)
+        self.a = np.zeros(env.action_space.shape, dtype=env.action_space.dtype) + a
+        self.b = np.zeros(env.action_space.shape, dtype=env.action_space.dtype) + b
```
pzhokhov: I think the action space has to be modified, otherwise, in the wrapped environment expressions like `env.step(env.action_space.sample())` may cause an error.

gym/wrappers/rescale_action.py
```diff
@@ -0,0 +1,24 @@
+import numpy as np
+
+import gym
+from gym import spaces
+
+
+class RescaleAction(gym.ActionWrapper):
+    r"""Rescale the continuous action space of the environment from a range of [a, b]. """
```
pzhokhov: The docstring sounds a little confusing to me. How about somewhat more wordy version:
```
Rescales the continuous action space of the environment to a range [a,b], such that
RescaleAction(env, a,b).action_space == Box(a,b)
```
?

zuoxingdong: @pzhokhov Thanks for the comment, updated accordingly.

gym/wrappers/rescale_action.py
zuoxingdong: Thanks @pzhokhov , now the change of action space is added.

zuoxingdong:
@pzhokhov Thanks a lot for the comments ! They are updated now.

PR #1492, opened 2019-05-14
[Wrappers]: FlattenObservation

zuoxingdong:


gym/wrappers/flatten_observation.py
```diff
@@ -0,0 +1,8 @@
+from gym.spaces import flatten
+from gym import ObservationWrapper
+
+
+class FlattenObservation(ObservationWrapper):
```
pzhokhov: If we wrap an environment with this wrapper, then `self.observation_space.shape` won't match the shape of observations returned by the env. I think it should also modify the observation space.

gym/wrappers/test_flatten_observation.py
```diff
@@ -0,0 +1,17 @@
+import pytest
+
+import gym
+from gym.wrappers import FlattenObservation
+
+
+@pytest.mark.parametrize('env_id', ['Pong-v0', 'SpaceInvaders-v0'])
```
pzhokhov: same spiel about atari envs in tests as in other wrapper PRs applies here as well

PR #1497, opened 2019-05-18
Fix hand reach v0 pickling

### LINKED ISSUE #1496: EzPickle used incorrectly on some environments
Some environments use `EzPickle` class incorrectly and thus the environment serialization does not result in an expected copy of the environment. An example of such class is the `HandReach-v0`: https://github.com/openai/gym/blob/a488c2fe2fbd4408ba9e193985ec2d86033cab6d/gym/envs/robotics/hand/reach.py#L66

I think the correct usage would be to move the call in the beginning of the `__init__`-method and call
```python
utils.EzPickle.__init__(
  distance_threshold=distance_threshold,
  n_substeps=n_substeps,
  relative_control=relative_control,
  initial_qpos=initial_qpos,
  reward_type=reward_type,
)
```
or alternatively
```python
utils.EzPickle.__init__(**locals())
```
Here's a simple script to verify the issue:
```python
import pickle

import gym


def main():
    env1 = gym.envs.make('HandReach-v0', distance_threshold=1e-6)
    env2 = pickle.loads(pickle.dumps(env1))

    assert env1.distance_threshold == env2.distance_threshold, (
        env1.distance_threshold, env2.distance_threshold)


if __name__ == '__main__':
    main()
```

Running it results in the following:
```
$ python ./tests/test_hand_copy.py
  File "./tests/test_hand_copy.py", line 15, in <module>
    main()
  File "./tests/test_hand_copy.py", line 11, in main
    env1.distance_threshold, env2.distance_threshold)
AssertionError: (1e-06, 0.01)
```

hartikainen:
Adds some tests cases for hand manipulation environments and fixes the `HandReach-v0` pickling. The other manipulations are not quite as easy to fix due to the inheritance between classes.

Partly fixes #1496


christopherhesse:
It looks like the tests are still failing, also they seem to use both unittest and pytest.  It looks like we only use pytest: https://github.com/openai/gym/blob/master/py.Dockerfile#L29 so probably unittest should not be used here.

PR #1498, opened 2019-05-19
Allow observation and action spaces to be customized in MujocoEnv

hartikainen:
Currently, environments that use dictionary observation spaces and extend `MujocoEnv` fail because the observation spaces are hard-coded to be of type `spaces.Box`. This PR will make things a bit more flexible by:
* Adding new _set_{observation,_action}_space methods for `MujocoEnv`,
  allowing customized spaces.
* Implementing automatic handling of dict observation spaces.

### DIFF
--- a/gym/envs/mujoco/mujoco_env.py
+++ b/gym/envs/mujoco/mujoco_env.py
@@ -1,11 +1,12 @@
+from collections import OrderedDict
 import os
+
 
 from gym import error, spaces
 from gym.utils import seeding
 import numpy as np
 from os import path
 import gym
-import six
 
 try:
     import mujoco_py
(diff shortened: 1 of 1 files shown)


gym/envs/mujoco/mujoco_env.py
```diff
@@ -40,20 +58,26 @@ def __init__(self, model_path, frame_skip, rgb_rendering_tracking=True):
...
+        action = self.action_space.sample()
+        observation, _reward, done, _info = self.step(action)
         assert not done
-        self.obs_dim = observation.size
 
+        self._set_observation_space(observation)
+
+        self.seed()
+
+    def _set_action_space(self):
```
pzhokhov: what is the purpose of putting this in a separate method vs keeping it in a constructor? Is it expected to be called or overridden in subclasses?

hartikainen: Yeah, I think it would be nice to provide the ability for the subclasses to override the default behavior of these functions.

PR #1499, opened 2019-05-20
Feature/pixel observation wrapper

### LINKED ISSUE #1374: Implement pixel observation wrapper
along the lines of discussion https://github.com/openai/gym/pull/740

hartikainen:
This implements a `PixelObservationWrapper` that allows observations to be augmented with pixel values. The wrapper by default uses the wrapped environment's render method (with `mode='rgb_array'`) to generate the pixels. The interface for the wrapper is very similar to [dm_control suite's pixel wrapper](https://github.com/deepmind/dm_control/blob/master/dm_control/suite/wrappers/pixels.py).

Closes #1374


gym/wrappers/pixel_observation.py
```diff
@@ -0,0 +1,111 @@
...
+        elif self._observation_is_dict:
+            self.observation_space = copy.deepcopy(wrapped_observation_space)
+        else:
+            self.observation_space = spaces.Dict()
+            self.observation_space.spaces[STATE_KEY] = wrapped_observation_space
+
+        # Extend observation space with pixels.
+        pixels = self.env.render(**render_kwargs)
+
+        if np.issubdtype(pixels.dtype, np.integer):
```
hartikainen: I'm not sure about these observation types. I guess typically the observation space bounds are (-inf, inf), but maybe in the pixel observations' case we could do better?

gym/envs/mujoco/mujoco_env.py
```diff
@@ -101,12 +101,22 @@ def do_simulation(self, ctrl, n_frames):
         for _ in range(n_frames):
             self.sim.step()
 
-    def render(self, mode='human', width=DEFAULT_SIZE, height=DEFAULT_SIZE):
+    def render(self,
+               mode='human',
+               width=DEFAULT_SIZE,
+               height=DEFAULT_SIZE,
+               camera_id=None,
```
hartikainen: I think it's important to be able to specify the camera name for the pixel observation, which is why added `camera_id` and `camera_name` as arguments here.

gym/envs/mujoco/mujoco_env.py
```diff
@@ -101,12 +101,22 @@ def do_simulation(self, ctrl, n_frames):
...
         if mode == 'rgb_array':
-            camera_id = None 
-            camera_name = 'track'
-            if self.rgb_rendering_tracking and camera_name in self.model.camera_names:
-                camera_id = self.model.camera_name2id(camera_name)
+            if camera_id is None or camera_name is None:
+                raise ValueError("Both `camera_id` and `camera_name` cannot be"
+                                 " specified at the same time.")
+
+            no_camera_specified = camera_name is None and camera_id is None
```
hartikainen: This should maintain the backwards compatibility. I don't know if it's needed though. Could we just get rid of the `rgb_rendering_tracking`?

gym/wrappers/pixel_observation.py
```diff
@@ -0,0 +1,111 @@
...
+        if pixels_only:
+            self.observation_space = spaces.Dict()
+        elif self._observation_is_dict:
+            self.observation_space = copy.deepcopy(wrapped_observation_space)
+        else:
+            self.observation_space = spaces.Dict()
+            self.observation_space.spaces[STATE_KEY] = wrapped_observation_space
+
+        # Extend observation space with pixels.
+        pixels = self.env.render(**render_kwargs)
```
hartikainen: Should this be
```python
pixels = self.env.get_pixel_observation(**render_kwargs)
```
instead and then `gym.Env.get_pixel_observation` could by default be:
```python
def get_pixel_observation(*args, **kwargs):
    pixels = self.render(*args, **kwargs)
```

This way the user could implement other ways of returning pixel observations.

gym/wrappers/pixel_observation.py
```diff
@@ -0,0 +1,111 @@
...
+    def _add_pixel_observation(self, observation):
+        if self._pixels_only:
+            observation = collections.OrderedDict()
+        elif self._observation_is_dict:
+            observation = type(observation)(observation)
+        else:
+            observation = collections.OrderedDict()
+            observation[STATE_KEY] = observation
+
+        pixels = self.env.get_pixels(**self._render_kwargs)
```
hartikainen: This one currently only supports a single pixel observation. In some cases we might want multiple observations (e.g. for multiple cameras). Maybe `get_pixels` should return a dictionary that gets merged to the observation. Something like:
```python
pixel_observation = self.env.get_pixels(**self._render_kwargs)
observation.update(pixel_observation)
```

gym/core.py
```diff
@@ -60,11 +60,14 @@ def step(self, action):
     def reset(self):
         """Resets the state of the environment and returns an initial observation.
 
-        Returns: 
+        Returns:
             observation (object): the initial observation.
         """
         raise NotImplementedError
 
+    def get_pixels(self, *args, **kwargs):
```
pzhokhov: why is this necessary (vs using render directly)?

gym/envs/mujoco/mujoco_env.py
pzhokhov: Get rid if `rgb_rendering_tracking` - i.e. use `camera_name='track'` instead? Yes, we can do that (probably will merit a new version release, but otherwise I am not opposed

gym/wrappers/pixel_observation.py
```diff
@@ -0,0 +1,111 @@
...
+
+from gym import spaces
+from gym import ObservationWrapper
+
+STATE_KEY = 'state'
+
+
+class PixelObservationWrapper(ObservationWrapper):
+    """Augment observations by pixel values."""
+    def __init__(self,
```
pzhokhov: I think a mode that does not use dictionary observations and returns only pixels would also be useful

pzhokhov:
Agree with @vitchyr on the first point about get_pixels vs render -  I don't see a reason a compelling reason to put a new function into the gym API - looks like every call to get_pixels can be substibuted by a call to render. My other concern is that while even though Dict observations are supported more and more, there is still little consistency, for instance, with Dict keys that correspond to observations/states/pixels; so it would be handy to also have a mode which produces only pixels as an image observation (not a Dict)

hartikainen:
Thanks for the comments @vitchyr and @pzhokhov.

> Overall, it seems odd to me that this wrapper re-uses the render function, but that this PR also defines a get_pixels interface. It seems like we want one or the other: either render is the only way to get images from the mujoco environments, or get_pixels and render are two completely functions with two isolated implementations (calling get_pixels doesn't trigger and of the rendering code). This distinction wouldn't be that important if it weren't for the fact that calling render has a number of side-effects (e.g. setting the self.viewer).

My initial implementation had the pixel wrapper to pass `render_kwargs` directly to the `env.render()` function, but I later change it to the `get_pixels` api because I thought it would allow a bit more flexibility for the wrapper. An example of the needed flexibility would be a case where the environment needs to provide 2 different images from two different cameras as the observation.

> calling render has a number of side-effects (e.g. setting the self.viewer)

This is actually a problem with the gym (mujoco) environment api and calling the `env.sim.render` itself shouldn't have any side effects (please correct me if I'm wrong here). If I understand correctly, the only side effect is the setting of `self.viewer` in the [_MujocoEnv._get_viewer](https://github.com/openai/gym/blob/master/gym/envs/mujoco/mujoco_env.py#L131-L141). If we think it's a problem, in my opinion,we should fix the viewer handling to not have side effects rather than force the implementation of the pixel wrapper.

> Also, I think it's troubling that PixelObservationWrapper.__init__ takes in render_kwargs but still needs to call self.env.get_pixels(**render_kwargs) to get the pixel size.

The only place where I use the shapes from the `self.env.get_pixels` is at the [time of setting the observation space in the wrapper](https://github.com/openai/gym/pull/1499/files#diff-6335cae9b0b38dbc4cc98629e3093918R87) (please correct me if you meant something else). I do agree that it's kinda weird that the two can not match currently. The right thing would probably be to just sanity check that the pixel shapes match what we asked for before observation space. Does that sounds good to you?

> Agree with @vitchyr on the first point about get_pixels vs render - I don't see a reason a compelling reason to put a new function into the gym API - looks like every call to get_pixels can be substibuted by a call to render.

This is a reasonable point, and I guess the answer here depends on the purpose of the pixel wrapper. Do we want the pixel wrapper to be the standard way of converting state-based environments into pixel-based environments? If so, then I think the api needs to be flexible enough to enable things like the above-mentioned multi-camera case, and I don't think directly calling `env.render` will suffice. If, on the other hand, the goal is just to provide a less flexible convenience wrapper that works for most of the simple use-cases, then I do agree that we should get rid of the `env.get_pixels`.

> My other concern is that while even though Dict observations are supported more and more, there is still little consistency, for instance, with Dict keys that correspond to observations/states/pixels; so it would be handy to also have a mode which produces only pixels as an image observation (not a Dict)

I can write support for this when I got some free time again.

gym/core.py
hartikainen: See https://github.com/openai/gym/pull/1499#issuecomment-495817496.

pzhokhov:
@hartikainen why would call to `render` (more precisely, multiple calls with different `camera` arguments not be sufficient in for a multi-camera use case? Something like:
```python
env = gym.make('Humanoid-v1')
multicamera_pixels_env = PixelObservationWrapper(env, cameras=['static', 'tracking'])

class PixelObservationWrapper(Wrapper)
  ...
  def _add_pixel_observation(self, observation)
      ...
     for camera in cameras:
       observation[camera] = self.env.render(camera=camera)
...
```

hartikainen:
@pzhokhov sorry for taking so long to get back to this. After using the pixel observation wrapper for a while, I do agree that the `get_pixels` api is unnecessary. I made the wrapper a bit more flexible but changing the `render_kwargs` to support dictionaries that map `pixel_key`s to `render_kwargs`. The `env.render` is then called multiple times with different `render_kwargs` and the pixels added to the observation with the corresponding key. Let me know what you think.

gym/envs/mujoco/mujoco_env.py
hartikainen: Done. I added a deprecation warning for it for now.

gym/wrappers/pixel_observation.py
hartikainen: I think this might not be necessary since you could use the `FlattenDictWrapper`. What do you think?

gym/envs/mujoco/mujoco_env.py
```diff
@@ -36,7 +36,14 @@ class MujocoEnv(gym.Env):
...
     """
 
-    def __init__(self, model_path, frame_skip, rgb_rendering_tracking=True):
+    def __init__(self, model_path, frame_skip, rgb_rendering_tracking=None):
+
+        if rgb_rendering_tracking is not None:
+            print("`rgb_rendering_tracking` is deprecated and will be removed"
+                  " in a future version of Gym. To maintain the old behavior,"
+                  " you can omit the `camera_id` and `camera_name` in the"
+                  " `env.render()` method.")
```
pzhokhov: this message not, technically, quite accurate - I if understand the logic in render method, omitting camera_id and camera_name in render arguments is the same as `rgb_rendering_tracking=True`. However, if the execution got to printing the message above, chances are `rgb_rendering_tracking=False` (because the opposite is default and does not need to be specified explicitly). 
I think we can delete the whole rgb_rendering_tracking logic entirely (and make a new version release out of this).

hartikainen: Sounds good to me! Changed in the latest commit.

PR #1513, opened 2019-06-02
Vectorized environments

tristandeleu:
One of the most useful feature in [`openai/baselines`](https://github.com/openai/baselines/) is the ability to run multiple environments in parallel with the [`SubprocVecEnv`](https://github.com/openai/baselines/blob/7c520852d9cf4eaaad326a3d548efc915dc60c10/baselines/common/vec_env/subproc_vec_env.py#L36) wrapper. However in order to benefit from this feature, it is necessary to install the whole baselines package, which includes some heavy dependencies like Tensorflow. To avoid this dependency, starting a new project often requires to [copy/paste the snippet of code for `SubprocVecEnv`](https://github.com/search?l=Python&q=SubprocVecEnv%2BVecEnv&type=Code). I am proposing to add this functionality in Gym, so that this feature can be used directly out of the box.

**TL;DR**
```python
import gym
env = gym.vector.make('CartPole-v1', num_envs=5)
observations = env.reset()
# array([[-0.04071225, -0.00272771,  0.04585911,  0.03397448],
#        [ 0.04807664, -0.00466015,  0.03836329, -0.04897803],
#        [ 0.03999503,  0.03953767, -0.02474664, -0.00865895],
#        [-0.03199483,  0.01612429, -0.03704762, -0.02390875],
#        [ 0.00967298,  0.01544605,  0.04391582,  0.0040252 ]],
#       dtype=float32)
```

 - Adding a base class `VectorEnv`, inheriting from `gym.Env`. This ensures that the vectorized environments are still valid instances of `gym.Env`. In particular, the `step` method still returns a tuple `(observations, rewards, dones, infos)`. However, the difference is that `rewards` and `dones` are now arrays (instead of scalars).
 - Adding `SyncVectorEnv` and `AsyncVectorEnv`. Roughly speaking, `SyncVectorEnv` is equivalent to [`DummyVecEnv`](https://github.com/openai/baselines/blob/7c520852d9cf4eaaad326a3d548efc915dc60c10/baselines/common/vec_env/dummy_vec_env.py#L5), and `AsyncVectorEnv` is a unified API for both [`SubprocVecEnv`](https://github.com/openai/baselines/blob/7c520852d9cf4eaaad326a3d548efc915dc60c10/baselines/common/vec_env/subproc_vec_env.py#L36) and [`ShmemVecEnv`](https://github.com/openai/baselines/blob/7c520852d9cf4eaaad326a3d548efc915dc60c10/baselines/common/vec_env/shmem_vec_env.py#L20).
   - `SyncVectorEnv` runs the environments serially.
   - `AsyncVectorEnv` uses `multiprocessing`, and runs the environments in parallel.
 - `AsyncVectorEnv` has the possibility to use shared memory across processes, to improve the efficiency of data transfer if observations are large (e.g. images). It supports all types of Gym spaces, including `Dict` and `Tuple`.
 - `AsyncVectorEnv` includes a low level control over the `reset` and `step` operations. In particular, these methods include an optional `timeout` argument, to shutdown the processes after a certain period, to avoid any hanging process.
 - The convenience function `gym.vector.make()` is the equivalent of `gym.make()`, with multiple copies of the environment wrapped in a vectorized environment.
```python
import gym

env = gym.vector.make('BreakoutDeterministic-v4', num_envs=8)
env.seed(0)
observations = env.reset()
# observations.shape: (8, 210, 160, 3)

actions = env.action_space.sample()
observations, rewards, dones, infos = env.step(actions)
# rewards: array([0., 0., 0., 0., 0., 0., 0., 0.])
# dones: array([False, False, False, False, False, False, False, False])
```

EDIT:
 - Added `max_retries` argument in `AsyncVectorEnv`. This automatically restarts processes that failed, up to a certain amount of maximum retries.

Bonus:
```python
import gym

env = gym.vector.make('HandReach-v0', num_envs=5)
env.observation_space
# Dict(achieved_goal:Box(5, 15), desired_goal:Box(5, 15), observation:Box(5, 63))
observations = env.reset()

observations.keys()
# odict_keys(['achieved_goal', 'desired_goal', 'observation'])
observations['observation'].shape
# (5, 63)
```

### DIFF
--- a/gym/vector/__init__.py
+++ b/gym/vector/__init__.py
@@ -1,0 +1,47 @@
+from gym.vector.async_vector_env import AsyncVectorEnv
+from gym.vector.sync_vector_env import SyncVectorEnv
+from gym.vector.vector_env import VectorEnv
+
+__all__ = ['AsyncVectorEnv', 'SyncVectorEnv', 'VectorEnv', 'make']
+
+def make(id, num_envs=1, asynchronous=True, **kwargs):
+    """Create a vectorized environment from multiple copies of an environment,
+    from its id
+
+    Parameters
+    ----------
+    id : str
+        The environment ID. This must be a valid ID from the registry.
+
+    num_envs : int
+        Number of copies of the environment. If `1`, then it returns an
+        unwrapped (i.e. non-vectorized) environment.
+
+    asynchronous : bool (default: `True`)
+        If `True`, wraps the environments in an `AsyncVectorEnv` (which uses 
+        `multiprocessing` to run the environments in parallel). If `False`,
+        wraps the environments in a `SyncVectorEnv`.
+
+    Returns
+    -------
+    env : `gym.vector.VectorEnv` instance
+        The vectorized environment.
+
+    Example
+    -------
+    >>> import gym
+    >>> env = gym.vector.make('CartPole-v1', 3)
+    >>> env.reset()
+    array([[-0.04456399,  0.04653909,  0.01326909, -0.02099827],
+           [ 0.03073904,  0.00145001, -0.03088818, -0.03131252],
+           [ 0.03468829,  0.01500225,  0.01230312,  0.01825218]],
+          dtype=float32)
+    """
+    from gym.envs import make as make_
+    def _make_env():
+        return make_(id, **kwargs)
+    if num_envs == 1:
+        return _make_env()
+    env_fns = [_make_env for _ in range(num_envs)]
+
+    return AsyncVectorEnv(env_fns) if asynchronous else SyncVectorEnv(env_fns)
--- a/gym/vector/async_vector_env.py
+++ b/gym/vector/async_vector_env.py
@@ -1,0 +1,347 @@
+import numpy as np
+import multiprocessing as mp
+import time
+
+from gym import logger
+from gym.vector.vector_env import VectorEnv
+from gym.error import (AlreadySteppingError, AlreadyResettingError,
+                       NotSteppingError, NotResettingError, ClosedEnvironmentError)
+from gym.vector.utils import (create_shared_memory, create_empty_array,
+                              write_to_shared_memory, read_from_shared_memory,
+                              concatenate, CloudpickleWrapper, clear_mpi_env_vars)
+
+
+class AsyncVectorEnv(VectorEnv):
+    """Vectorized environment that runs multiple environments in parallel. It
+    uses `multiprocessing` processes, and pipes for communication.
+
+    Parameters
+    ----------
+    env_fns : iterable of callable
+        Functions that create the environments.
+
+    observation_space : `gym.spaces.Space` instance, optional
+        Observation space of a single environment. If `None`, then the
+        observation space of the first environment is taken.
+
+    action_space : `gym.spaces.Space` instance, optional
+        Action space of a single environment. If `None`, then the action space
+        of the first environment is taken.
+
+    shared_memory : bool (default: `True`)
+        If `True`, then the observations from the worker processes are
+        communicated back through shared variables. This can improve the
+        efficiency if the observations are large (e.g. images).
+
+    copy : bool (default: `True`)
+        If `True`, then the `reset` and `step` methods return a copy of the
+        observations.
+
+    context : str, optional
+        Context for multiprocessing. If `None`, then the default context is used.
+    """
+    def __init__(self, env_fns, observation_space=None, action_space=None,
+                 shared_memory=True, copy=True, context=None):
+        ctx = mp.get_context(context)
+        self.env_fns = env_fns
+        self.shared_memory = shared_memory
+        self.copy = copy
+
+        if (observation_space is None) or (action_space is None):
+            dummy_env = env_fns[0]()
+            observation_space = observation_space or dummy_env.observation_space
+            action_space = action_space or dummy_env.action_space
+            dummy_env.close()
+            del dummy_env
+        super(AsyncVectorEnv, self).__init__(num_envs=len(env_fns),
+            observation_space=observation_space, action_space=action_space)
+
+        if self.shared_memory:
+            _obs_buffer = create_shared_memory(self.single_observation_space,
+                n=self.num_envs)
+            self.observations = read_from_shared_memory(_obs_buffer,
+                self.single_observation_space, n=self.num_envs)
+        else:
+            _obs_buffer = None
+            self.observations = create_empty_array(self.single_observation_space,
+                n=self.num_envs, fn=np.empty)
+
+        self.parent_pipes, self.processes = [], []
+        target = _worker_shared_memory if self.shared_memory else _worker
+        with clear_mpi_env_vars():
+            for idx, env_fn in enumerate(self.env_fns):
+                parent_pipe, child_pipe = ctx.Pipe()
+                process = ctx.Process(target=target,
+                    name='Worker<{0}>-{1}'.format(type(self).__name__, idx),
+                    args=(idx, CloudpickleWrapper(env_fn), child_pipe,
+                    parent_pipe, _obs_buffer))
+
+                self.parent_pipes.append(parent_pipe)
+                self.processes.append(process)
+
+                process.deamon = True
+                process.start()
+                child_pipe.close()
+
+        self.waiting_reset = False
+        self.waiting_step = False
+
+    def seed(self, seeds=None):
+        self._assert_is_running()
+        if seeds is None:
+            seeds = [None for _ in range(self.num_envs)]
+        if isinstance(seeds, int):
+            seeds = [seeds + i for i in range(self.num_envs)]
+        assert len(seeds) == self.num_envs
+
+        for pipe, seed in zip(self.parent_pipes, seeds):
+            pipe.send(('seed', seed))
+        for pipe in self.parent_pipes:
+            pipe.recv()
+
+    def reset_async(self):
+        self._assert_is_running()
+        if self.waiting_step:
+            logger.warn('Calling `reset_async` while waiting for a pending '
+                'call to `step` to complete. Cancelling the previous call to '
+                '`step`...')
+            try:
+                self.step_wait(timeout=0)
+            except mp.TimeoutError:
+                pass
+
+        if self.waiting_reset:
+            logger.error('Calling `reset_async` while waiting for a pending '
+                'call to `reset` to complete. Closing `{0}`...'.format(self))
+            self.close(terminate=True)
+            raise AlreadyResettingError('Calling `reset_async` while waiting '
+                'for a pending call to `reset` to complete.')
+
+        for pipe in self.parent_pipes:
+            pipe.send(('reset', None))
+        self.waiting_reset = True
+
+    def reset_wait(self, timeout=None):
+        """
+        Parameters
+        ----------
+        timeout : int, optional
+            Number of seconds before the call to `reset_wait` times out. If
+            `None`, the call to `reset_wait` never times out.
+
+        Returns
+        -------
+        observations : sample from `observation_space`
+            A batch of observations from the vectorized environment.
+        """
+        self._assert_is_running()
+        if self.waiting_step or (not self.waiting_reset):
+            logger.error('Calling `reset_wait` without any prior call to '
+                '`reset_async`. Closing `{0}`...'.format(type(self).__name__))
+            self.close(terminate=True)
+            raise NotResettingError('Calling `reset_wait` without any prior '
+                'call to `reset_async`.')
+
+        if not self.poll(timeout):
+            self.waiting_reset = False
+            raise mp.TimeoutError('The call to `reset_wait` has timed out after '
+                '{0} second{1}.'.format(timeout, 's' if timeout > 1 else ''))
+
+        observations_list = [pipe.recv() for pipe in self.parent_pipes]
+        self.waiting_reset = False
+
+        if not self.shared_memory:
+            concatenate(observations_list, self.observations,
+                self.single_observation_space)
+
+        return np.copy(self.observations) if self.copy else self.observations
+
+    def step_async(self, actions):
+        """
+        Parameters
+        ----------
+        actions : iterable of samples from `action_space`
+            List of actions.
+        """
+        self._assert_is_running()
+        if self.waiting_reset:
+            logger.warn('Calling `step_async` while waiting for a pending '
+                'call to `reset` to complete. Waiting for the previous call '
+                'to `reset`...')
+            try:
+                self.reset_wait()
+            except mp.TimeoutError:
+                pass
+
+        if self.waiting_step:
+            logger.error('Calling `step_async` while waiting for a pending '
+                'call to `step` to complete. Closing `{0}`...'.format(
+                type(self).__name__))
+            self.close(terminate=True)
+            raise AlreadySteppingError('Calling `step_async` while waiting for '
+                'a pending call to `step` to complete.')
+
+        for pipe, action in zip(self.parent_pipes, actions):
+            pipe.send(('step', action))
+        self.waiting_step = True
+
+    def step_wait(self, timeout=None):
+        """
+        Parameters
+        ----------
+        timeout : int, optional
+            Number of seconds before the call to `step_wait` times out. If
+            `None`, the call to `step_wait` never times out.
+
+        Returns
+        -------
+        observations : sample from `observation_space`
+            A batch of observations from the vectorized environment.
+
+        rewards : `np.ndarray` instance (dtype `np.float_`)
+            A vector of rewards from the vectorized environment.
+
+        dones : `np.ndarray` instance (dtype `np.bool_`)
+            A vector whose entries indicate whether the episode has ended.
+
+        infos : list of dict
+            A list of auxiliary diagnostic informations.
+        """
+        self._assert_is_running()
+        if self.waiting_reset or not self.waiting_step:
+            logger.error('Calling `step_wait` without any prior call to '
+                '`step_async`. Closing `{0}`...'.format(type(self).__name__))
+            self.close(terminate=True)
+            raise NotSteppingError('Calling `step_wait` without any prior call '
+                'to `step_async`.')
+
+        if not self.poll(timeout):
+            self.waiting_step = False
+            raise mp.TimeoutError('The call to `step_wait` has timed out after '
+                '{0} second{1}.'.format(timeout, 's' if timeout > 1 else ''))
+
+        results = [pipe.recv() for pipe in self.parent_pipes]
+        self.waiting_step = False
+        observations_list, rewards, dones, infos = zip(*results)
+
+        if not self.shared_memory:
+            concatenate(observations_list, self.observations,
+                self.single_observation_space)
+
+        return (np.copy(self.observations) if self.copy else self.observations,
+                np.array(rewards), np.array(dones, dtype=np.bool_), infos)
+
+    def poll(self, timeout=None):
+        self._assert_is_running()
+        if timeout is not None:
+            end_time = time.time() + timeout
+        delta = None
+        for pipe in self.parent_pipes:
+            if timeout is not None:
+                delta = max(end_time - time.time(), 0)
+            if not pipe.poll(delta):
+                break
+        else:
+            return True
+        return False
+
+    def close_extras(self, timeout=None, terminate=False):
+        timeout = 0 if terminate else timeout
+        try:
+            if self.waiting_reset:
+                logger.warn('Calling `close` while waiting for a pending '
+                    'call to `reset` to complete.')
+                self.reset_wait(timeout)
+
+            if self.waiting_step:
+                logger.warn('Calling `close` while waiting for a pending '
+                    'call to `step` to complete.')
+                self.step_wait(timeout)
+        except mp.TimeoutError:
+            terminate = True
+
+        if terminate:
+            for process in self.processes:
+                if process.is_alive():
+                    process.terminate()
+        else:
+            for pipe in self.parent_pipes:
+                pipe.send(('close', None))
+            for pipe in self.parent_pipes:
+                pipe.recv()
+
+        for pipe in self.parent_pipes:
+            pipe.close()
+        for process in self.processes:
+            process.join()
+
+    def _assert_is_running(self):
+        if self.closed:
+            raise ClosedEnvironmentError('Trying to operate on `{0}`, after a '
+                'call to `close()`.'.format(type(self).__name__))
+
+    def __del__(self):
+        self.close(terminate=True)
+
+
+def _worker(index, env_fn, pipe, parent_pipe, shared_memory):
+    assert shared_memory is None
+    env = env_fn()
+    parent_pipe.close()
+    try:
+        while True:
+            command, data = pipe.recv()
+            if command == 'reset':
+                observation = env.reset()
+                pipe.send(observation)
+            elif command == 'step':
+                observation, reward, done, info = env.step(data)
+                if done:
+                    observation = env.reset()
+                pipe.send((observation, reward, done, info))
+            elif command == 'seed':
+                env.seed(data)
+                pipe.send(None)
+            elif command == 'close':
+                pipe.send(None)
+                break
+            else:
+                raise RuntimeError('Received unknown command `{0}`. Must '
+                    'be one of {`reset`, `step`, `seed`, `close`}.'.format(command))
+    except KeyboardInterrupt:
+        pass
+    finally:
+        env.close()
+
+
+def _worker_shared_memory(index, env_fn, pipe, parent_pipe, shared_memory):
+    assert shared_memory is not None
+    env = env_fn()
+    observation_space = env.observation_space
+    parent_pipe.close()
+    try:
+        while True:
+            command, data = pipe.recv()
+            if command == 'reset':
+                observation = env.reset()
+                write_to_shared_memory(index, observation, shared_memory,
+                                       observation_space)
+                pipe.send(None)
+            elif command == 'step':
+                observation, reward, done, info = env.step(data)
+                write_to_shared_memory(index, observation, shared_memory,
+                                       observation_space)
+                pipe.send((None, reward, done, info))
+            elif command == 'seed':
+                env.seed(data)
+                pipe.send(None)
+            elif command == 'close':
+                pipe.send(None)
+                break
+            else:
+                raise RuntimeError('Received unknown command `{0}`. Must '
+                    'be one of {`reset`, `step`, `seed`, `close`}.'.format(command))
+    except KeyboardInterrupt:
+        pass
+    finally:
+        env.close()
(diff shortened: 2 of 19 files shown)


zuoxingdong:
@tristandeleu That looks really great ! One potential issue with proposed API to me is that what if the user would like to wrap the "atomic" environment at first before wrapping them into a vectorized environments ?

tristandeleu:
Thank you! The `gym.vector.make` function is meant for very simple cases, where you don't want to wrap the atomic environments. If you want to use wrappers around the environments, you can still use the lower level `SyncVectorEnv` and `AsyncVectorEnv`. For example:
```python
import gym
from gym.vector import AsyncVectorEnv
from gym.wrappers import AtariPreprocessing

def make_env(env_id):
    def _make():
        env = gym.make(env_id)
        env = AtariPreprocessing(env)
        return env
    return _make

env_fns = [make_env('BreakoutDeterministic-v4') for _ in range(5)]
env = AsyncVectorEnv(env_fns)
observations = env.reset()
# observations.shape: (5, 84, 84)
```

tristandeleu:
I added the `episodic` argument to both `SyncVectorEnv` and `AsyncVectorEnv` to override the default behavior of calling `reset` after an environment finishes an episode. The `episodic` argument is also available in the `gym.vector.make` function. This allows you to do something like
```python
import gym
import numpy as np

env = gym.vector.make('CartPole-v1', num_envs=8, episodic=True)
observations = env.reset()
dones = np.zeros(env.num_envs, dtype=np.bool_)

while not dones.all():
    actions = env.action_space.sample()
    observations, rewards, dones, infos = env.step(actions)
```

pzhokhov:
@tristandeleu Thanks, this looks great! In fact we have bounced this idea internally for a while, great to see the initiative taken by the community :)  Given the size of the PR, however, I'd like to take more time to read through it carefully; also putting @christopherhesse into the loop.

christopherhesse:
Having a standard vectorized environment in gym would be really nice. I'll write a few comments on this PR.

gym/vector/__init__.py
```diff
@@ -0,0 +1,56 @@
+from gym.vector.async_vector_env import AsyncVectorEnv
+from gym.vector.sync_vector_env import SyncVectorEnv
+from gym.vector.vector_env import VectorEnv
+
+__all__ = ['AsyncVectorEnv', 'SyncVectorEnv', 'VectorEnv', 'make']
+
+def make(id, num_envs=1, asynchronous=True, episodic=False, **kwargs):
```
christopherhesse: When is episodic=True used?  I can't reset individual environments, so I'd have to reset them all at once right?

gym/vector/async_vector_env.py
```diff
@@ -0,0 +1,482 @@
...
+
+    observation_space : `gym.spaces.Space` instance, optional
+        Observation space of a single environment. If `None`, then the
+        observation space of the first environment is taken.
+
+    action_space : `gym.spaces.Space` instance, optional
+        Action space of a single environment. If `None`, then the action space
+        of the first environment is taken.
+
+    shared_memory : bool (default: `True`)
```
christopherhesse: combining shared memory vec env with subproc is great!

gym/vector/async_vector_env.py
```diff
@@ -0,0 +1,482 @@
...
+        Only available in Python 3.
+    """
+    def __init__(self, env_fns, observation_space=None, action_space=None,
+                 shared_memory=True, copy=True, max_retries=0,
+                 episodic=False, context=None):
+        try:
+            self.ctx = mp.get_context(context)
+        except AttributeError:
+            logger.warn('Context switching for `multiprocessing` is not '
+                'available in Python 2. Using the default context.')
```
christopherhesse: Python 2.7 compat, nice!

gym/vector/async_vector_env.py
```diff
@@ -0,0 +1,482 @@
...
+            self.ctx = mp
+        self.env_fns = env_fns
+        self.shared_memory = shared_memory
+        self.copy = copy
+        self.max_retries = max_retries
+        self.episodic = episodic
+        self._num_retries = 0
+
+        if (observation_space is None) or (action_space is None):
+            dummy_env = env_fns[0]()
```
christopherhesse: Creating the dummy_env can be problematic, it would be better to use the subproc approach where the child is queried for the spaces: https://github.com/openai/baselines/blob/master/baselines/common/vec_env/subproc_vec_env.py#L61

gym/vector/async_vector_env.py
```diff
@@ -0,0 +1,482 @@
...
+
+        for pipe, seed in zip(self.parent_pipes, seeds):
+            pipe.send(('seed', seed))
+        for pipe in self.parent_pipes:
+            pipe.recv()
+
+    def reset_async(self):
+        self._assert_is_running()
+        if self.waiting_step:
+            logger.warn('Calling `reset_async` while waiting for a pending '
```
christopherhesse: Is there any reason to not make this a fatal error?

gym/vector/async_vector_env.py
```diff
@@ -0,0 +1,482 @@
...
+                '`step`...')
+            try:
+                self.step_wait(timeout=0)
+            except mp.TimeoutError:
+                pass
+
+        if self.waiting_reset:
+            logger.error('Calling `reset_async` while waiting for a pending '
+                'call to `reset` to complete. Closing `{0}`...'.format(self))
+            self.close(terminate=True)
```
christopherhesse: Why is closed called here? Shouldn't that be the job of the caller?

gym/vector/async_vector_env.py
```diff
@@ -0,0 +1,482 @@
...
+        if self.waiting_step or (not self.waiting_reset):
+            logger.error('Calling `reset_wait` without any prior call to '
+                '`reset_async`. Closing `{0}`...'.format(type(self).__name__))
+            self.close(terminate=True)
+            raise NotResettingError('Calling `reset_wait` without any prior '
+                'call to `reset_async`.')
+
+        if not self.poll(timeout):
+            self.waiting_reset = False
+            raise mp.TimeoutError('The call to `reset_wait` has timed out after '
```
christopherhesse: If this is going to use TimeoutError, why not use the python builtin instead of the multiprocessing one?

gym/vector/async_vector_env.py
christopherhesse: This also applies to all the other cases where it looks like the user called the methods out of order.  Rather than trying to fix it, this should just throw an exception so the user can fix their code.

gym/vector/async_vector_env.py
```diff
@@ -0,0 +1,482 @@
...
+            A vector of rewards from the vectorized environment.
+
+        dones : `np.ndarray` instance (dtype `np.bool_`)
+            A vector whose entries indicate whether the episode has ended.
+
+        infos : list of dict
+            A list of auxiliary diagnostic informations.
+        """
+        self._assert_is_running()
+        if self.waiting_reset or not self.waiting_step:
```
christopherhesse: rather than dealing with a bunch of booleans, you can use a single state variable like this: https://github.com/cshesse/libenv/blob/master/libenv/libenv.py#L102

gym/vector/async_vector_env.py
```diff
@@ -0,0 +1,482 @@
...
+    def _restart_if_errors(self, reset=True):
+        if not self.error_queue.empty():
+            while not self.error_queue.empty():
+                index, exctype, value = self.error_queue.get()
+                logger.error('Received the following error from Worker-{0}: '
+                    '{1}: {2}'.format(index, exctype.__name__, value))
+
+                if (self.max_retries is not None) \
+                        and (self._num_retries >= self.max_retries):
+                    for pipe in self.parent_pipes:
```
christopherhesse: Why does this read from all the parent pipes?

gym/vector/async_vector_env.py
```diff
@@ -0,0 +1,482 @@
...
+        for pipe in self.parent_pipes:
+            if timeout is not None:
+                delta = max(end_time - time.time(), 0)
+            if not pipe.poll(delta):
+                break
+        else:
+            return True
+        return False
+
+    def close_extras(self, timeout=None, terminate=False):
```
christopherhesse: The existing subprocvecenv doesn't seem to need the `terminate` or `timeout` arguments.  While I could see an argument for the timeout one, why add `terminate`?

gym/vector/async_vector_env.py
```diff
@@ -0,0 +1,482 @@
...
+                            observation = env.reset()
+                pipe.send((observation, reward, done, info))
+            elif command == 'seed':
+                env.seed(data)
+                pipe.send(None)
+            elif command == 'close':
+                pipe.send(None)
+                break
+            elif command == '_check_observation_space':
+                try:
```
christopherhesse: This seems odd, can this check be improved?

gym/vector/async_vector_env.py
```diff
@@ -0,0 +1,482 @@
...
+                    pipe.send(observation)
+                else:
+                    infos = {'AsyncVectorEnv.restart': True}
+                    pipe.send((observation, 0., True, infos))
+            else:
+                raise RuntimeError('Received unknown command `{0}`. Must '
+                    'be one of {`reset`, `step`, `seed`, `close`, '
+                    '`_check_observation_space`, `_restart`}.'.format(command))
+    except Exception:
+        import sys
```
christopherhesse: Imports should go at the top of the file

gym/vector/sync_vector_env.py
```diff
@@ -0,0 +1,113 @@
...
+        self.episodic = episodic
+
+        if (observation_space is None) or (action_space is None):
+            observation_space = observation_space or self.envs[0].observation_space
+            action_space = action_space or self.envs[0].action_space
+        super(SyncVectorEnv, self).__init__(num_envs=len(env_fns),
+            observation_space=observation_space, action_space=action_space)
+
+        self._check_observation_spaces()
+        self.observations = create_empty_array(self.single_observation_space,
```
christopherhesse: Please use numpy.zeros instead of numpy.empty since empty does not zero out memory.

gym/vector/sync_vector_env.py
```diff
@@ -0,0 +1,113 @@
...
+
+        return (np.copy(self.observations) if self.copy else self.observations,
+            np.copy(self._rewards), np.copy(self._dones), infos)
+
+    def close_extras(self):
+        for env in self.envs:
+            env.close()
+
+    def _check_observation_spaces(self):
+        from gym.spaces import Box
```
christopherhesse: Imports at the top of the file please.

gym/vector/sync_vector_env.py
```diff
@@ -0,0 +1,113 @@
...
+        for env in self.envs:
+            observation_space = env.observation_space
+            # Equality between Box spaces does not check for shape equality
```
christopherhesse: This seems to duplicate code in AsyncVecEnv.  Maybe Box space equality should be fixed instead of hacking around it.

gym/vector/sync_vector_env.py
```diff
@@ -0,0 +1,113 @@
...
+            if isinstance(observation_space, Box) \
+                    and isinstance(self.single_observation_space, Box) \
+                    and observation_space.shape != self.single_observation_space.shape:
+                break
+            if not (observation_space == self.single_observation_space):
+                break
+        else:
+            return True
+        self.close()
```
christopherhesse: Seems like the user should catch exceptions and close environments rather than them automatically closing themselves.

gym/vector/vector_env.py
```diff
@@ -0,0 +1,70 @@
...
+        self.num_envs = num_envs
+        self.observation_space = batch_space(observation_space, n=num_envs)
+        self.action_space = Tuple((action_space,) * num_envs)
+
+        self.closed = False
+        self.viewer = None
+
+        # The observation and action spaces of a single environment are
+        # kepts in separate properties
+        self.single_observation_space = observation_space
```
christopherhesse: Nice

setup.py
```diff
@@ -31,7 +31,7 @@
                 if package.startswith('gym')],
       zip_safe=False,
       install_requires=[
-          'scipy', 'numpy>=1.10.4', 'six', 'pyglet>=1.2.0',
+          'scipy', 'numpy>=1.10.4', 'six', 'pyglet>=1.2.0', 'cloudpickle'
```
christopherhesse: This should specify `cloudpickle~=<version number>` to make sure it doesn't randomly break later.

gym/vector/async_vector_env.py
christopherhesse: Actually, what's the argument for timeout as well?  We haven't needed it in baselines.

christopherhesse:
Overall, this looks good, it simplifies some things but adds some complexity over the baselines implementation and it's not clear that the extra complexity is necessary.

gym/vector/async_vector_env.py
```diff
@@ -0,0 +1,482 @@
...
+        observations_list, rewards, dones, infos = zip(*results)
+
+        if not self.shared_memory:
+            concatenate(observations_list, self.observations,
+                self.single_observation_space)
+
+        return (deepcopy(self.observations) if self.copy else self.observations,
+                np.array(rewards), np.array(dones, dtype=np.bool_), infos)
+
+    def poll(self, timeout=None):
```
christopherhesse: Why is this method in the public api of the class?

gym/vector/vector_env.py
```diff
@@ -0,0 +1,70 @@
...
+        pass
+
+    def step_wait(self, **kwargs):
+        raise NotImplementedError()
+
+    def step(self, actions):
+        self.step_async(actions)
+        return self.step_wait()
+
+    def close_extras(self):
```
christopherhesse: While close_extras is in the baselines API, I'm not certain it's much better than putting:

```
if self.closed:
    return
<close stuff>
super().close()
```

In subclasses.  Is close_extras actually that much better?

christopherhesse:
Also it looks like the existing tests did not make it in this PR: https://github.com/openai/baselines/blob/master/baselines/common/vec_env/test_vec_env.py

It would be nice to have those here as well.

gym/vector/async_vector_env.py
tristandeleu: This was addressed in #1519, I will fix this now that this PR has been merged

gym/vector/sync_vector_env.py
tristandeleu: Addressed in #1519

gym/vector/async_vector_env.py
tristandeleu: For Python 2 compatibility, Python 2 does not have `TimeoutError`

gym/vector/async_vector_env.py
tristandeleu: No reason not to make it fatal, I will throw an exception

gym/vector/__init__.py
tristandeleu: Yes, `episodic=True` means that the individual environments do not reset automatically after finishing an episode. You get only a single episode of each environment, and you are responsible for resetting the environment yourself. This is meant to cover the frequent pattern
```python
while not done
    env.step(action)
```

gym/vector/async_vector_env.py
tristandeleu: No particular reason to have it public, I will make it private

gym/vector/async_vector_env.py
tristandeleu: The `observation_space` is required to create the shared memory, and the shared memory is required for starting the processes. If we don't create a `dummy_env` in the main process, we would have to start a (dummy) process to get the `observation_space`, and then closing it to create the "real" processes. We wouldn't be able to use that dummy process for further operations (`reset`/`step`/etc...) because it wouldn't have the shared memory.

gym/vector/async_vector_env.py
tristandeleu: The reason why I was calling `close` here (and elsewhere) was to avoid any hanging process if the user wasn't careful in closing the environment himself. But leaving the responsibility up to the user is also fine for me.

gym/vector/async_vector_env.py
tristandeleu: The idea behind `timeout` is to leave `close` the possibility to close the individual environment gracefully, but doesn't have to wait for too long (for example, if something went wrong in one of the individual environment that leaves it hanging forever). `terminate` is the extreme version of it, where the user wants to close the environment without having to wait at all.

gym/vector/__init__.py
christopherhesse: How does that cover that pattern?  Seems like I want episodic=True in that case or else I'll step past the end of some environments.

gym/vector/async_vector_env.py
christopherhesse: Yeah it looks like multiprocessing can't send shared memory from the child process to the parent :/

gym/vector/async_vector_env.py
christopherhesse: Sorry, that wasn't a very clear question.  I realize why they're there, it's just I haven't used an environment where they were necessary.  It seems like something that would be handled by a wrapper instead of the environment vectorizer.  Do you have a case where this is actually used?

christopherhesse:
Looks great!  Thanks for updating the PR!  I should have time to review more on friday.  

I noticed that `create_empty_array` takes `np.empty` as a parameter, is there a reason not to have this just always be `np.zeros` and remove the argument?  I don't think there are any advantages to using `np.empty`, and programs will behave inconsistently if someone forgets to initialize memory.

I marked all but a few of the open comments as resolved.

araffin:
Hello,
I have an additional suggestion for the base class.
In [stable-baselines](https://github.com/hill-a/stable-baselines), we found it quite useful to have access to individual environments, even after the VecEnv creation. What do you think of adding:
- `get_attr(self, attr_name, indices=None)`, a method for accessing specific envs attributes (filtered by the `indices` keyword 
- `set_attr(self, attr_name, value, indices=None)`, counterpart of `get_attr`
- `env_method(self, method_name, *method_args, indices=None, **method_kwargs):`, that allows to calls method of a vectorized environement.

See [documentation](https://stable-baselines.readthedocs.io/en/master/guide/vec_envs.html#stable_baselines.common.vec_env.SubprocVecEnv.env_method), [code](https://github.com/hill-a/stable-baselines/blob/master/stable_baselines/common/vec_env/base_vec_env.py) and [tests](https://github.com/hill-a/stable-baselines/blob/master/tests/test_vec_envs.py).

gym/vector/async_vector_env.py
tristandeleu: I never had to use it, but I feel like leaving the user the ability to have a finer control over this if they need to might be a good idea. I think it makes sense to have it in the base class for `AsyncVectorEnv`, especially for handling termination of processes, but I'm completely open to a wrapper-based solution if you have an idea.

gym/vector/__init__.py
tristandeleu: Yes, in that case you'd like `episodic=True`. Here is an example
```python
import gym
import numpy as np

env = gym.vector.make('CubeCrash-v0', 5, episodic=True)
observations = env.reset()
dones = np.zeros(env.num_envs, dtype=np.bool_)
# This terminates after getting 1 episode per environment
while not dones.all():
    actions = env.action_space.sample()
    observations, rewards, dones, _ = env.step(actions)

env = gym.vector.make('CubeCrash-v0', 5, episodic=False)
observations = env.reset()
dones = np.zeros(env.num_envs, dtype=np.bool_)
# This doesn't terminate
while not dones.all():
    actions = env.action_space.sample()
    observations, rewards, dones, _ = env.step(actions)
```

This allows you to either
 - Get a fixed amount of episodes. Let's say you have 5 environments and you want 20 episodes overall, with `episodic=True` you can just loop 4 times through this (and you are responsible for calling `reset`):
```python
env = gym.vector.make('CubeCrash-v0', 5, episodic=True)
for _ in range(4):
    observations = env.reset()
    dones = np.zeros(env.num_envs, dtype=np.bool_)

    while not dones.all():
        actions = env.action_space.sample()
        observations, rewards, dones, _ = env.step(actions)
```
 - Or get an unbounded amount of episodes, without having to reset the environment yourself. This is the behavior you get in baselines, and is the default one here.

tristandeleu:
As suggested by @araffin, I added `call`, `get_attr` and `set_attr`. However it doesn't have the `indices` argument, and calls them for all environments. Adding `indices` might makes things more complex, do you have a case where you'd like to call these functions for only a subset of the environments?

araffin:
>adding indices might make things more complex, do you have a case where you'd like to call these functions for only a subset of the environments?

I agree that will add a bit of complexity (in fact, not that much, see [here](https://github.com/hill-a/stable-baselines/blob/master/stable_baselines/common/vec_env/base_vec_env.py#L169)) and that the use cases are limited.

Among them, I see mainly two:
- setting different random seed / env parameters for each env, in fact that what is done in the recent [Adversarial Policies: Attacking Deep Reinforcement Learning](https://github.com/HumanCompatibleAI/adversarial-policies/blob/99700aab22f99f8353dc74b0ddaf8e5861ff34a5/src/modelfree/training/lookback.py#L216) [Paper](https://arxiv.org/abs/1905.10615)
- debugging purposes

That's only a suggestion, I would let @christopherhesse  and @pzhokhov decide for that.

gym/vector/async_vector_env.py
```diff
@@ -0,0 +1,594 @@
...
+        """
+        Parameters
+        ----------
+        seeds : list of int, or int, optional
+            Random seed for each individual environment. If `seeds` is a list of
+            length `num_envs`, then the items of the list are chosen as random
+            seeds. If `seeds` is an int, then each environment uses the random
+            seed `seeds + n`, where `n` is the index of the environment (between
+            `0` and `num_envs - 1`).
+        """
```
tristandeleu: @araffin The `seed` and `set_attr` methods can take lists of seeds/parameters to set different values for each individual environment. However this still requires the list to have the same length as the number of environments.

christopherhesse:
Thanks again for all the work on this PR!  I think we should @araffin's suggestion seems good, but I think that should be a separate PR after this one is merged.  The scope of this PR already seems a bit large.

gym/vector/async_vector_env.py
```diff
@@ -0,0 +1,594 @@
...
+                    pipe.send(function)
+            elif command == '_setattr':
+                name, value = data
+                setattr(env, name, value)
+                pipe.send(None)
+            elif command == '_check_observation_space':
+                pipe.send(data == env.observation_space)
+            elif command == '_restart':
+                observation = env.reset()
+                episode_done = True
```
pzhokhov: why is `episode_done = True` here? It seems to me that after restart (and calling env.reset()) it should be False

gym/vector/__init__.py
christopherhesse: I think this is a little dangerous since it's easy to confuse the past-done "undefined" observations with the good ones in this case.  What we normally do is if you want 20 episodes you do `episodic=False` and then count the number of dones (`done_count += np.sum(dones)` for instance).   Any thoughts @pzhokhov ?

gym/vector/sync_vector_env.py
```diff
@@ -0,0 +1,200 @@
...
+            A vector of rewards from the vectorized environment.
+
+        dones : `np.ndarray` instance (dtype `np.bool_`)
+            A vector whose entries indicate whether the episode has ended.
+
+        infos : list of dict
+            A list of auxiliary diagnostic informations.
+        """
+        observations, infos = [], []
+        for i, action in enumerate(actions):
```
christopherhesse: This should likely be `for i in range(self.num_envs)` to be consistent with `reset`

gym/vector/async_vector_env.py
tristandeleu: This makes sense, it should probably only set `episode_done = True` if we are calling `step`.

gym/vector/sync_vector_env.py
tristandeleu: I changed `reset` instead (as well as `call`) to be more pythonic

gym/vector/async_vector_env.py
```diff
@@ -0,0 +1,594 @@
...
+            if command == 'reset':
+                observation = env.reset()
+                write_to_shared_memory(index, observation, shared_memory,
+                                       observation_space)
+                episode_done = False
+                pipe.send(None)
+            elif command == 'step':
+                if episodic and episode_done:
+                    observation = _zero_observation
+                    reward, done, info = 0., True, {}
```
pzhokhov: I think this can create subtle bugs downstream - for instance, a) if the environment produces something useful in info dictionary, and consumer relies on that, after episode end the consumer will complain about lack of right keys in info dictionary or b) reward / observation statistic computation can be skewed by this.

gym/vector/__init__.py
pzhokhov: The episodic=True seems dangerous for me as well (left a separate comment in the code about that); while I can see the cases where it is useful, I think proper implementation will require env-specific details (like what is an ok observation after done, what is in the info dictionary).

gym/vector/async_vector_env.py
tristandeleu: I'd argue that the user shouldn't rely on specific keys in the info dictionary, nor the observation, after the episode ends because the results are undefined (as mentioned [here](https://github.com/openai/gym/blob/13f32c7689b9c8738d6db7f6b850ec216cb4d781/gym/core.py#L55)). But in the current state, I agree that reward and observation statistics might be skewed. This could be handled by having a key in the info dictionary for all steps after the episode ends (for now, `AsyncVectorEnv.end_episode` exists only for the final step of the episode)?

tristandeleu:
One use case for `episodic=True` is in meta-RL, where you only want to sample a batch of a few trajectories, as opposed to feeding your trajectories to a sampler/experience replay. Maybe it can be removed from `gym.vector.make` to avoid confusion, but leaving it in `AsyncVectorEnv` and `SyncVectorEnv` for advanced usage. `episodes=<int>` could eventually work.

Removing the restart and episodic features might be difficult through rebase, since this is a big PR. Any chance of getting both in this PR (after a possible rework)?

tristandeleu:
I have removed the episodic and automatic restart features from the current PR, and have them ready for future PRs once this one is merged.

christopherhesse:
Awesome, thanks @tristandeleu!  It looks to me that the only remaining item is to separate `call_wait/call_async/call/get_attr/set_attr` into a separate PR.  Sorry again for the long cycle here, but it's a rather large PR.  If everything on this PR looks good on Friday, hopefully we can merge it then.

The docstrings and tests look great btw.

tristandeleu:
I removed `call`/`get_attr`/`set_attr` from the current PR, and this is also ready for another PR once this one is merged.

pzhokhov:
LGTM! Thanks for the all the efforts, this looks very nice!

christopherhesse:
Thanks again @tristandeleu, having vectorized envs in gym is nice, and the cleanup over what baselines did is also good.

christopherhesse:
I filed some issues I found with this: https://github.com/openai/gym/labels/VectorEnv

PR #1480, opened 2019-05-12
Added pendulumV2 with SI units

abhinavsagar:

PR #1515, opened 2019-06-02
Fix gravity constant

abhinavsagar:
Continued from [here](https://github.com/openai/gym/pull/1480).


abhinavsagar:
@pzhokhov @christopherhesse Need some help here.

gym/envs/classic_control/pendulumV2.py
```diff
@@ -0,0 +1,92 @@
...
+from os import path
+
+# This is similar to pendulum.py with the only change that the units are expressed in SI this time.
+
+class PendulumEnv(gym.Env):
+    metadata = {
+        'render.modes' : ['human', 'rgb_array'],
+        'video.frames_per_second' : 30
+    }
+
```
WalterPichl: you can do clearer code

gym/envs/classic_control/pendulum.py
```diff
@@ -26,7 +26,7 @@ def seed(self, seed=None):
         self.np_random, seed = seeding.np_random(seed)
         return [seed]
 
-    def step(self,u):
+    def step(self, u, **kwargs):
```
pzhokhov: I am not a fan of changing `step` call signature - basically, any algorithm that works does not know about this change will not be able to use the new version. 
IMO, a better way of doing this will be the following:
change the constructor to take g:
```python
 def __init__(self, g=10.0):
    self.g = g
    ...
```
This way, you can  create a Pendulum with g=9.8 via:
```python
gym.make('Pendulum-v0', g=9.8)
```
(and make sure in the code below at line 32 `g=10.` is replaced with `g=self.g`)

gym/envs/classic_control/pendulum.py
```diff
@@ -45,6 +45,8 @@ def step(self,u):
         self.state = np.array([newth, newthdot])
         return self._get_obs(), -costs, False, {}
 
+    step(g = 9.8)
```
pzhokhov: this line is failing in travis, I don't think it should be here

abhinavsagar:
@pzhokhov Done the changes.

abhinavsagar:
@pzhokhov Done the changes.

Issue #1523, opened 2019-06-06
Document EnvSpec


lobachevzky:
The latest commit (https://github.com/openai/gym/commit/85a5372a19c0f35db2410e586cc9a32c4d94bf1a#diff-fd699eda20bea2c460f55cb0c2a22702) makes `TimeLimit` compatible only with environments that define `self.spec` as an `EnvSpec` instance. Please add some documentation for how gym expects this value to be defined in custom environments. Currently I am just using 
```
self.spec = EnvSpec()
```
and hoping for the best.

christopherhesse:
This is likely a bug in `TimeLimit`, any interest in filing a PR that checks if `env.spec` is None before trying to access properties on it?

christopherhesse:
Should be fixed by https://github.com/openai/gym/pull/1525, thanks for the PR!

PR #1525, opened 2019-06-08
Check if env has `spec` attribute before assigning to it.

lobachevzky:
See https://github.com/openai/gym/issues/1523.


christopherhesse:
Thanks!

PR #1527, opened 2019-06-10
Added velocity goal constraint in mountaincarV2

abhinavsagar:


gym/envs/__init__.py
```diff
@@ -78,6 +78,13 @@
     reward_threshold=90.0,
 )
 
+register(
```
pzhokhov: when registered this way, `MountainCarContinuous-v1` and `MountainCarContinuous-v0` are the exact same environment

gym/envs/classic_control/continuous_mountain_carV2.py
```diff
@@ -0,0 +1,146 @@
...
+
+        velocity += force*self.power -0.0025 * math.cos(3*position)
+        if (velocity > self.max_speed): velocity = self.max_speed
+        if (velocity < -self.max_speed): velocity = -self.max_speed
+        position += velocity
+        if (position > self.max_position): position = self.max_position
+        if (position < self.min_position): position = self.min_position
+        if (position==self.min_position and velocity<0): velocity = 0
+
+        done = bool(position >= self.goal_position and velocity >= self.goal_velocity)
```
pzhokhov: the only difference between this env and Continuous_MountainCarEnv is this goal_velocity, is that correct? 
how about we just add a new kwarg to Continuous_MountainCarEnv, with None or negative value indicating that there is no goal velocity constraint?

PR #1529, opened 2019-06-11
Added optional double down action to blackjack

chisness:
Extends the game from just hit/stick to the 3rd action of doubling down, which means doubling the bet and getting only 1 additional card

### DIFF
--- a/gym/envs/toy_text/blackjack.py
+++ b/gym/envs/toy_text/blackjack.py
@@ -31,7 +31,7 @@ def is_bust(hand):  # Is this hand a bust?
     return sum_hand(hand) > 21
 
 
-def score(hand):  # What is the score of this hand (0 if bust)
+def score(hand):  # What is the score of this hand (0 if bust)?
     return 0 if is_bust(hand) else sum_hand(hand)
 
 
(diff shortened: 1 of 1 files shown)


gym/envs/toy_text/blackjack.py
```diff
@@ -81,6 +84,8 @@ def __init__(self, natural=False):
         # Flag to payout 1.5 on a "natural" blackjack win, like casino rules
         # Ref: http://www.bicyclecards.com/how-to-play/blackjack/
         self.natural = natural
+        #Flag for allowing doubling down
```
pzhokhov: this comment should rather be in the docstring (could you also describe what natural stands for while at it then ^_^)

gym/envs/toy_text/blackjack.py
```diff
@@ -90,15 +95,25 @@ def seed(self, seed=None):
 
     def step(self, action):
         assert self.action_space.contains(action)
-        if action:  # hit: add a card to players hand and return
+        if self.double_down:
```
pzhokhov: minor nitpick - given the `assert self.action_space.contains(action)` above, if action == 2, then self.double_down should always be true, so the outer if statement is unnecessary here

chisness: Thanks, fixed in new commit

gym/envs/toy_text/blackjack.py
chisness: Thanks, updated in new commit

chisness:
@pzhokhov I made those changes and a couple of other small ones in the new commit. I showed a test of 100 double_down actions to show the reward is always {-2, 0, 2} (except the case of natural Blackjack when natural=True, which gives an immediate reward of 1.5).

christopherhesse:
Do we really want to extend these built-in environments instead of having people make their own environments outside the gym repo @pzhokhov?

jkterry1:
Closing per #2259

jkterry1:
That was a typo sorry, I'm closing this after discussion with @cpnota because this is intended to match the simplified blackjack game in Barto and Sutton

PR #1534, opened 2019-06-14
done=true when successfully achieve the goal in robot_env

lilianweng:
It feels more intuitive to return `done=true` when successfully achieve the goal in robot_env.
Or we keep this behavior but mention that in the documentation page: https://gym.openai.com/envs/#robotics?


lilianweng:
solved offline.

YeeCY:
What's the discussion result? The `done=False` creates an infinity episode which is confusing when we sample from the environment.

PR #1535, opened 2019-06-15
ALE 0.6 - modes and difficulties for Atari games

JesseFarebro:
This PR along with https://github.com/openai/atari-py/pull/49 introduces modes and difficulties for Atari 2600 games in Gym. This was introduced in version 0.6 of the ALE and also outlined in the paper ["Revisiting the Arcade Learning Environment: Evaluation Protocols and Open Problems for General Agents"](https://arxiv.org/abs/1709.06009) by Machado et al.

I had previously discussed versioning in  https://github.com/openai/atari-py/pull/49 with @christopherhesse as this PR would rely on a new API in v0.6 which allows you to query and set the mode and/or difficulty.

The other thing to note in this PR is the removal of `self.game_path`. I don't understand why you call `ale.loadROM` in the seed function. To my knowledge, you don't need to set the random seed before loading the ROM. The seed only affects the RNG for sticky actions. In fact, most examples in the ALE repository set the seed after loading the ROM.
I had to make this change as it seems `seed` gets called twice and the second call would override any previously defined modes or difficulties (each call to `loadROM` resets the mode and difficulty to the default of 0).

To get this PR merged I think we should discuss:

1) <del>How to handle versioning in Gym with respect to `atari-py`. It's probably necessary to check for `setDifficulty` and `setMode` in `atari-py`. If the newer API isn't found we can throw an error asking the user to upgrade `atari-py`.</del>
2) <del>Gain some clarity on the `ale.loadRom` call in `seed`. I don't see this breaking anything but it would be nice to get some clarity on why this existed in the first place.</del>

### DIFF
--- a/gym/envs/atari/atari_env.py
+++ b/gym/envs/atari/atari_env.py
@@ -26,6 +26,8 @@ class AtariEnv(gym.Env, utils.EzPickle):
     def __init__(
             self,
             game='pong',
+            mode=0,
+            difficulty=0,
             obs_type='ram',
             frameskip=(2, 5),
             repeat_action_probability=0.,
@@ -36,12 +38,18 @@ class AtariEnv(gym.Env, utils.EzPickle):
         utils.EzPickle.__init__(
                 self,
                 game,
+                mode,
+                difficulty,
                 obs_type,
                 frameskip,
                 repeat_action_probability)
         assert obs_type in ('ram', 'image')
 
+        self.game = game
         self.game_path = atari_py.get_game_path(game)
+        self.game_mode = mode
+        self.game_difficulty = difficulty
+
         if not os.path.exists(self.game_path):
             msg = 'You asked for game %s but path %s does not exist'
             raise IOError(msg % (game, self.game_path))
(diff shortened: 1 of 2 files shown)


christopherhesse:
For 1), why not just require `atari-py~=0.2.0` in gym?

JesseFarebro:
@christopherhesse I agree, I'm not too familiar with pip's upgrade strategy but if we specify `atari-py~=0.2.0` in gym will it indeed upgrade `atari-py` even though the major version hasn't changed?

We'll have to bump `atari-py` then I can reflect that change in this PR.

christopherhesse:
Yeah it should work, the major version for a pre 1.0 release should be the second digit.  For that requirement https://www.python.org/dev/peps/pep-0440/#compatible-release indicates that it would require `atari-py>=0.2.0, atari-py==0.2.*`.

JesseFarebro:
Removing `loadROM` from seed ends up breaking determinism (I can see how since the ALE manages Stella's RNG after a `loadROM` call). Without looking too much more into it I just figured I'd leave the `loadROM` call in `seed`. This way I just set the mode/difficulty in `seed`.

I have tested everything and it's working as I expect. This is ready for someone to take a look at now.

pzhokhov:
LGTM, let's merge this, I'll bump version and release the new pypi package.

JesseFarebro:
Thanks for everything @pzhokhov and @christopherhesse!

PR #1536, opened 2019-06-15
Removed image links that are not working

abhinavsagar:


pzhokhov:
Thanks for pointing it out! I'd would rather prefer to update them instead of deleting: #1547

pzhokhov:
closing this PR; links are fixed now

PR #1547, opened 2019-06-21
update links to robotics images

pzhokhov:


pzhokhov:
yeah... I feel lazy about downloading them and putting them in the repo though, let's wait till it is broken again :)

PR #1552, opened 2019-06-22
Separate tests for single and action_space

tristandeleu:
Separate `test_step_...` into two tests, depending on how the actions are created (either from a list of `env.single_action_space.sample()`, or with `env.action_space.sample()`).


tristandeleu:
Fixed it, it is now using `@pytest.mark.parametrize` on whether to use the single action space or not.

christopherhesse:
Thanks!  I think it would also make sense to combine the Async/Sync tests, but that can wait for some other PR.

Issue #1542, opened 2019-06-21
async env with spawn context handles errors oddly


christopherhesse:
The following code is incorrect, but fails in an odd way:

```
import gym
import numpy as np


def main():
    def make_env():
        return gym.make("PongNoFrameskip-v4")

    venv = gym.vector.AsyncVectorEnv([make_env] * 3, context="spawn")
    venv.reset()
    while True:
        obs, rews, dones, infos = venv.step(np.array([venv.action_space.sample() for _ in range(venv.num_envs)]))


if __name__ == "__main__":
    main()
```

This produces:

```
ERROR: Received the following error from Worker-2: TypeError: only size-1 arrays can be converted to Python scalars
ERROR: Shutting down Worker-2.
ERROR: Received the following error from Worker-0: TypeError: only size-1 arrays can be converted to Python scalars
ERROR: Shutting down Worker-0.
ERROR: Received the following error from Worker-1: TypeError: only size-1 arrays can be converted to Python scalars
ERROR: Shutting down Worker-1.
ERROR: Raising the last exception back to the main process.
Traceback (most recent call last):
  File "/Users/csh/miniconda3/envs/py37/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/Users/csh/miniconda3/envs/py37/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/csh/agi/sandbox/oneoff/try_vec_env.py", line 45, in <module>
    main()
  File "/Users/csh/agi/sandbox/oneoff/try_vec_env.py", line 41, in main
    obs, rews, dones, infos = venv.step(np.array([venv.action_space.sample() for _ in range(venv.num_envs)]))
  File "/Users/csh/agi/gym/gym/vector/vector_env.py", line 54, in step
    return self.step_wait()
  File "/Users/csh/agi/gym/gym/vector/async_vector_env.py", line 228, in step_wait
    self._raise_if_errors()
  File "/Users/csh/agi/gym/gym/vector/async_vector_env.py", line 327, in _raise_if_errors
    raise exctype(value)
TypeError: only size-1 arrays can be converted to Python scalars
/Users/csh/agi/gym/gym/logger.py:30: UserWarning: WARN: Calling `close` while waiting for a pending call to `step` to complete.
  warnings.warn(colorize('%s: %s'%('WARN', msg % args), 'yellow'))
[...]

Issue #1546, opened 2019-06-21
VectorEnv close() uses logger which performs late import of `six` resulting in odd error message


christopherhesse:
```
from gym.vector import make

venv = make("PongNoFrameskip-v4", num_envs=3)
obs, rews, dones, infos = venv.step(venv.action_space.sample())
print(obs.shape, rews, dones, infos)
```

This script sometimes produces an extra exception:

```
ERROR: Received the following error from Worker-0: AssertionError: Cannot call env.step() before calling reset()
ERROR: Shutting down Worker-0.
ERROR: Received the following error from Worker-2: AssertionError: Cannot call env.step() before calling reset()
ERROR: Shutting down Worker-2.
ERROR: Received the following error from Worker-1: AssertionError: Cannot call env.step() before calling reset()
ERROR: Shutting down Worker-1.
ERROR: Raising the last exception back to the main process.
Traceback (most recent call last):
  File "/Users/csh/miniconda3/envs/py37/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/Users/csh/miniconda3/envs/py37/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/csh/agi/sandbox/oneoff/try_vec_env.py", line 4, in <module>
    obs, rews, dones, infos = venv.step(venv.action_space.sample())
  File "/Users/csh/agi/gym/gym/vector/vector_env.py", line 54, in step
    return self.step_wait()
  File "/Users/csh/agi/gym/gym/vector/async_vector_env.py", line 228, in step_wait
    self._raise_if_errors()
  File "/Users/csh/agi/gym/gym/vector/async_vector_env.py", line 327, in _raise_if_errors
    raise exctype(value)
AssertionError: Cannot call env.step() before calling reset()
Exception ignored in: <function AsyncVectorEnv.__del__ at 0x111340950>
Traceback (most recent call last):
  File "/Users/csh/agi/gym/gym/vector/async_vector_env.py", line 332, in __del__
  File "/Users/csh/agi/gym/gym/vector/async_vector_env.py", line 263, in close
  File "/Users/csh/agi/gym/gym/logger.py", line 30, in warn
  File "/Users/csh/agi/gym/gym/utils/colorize.py", line 27, in colorize
ModuleNotFoundError: import of six halted; None in sys.modules
```

There's an extra `ModuleNotFoundError` due to a late import of `six` in `colorize`.

[...]

PR #1556, opened 2019-06-23
Move _raise_if_errors after recv in AsyncVectorEnv

tristandeleu:
There is a race condition in `AsyncVectorEnv`, most likely due to `_raise_if_errors` being called before the output of the workers is received: if the worker raises an error after `_raise_if_errors` is called, but before `pipe.recv()`, then the overall result contains spurious `None`, which cannot be iterated over anymore (see https://github.com/openai/gym/issues/1546#issuecomment-504585415).

### DIFF
--- a/gym/vector/async_vector_env.py
+++ b/gym/vector/async_vector_env.py
@@ -129,8 +129,8 @@ class AsyncVectorEnv(VectorEnv):
 
         for pipe, seed in zip(self.parent_pipes, seeds):
             pipe.send(('seed', seed))
-        for pipe in self.parent_pipes:
-            pipe.recv()
+        _, successes = zip(*[pipe.recv() for pipe in self.parent_pipes])
+        self._raise_if_errors(successes)
 
     def reset_async(self):
         self._assert_is_running()
@@ -166,13 +166,12 @@ class AsyncVectorEnv(VectorEnv):
             raise mp.TimeoutError('The call to `reset_wait` has timed out after '
                 '{0} second{1}.'.format(timeout, 's' if timeout > 1 else ''))
 
-        observations_list = [pipe.recv() for pipe in self.parent_pipes]
-        self._raise_if_errors()
+        results, successes = zip(*[pipe.recv() for pipe in self.parent_pipes])
+        self._raise_if_errors(successes)
         self._state = AsyncState.DEFAULT
 
         if not self.shared_memory:
-            concatenate(observations_list, self.observations,
-                self.single_observation_space)
+            concatenate(results, self.observations, self.single_observation_space)
 
         return deepcopy(self.observations) if self.copy else self.observations
 
@@ -225,8 +224,8 @@ class AsyncVectorEnv(VectorEnv):
             raise mp.TimeoutError('The call to `step_wait` has timed out after '
                 '{0} second{1}.'.format(timeout, 's' if timeout > 1 else ''))
 
-        results = [pipe.recv() for pipe in self.parent_pipes]
-        self._raise_if_errors()
+        results, successes = zip(*[pipe.recv() for pipe in self.parent_pipes])
+        self._raise_if_errors(successes)
         self._state = AsyncState.DEFAULT
         observations_list, rewards, dones, infos = zip(*results)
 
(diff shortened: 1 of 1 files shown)


gym/vector/async_vector_env.py
```diff
@@ -276,7 +275,7 @@ def close(self, timeout=None, terminate=False):
                     pipe.send(('close', None))
             for pipe in self.parent_pipes:
                 if not pipe.closed:
-                    pipe.recv()
+                    _, success = pipe.recv()
```
christopherhesse: Was this value supposed to be used anywhere?

christopherhesse: Seems like it should be used or the change reverted.

christopherhesse:
Thanks for fixing this!

gym/vector/async_vector_env.py
tristandeleu: Fixed it, this was not used anywhere

christopherhesse:
This says it fixes #1542, but I think #1542 mentions two issues, and this only solves the second one, is that correct @tristandeleu?

tristandeleu:
This actually should fix both. The first one was an issue with `pipe` being `None` in `_poll`, which is also part of the fix in this PR. The second one is handled by the new "success" flag returned by the workers.

christopherhesse:
I ran the script there and it still has this error:

```
ERROR: Received the following error from Worker-1: AssertionError: Cannot call env.step() before calling reset()
ERROR: Shutting down Worker-1.
ERROR: Received the following error from Worker-2: AssertionError: Cannot call env.step() before calling reset()
ERROR: Shutting down Worker-2.
ERROR: Received the following error from Worker-0: AssertionError: Cannot call env.step() before calling reset()
ERROR: Shutting down Worker-0.
ERROR: Raising the last exception back to the main process.
  File "test.py", line 4, in <module>
    obs, rews, dones, infos = venv.step(venv.action_space.sample())
  File "/Users/csh/agi/gym/gym/vector/vector_env.py", line 54, in step
    return self.step_wait()
  File "/Users/csh/agi/gym/gym/vector/async_vector_env.py", line 228, in step_wait
    self._raise_if_errors(successes)
  File "/Users/csh/agi/gym/gym/vector/async_vector_env.py", line 333, in _raise_if_errors
    raise exctype(value)
AssertionError: Cannot call env.step() before calling reset()
Exception ignored in: <function AsyncVectorEnv.__del__ at 0x10d741ae8>
  File "/Users/csh/agi/gym/gym/vector/async_vector_env.py", line 338, in __del__
  File "/Users/csh/agi/gym/gym/vector/async_vector_env.py", line 262, in close
  File "/Users/csh/agi/gym/gym/logger.py", line 30, in warn
  File "/Users/csh/agi/gym/gym/utils/colorize.py", line 27, in colorize
ModuleNotFoundError: import of six halted; None in sys.modules
```

This isn't the sole fault of vectorenv I think, but it looks like it's using logger during close, and logger does a late import which fails because the interpreter is shutting down.

christopherhesse:
Oh sorry about that, I said 1542 but I actually meant this issue: https://github.com/openai/gym/issues/1546

tristandeleu:
Yes for #1546 it only fixes the second issue in the comment. For the first issue (and this one) I also think it is unrelated to `AsyncVectorEnv`. One fix would be to remove the logger from `close`, but this seems like a weak patch.

christopherhesse:
I also checked out this branch against https://github.com/openai/gym/issues/1542 and still get an odd error there:

```
  File "test.py", line 16, in <module>
    main()
  File "test.py", line 12, in main
    obs, rews, dones, infos = venv.step(np.array([venv.action_space.sample() for _ in range(venv.num_envs)]))
  File "/Users/csh/agi/gym/gym/vector/vector_env.py", line 54, in step
    return self.step_wait()
  File "/Users/csh/agi/gym/gym/vector/async_vector_env.py", line 228, in step_wait
    self._raise_if_errors(successes)
  File "/Users/csh/agi/gym/gym/vector/async_vector_env.py", line 333, in _raise_if_errors
    raise exctype(value)
TypeError: only size-1 arrays can be converted to Python scalars
/Users/csh/agi/gym/gym/logger.py:30: UserWarning: WARN: Calling `close` while waiting for a pending call to `step` to complete.
  warnings.warn(colorize('%s: %s'%('WARN', msg % args), 'yellow'))
Exception ignored in: <function AsyncVectorEnv.__del__ at 0x117229ae8>
  File "/Users/csh/agi/gym/gym/vector/async_vector_env.py", line 338, in __del__
    self.close(terminate=True)
  File "/Users/csh/agi/gym/gym/vector/async_vector_env.py", line 281, in close
    pipe.close()
AttributeError: 'NoneType' object has no attribute 'close'
```

christopherhesse:
I'll split out 1546 then

christopherhesse:
Great, thanks!  I'll wait for tests to pass then merge this.

PR #1569, opened 2019-06-27
Recursively check for nested Dict in observation space

lbertge:
I've encountered observations of the following format:
```
Dict({
    "key1": "Box()",
    "key2": {
            "subkey1": "Box()",
            "subkey2": "Box()",
             ...
    }
})
```
and currently, the `FlattenDictWrapper` class is unable to handle such a case, since it assumes that all values in the root `Dict` have a `shape` field (`Box`, `Discrete`, etc).

The following changes recursively check each value to see if it's a nested `Dict`, and returns a size which is the aggregated sum of all valid `gym.Space` objects with a shape.

On line [28](https://github.com/openai/gym/compare/master...lbertge:nested-obs-space?expand=1#diff-db310eea40a35ab2276687acee3f20eaR28) I have to convert the result into `int64` because it's possible that sometimes the shape is `()` (a singleton?) and `np.prod(()) = 1.0`; the Box constructor does not like a shape like `(1.0, )`. So I'm not sure if this is an improvement, but I thought it'd be important to point out.

### DIFF
--- a/gym/wrappers/dict.py
+++ b/gym/wrappers/dict.py
@@ -12,6 +12,7 @@ class FlattenDictWrapper(gym.ObservationWrapper):
     def __init__(self, env, dict_keys):
         super(FlattenDictWrapper, self).__init__(env)
         self.dict_keys = dict_keys
+        self.dtype = None
 
         # Figure out observation_space dimension.
         size = self.get_dict_size(self.env.observation_space.spaces, dict_keys)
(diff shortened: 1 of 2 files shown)


lbertge:
Comments:
1. Added tests for nested dicts, dicts with tuples, dicts with tuples with dicts, non-Box spaces, mismatched dtypes
2. For now, I've done as you asked, with restrictions on Box and dtype. In the future I was hoping that this class could support mixed Discrete/Box/Space types, since it's in my use case. I'm sure this has been suggested canonically - the docstrings under the `spaces.Dict` class gives an example of an observation space with different subclasses of `Space`, so it makes sense to me that `FlattenDictWrapper` should be compatible with this vision. I guess, let me know if you want to continue this discussion elsewhere - I'm happy to put in the remaining work and finish it up. 
3. I hope this is what you mean by support for tuples. This allows the wrapper class to work with nested `spaces.Tuple` inside of `spaces.Dict`, and vice versa.

pzhokhov:
Merging this. As for mixed Box/Discrete spaces: 
1) the fact that Dict spaces support them does not mean it is obvious how to flatten them together
2) I think it is rather possible; we can have a flag / function that converts discrete observation into a float vector (which can be a one-hot encoding or an integer value).

PR #1570, opened 2019-06-28
Allow `gym.vector.make` to have vectorized environment for single environment

zuoxingdong:
This can be very convenient for the user who deals with neural network frameworks, single vectorized environment, the observation can be easily served with batch size also for the actions..


christopherhesse:
It does seem inconsistent to return a single environment, which has a different interface, when using `vector.make`, @tristandeleu does this seem like a good change to you?

tristandeleu:
I think this could be a good change. I had this special case for `num_envs=1` because I thought you shouldn't need a vectorized environment in the first place. But this would indeed make things more consistent.

PR #1571, opened 2019-06-28
Support `len(env)` for `gym.vector.make` environment

zuoxingdong:


christopherhesse:
I think the `num_envs` property already handles this.  Make the vectorenv have a list like interface seems interesting, but I'm not certain it belongs here.

PR #1572, opened 2019-06-28
Allow indexing (`env[3]`) for vectorized environment

zuoxingdong:


christopherhesse:
Can't the user just access `.envs` themselves here?  Not sure why this is that desireable.  Also I don't think it works with async vector env.

zuoxingdong:
@christopherhesse , yup, I agree with you, it is better to be consistent. Closing PR now.

PR #1573, opened 2019-06-28
Allow atomic transformation (sequence of wrapping) for vectorized environment

### LINKED ISSUE #1560: Support wrapper transformations to VecEnv
@tristandeleu @pzhokhov @christopherhesse It might be rather convenient for the user experience to provide a list of wrapper transformations for the atomic environments when creating vectorized environment, e.g.

```python
transforms = [AtariPreprocessing, SignReward, ...]
env = gym.vector.make('Pong-v0', 16, True, transforms=transforms)
```
For additional arguments, the user is required to use `partial()` to define them within the transform list. So that each internal environment is wrapped according to the transformation list.

zuoxingdong:
Fix #1560


pzhokhov:
I like the idea and the implementation, but the name `atomic_transformation` seems a bit confusing. Maybe 'wrappers' and allow it to be either callable or an iterable of callables?

zuoxingdong:
@pzhokhov Thanks a lot for the comment. Now it is updated. The wrappers argument could be either a single wrapper or an Iterable of wrappers.

gym/vector/__init__.py
```diff
@@ -39,7 +45,14 @@ def make(id, num_envs=1, asynchronous=True, **kwargs):
     """
     from gym.envs import make as make_
     def _make_env():
-        return make_(id, **kwargs)
+        env = make_(id, **kwargs)
+        if wrappers is not None:
+            if callable(wrappers):
+                env = wrappers(env)
+            elif isinstance(wrappers, Iterable) and all([callable(w) for w in wrappers]):
```
pzhokhov: looks like `collections.abc.Iterable` is not available in python 2.7 (and either way I think silently ignoring the argument if it is not an iterable or callable is not great, let's just try iterating and let python throw an error if it is not possible

pzhokhov: (that's what fails the travis test)

zuoxingdong: @pzhokhov I've got a work-around import for both Python 2 and 3. And when it's neither callable or Iterable, a NotImplementedError is raised, because we do want the wrappers to be either a single one or a sequence of wrappers that would apply on top of the environment.

PR #1574, opened 2019-06-28
add spec property to wrapper to be forwarded to wrapped env. Fixes 1554

pzhokhov:


gym/core.py
```diff
@@ -209,13 +209,16 @@ def __init__(self, env):
...
         self.reward_range = self.env.reward_range
         self.metadata = self.env.metadata
-        self.spec = getattr(self.env, 'spec', None)
 
     def __getattr__(self, name):
         if name.startswith('_'):
             raise AttributeError("attempted to get missing private attribute '{}'".format(name))
         return getattr(self.env, name)
 
+    @property
```
christopherhesse: How is this different from just not having the property decorator?

PR #1579, opened 2019-07-02
Update spaces.py method definition

sritee:
No longer sampling uniformly, after #1567


DavidNKraemer:
Good catch. I must have missed that.

PR #1581, opened 2019-07-03
Recursively ravel nested observations

lbertge:
Relies on #1569. This is definitely related but I wasn't sure about piling up on the PR so I made a separate one. 

`FlattenDictWrapper` lacks a way to convert a nested observation into a flattened vector with the same shape as its `observation_space`. This commit handles those cases so nested dicts can be properly used. I'm also adding more tests to the test suite.

Since #1569 should not cause any regressions, this does not have to be merged immediately. However interacting (stepping, sampling) with a nested-observation env will still break until merged.


gym/wrappers/dict.py
```diff
@@ -50,9 +39,37 @@ def get_box_size(self, space):
...
+    def _size(self, space):
+        if isinstance(space, gym.spaces.Dict):
+            return self.get_dict_size(space.spaces, space.spaces.keys())
+        elif isinstance(space, gym.spaces.Tuple):
+            return self.get_tuple_size(space.spaces)
+        else:
+            return self.get_box_size(space)
+
     def observation(self, observation):
+        print(observation)
```
lbertge: will remove

Issue #1588, opened 2019-07-06
classical_control Rending is Incompatible with Pyglet 1.4.0


wphuhn:
When running `examples/agent/cem.py` using the latest version of pyglet (1.4.0, released 5 July 2019), the script terminates early with the following output:

```
$ python3 ./cem.py 
INFO: Making new env: CartPole-v0
INFO: Clearing 3 monitor files from previous run (because force=True was provided)
INFO: Starting new video recorder writing to /tmp/cem-agent-results/openaigym.video.0.21090.video000000.mp4
Traceback (most recent call last):
  File "./cem.py", line 83, in <module>
    cem(noisy_evaluation, np.zeros(env.observation_space.shape[0]+1), **params)):
  File "./cem.py", line 28, in cem
    ys = np.array([f(th) for th in ths])
  File "./cem.py", line 28, in <listcomp>
    ys = np.array([f(th) for th in ths])
  File "./cem.py", line 78, in noisy_evaluation
    rew, T = do_rollout(agent, env, num_steps)
  File "./cem.py", line 37, in do_rollout
    ob = env.reset()
  File "/home/will/opt/OpenAI_Gym/gym/wrappers/monitor.py", line 39, in reset
    self._after_reset(observation)
  File "/home/will/opt/OpenAI_Gym/gym/wrappers/monitor.py", line 188, in _after_reset
    self.reset_video_recorder()
  File "/home/will/opt/OpenAI_Gym/gym/wrappers/monitor.py", line 209, in reset_video_recorder
    self.video_recorder.capture_frame()
  File "/home/will/opt/OpenAI_Gym/gym/wrappers/monitoring/video_recorder.py", line 101, in capture_frame
    frame = self.env.render(mode=render_mode)
  File "/home/will/opt/OpenAI_Gym/gym/core.py", line 233, in render
    return self.env.render(mode, **kwargs)
  File "/home/will/opt/OpenAI_Gym/gym/envs/classic_control/cartpole.py", line 188, in render
    return self.viewer.render(return_rgb_array = mode=='rgb_array')
  File "/home/will/opt/OpenAI_Gym/gym/envs/classic_control/rendering.py", line 105, in render
    arr = np.frombuffer(image_data.data, dtype=np.uint8)
AttributeError: 'ImageData' object has no attribute 'data'
INFO: Finished writing results. You can upload them to the scoreboard via gym.upload('/tmp/cem-agent-results')
```

[...]

PR #1589, opened 2019-07-06
Update classical_control Rendering for Pyglet API

wphuhn:
This commit fixes issue #1438 and #1588 by adding support for
pre-1.4.0 and 1.4.0+ versions of Pyglet.

### DIFF
--- a/gym/envs/box2d/car_racing.py
+++ b/gym/envs/box2d/car_racing.py
@@ -403,7 +403,11 @@ class CarRacing(gym.Env, EzPickle):
             return self.viewer.isopen
 
         image_data = pyglet.image.get_buffer_manager().get_color_buffer().get_image_data()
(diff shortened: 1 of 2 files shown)


christopherhesse:
Thanks for the PR!  I think it might be simpler to change setup.py to require `pyglet~=1.3.2` though, and then later update all the code at once.  Does that seem reasonable?

christopherhesse:
I updated gym master to use pyglet <=1.3.2, so if you want to update your PR, it could require `pyglet~=1.4.0` and avoid the use of hasattr

wphuhn:
I feel that, with the latest changes to gym/master, this PR is no longer necessary.  I'll just close it out, if there's no objections.

wphuhn:
OK, I'm closing this and issue #1588  out.

PR #1590, opened 2019-07-07
Add CartPoleSwingUp environment

angelolovatto:
A simple continuous control benchmark like Pendulum, but with a not
as trivial solution. It also does not require external library dependencies like
Mujoco and Box2D. The task has been used in recent work and also in the real world:
`https://weightagnostic.github.io`
`http://proceedings.mlr.press/v48/duan16.html`
`https://www.youtube.com/watch?v=XiigTGKZfks`

CartPoleSwingUp based heavily on the following implementation:
`https://github.com/hardmaru/estool/blob/master/custom_envs/cartpole_swingup.py`
This implementation is a refactored version of the original, with the intent of
making it more readable and easier to modify and extend.

Signed-off-by: Ângelo Lovatto <angelolovatto@gmail.com>

### DIFF
--- a/gym/envs/__init__.py
+++ b/gym/envs/__init__.py
@@ -91,6 +91,12 @@ register(
     max_episode_steps=500,
 )
 
+register(
+    id='CartPoleSwingUp-v0',
+    entry_point='gym.envs.classic_control:CartPoleSwingUpEnv',
+    max_episode_steps=500,
+)
+
 # Box2d
 # ----------------------------------------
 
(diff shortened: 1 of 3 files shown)


christopherhesse:
I'm afraid we are not adding new environments to the gym library itself, but you're free to make a package for your environment and users can instantiate the environment using `gym.make` as detailed here: https://github.com/openai/gym/blob/master/docs/creating-environments.md

Thanks for the PR though!

PR #1597, opened 2019-07-12
Expose entry_point on EnvSpec

JesseFarebro:
There may be some use cases where exposing `entry_point` could be useful. For example, verifying the environment is a valid Atari environment by checking if `entry_point` starts with `gym.envs.atari`.

This makes the access of `entry_point` not feel as nasty as calling `spec._entry_point`.

### DIFF
--- a/gym/envs/registration.py
+++ b/gym/envs/registration.py
@@ -42,6 +42,7 @@ class EnvSpec(object):
         self.reward_threshold = reward_threshold
         # Environment properties
         self.nondeterministic = nondeterministic
+        self.entry_point = entry_point
 
         if tags is None:
             tags = {}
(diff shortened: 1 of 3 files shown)


gym/envs/registration.py
```diff
@@ -77,6 +77,10 @@ def make(self, **kwargs):
 
         return env
 
+    @property
+    def entry_point(self):
+        return self._entry_point
```
pzhokhov: if we were to just rename `_entry_point` field to `entry_point`, would that not do the trick?

JesseFarebro: Sure! Wasn't sure what convention you wanted. Thought leaving it this way could imply to an onlooker that `_entry_point` isn't meant to be set apart from the construction of `EnvSpec`.

JesseFarebro: There is also tests and perhaps some people already using `_entry_point`. It's not a problem to change the tests to use `entry_point`.

christopherhesse: If someone outside gym is accessing `_entry_point`, that's basically guaranteed to break at some point.

JesseFarebro: Sounds good @christopherhesse. I update the PR with these changes.

christopherhesse:
Thanks!

AGSNicolaMarino:
I think this commit messed up the interface with baselines, which still evaluates the type of the environment as:
    env_type = env._entry_point.split(':')[0].split('.')[-1]

It would have been better to make a property to extract the attribute, so that other classes need no change.

PR #1599, opened 2019-07-12
don't create scalar env from vector.make

christopherhesse:
@tristandeleu it seems odd to me to return two different object types from this function, scalar envs and vector envs.  Was there a reason to return a scalar env when `num_envs=1`?


tristandeleu:
The motivation was to avoid the overhead of having a vectorized environment if you only want a single environment. From a typing point of view I agree it is not ideal to have `gym.vector.make` returning either an `Env` or a `VectorEnv`. I am all for removing this special case `num_envs=1`!

This is also addressed in #1570

christopherhesse:
Oh good call, I forgot about that.

PR #1600, opened 2019-07-12
Add call, get_attr and set_attr methods to VectorEnv

tristandeleu:
As suggested in #1513, this PR adds 3 new methods to `VectorEnv`:
 - `call(name, *args, **kwargs)`, which calls `get_attr` in each individual environment, and returns either its property `name` (in which case `*args` and `**kwargs` are ignored), or the result of `name(*args, **kwargs)`.
 - `get_attr(name)`, which is syntactic sugar for a call to `call` to get a property of the individual environments.
 - `set_attr(name, values)`, which alls `setattr` on each individual environment.

### DIFF
--- a/gym/vector/async_vector_env.py
+++ b/gym/vector/async_vector_env.py
@@ -20,6 +20,7 @@ class AsyncState(Enum):
     DEFAULT = 'default'
     WAITING_RESET = 'reset'
     WAITING_STEP = 'step'
+    WAITING_CALL = 'call'
 
 
 class AsyncVectorEnv(VectorEnv):
@@ -241,6 +242,80 @@ class AsyncVectorEnv(VectorEnv):
         return (deepcopy(self.observations) if self.copy else self.observations,
                 np.array(rewards), np.array(dones, dtype=np.bool_), infos)
 
+    def call_async(self, name, *args, **kwargs):
+        """
+        Parameters
+        ----------
+        name : string
+            Name of the method or property to call.
+        """
+        self._assert_is_running()
+        if self._state != AsyncState.DEFAULT:
+            raise AlreadyPendingCallError('Calling `call_async` while waiting '
+                'for a pending call to `{0}` to complete.'.format(
+                self._state.value), self._state.value)
+
+        for pipe in self.parent_pipes:
+            pipe.send(('_call', (name, args, kwargs)))
+        self._state = AsyncState.WAITING_CALL
+
+    def call_wait(self, timeout=None):
+        """
+        Parameters
+        ----------
+        timeout : int or float, optional
+            Number of seconds before the call to `step_wait` times out. If
+            `None`, the call to `step_wait` never times out.
+
+        Returns
+        -------
+        results : list
+            List of the results of the individual calls to the method or
+            property for each environment.
+        """
+        self._assert_is_running()
+        if self._state != AsyncState.WAITING_CALL:
+            raise NoAsyncCallError('Calling `call_wait` without any prior call '
+                'to `call_async`.', AsyncState.WAITING_CALL.value)
+
+        if not self._poll(timeout):
+            self._state = AsyncState.DEFAULT
+            raise mp.TimeoutError('The call to `call_wait` has timed out after '
+                '{0} second{1}.'.format(timeout, 's' if timeout > 1 else ''))
+
+        results, successes = zip(*[pipe.recv() for pipe in self.parent_pipes])
+        self._raise_if_errors(successes)
+        self._state = AsyncState.DEFAULT
+
+        return results
+
+    def set_attr(self, name, values):
+        """
+        Parameters
+        ----------
+        name : string
+            Name of the property to be set in each individual environment.
+
+        values : list or object
+            Values of the property to bet set to. If `values` is a list, then
+            it corresponds to the values for each individual environment,
+            otherwise a single value is set for all environments.
+        """
+        self._assert_is_running()
+        if not isinstance(values, list):
+            values = [values for _ in range(self.num_envs)]
+        assert len(values) == self.num_envs
+
+        if self._state != AsyncState.DEFAULT:
+            raise AlreadyPendingCallError('Calling `set_attr` while waiting '
+                'for a pending call to `{0}` to complete.'.format(
+                self._state.value), self._state.value)
+
+        for pipe, value in zip(self.parent_pipes, values):
+            pipe.send(('_setattr', (name, value)))
+        _, successes = zip(*[pipe.recv() for pipe in self.parent_pipes])
+        self._raise_if_errors(successes)
+
     def close_extras(self, timeout=None, terminate=False):
         """
         Parameters
@@ -352,12 +427,26 @@ def _worker(index, env_fn, pipe, parent_pipe, shared_memory, error_queue):
             elif command == 'close':
                 pipe.send((None, True))
                 break
+            elif command == '_call':
+                name, args, kwargs = data
+                if name in ['reset', 'step', 'seed', 'close']:
+                    raise ValueError('Trying to call function `{0}` with '
+                        '`_call`. Use `{0}` directly instead.'.format(name))
+                function = getattr(env, name)
+                if callable(function):
+                    pipe.send((function(*args, **kwargs), True))
+                else:
+                    pipe.send((function, True))
+            elif command == '_setattr':
+                name, value = data
+                setattr(env, name, value)
+                pipe.send((None, True))
             elif command == '_check_observation_space':
                 pipe.send((data == env.observation_space, True))
             else:
                 raise RuntimeError('Received unknown command `{0}`. Must '
-                    'be one of {`reset`, `step`, `seed`, `close`, '
-                    '`_check_observation_space`}.'.format(command))
+                    'be one of {`reset`, `step`, `seed`, `close`, `_call`, '
+                    '`_setattr`, `_check_observation_space`}.'.format(command))
     except (KeyboardInterrupt, Exception):
         error_queue.put((index,) + sys.exc_info()[:2])
         pipe.send((None, False))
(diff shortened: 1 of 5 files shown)


namheegordonkim:
Is this getting merged anytime soon?

tristandeleu:
@pzhokhov I have rebased on master to remove any merge conflict. This PR is now ready to be merged or closed.

gym/vector/async_vector_env.py
```diff
@@ -420,13 +503,29 @@ def _worker(index, env_fn, pipe, parent_pipe, shared_memory, error_queue):
             elif command == "close":
                 pipe.send((None, True))
                 break
+            elif command == "_call":
+                name, args, kwargs = data
+                if name in ["reset", "step", "seed", "close"]:
+                    raise ValueError(
+                        "Trying to call function `{0}` with "
+                        "`_call`. Use `{0}` directly instead.".format(name)
```
FirefoxMetzger, line 511:
```suggestion
                        f"Trying to call function `{name}` with "
                        f"`_call`. Use `{name}` directly instead."
```

gym/vector/async_vector_env.py
```diff
@@ -463,13 +562,29 @@ def _worker_shared_memory(index, env_fn, pipe, parent_pipe, shared_memory, error
...
+                pipe.send((None, True))
             elif command == "_check_observation_space":
                 pipe.send((data == observation_space, True))
             else:
                 raise RuntimeError(
                     "Received unknown command `{0}`. Must "
-                    "be one of {`reset`, `step`, `seed`, `close`, "
-                    "`_check_observation_space`}.".format(command)
+                    "be one of {`reset`, `step`, `seed`, `close`, `_call`, "
+                    "`_setattr`, `_check_observation_space`}.".format(command)
```
FirefoxMetzger, line 587: I can't suggest a format string here, since the relevant line was only partially modified, but I'd probably use it here, too.

gym/vector/sync_vector_env.py
```diff
@@ -95,6 +95,48 @@ def step_wait(self):
             infos,
         )
 
+    def call(self, name, *args, **kwargs):
+        """
+        Parameters
+        ----------
+        name : string
+            Name of the method or property to call.
```
FirefoxMetzger, line 103: Similar to above, would it be worthwhile to document args and kwargs?

gym/vector/sync_vector_env.py
```diff
@@ -95,6 +95,48 @@ def step_wait(self):
...
+            Name of the property to be set in each individual environment.
+
+        values : list or object
+            Values of the property to bet set to. If `values` is a list, then
+            it corresponds to the values for each individual environment,
+            otherwise a single value is set for all environments.
+        """
+        if not isinstance(values, list):
+            values = [values for _ in range(self.num_envs)]
+        assert len(values) == self.num_envs
```
FirefoxMetzger, line 135: Similar to above, are we worried about other iterables?

FirefoxMetzger (commented):
Nice work 🚀 I left some minor comments that you may choose to accept or reject depending on what you think the best solution is here.

gym/vector/async_vector_env.py
```diff
@@ -301,6 +302,88 @@ def step_wait(self, timeout=None):
...
+        Parameters
+        ----------
+        name : string
+            Name of the method or property to call.
+        """
+        self._assert_is_running()
+        if self._state != AsyncState.DEFAULT:
+            raise AlreadyPendingCallError(
+                "Calling `call_async` while waiting "
+                "for a pending call to `{0}` to complete.".format(self._state.value),
```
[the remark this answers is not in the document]
tristandeleu: Indeed f-strings are much nicer! At the time CI was testing also for versions 2.7 and 3.5, that shows how old this PR is.

gym/vector/async_vector_env.py
```diff
@@ -301,6 +302,88 @@ def step_wait(self, timeout=None):
...
+        results : list
+            List of the results of the individual calls to the method or
+            property for each environment.
+        """
+        self._assert_is_running()
+        if self._state != AsyncState.WAITING_CALL:
+            raise NoAsyncCallError(
+                "Calling `call_wait` without any prior call " "to `call_async`.",
+                AsyncState.WAITING_CALL.value,
+            )
```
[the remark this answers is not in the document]
tristandeleu: This is matching the rest of the API of `AsyncVectorEnv` (`reset_wait` and `step_wait`), but this is not a future-based one because the corresponding `*_async` methods do not return any future. The idea was to be as conservative as possible to avoid anything going wrong, but it could be swapped with a warning/empty list.

Personally, I like the option of raising an exception that can be caught by the user later on better, but I'm open to this change.

gym/vector/async_vector_env.py
```diff
@@ -301,6 +302,88 @@ def step_wait(self, timeout=None):
...
+            Name of the property to be set in each individual environment.
+
+        values : list or object
+            Values of the property to bet set to. If `values` is a list, then
+            it corresponds to the values for each individual environment,
+            otherwise a single value is set for all environments.
+        """
+        self._assert_is_running()
+        if not isinstance(values, list):
+            values = [values for _ in range(self.num_envs)]
```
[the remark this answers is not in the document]
tristandeleu: I don't think allowing general iterables is a good idea, since we want to control the number of elements in `values`, which you can't do with a general iterable (e.g. a generator, which doesn't have a `__len__` implemented). Again this is a conservative design, but I think being precise in the docstring about the kind of inputs `set_attr` can take (a list), then the user can adapt if they have an iterable and do something like
```python
envs.set_attr(list(iterable))
```

gym/vector/async_vector_env.py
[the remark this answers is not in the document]
FirefoxMetzger: Depends if you see AsyncVectorEnv as a state machine (in which case it may be invalid to call_wait if the environment is not in state `WAITING_CALL`, or if you see it as a parallel computing fan-out/fan-in. In the former case, I think the exception is better (we attempted an invalid action in the current state), in the latter a simple return is better (no work was scheduled, so we don't need to wait for any work to be completed).

gym/vector/async_vector_env.py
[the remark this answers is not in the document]
FirefoxMetzger: I see, perhaps it is still worthwhile to consider allowing `tuple` input given their speed and memory advantages?

gym/vector/async_vector_env.py
```diff
@@ -308,12 +306,15 @@ def call_async(self, name, *args, **kwargs):
         ----------
         name : string
             Name of the method or property to call.
+
+        args, kwargs :
+            Arguments and keyword arguments to apply to the method call.
```
FirefoxMetzger, line 311:
```suggestion
        args, kwargs :
            Arguments and keyword arguments to apply to the method call.
```

Nitpick: remove extra whitespace.

gym/vector/async_vector_env.py
```diff
@@ -362,20 +362,25 @@ def set_attr(self, name, values):
         name : string
             Name of the property to be set in each individual environment.
 
-        values : list or object
-            Values of the property to bet set to. If `values` is a list, then
+        values : list of object
```
FirefoxMetzger, line 365: Should this be `or` like before (suggested by the description and by the code), or `of` as changed to now? If the latter, I think the description should be updated to reflect that change. Should the source code raise an exception on non-list types, too?

gym/vector/async_vector_env.py
```diff
@@ -362,20 +362,25 @@ def set_attr(self, name, values):
...
             it corresponds to the values for each individual environment,
             otherwise a single value is set for all environments.
         """
         self._assert_is_running()
         if not isinstance(values, list):
             values = [values for _ in range(self.num_envs)]
-        assert len(values) == self.num_envs
+        if len(values) != self.num_envs:
+            raise ValueError(
+                "The values must be a list of length the number "
```
FirefoxMetzger, line 375:
```suggestion
                "Values must be a list with length equal to the number "
```

gym/vector/sync_vector_env.py
```diff
@@ -125,14 +128,19 @@ def set_attr(self, name, values):
         name : string
             Name of the property to be set in each individual environment.
 
-        values : list or object
-            Values of the property to bet set to. If `values` is a list, then
+        values : list of object
```
FirefoxMetzger, line 131: Same question as previously; is the `of` here intentional?

FirefoxMetzger (commented):
Some updates based on the changes you added :)

gym/vector/async_vector_env.py
tristandeleu: No this was a mistake indeed, I thought it was a typo but it is list *or* object.

gym/vector/async_vector_env.py
tristandeleu: I didn't know what was the change you suggested (where the extra whitespace was), but I changed the docstring to match more closely the [Numpy docstring guidelines](https://numpydoc.readthedocs.io/en/latest/format.html#parameters). I also moved the docstrings from `SyncVectorEnv` to the base class `VectorEnv` to match what was done for the other methods.

gym/vector/async_vector_env.py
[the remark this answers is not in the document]
tristandeleu: The idea was indeed something more like a state machine, following [the design of `SubprocVecEnv` in baselines](https://github.com/openai/baselines/blob/ea25b9e8b234e6ee1bca43083f8f3cf974143998/baselines/common/vec_env/subproc_vec_env.py#L80) (which was the source of inspiration for `AsyncVectorEnv`, and our discussion in the original PR https://github.com/openai/gym/pull/1513#discussion_r291767318.

Actually looking back at the original PR, I initially had it with a simple warning, but it was suggested by the maintainers to make it fatal https://github.com/openai/gym/pull/1513#discussion_r291765948.

gym/vector/async_vector_env.py
FirefoxMetzger: it's really just a small thing; there was a leading newline (`\n`) above `args, kwargs` that separated it from the previous parameter. If you look closely the suggestion changes 3 lines of code into 2 lines of code.

tristandeleu:
What is the status of this PR? Any timeline for this PR getting merged? @jkterry1

vwxyzjn:
@jkterry1 this PR makes sense to me. In gym-microrts we use this feature to get invalid action masks. 

Everything in the PR looks good to me except the testing environment might need to be replaced with `CartPole-v1`. `CubeCrush-v0` has been deprecated in #2553.

jkterry1:
There's also merge conflicts. If you're happy with it I'll merge it once tests and merge conflicts are fixed.

vwxyzjn:
LGTM

PR #1601, opened 2019-07-12
Add episodic argument to VectorEnv

tristandeleu:
This PR adds an "episodic" mode for `VectorEnv`. In episodic mode, the environment does not call automatically `reset` once the episode of an environment is `done=True`. The user is responsible to call `reset` herself. This would allow the following pattern:
```python
import gym
import numpy as np

def make_env():
    return gym.make('CartPole-v1')

env_fns = [make_env for _ in range(8)]
env = gym.vector.AsyncVectorEnv(env_fns, episodic=True)
observations = env.reset()
dones = np.zeros(env.num_envs, dtype=np.bool_)

while not dones.all():
    actions = env.action_space.sample()
    observations, rewards, dones, infos = env.step(actions)
```

The episodic mode is useful in case one wants to only sample a fixed amount of episodes, rather than continuously sample episodes. Once an episode is over, the environment adds a blank observation with 0 reward and `done=True`, to keep batches of observations/rewards/dones consistent. Note that the default behavior is `episodic=False`, where `reset` is called automatically.

I would like to discuss this PR in greater details before going with it.

### DIFF
--- a/gym/vector/async_vector_env.py
+++ b/gym/vector/async_vector_env.py
@@ -48,12 +48,18 @@ class AsyncVectorEnv(VectorEnv):
         If `True`, then the `reset` and `step` methods return a copy of the
         observations.
 
+    episodic : bool (default: `False`)
+        If `True`, then the environments run for a single episode (until
+        `done=True`), and subsequent calls to `step` have an unexpected
+        behaviour. If `False`, then the environments call `reset` at the end of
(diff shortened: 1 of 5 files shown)


tristandeleu:
Closing this PR because this can be implemented as a custom worker for `AsyncVectorEnv`.

PR #1602, opened 2019-07-12
Add automatic restart of failing processes in AsyncVectorEnv

tristandeleu:
This PR adds an option to automatically restart processes in `AsyncVectorEnv` if one of the processes failed, up to a `max_retries` number of retries. The goal is to avoid the vectorized environment to fail if there is an issue with one environment. Failures may happen, and it would can be frustrating to have a whole run killed by some failing environment.

This PR has gone through a significant rework since #1513, to account for the more recent changes added, so it would probably need a more careful review. I think it could be a nice feature to have, but since I have never faced that situation of a failing environment myself, I would be happy to have any feedback and discuss this change!

### DIFF
--- a/gym/vector/async_vector_env.py
+++ b/gym/vector/async_vector_env.py
@@ -48,21 +48,27 @@ class AsyncVectorEnv(VectorEnv):
         If `True`, then the `reset` and `step` methods return a copy of the
         observations.
 
+    max_retries : int (default: 0)
+        Maximum number of times `AsyncVectorEnv` tries to restart a process if
+        it fails. If `None`, then it always tries to restart a failing process.
+
     context : str, optional
         Context for multiprocessing. If `None`, then the default context is used.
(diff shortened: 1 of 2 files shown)


christopherhesse:
Thanks for the PR!  I'd like to wait until there are some users of this feature before merging it though.  We haven't needed this with our VecEnv implementation, and it's not clear to me that this should exist here instead of as a wrapper.

tristandeleu:
I agree, ideally it would make more sense to have this feature as a wrapper around `AsyncVectorEnv`. But this requires a number of backend changes (like the `_restart` function in the workers, `obs_buffer` and `ctx` as properties of `AsyncVectorEnv`), which would a priori be outside of the scope of what a wrapper would be able to do.

That being said, I also agree that we should wait before merging this PR. It is unclear that it would actually be useful at all.

PR #1605, opened 2019-07-15
fix python patch version

pzhokhov:
fixes failing tests on master - looks like something is off in newer versions of python docker images and opengl rendering for mujoco tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,9 @@ services:
 env:
   # - UBUNTU_VER=14.04 - problems with atari-py
   - PY_VER=2.7
-  - PY_VER=3.5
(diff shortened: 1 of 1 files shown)


pzhokhov:
seems to be related to https://github.com/mapbox/mapbox-gl-native/issues/12223 and https://github.com/glfw/glfw/issues/833

pzhokhov:
I think the problem with that is that mujoco does not actually use the system-level glfw, instead, it comes packaged with glfw binaries. I suspect the alternative solution is to replace those after mujoco installation. Neither fixing python patch version nor replacing the binaries strikes me as a particularly elegant, but python version seems like a lesser of two evils.

PR #1606, opened 2019-07-15
Update supported os sections

pzhokhov:


pzhokhov:
@abhinavsagar oh right, that's why you made the original PR in the first place. Sorry, should be fixed now

Issue #1598, opened 2019-07-12
env.render(mode='rgb_array') fails now


ZurMaD:
Hi Openai Team, I was using the env command without errors weeks ago but now I get this error

![imagen](https://user-images.githubusercontent.com/28235457/61154296-58f40880-a4b3-11e9-8eb2-4bf79b71f68f.png)

An stops in this file

https://github.com/openai/gym/blob/f33408c4fd85114baa7b1938b127c0254e8df536/gym/envs/classic_control/rendering.py#L90-L116

christopherhesse:
This should be fixed in `gym==0.13.1` by https://github.com/openai/gym/pull/1591.  pyglet was updated and broke gym.

ZurMaD:
Thanks.

PR #1607, opened 2019-07-18
replace data with get_data()

cprogrammer1994:
Description:

The `env.render('rgb_array')` is broken.

Test code:

```py
import gym
env = gym.make('Pendulum-v0')
env.reset()
env.render('rgb_array')
```

Error:

```
AttributeError: 'ImageData' object has no attribute 'data'
```

I have `gym==0.13.0` and `pyglet==1.4.0b1`


cprogrammer1994:
Installing `gym==0.13.1` installs an older `pyglet` version, so the problem is fixed already.
closing...

link https://github.com/openai/gym/issues/1598

PR #1626, opened 2019-08-01
[Update registration.py]: Clean-up of unused arguments

zuoxingdong:


wookayin:
Unfortunately, this breaks some public APIs.

PR #1628, opened 2019-08-02
[Wrapper]: RecordEpisodeStatistics

zuoxingdong:


gym/wrappers/record_episode_statistics.py
```diff
@@ -0,0 +1,34 @@
...
+
+import gym
+
+
+class RecordEpisodeStatistics(gym.Wrapper):
+    def __init__(self, env, deque_size=100):
+        super(RecordEpisodeStatistics, self).__init__(env)
+        self.t0 = time.time()  # TODO: use perf_counter when gym removes Python 2 support
+        self.episode_return = 0.0
+        self.episode_horizon = 0
```
pzhokhov: would episode_length not be more descriptive?

pzhokhov:
this seems very similar to baselines Monitor class (except for saving data to a file); however, uses completely different names and so is incompatible. @zuoxingdong do you think it makes sense to change the info dict keys for the sake of compatibility?

gym/wrappers/record_episode_statistics.py
zuoxingdong: @pzhokhov Indeed, it is more descriptive. Thanks!

zuoxingdong:
@pzhokhov Thanks a lot for your comments. I've made the change accordingly.

PR #1632, opened 2019-08-02
[gym.vector]: add `last_observation` to `info`

zuoxingdong:


pzhokhov:
this is an interesting idea.. could you elaborate on when would the last observation be useful? 
Also, there is a technical problem with this - the info dicts are not batched the same way observations are; so last_observation will have different format from all other observations.

zuoxingdong:
@pzhokhov It might be very useful when we want to have an additional forward pass to get the value function for the last observation so that it could be used for bootstrapping, e.g. in policy gradient context.

zuoxingdong:
for the format in the info dicts, would you recommend to add a batch dimension 1 for each info dictionary?

A use case of that could be e.g.
```python
np.concatenate([info['last_observation'] for info in infos], 0)
```

pzhokhov:
For value function bootstrapping you normally would not want to bootstrap the value of the last state in the episode - because regardless of what value function estimate is, the reward to go from the last state is 0.  Arguably. you may want to try to learn the value function that understands that last state is last and returns 0. 
As for batch dimension - yes, that would work, I think ideally we should batch leaves of the info dicts for all the fields (not only for last observation), something like

```python
def batch_leaves(infos):
  assert isinstance(infos, (list, tuple))
  if isinstance(infos[0], dict):
     return {k: batch_leaves([i[k] for i in infos]) for k in infos[0].keys()}
  elif isinstance(infos[0], list):
     return [batch_leaves([i[k] for i in infos]) for k in range(len(info[0]))]
  elif isinstance(infos[0], numpy.ndarray): 
     return np.concatenate(infos, axis=0)
  else: 
     return infos
```

tristandeleu:
One issue with batching infos in general is that there is no guarantee that the infos returned by the different environments have the same key. For example in this case, only the infos corresponding to the end of episodes will have the `last_observation` key available, so out of the box, `np.concatenate` or `batch_leaves(infos)` would fail.

This is not an issue we have with observations and actions because in those cases, we do have a fixed template given by the `observation_space`/`action_space`.

For those reasons, you won't be able to batch the last observations in such a way that it has the same shape as a regular batched observation. However, something like
```python
from gym.vector.utils import create_empty_array, concatenate

zero_observation = create_empty_array(envs.single_observation_space, n=None)
last_observations = concatenate(
    [info.get('last_observation', zero_observation) for info in infos],
    create_empty_array(envs.single_observation_space, n=envs.num_envs),
    envs.single_observation_space
)
```
should probably work. It fills the other observations with zeros, and you can retrieve which ones are the "real" last observations with `dones`.

vwxyzjn:
Hey, I am following up with this PR. This feature should be added to gym. It is useful in DQN where the terminal observation is important for the replay buffer.

@zuoxingdong would you mind changing the key name to `terminal_observation` instead? `terminal_observation` seems to be a better term. The PR otherwise looks good to me. I also don't feel it's necessary in ensuring the `terminal_observation` to have the same dimension as `obs`.

@tristandeleu @pzhokhov I do something like the following with SB3's vector wrapper. See the code [here](https://github.com/vwxyzjn/cleanrl/blob/e98726da159796163e8be7a80f0c9421bcfd6d49/cleanrl/atari/dqn_atari.py)

```
        # TRY NOT TO MODIFY: save data to reply buffer; handle `terminal_observation`
        real_next_obs = next_obs.copy()
        for idx, d in enumerate(dones):
            if d:
                real_next_obs[idx] = infos[idx]["terminal_observation"]
        rb.add(obs, real_next_obs, actions, rewards, dones)
```

PR #1633, opened 2019-08-02
[Update vector_env.py]: add `VectorEnvWrapper`

zuoxingdong:


gym/vector/vector_env.py
```diff
@@ -57,3 +57,65 @@ def __del__(self):
...
+        return self.env.step(actions)
+
+    def get_images(self):
+        return self.env.get_images()
+
+    def render(self, mode='human', **kwargs):
+        return self.env.render(mode, **kwargs)
+
+    def close(self, **kwargs):
+        return self.env.close(**kwargs)
```
pzhokhov: AFAIU all these don't need to be de defined, because the attribute forwarding  in `__getattar__` will take care of them.

gym/vector/vector_env.py
```diff
@@ -57,3 +57,65 @@ def __del__(self):
...
+            raise AttributeError("attempted to get missing private attribute '{}'".format(name))
+        return getattr(self.env, name)
+
+    @property
+    def spec(self):
+        return self.env.spec
+
+    @classmethod
+    def class_name(cls):
+        return cls.__name__
```
pzhokhov: I know there is a similar method in the Wrapper class, but it seems very redundant (in both cases)

gym/vector/vector_env.py
```diff
@@ -57,3 +57,65 @@ def __del__(self):
...
+    """
+    def __init__(self, env):
+        assert isinstance(env, VectorEnv)
+        self.env = env
+        self.metadata = env.metadata
+        self.num_envs = env.num_envs
+        self.observation_space = env.observation_space
+        self.action_space = env.action_space
+        self.single_observation_space = env.single_observation_space
+        self.single_action_space = env.single_action_space
```
pzhokhov: these can also be forwarded by the attribute forwarding

zuoxingdong:
Thanks @pzhokhov for pointing this out. I do indeed agree with you for the code redundancy. Now it's updated accordingly. 

I'll send another PR for similar refactoring in `core.py`

PR #1634, opened 2019-08-02
[gym.vector]: allow access to `reward_range` and `spec`

zuoxingdong:


pzhokhov:
To my experience, the reward_range and spec are pretty much never used; so I think it is better to gradually push it out of the API, rather than propagating. Please comment if there is a strong use case why those are needed in VectorEnv, otherwise, closing this

PR #1635, opened 2019-08-02
[Wrapper]: add `NormalizeObservation` and `NormalizeReward`

zuoxingdong:


pzhokhov:
observation / reward normalization is a bit of contested topic - on one hand, it can greatly help learning across a variety of environments, on the other hand, it makes the environment stateful (i.e. reset does not return the environment quite to a clean state), agent become hard to serialize (because one has to save state of the environment along with the state of the agent) and training on environments with sparse rewards can be completely messed up. As such, I don't think this should a part of the gym.

PR #1647, opened 2019-08-13
fix hand serialization tests

pzhokhov:
by moving EzPickle inheritance to the leaf classes and passing correct arguments to EzPickle


pzhokhov:
yeah, there was a PR that added these tests. It allegedly fixed the serialization too, however, what I did not catch is that the PR does not run mujoco tests (because of the key injection), so they only ran and failed once in master.

PR #1660, opened 2019-08-26
Added an environment for the board game Go

aigagror:


aigagror:
Hello can someone the OpenAI team say whether this environment can be added or not?

christopherhesse:
Looks good, could you add it to the bottom of the list?  I think we've been adding them there until we have some better method for sorting them.

aigagror:
Ok I did that

aigagror:
On a side note, I was wondering if third party environments every get added to the official list. If so, how does that process work?

pzhokhov:
@aigagror thanks!  What do you mean by the official list? To the gym itself, to the gym website (http://gym.openai.com)? If the former (to the gym itself), then basically that does not happen anymore, in fact, we are trying to slowly move environments out of the gym (and only keep Env interface, wrappers and tools). If the latter (to the gym website), then the link to the third-party environments is already there.

PR #1668, opened 2019-09-04
Add text label while rendering environment

Tony-Feng:
Add text label in `rendering.py` from folder `classic_control`

### DIFF
--- a/gym/envs/classic_control/rendering.py
+++ b/gym/envs/classic_control/rendering.py
@@ -139,6 +139,12 @@ class Viewer(object):
         _add_attrs(geom, attrs)
         self.add_onetime(geom)
         return geom
+    
+    def draw_text(self, text, position=(0, 0), **attrs):
+        geom = Text(text, position)
+        _add_attrs(geom, attrs)
+        self.add_onetime(geom)
+        return geom
 
     def get_array(self):
         self.window.flip()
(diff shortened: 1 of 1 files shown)


pzhokhov:
what's the usecase for this change?

Tony-Feng:
```python3
import gym
import gym.envs.classic_control.rendering as rendering

class Env(gym.Env):
    def __init__(self):
        self.viewer = rendering.Viewer(900, 900)
        self.viewer.set_bounds(*(-10, 10, -10, 10))
    
    def display(self):
        self.viewer.draw_text("Hello World!")
        return self.viewer.render()
        
    def close(self):
        self.viewer.close()
        
if __name__ == "__main__":
    e = Env()
    for _ in range(200):
        e.display()
    e.close()
```
Thank you for your response. Here is the usecase. It will display `Hello World!` in the center.

pzhokhov:
right, but it is not used in any of the environments currently existing in the gym, correct? I'd say if it is not used let's not include it into the gym. IMO if you are planning to use it for an external (with respect to gym) environment, the codebase of such environment would be a better place for this change.

pzhokhov:
Also, as @christopherhesse points out, it is possible to use the rest of the functionality of classic control rendering by subclassing Viewer class in your codebase. Closing this.

PR #1669, opened 2019-09-05
[Delete dict.py]: replace with FilterObservation + FlattenObservation

zuoxingdong:
The functionality in this wrapper can be achieved by `FilterObservation` + `FlattenObservation`, thus it is to be removed to reduce redundancy.


pzhokhov:
test failures are related to the bug fixed here: https://github.com/openai/gym/pull/1710

PR #1674, opened 2019-09-09
Mountain car mppi

eeching:


pzhokhov:
Hi @eeching and thanks for the PR! I don't think it belongs in the gym though - as you notice, we don't have any control algorithms here, only the environments and wrappers / tools.  As such, I am closing this PR.

PR #1677, opened 2019-09-11
Updated registration.py for registering custom task_env registration

saifullah3396:
Previously the load function was giving attribute errors on recursive import calls of task environments such as:
AttributeError: 'module' object has no attribute 'task_envs.turtlebot3.turtlebot3_world.TurtleBot3WorldEnv'.
The update simply allows the registration of the task environment TurtleBot3WorldEnv with consecutive getattr() calls.

### DIFF
--- a/gym/envs/registration.py
+++ b/gym/envs/registration.py
@@ -15,7 +15,7 @@ env_id_re = re.compile(r'^(?:[\w:-]+\/)?([\w:.-]+)-v(\d+)$')
 def load(name):
     mod_name, attr_name = name.split(":")
     mod = importlib.import_module(mod_name)
-    fn = getattr(mod, attr_name)
+    fn = reduce(getattr, attr_name.split("."), mod)
     return fn
 
 


christopherhesse:
Do you have an example when this is necessary?  I can't think of any case where `attr_name` should have the `.` symbol in it.

saifullah3396:
Well, this problem occurred when I was working on the setting up the openai_ros project at http://wiki.ros.org/openai_ros. In the task environments of openai_ros, whenever they register a custom environment to gym they call a function like this:
```
register(
            id=task_env,
            entry_point='openai_ros:task_envs.turtlebot3.turtlebot3_world.TurtleBot3WorldEnv'
        ) 

```
That's where the error was arising... python was unable to parse the attribute 'task_envs.turtlebot3.turtlebot3_world.TurtleBot3WorldEnv'.
You can see for yourself at: https://bitbucket.org/theconstructcore/openai_ros/src/version2/openai_ros/src/openai_ros/task_envs/task_envs_list.py.

pzhokhov:
@saifullah3396 sorry about the lag. I think in your case the entry point syntax is somewhat off - if the `TurtleBot3WorldEnv` class is in the module `task_envs.turtlebot3.turtlebot3_world` of package `openai_ros` (I looked briefly at the code you linked and it seemed to be the case), then it `entry_point` should rather be
`openai_ros.task_envs.turtlebot3.turtlebot3_world:TurtleBot3Env`. In other words, it is the name of the class / callable that is separated from the module name by the semicolon, not the package from the module.

saifullah3396:
@pzhokhov it's alright and thank you very much for explaining the problem. I guess I should have raised the issue on openai_ros repository instead of on gym since I was just trying to build their code and this problem occurred. I will try the syntax you've provided here and see if it works fine!

Issue #1666, opened 2019-08-30
env.close() not working as intended?


pkoppol:
Hello,

I recently installed mujoco_py (2.0.2.5) on an Ubuntu 18.04 system, and was playing around with the FetchReach environment when I noticed that env.close() was not properly closing the render window. 

Instead, the window would freeze and become unresponsive, causing me to force-quit and terminate my program early. I believe that adding:

```
if self.viewer is not None:
               glfw.destroy_window(self.viewer.window)
               self.viewer = None
```
will resolve this issue. 

As I'm writing this, I found a related issue (#1000) where someone encountered a similar problem and surfaced the same solution about a year ago. I'm wondering if there's a reason that this hasn't been addressed, or if it makes sense to go ahead with a pull request.

Thanks!

christopherhesse:
I don't know of any reason, please provide a pull request if this change doesn't break anything else.

christopherhesse:
From looking more closely at the PR, I think this is a bug in `mujoco-py`, so you should file the issue/PR there, sorry for the confusion!

PR #1678, opened 2019-09-11
Fix env.close() viewer issue

pkoppol:
Without glfw.destroy_window, the viewer window freezes and becomes unresponsive. Adds fix proposed in #1666

### DIFF
--- a/gym/envs/robotics/robot_env.py
+++ b/gym/envs/robotics/robot_env.py
@@ -1,5 +1,6 @@
 import os
 import copy
+import glfw
 import numpy as np
 
 import gym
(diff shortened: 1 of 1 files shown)


christopherhesse:
This seems unlikely to be the correct way to do this, since we use `pyglet` and not `glfw`.   What about calling the `viewer`'s `.close()` function? https://github.com/openai/gym/blob/master/gym/envs/classic_control/rendering.py#L70

I don't see any other way to close a window in `pyglet` besides the one used in `viewer.close()`.

christopherhesse:
gym only has an optional dependency on `glfw`.  This could be fixed by doing a late import of `glfw`, but this seems like something that should be fixed in `MjViewer`.    Could you change `MjViewer` such that it has `close()` method that is called in `__del__`?  https://github.com/openai/mujoco-py/blob/41a5fc26eaeda0bb88aa7ba8e20606c411879381/mujoco_py/mjviewer.py#L15

christopherhesse:
Going to close this since it looks like a mujoco issue.

Issue #1679, opened 2019-09-17
Frame Normalization in AtariPreProcessing


koulanurag:
Hi,
I guess, it would be useful to include [frame normalization](https://github.com/openai/baselines/blob/master/baselines/common/atari_wrappers.py#L219) as well in the [AtariPreProcessing](https://github.com/openai/gym/blob/master/gym/wrappers/atari_preprocessing.py).

This is also included optionally in [wrap_deepmind](https://github.com/openai/baselines/blob/master/baselines/common/atari_wrappers.py#L275) method

koulanurag:
@zuoxingdong @pzhokhov. This might be relevant to you.

christopherhesse:
Looks like this was merged https://github.com/openai/gym/pull/1680

PR #1680, opened 2019-09-17
Adding Frame Normalization in Atari Pre-Processing

koulanurag:
related to #1679


koulanurag:
#1680

pzhokhov:
I like this, but please add a unit test

zuoxingdong:
Hi @koulanurag @pzhokhov , it might be nice to add some warning in docstring about `Scale observation`, because if it's turned on, then the benefits of `FrameStack` memory optimization will break. So at least provide this warning in the documentation might be nice for the users.

koulanurag:
Hi @pzhokhov @zuoxingdong 

I have added the unit test cases and the required message in the docstring.

PR #1706, opened 2019-10-09
vector env updates

pzhokhov:
AsyncVectorEnv updates: allow `daemon` flag to be on or off, allow custom workers (for messing with reset logic in a vecenv)
SyncVectorEnv - update API with `step_async` and `step_wait` methods for compatibility

### DIFF
--- a/gym/vector/async_vector_env.py
+++ b/gym/vector/async_vector_env.py
@@ -51,6 +51,20 @@ class AsyncVectorEnv(VectorEnv):
     context : str, optional
         Context for multiprocessing. If `None`, then the default context is used.
         Only available in Python 3.
+
+    daemon : bool (default: `True`)
+        If `True`, then subprocesses have `daemon` flag turned on; that is, they
+        will quit if the head process quits. However, `daemon=True` prevents
+        subprocesses to spawn children, so for some environments you may want
+        to have it set to `False`
+
+    worker : function, optional
+        WARNING - advanced mode option! If set, then use that worker in a subprocess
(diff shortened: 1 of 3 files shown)


gym/vector/sync_vector_env.py
```diff
@@ -103,8 +104,15 @@ def step(self, actions):
         infos : list of dict
             A list of auxiliary diagnostic informations.
         """
+        self._actions = actions
```
christopherhesse: for consistency, this should just be self.step_async

christopherhesse: actually you should remove self.step since it's defined in the parent class

gym/vector/async_vector_env.py
```diff
@@ -53,7 +53,7 @@ class AsyncVectorEnv(VectorEnv):
         Only available in Python 3.
     """
     def __init__(self, env_fns, observation_space=None, action_space=None,
-                 shared_memory=True, copy=True, context=None):
+                 shared_memory=True, copy=True, context=None, daemon=True, worker=None):
```
christopherhesse: When do you want to override the worker?

pzhokhov: currently, worker calls reset every time environment returns done, I want to override worker to replace that logic with our AsyncResetOnDone logic

pzhokhov: The irony is that there was a PR with a very similar change in baselines, and I closed it because I was not motivated enough by the example :)

PR #1713, opened 2019-10-18
Update setup.py

modanesh:
Just added opencv-python to be installed as a requirement. The new atari_preprocessing.py wrapper needs it.


pzhokhov:
I am having second thoughts about this - given that opencv-python is specified in `gym[atari]` requirements, I don't think it is necessary to put it in the core requirements after all

modanesh:
I know, but then there will an error after using the new atari_preprocessing.py. This way it's more straight forward.

pzhokhov:
I agree that the current way is more concise, however, as pointed out in #1762 , on some platforms that does not work :) I made a PR to revert this change, and added an error message to atari_preprocessing.py in case opencv-python is not installed (#1764)

PR #1717, opened 2019-10-19
Fixed bug in CliffWalking env, cliff states were no terminal states. Memory error.

d4vidbiertmpl:
Hi,
in the current CliffWalking environment, the cliff states are not considered terminal states, which is not how the problem is defined for example in Sutton & Barto. Also, this causes memory errors as the agent takes 7660 steps (and receives -89434 reward) in the first episode. 

My changes fix this issue. I tried to be as pythonic as possible.

Cheers, 
David.

### DIFF
--- a/gym/envs/toy_text/cliffwalking.py
+++ b/gym/envs/toy_text/cliffwalking.py
@@ -41,6 +41,9 @@ class CliffWalkingEnv(discrete.DiscreteEnv):
         self._cliff = np.zeros(self.shape, dtype=np.bool)
         self._cliff[3, 1:-1] = True
 
+        _cliff_coords = np.argwhere(self._cliff)
+        self.terminal_states = np.vstack((_cliff_coords, np.asarray([self.shape[0] - 1, self.shape[1] - 1])))
+
         # Calculate transition probabilities and rewards
         P = {}
         for s in range(nS):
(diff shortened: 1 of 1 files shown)


christopherhesse:
From reading the text, it's not obvious to me that the episode ends when reaching a cliff square.  Does it say that in the text?  If the text is ambiguous I'd prefer to keep the current interpretation.

d4vidbiertmpl:
Hi Christopher,

My source is the book "Reinforcement Learning: An Introduction" by Richard S. Sutton and Andrew G. Barto (http://incompleteideas.net/book/the-book.html). 

The cliff world is described in example 6.6 on page 132, where it says: "Reward is -1 on all
transitions except those into the region marked “The Cliff”. Stepping into this region incurs a reward of -100 and sends the agent instantly
back to the start."

From my understanding this is the official definition of this example and the correct way to implement it.

pzhokhov:
Here's a diagram from the respective chapter: 
Screen Shot 2019-11-01 at 3 54 22 PM
it has transitions from cliff to start marked explicitly as arrows with R=-100, but transitions from goal to start not marked.
Based on that, I agree with @christopherhesse that walking off the cliff causes teleportation to start rather than the end of the episode. I also agree with @d4vidbiertmpl that the end of the episode with reward -100 seems more natural... however, facing this uncertainty, I am inclined to not make the change. Closing this.

PR #1723, opened 2019-10-28
[Update core.py]: Remove redundancy

zuoxingdong:
according to the discussion in #1633

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -30,14 +30,15 @@ class Env(object):
     non-underscored versions are wrapper methods to which we may add
     functionality over time.
     """
-    # Set this in SOME subclasses
-    metadata = {'render.modes': []}
-    reward_range = (-float('inf'), float('inf'))
-    spec = None
+    def __init__(self):
+        # Set this in SOME subclasses
+        self.metadata = {'render.modes': []}
+        self.reward_range = (-float('inf'), float('inf'))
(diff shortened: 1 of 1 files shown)


gym/core.py
```diff
@@ -207,52 +207,19 @@ class Wrapper(Env):
...
-
-    def compute_reward(self, achieved_goal, desired_goal, info):
-        return self.env.compute_reward(achieved_goal, desired_goal, info)
+    def unwrapped(self):
+        return self.env.unwrapped
 
     def __str__(self):
         return '<{}{}>'.format(type(self).__name__, self.env)
 
-    def __repr__(self):
```
pzhokhov: should we rather delete __str__ method? To my understanding. if __str__ is not specified, `str(x)` will call x.__repr__(), but if `__repr__` is not specified, `repr(x)` will *not* call `x.__str__()`

zuoxingdong: @pzhokhov Thanks for pointing this out. You are right. Also according to the discussion [here](https://stackoverflow.com/a/2626364/993191), it is better to implement `__repr__` and let contains default behavior for `__str__ = __repr__`

justinkterry:
@pzhokhov the change on line 220-251 removes key functionality, and the other changes don't do anything of note. You may want to close this.

Issue #1724, opened 2019-10-28
Assert 'NoFrameSkip' in AtariPreprocessing


koulanurag:
I guess , we must assert 'NoFrameSkip'  over [here](https://github.com/openai/gym/blob/master/gym/wrappers/atari_preprocessing.py#L41), otherwise, the specified environment may already have some in-built frame-skipping. For eg: 'Pong-v0'

pzhokhov:
let's move the discussion into the PR

PR #1725, opened 2019-10-28
added assertion of 'NoFrameskip' in AtariPreprocessing

koulanurag:
relevant to #1724


koulanurag:
@pzhokhov Thanks for pointing it out. I have corrected the assertion.

Also, added an assertion for noop_max. Though, It's validated in [here](https://github.com/openai/gym/blob/master/gym/wrappers/atari_preprocessing.py#L101). I guess It's better to assert to avoid any confusion.

koulanurag:
Thank You :)

PR #1728, opened 2019-10-30
Update Readme.md

jmuskaan72:
Added instructions for Windows OS installer of gym toolkit.


christopherhesse:
I suspect that anaconda is not necessary to install gym on windows.  It might be nice for some of the environments, but in that case miniconda is likely a better choice for most users.

PR #1730, opened 2019-11-01
fix sampling bug

### LINKED ISSUE #1729: Sampling Bug
Gym Version: 0.15.3
issue: Box samples numbers above the `high` parameter.

```
from gym.spaces import Box
observation_space = Box(low=-3, high=-1, shape=(9,), dtype='int')
print(observation_space.sample())
>> [ 0 -2  0 -2  0 -1  0 -2  0]
```
The current implementation samples float numbers from uniform distribution of [`low`, `high`] and then converts the resulting samples to desired `dtype`. This runs into the problem of sampling `low` parameter very rarely(and not uniformly) when `dtype` is `int`(as we are converting the floats back to int which results in ceil operation in case of negative numbers) i.e in the above example -3 is almost never sampled as most of the low sampled floats like -2.85, -2.9 get converted to -2.
https://github.com/openai/gym/blob/0cd9266d986d470ed9c0dd87a41cd680b65cfe1c/gym/spaces/box.py#L93-L97

gowtham1997:
Fixes #1729.

I'm clipping the sampled values to be between [`low`, `high`] and also changing `low` while sampling  so that when we convert the float samples back to dtype int, the `low` param gets sampled uniformly

### DIFF
--- a/gym/spaces/box.py
+++ b/gym/spaces/box.py
@@ -70,6 +70,8 @@ class Box(Space):
         """
         high = self.high if self.dtype.kind == 'f' \
                 else self.high.astype('int64') + 1
+        low = self.low if self.dtype.kind == 'f' \
+                else self.low.astype('int64') - 1
         sample = np.empty(self.shape)
 
         # Masking arrays which classify the coordinates according to interval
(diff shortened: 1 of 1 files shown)


gowtham1997:
@pzhokhov Thanks for pointing this out :). I'll update the PR 
```
sample[bounded] = self.np_random.uniform(low=self.low[bounded], 
                                            high=high[bounded],
                                            size=bounded[bounded].shape)
if self.dtype.kind == 'i':
     sample = np.floor(sample)

return sample.astype(self.dtype)
```
^ Just to be clear, I only need to do floor if the dtype is int right?

Issue #1683, opened 2019-09-18
Environment reset() should also return "info", like step() does.


keskival:
To be symmetric with the observation returned from the step() method, I think reset() method should also return info for arbitrary information about the game state.

This would be useful for certain games where non-Space compliant information about the game state could be returned, especially for the subprocess vectorized environments where the environment internal state is not otherwise easily accessible by the agent.

Sure, in principle all observed state should be projected to the observation space, but that is not always possible, which is why I assume returned "info" exists in the step() method in the first place.

christopherhesse:
The lack of symmetry is odd.  But if we look at `step()` as actually `obs = observe()` and `rew, done, info = act()` combined, then `reset()` is `reset()` (which has no return value) + `observe()`.

I think changing this might break the current implementation of vecenvs, since a reset step would need to return (pre-reset info, post-reset info).

christopherhesse:
`info` is generally not meant to be part of the observation.  If you can't represent your observation using a `gym.Space` you can just return some other object and not obey the space interface (though of course this means that anything that requires a valid space to work will not work with your environment).

keskival:
Thank you for the quick response!

I appreciate the clarification of the purpose of the info return item. Currently the documentation states as follows:

"info (dict): diagnostic information useful for debugging. It can sometimes be useful for learning (for example, it might contain the raw probabilities behind the environment’s last state change). However, official evaluations of your agent are not allowed to use this for learning."

This is given in association to other observation information, with observation, reward and done. But as you say, it specifically mentions state change which associates it to act.

I still think the interfaces should be symmetric, or that info should at least be retrievable from the initial state through the vectorized environment interface, because there is no other good way to access such debugging relevant state information in the environment when one does not want to incorporate it to the observation space for various reasons.

pzhokhov:
How about `__getattr__` method in vectorized env, that could query all sub-environments and returns a list of results?

pzhokhov:
in fact, I think there is already a PR to that extent: https://github.com/openai/gym/pull/1600/files (maybe it is reasonable to limit its scope to read-only properties)

jkterry1:
@vwxyzjn @cpnota opinions on what to do with this?

cpnota:
I agree that it should have been designed to return `info` to begin with, but obviously at this point changing the core API is a bad idea.

Regarding #1600 , it seems potentially useful but I have no strong opinions.

vwxyzjn:
[...]

PR #1746, opened 2019-11-21
[Wrappers]: TimeStepEnv

zuoxingdong:
It might be helpful to support this wrapper to enforce output consistency between `step` and `reset`. Implemented as a wrapper in order not to break the current gym API. 

Some discussions in #1683
@christopherhesse @keskival

### DIFF
--- a/gym/wrappers/__init__.py
+++ b/gym/wrappers/__init__.py
@@ -13,3 +13,4 @@ from gym.wrappers.transform_reward import TransformReward
 from gym.wrappers.resize_observation import ResizeObservation
 from gym.wrappers.clip_action import ClipAction
 from gym.wrappers.record_episode_statistics import RecordEpisodeStatistics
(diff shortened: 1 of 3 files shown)


pzhokhov:
as I am late to the party in #1683 , I'll defer the final decision on this to @christopherhesse , however, my take is that this:
1. Is not really necessary - as @christopherhesse pointed out, when implementing a new gym environment, there is nothing that prevents implementer from returning arbitrary structure on reset(); or better yet, adding a special access method. 
2. Does not solve the problem outlined in #1683 as the biggest one - does not really work with vecenvs / vectorenvs. 
3. Causes confusion about / creep of what is the correct gym env api - should the environments return TimeStep on the step, or (observation, reward, done, first) tuple?

christopherhesse:
Thanks for the PR @zuoxingdong, but I don't see a compelling reason to add this.

PR #1748, opened 2019-11-23
Respect the order of keys in a Dict's observation space when flattening

dwiel:
Prior to this change, during flattening, the order of the key/values in the observation was used instead of the order in the Dict's observation space. unflatten already uses the order specified by the Dict's observation space.


dwiel:
I have added unit tests

PR #1754, opened 2019-11-27
Fully automated dev setup + online live demo

jankeromnes:
Hello! I work on gitpod.io, an online IDE based on VS Code, and I automated your dev setup with a small script and Dockerfile.

This allows anyone to quickly get a working dev environment for `gym`, simply by clicking on a link like this:

[https://gitpod.io/#https://github.com/openai/gym](https://gitpod.io/#https://github.com/jankeromnes/gym/blob/master/examples/agents/cem.py)

It will give you an IDE + Linux container where:
- `gym` is already cloned
- dependencies are already installed
- the `cem.py` demo runs inside a preview tab, so that you can play with it

Here is what it looks like:
Screenshot 2019-11-27 at 17 35 03


pzhokhov:
This looks interesting, however, I would prefer to keep gym ide-agnostic... Also, to my understanding, gym is rarely used by itself, and mostly as a dependency for some other projects that train on or otherwise use the gym environments or API, and in that setup having one version of containerized dependencies does not make much sense (because the actual project under development will have its own requirements as well). That being said, I'd be happy to refer to gitpod.io in the readme. Would gitpod work if there is no settings files (.yml and .Dockerfile files) in the gym repo?

jankeromnes:
Hi @pzhokhov, many thanks for the feedback!

> Also, to my understanding, gym is rarely used by itself, and mostly as a dependency for some other projects that train on

Ah, interesting. I originally automated the `gym` setup to help a Gitpod user who was [asking about this](https://spectrum.chat/gitpod/general/has-anyone-managed-to-get-open-ai-gym-working-with-gitpod~c317c6b8-5b2a-4f88-bcf2-32da58fcb721) in our community forum, and thought I'd contribute this setup here in case it's useful to others too.

> That being said, I'd be happy to refer to gitpod.io in the readme. Would gitpod work if there is no settings files (.yml and .Dockerfile files) in the gym repo?

Thank you. If you prefer to have the environment files outside of the `gym` repository, Gitpod also supports external configurations via the [definitely-gp](https://github.com/gitpod-io/definitely-gp) repository. I'm happy to move the `.yml` and `.Dockerfile` files over there, to keep this repository ide-agnostic.

PR #1761, opened 2019-12-08
Fix Box sampling bug

PVShutov:
In case of up-bounded `(-oo, b]` sampling the shifted negative exponential distribution had expected mean value equal to `-1-b` but we want the mean value to have been expected equal to `b-1` in this case


DavidNKraemer:
I mentioned in PR #1567 that we should probably have a smarter way to sample from the noncompact boxes, but I really don't have a sense for why one distribution is more natural than another. But @pzhokhov , I think you're right that modifying the scale parameter based on the `low` / `high` value might be a smart idea.

PR #1789, opened 2020-01-18
Fix BipedalWalker - now returns closest lidar trace instead of furthest

chozabu:
Finding this bug makes me even more impressed anyone has solved BipedalWalkerHardcore-v2 - it seems the observations from lidar have been inconsistent and incorrect, returning the furthest hit result instead of closest.

In screenshots below, I've tweaked the lidar drawing routine to draw last (in front of terrain and objects), and draw every trace each frame to more clearly see what's happening.

Before fix - lidar traces through ground, and hits the side of a pit, giving the agent the impression of a "phantom canyon" in front of the pit, that only appears as it approaches the pit: https://i.imgur.com/XKPnRTR.png

After fix - lidar is stopped by terrain, even when another object is behind it: https://i.imgur.com/Gg8B5BD.png

It feels very counter-intuitive for Box2Ds raycast to return furthest point first, but when I (double) checked the PyBox2D docs, I found that is indeed how it works: https://github.com/pybox2d/pybox2d/wiki/manual

> ... By returning
    0, you set the ray length to zero. By returning the current fraction, you proceed
    to find the **closest point**.

@olegklimov Tagging you since your name pops up the most in relation to BipedalWalker-v2

I've just restarted my models training with this fix applied, and been watching the traces like a hawk - worth noting it still seems to be imperfect sometimes - slightly penetrating a surface, but still looks much more accurate.

### DIFF
--- a/gym/envs/__init__.py
+++ b/gym/envs/__init__.py
@@ -109,14 +109,14 @@ register(
 )
 
 register(
-    id='BipedalWalker-v2',
+    id='BipedalWalker-v3',
     entry_point='gym.envs.box2d:BipedalWalker',
     max_episode_steps=1600,
     reward_threshold=300,
 )
 
 register(
-    id='BipedalWalkerHardcore-v2',
+    id='BipedalWalkerHardcore-v3',
     entry_point='gym.envs.box2d:BipedalWalkerHardcore',
     max_episode_steps=2000,
     reward_threshold=300,
--- a/gym/envs/box2d/bipedal_walker.py
+++ b/gym/envs/box2d/bipedal_walker.py
@@ -365,7 +365,7 @@ class BipedalWalker(gym.Env, EzPickle):
         class LidarCallback(Box2D.b2.rayCastCallback):
             def ReportFixture(self, fixture, point, normal, fraction):
                 if (fixture.filterData.categoryBits & 1) == 0:
-                    return 1
+                    return -1
                 self.p2 = point
                 self.fraction = fraction
                 return fraction
--- a/gym/envs/box2d/bipedal_walker.py
+++ b/gym/envs/box2d/bipedal_walker.py
@@ -368,7 +368,7 @@ class BipedalWalker(gym.Env, EzPickle):
                     return 1
                 self.p2 = point
                 self.fraction = fraction
-                return 0
+                return fraction
         self.lidar = [LidarCallback() for _ in range(10)]
 
         return self.step(np.array([0,0,0,0]))[0]


olegklimov:
Cool! 

But are you sure it was like this (furthest point) all the time? Yes, a lot of people adopted the code and it worked to them. Maybe it was a change in Box2d at some point? Maybe we need to detect the version or find the minimum?

Also why `return fraction`, does it have any significance apart from returning 0 or 1?

chozabu:
It looks like it has always been that way - the docs page linked above was last changed in 2015 - before bipedalwalkers first commit in this repo.

Looking through a few videos of good solutions frame-by-frame they seem to have the same issue: https://github.com/1110589721/BipedalWalkerHardcore-v2 is a fantastic solution - but near the start of the attached gif you can see it trace through the floor, and hit the "outside edge" of a pit. going through frame-by-frame in GIMP also makes it possible to capture this image: https://i.imgur.com/tSnC62L.png

Looking closely at the gif provided by the writeup from the top of the leaderboard:
https://mp.weixin.qq.com/s/8vgLGcpsWkF89ma7T2twRA we can see this issue still happens: https://gyazo.com/ea32df04ce9866753c1614b981c4d892.
Based on automatic translation - it sounds like they clip the minimum reward to 0, thinking the "done" signal is strong enough, but perhaps more importantly it helps prevent the agent from freaking out when it looks like there is an unusually large pit just ahead.

It seems to me that solutions to BipedalWalkerHardcore-v2 have not just learned to deal with the complex environment - but advanced a step ahead, and are able to deal with the complex environment _and sensory hallucinations_ causing them to jump at the slightest hint of a cube, and keep running even when it looks like the ground is not visible below their feet, relying more on the touch sensor than the lidar, or perhaps recognising the difference in "shape" between a real pit and a "fake pit" (A real pit has a floor)

as for the significance of returning "fraction" from: https://github.com/behdad/box2d/blob/a5d2713a5b05f132a7858e58c14b338bb5c31eb3/Box2D/Testbed/Tests/RayCast.h#L54

it is mentioned:

> 		// By returning the current fraction, we instruct the calling code to clip the ray and
> 		// continue the ray-cast to the next fixture. WARNING: do not assume that fixtures
> 		// are reported in order. However, by clipping, we can always get the closest fixture.

Personally, I've been working with bipedal walker on and off for a couple of months and having a great time - Even though I thought the lidar looked like it may have been incorrect when flashing past a couple of times, this thought/feeling was not strong enough for me to investigate.
I only became seriously aware of the issue when adding a network to predict the next state, and as part of that made the lidar rendering "always on".

olegklimov:
OK, you've convinced me!

Let's run this by @pzhokhov and see what he says. Should we merge this and bump the env version?

chozabu:
Glad this has not been merged yet - I've spotted (and fixed) another issue in this area was still happening in my local copy: https://i.imgur.com/EcAqs5b.png

After triple checking the docs - I've just submitted another minor tweak (returning `-1` instead of `1` for an object that should be ignored) - it seems without this, a trace intersecting with the walkers legs would be incorrect in some situations (made easier to spot with feet and extra traces).

chozabu:
@pzhokhov @olegklimov versions bumped

PR #1791, opened 2020-01-22
Add usage for MultiBinary space

priyankasrs:
All spaces come with a definition and an example usage except for Multibinary(). Adding usage provides uniformity in documentation and ease of understanding.

### DIFF
--- a/gym/spaces/multi_binary.py
+++ b/gym/spaces/multi_binary.py
@@ -3,6 +3,17 @@ from .space import Space
 
 
 class MultiBinary(Space):
+    '''
+    An n-dimensional binary space. 
+
+    The argument to MultiBinary defines n.
+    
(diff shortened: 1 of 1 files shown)


gym/spaces/multi_binary.py
```diff
@@ -3,6 +3,17 @@
...
 class MultiBinary(Space):
+    '''
+    An n-dimensional binary space. 
+
+    The argument to MultiBinary defines n.
+    
+    Example Usage:
+    
+    self.observation_space = spaces.MultiBinary(5)
+
```
pzhokhov: looks good, but could you also add an example of what would self.observation_space.sample() return?

priyankasrs: @pzhokhov made the changes.

PR #1796, opened 2020-02-01
AtariPreprocessing: return 3-D array when grayscale_obs=True

brett-daley:
The AtariPreprocessing wrapper returns an array of shape `(84, 84, 3)` for RGB observations but only `(84, 84)` for grayscale observations. This means users have to conditionally add a channel axis to observations (according to the color scale) when feeding them into a convolutional network. This is inconvenient and affects all users who want to use this wrapper for deep RL. This PR changes the returned shape for grayscale observations to `(84, 84, 1)` to make the dimensions consistent in all cases.


justinkterry:
@pzhokhov can this be merged?

brett-daley:
@pzhokhov I agree that it's a good idea to make the current implementation the default. I'll work on an update but might be a bit delayed due to the impending NeurIPS deadline.

brett-daley:
@pzhokhov I added a kwarg `grayscale_newaxis=False` that allows users to add a 3rd axis to grayscale observations if desired, but retains the original 2-D behavior by default. The unit test has also been updated accordingly.

brett-daley:
@pzhokhov thanks! Good catch regarding the observation space. It's fixed now, and I also updated the tests to check the shape in the future.

justinkterry:
@brett-daley You might want to take a look at this: https://github.com/PettingZoo-Team/SuperSuit

It's a library of Gym wrappers for things like this that's arguably much nicer than those included by default in Gym.

brett-daley:
@justinkterry Thanks, I'll check it out! At this point, I'm just trying to close this PR for the community, as I haven't been doing much with Atari games recently -- but it's good to know for the future.

PR #1799, opened 2020-02-03
made fetchenv sample goal flexible

YangRui2015:
I'm doing curriculum learning on FetchEnvs but found that the target range of FetchEnvs is set fixed here:
`goal = self.initial_gripper_xpos[:3] + self.np_random.uniform(-0.15, 0.15, size=3)`

It's better set to `goal = self.initial_gripper_xpos[:3] + self.np_random.uniform(-self.target_range, self.target_range, size=3)`, thus we can change the range of sample goals by initializing the environment. After updating this, the environment can be more compatible.

And I have also checked that HandEnvs don't have such problem.


bycn:
Might want to clip target_range so that the goal is reachable

YangRui2015:
> Might want to clip target_range so that the goal is reachable

Yes, so I think this should not be set fixed.

PR #1815, opened 2020-02-19
Update README.md

dA505819:
Added italics to some important words in sentence to enhance readability(I missed them but those words shouldn't be missed).
Added comma wherever necessary.


dA505819:
Thankyou!

PR #1816, opened 2020-02-19
Pick up a newer version of cloudpickle.

tvalentyn:
Early versions of cloudpickle have an inconvenient side-effect that makes `cloudpickle<1.3.0`, and libraries that depend on these versions, hard to use together with libraries that use different picklers.

See: https://github.com/cloudpipe/cloudpickle/issues/82

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ setup(name='gym',
                 if package.startswith('gym')],
       zip_safe=False,
       install_requires=[
-          'scipy', 'numpy>=1.10.4', 'six', 'pyglet>=1.4.0,<=1.5.0', 'cloudpickle~=1.2.0',
+          'scipy', 'numpy>=1.10.4', 'six', 'pyglet>=1.4.0,<=1.5.0', 'cloudpickle~=1.3.0',
           'enum34~=1.1.6;python_version<"3.4"',
       ],
       extras_require=extras,


tvalentyn:
Tests are failing with the same error on https://github.com/openai/gym/pull/1815, so that looks unrelated.

tvalentyn:
retest this please

tvalentyn:
@christopherhesse @pzhokhov is there a way to re-run the CI?

pzhokhov:
> @christopherhesse @pzhokhov is there a way to re-run the CI?
Yes, but simply re-running it on the same commit will likely lead to the same failure. I think a better idea is to merge the latest upstream master, and then push the result. It will trigger travis as well

tvalentyn:
I see, will rebase. I wasn't sure how your precommits are configured - in Apache Beam we rebase PRs on top of master before CI is running the tests.

tvalentyn:
When do we expect the next release of gym?
Thanks!

pzhokhov:
Hey @tvalentyn! Could you elaborate a little on the use case in which version 1.3 is required? Specifically, we are thinking about relaxing the requirement to be `>=1.2.0, <1.4.0` instead (which means the latest 1.3 version will be installed if no other version is found). Does that break your usecase?

tvalentyn:
That should not break my usecase, but I thought you had some reason to fix a narrow range of cloudpickle. In theory, a problem may arise if a payload is pickled and unpickled using different versions of cloudpickle. I've seen this problem with dill pickler (https://github.com/uqfoundation/dill/issues/341). If usage of your library does not involve transmitting pickled payloads across machines, you might not need to worry about about setting a narrow range.

pzhokhov:
https://github.com/openai/gym/pull/1836

PR #1817, opened 2020-02-21
Video recorder: distinguish input and output fps

eleurent:
Common video players such as VLC, or Quicktime struggle to play videos with low framerate, which is problematic for environments where the policy frequency is low (e.g. 1 Hz).
See: https://superuser.com/a/601916

This commit sets:
- The input image sequence framerate to the metadata 'video.frames_per_second'
- The output video framerate to at least 30 fps

### DIFF
--- a/gym/wrappers/monitoring/video_recorder.py
+++ b/gym/wrappers/monitoring/video_recorder.py
@@ -270,18 +270,19 @@ class ImageEncoder(object):
                      '-nostats',
                      '-loglevel', 'error', # suppress warnings
                      '-y',
-                     '-r', '%d' % self.frames_per_sec,
 
                      # input
                      '-f', 'rawvideo',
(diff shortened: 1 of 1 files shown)


pzhokhov:
this is not a bad idea, but I think the min output framerate needs to be configurable, and by default be equal to the self.frames_per_sec , for two reasons: 
- if the player supports low fps, this change may blow up video file sizes unnecessarily (although maybe video compression will take care of most of that)
- strobe issues if the environment runs at, say, 29 Hz, and the video is saved at 30 Hz (this is, admittedly, a little contrived, and maybe ffmpeg is smart and takes care of that too).

eleurent:
Thanks for raising these points, I agree that having a configurable output framerate is a better and safer approach. I'll update the PR.

PR #1819, opened 2020-02-21
Drop python2 support

pzhokhov:


christopherhesse:
Also I'm a little worried that the 3.8 tests are segfaulting, can just the mujoco ones be disabled?

zuoxingdong:
Should we remove importing that involves `from __future__ import *`?

pzhokhov:
@christopherhesse 
> I noticed you didn't add 3.8 to the classifiers: 'Programming Language :: Python :: 3.7', (not sure that does anything)
 - that's because the tests for 3.8 are segfaulting :) I think it only adds a category in pypi, nothing else. I'll disable mujoco and add 3.8 tests

pzhokhov:
@christopherhesse I think removing six should be its own, bigger PR, because it is used in a bunch of places. Created an issue (https://github.com/openai/gym/issues/1823) for that.

pzhokhov:
@zuoxingdong 
> Should we remove importing that involves from __future__ import *?
yeah, but let's do that in a separate PR (maybe the same one that removes six - I added removal of these imports to #1823 issue as well)

setup.py
```diff
@@ -45,11 +45,10 @@
       tests_require=['pytest', 'mock'],
       python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*',
```
tvalentyn: I think you'd want to update `python_requires` as well, otherwise as soon as you add Python code that is not Py2-compatible, you will break Python 2 users who don't set an upper bound on your dependency (`gym>=0.10`), or expect compatibility w.r.t. semantic versioning (`gym>=0.15, <1.0` for example)  - note that this is a minor version change.

pzhokhov: good point, thanks!

pzhokhov: https://github.com/openai/gym/pull/1827

tvalentyn: Hey @cclauss , I've seen your drive-by comments on quite a few projects re: Python 3 migration. FYI - this is a common error when dropping Py2. Another project I saw this: https://github.com/hamcrest/PyHamcrest/issues/131. Perhaps this is something to keep an eye on in the projects that you come across.

cclauss: I would use the [`python_requires=">=3.5"`](https://github.com/psf/black/blob/master/setup.py#L64) approach when dropping legacy Python.

tvalentyn: Yeah, fully agree; I meant that more projects should be following this pattern, and I thought perhaps you are in the position to raise awareness of the proper way to drop Py2 support in other projects.

PR #1829, opened 2020-02-29
Update the docstring in envs/toy_text/blackjack.py

sorryformyself:
In "Example 5.1 in Reinforcement Learning: An Introduction by Sutton and Barto", there says

`One of the dealer’s cards is face
up and the other is face down`

which doesn't even mention the player. 

In today's blackjack games, the player have two face up card in virtually all of those games. So it's better to make it clear here to avoid confusion. That's why I change a little bit docstring here.

### DIFF
--- a/gym/envs/toy_text/blackjack.py
+++ b/gym/envs/toy_text/blackjack.py
@@ -49,7 +49,7 @@ class BlackjackEnv(gym.Env):
     Aces can either count as 11 or 1, and it's called 'usable' at 11.
     This game is placed with an infinite deck (or with replacement).
     The game starts with dealer having one face up and one face down card, while
-    player having two face up card. (Virtually for all Blackjack games today).
+    player having two face up cards. (Virtually for all Blackjack games today).
 
     The player can request additional cards (hit=1) until they decide to stop
     (stick=0) or exceed 21 (bust).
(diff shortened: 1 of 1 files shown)


gym/envs/toy_text/blackjack.py
```diff
@@ -48,8 +48,8 @@ class BlackjackEnv(gym.Env):
     Face cards (Jack, Queen, King) have point value 10.
     Aces can either count as 11 or 1, and it's called 'usable' at 11.
     This game is placed with an infinite deck (or with replacement).
-    The game starts with each (player and dealer) having one face up and one
-    face down card.
+    The game starts with dealer having one face up and one face down card, while
+    player having two face up card. (Virtually for all Blackjack games today).
```
pzhokhov: typo `card` -> `cards`?

pzhokhov:
Makes sense, dealer's strategy does not change based on player's cards, so player may as well have two open cards at the beginning. LGTM (other than the typo)

gym/envs/toy_text/blackjack.py
sorryformyself: yeah that's just a typo, sorry for my mistake.

sorryformyself: > typo `card` -> `cards`?

yeah that's just a typo, sorry for my mistake.

PR #1836, opened 2020-03-04
expand cloudpickle range

pzhokhov:


jachiam:
LGTM!

araffin:
@pzhokhov is there a reason to cap the version to 1.4.0?
Could we extend it to more recent versions?

pzhokhov:
I suspect the main reason was that 1.3 was the latest minor version at the time of this PR, and we did not want to leave upper version open - in case the cloudpickle API were to change in later releases.

pzhokhov:
it seems like all our tests are passing with the latest (1.6.0) cloudpickle, if that's a desired change, we can merge it into master

PR #1853, opened 2020-03-26
placed texture creation before parameter setting

Sh0cktr4p:
texture parameter changes only have an effect if texture was already created.
In current version, scaled-up images thus are interpolated linearly and wrap around edges.
=> Switched get_texture() and glTexParameteri() statements

### DIFF
--- a/gym/envs/classic_control/rendering.py
+++ b/gym/envs/classic_control/rendering.py
@@ -349,9 +349,9 @@ class SimpleImageViewer(object):
         assert len(arr.shape) == 3, "You passed in an image with the wrong number shape"
         image = pyglet.image.ImageData(arr.shape[1], arr.shape[0], 
             'RGB', arr.tobytes(), pitch=arr.shape[1]*-3)
+        texture = image.get_texture()
         gl.glTexParameteri(gl.GL_TEXTURE_2D, 
             gl.GL_TEXTURE_MAG_FILTER, gl.GL_NEAREST)
-        texture = image.get_texture()
         texture.width = self.width
         texture.height = self.height
         self.window.clear()


pzhokhov:
Thanks @Sh0cktr4p! Does this mean 
`gl.glTexParameteri(gl.GL_TEXTURE_2D,  gl.GL_TEXTURE_MAG_FILTER, gl.GL_NEAREST)`
line is not actually doing anything? 
Could you provide a snippet that demonstrates that?

Sh0cktr4p:
In the current state, the width and height of the created window is determined by the dimensions of the supplied rgb data array. The only case the image would be magnified in the current state is when the window is manually resized while open.
However, in this case a linear interpolation is performed with interpolated values wrapping around the image borders. This is not an intended behavior, as the line `gl.glTexParameteri(gl.GL_TEXTURE_2D, gl.GL_TEXTURE_MAG_FILTER, gl.GL_NEAREST)`
explicitly states that in this case nearest neighbor interpolation should be used instead. Acquiring the texture with `texture = image.get_texture()` before setting the texture parameters fixes this issue.

In the supplied script, a checkerboard image is being rendered. When the image is resized manually, the described issue (linear interpolation, edge wrapping) becomes apparent.

[demo.zip](https://github.com/openai/gym/files/4466265/demo.zip)

pzhokhov:
demo segfaults both in master branch and with the proposed changes :/ 
```
  File "_ctypes/callbacks.c", line 232, in 'calling callback function'
  File "/Users/peterz/anaconda3/envs/gym/lib/python3.7/site-packages/pyglet/libs/darwin/cocoapy/runtime.py", line 1121, in objc_method
    result = f(py_self, *args)
  File "/Users/peterz/anaconda3/envs/gym/lib/python3.7/site-packages/pyglet/window/cocoa/pyglet_window.py", line 64, in nextEventMatchingMask_untilDate_inMode_dequeue_
    if self.inLiveResize():
  File "/Users/peterz/anaconda3/envs/gym/lib/python3.7/site-packages/pyglet/libs/darwin/cocoapy/runtime.py", line 785, in __call__
    return self.method(self.objc_id, *args)
  File "/Users/peterz/anaconda3/envs/gym/lib/python3.7/site-packages/pyglet/libs/darwin/cocoapy/runtime.py", line 755, in __call__
    result = f(objc_id, self.selector, *args)
ctypes.ArgumentError: ("argument 1: <class 'RecursionError'>: maximum recursion depth exceeded in __instancecheck__", "selector = b'inLiveResize'", "argtypes =[<class 'ctypes.c_void_p'>, <class 'ctypes.c_void_p'>]", "encoding = b'c16@0:8'")
[1]    60462 segmentation fault  python ~/Downloads/demo.py
```

pzhokhov:
(at the attempt to resize the image manually, that is)

pzhokhov:
also, I tried a modified version of your snippet - that changes the viewer width: 

```python
import numpy as np
import time


def create_checkerboard(size):
    arr = np.arange(size*size * 4).reshape(size * 2, size * 2, 1)
    return np.tile(np.where((arr % (2 * size) < size) ^ (arr // (2 * size) < size), 255, 0).astype(np.uint8), (4, 4, 3)) 

dim = 8 

v = rendering.SimpleImageViewer()

while True:
    v.imshow(create_checkerboard(dim))
    time.sleep(0.2)
    v.width += 1
```
results seem to be the same in master and in your branch...
I have little expertise with opengl, but maybe @christopherhesse has an opinion on this?

crypt096 (approved):

jkterry1:
Reviewer: @kcorder

jkterry1:
Closed by #2312

PR #1867, opened 2020-04-07
Clean up continuous mountain car environment

instance01:
Improve readability and mostly adhere to PEP-8.
I left the one-line ifs (l.77-81) as is.


pzhokhov:
we should probably standardize (and check in CI) this across the entire gym at some point.

PR #1877, opened 2020-04-11
Clean up pendulum environment.

instance01:
Mostly PEP-8, left the equations as is.
I removed the pylint comment. I don't think E1111 applies here anyways since np.clip returns something.


instance01:
@pzhokhov
Whenever I am stuck on something (on other projects) and bored I just clean up something in this project. I hope this is fine (in terms of more PRs coming up), else please let me know.
You had the idea to set up project-wide linting (#1867). Maybe at some point I will do that and create a PR. But for now, for me it is easier to just clean random stuff.

PR #1884, opened 2020-04-15
Improve observation space of FlattenObservation wrapper

troiganto:
The wrapper `gym.wrappers.FlattenObservation` turns observations of any types (integers, tuples, dicts, etc.) into flat, 1D NumPy arrays. This obviously also changes the observation space: it becomes a `gym.spaces.Box` whose boundaries also are 1D arrays.

One weakness of this is that the `Box`'s boundaries are set generically to `-inf` and `+inf`, ignoring the original boundaries.

This PR fixes this by adding a function `flatten_space()` to `gym.spaces.utils`, where `flatdim()` and `flatten()` already live. The new code follows the same logic, but also takes the original boundaries into account.

While I was there, I also added some docstrings and unit tests.

Please let me know if there is anything you'd like me to change and thank you for maintaining this project!

### DIFF
--- a/gym/spaces/__init__.py
+++ b/gym/spaces/__init__.py
@@ -7,7 +7,8 @@ from gym.spaces.tuple import Tuple
 from gym.spaces.dict import Dict
 
 from gym.spaces.utils import flatdim
+from gym.spaces.utils import flatten_space
 from gym.spaces.utils import flatten
 from gym.spaces.utils import unflatten
 
-__all__ = ["Space", "Box", "Discrete", "MultiDiscrete", "MultiBinary", "Tuple", "Dict", "flatdim", "flatten", "unflatten"]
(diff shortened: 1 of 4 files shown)


troiganto:
The test failure in 0565123 actually unearthed a bug: `gym.spaces.utils.unflatten()` returns a regular dict for `Dict` spaces, but `Dict` spaces consistently use `OrderedDict` instead.

I've made a commit to use an `OrderedDict` instead and also verify item order in the unit test. Let me know if that's alright.

pzhokhov:
Thanks for the PR @troiganto! The logic looks good, but I think the docstring for flatten_space would be clearer with some examples. Also, do you think it would be useful to include some equalities into the docstring as well? Like `flatten(space.sample) in flatten_space(space)`?

troiganto:
Sorry for the radio silence these past two weeks, I've been a bit wrapped up with work! Thank you for the review and suggestions.

I've added three examples to the docstring and added the invariant `flatten(sample) in flatten_space(space)` to all of them. I've also run doctest locally to ensure that the code is actually correct.

I've also taken the opportunity to rebase the PR to the latest master, in case this is desirable.

PR #1889, opened 2020-04-22
fix "Could not create GL context"

shashikg:
This PR may refer to this issue:  #1620 

I found the same issue while working with multiple displays. I found that probably by default pyglet assumes the same GL context for each of the displays while creating the window unless you specifically mention it. The below colab notebook demonstrates the issue:
[https://colab.research.google.com/drive/1-3VuCaWNbFagYQvUOPHNEJMo9ErRxR_0](https://colab.research.google.com/drive/1-3VuCaWNbFagYQvUOPHNEJMo9ErRxR_0)

### DIFF
--- a/gym/envs/classic_control/rendering.py
+++ b/gym/envs/classic_control/rendering.py
@@ -43,7 +43,9 @@ def get_display(spec):
     Pyglet only supports multiple Displays on Linux.
     """
     if spec is None:
-        return None
+        return pyglet.canvas.get_display()
+        # returns already available pyglet_display,
+        # if there is no pyglet display available then it creates one
     elif isinstance(spec, str):
         return pyglet.canvas.Display(spec)
     else:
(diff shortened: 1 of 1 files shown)


christopherhesse:
Good catch, is there any reason to not create the context all the time?

shashikg:
I made some changes in `get_display()` to create the context all the time please check if that's okay?

So, I was not sure why the `get_display()` function was written such that it should return `None` in case of `None` spec. 
https://github.com/openai/gym/blob/2c53496e2c0aa4bc8abae5bd753a7bf30866651a/gym/envs/classic_control/rendering.py#L39-L46

For creating the right context, you need the display which will be used by `pyglet.window.Window`

christopherhesse:
Thanks!  Why was this changed for `Viewer` but not `SimpleImageViewer`?

shashikg:
Ohh I did not see if this is also used in any other module.. Should I create another pull request with the changes?

pzhokhov:
@shashikg yes please, I think that would be best

shashikg:
@pzhokhov https://github.com/openai/gym/pull/1916

PR #1892, opened 2020-04-24
added openmodelica-microgrid-gym

stheid:


christopherhesse:
I moved this to the bottom since we're just doing that for now until we have a better way to order environments.  Thanks for the PR!

christopherhesse:
Github got real confused, merged in https://github.com/openai/gym/pull/1895 instead

PR #1899, opened 2020-04-27
[IDEA] Adding agents and managers interfaces

pierluigi-failla:
The general rationale here is to have three main pieces: **environments** (which describe actions, spaces, etc), **agents** (which implements the logic to pick an action) and **managers** (which takes care of running the agent/s in a given enviroment).

I tried to keep the change as simple as possible and this is not intended to be a final version at least at the current moment. 

_Short summary_:

1. introduce module 'agents'. The idea is to give a standard interface for agents. 'RandomAgent' already present in the examples has been rewritten with this pattern in mind.
2. introduce module 'managers' which take care of running a given agent on a given env. Take a look to 'examples/example##'.
3. add a lazy implementation of epsilon greedy agent.
4. add a tictactoe env and a OneVsOne manager, so that it is possible to have 2 agents running on the same env (alternating each other). take a look to 'examples/example03'


pzhokhov:
@pierluigi-failla thanks for the PR! I think unifying agents and the running agents in the environments goes outside of the scope of gym.  To my experience, unifying agents and runners is a whole other can of worms - I don't think there exists a single design that is useful yet compatible with everyone's approaches to distributed/parallelized rollouts and optimization (running environments in parallel, making batches for optimizer) etc.
That being said, the choices made by this PR look very reasonable, and I'd be happy to reference this in our readme if you choose to make it a separate project.

PR #1906, opened 2020-05-08
Rewrite LazyFrames.__getitem__ to only decompress if needed

KristianHolsheimer:
This is a relatively small PR. It adds the `LazyFrames.last_frame` property for efficient extraction of the last frame contained in a `LazyFrames` frame stack.

This is particularly useful when you want to insert frames into an experience-replay buffer without duplication.

A counter-argument against adding this property might be to say that this functionality is already covered by `__getitem__`, e.g. `lazy_frames[-1]`. The added benefit of using `last_frame` instead is that it skips the `__array__` call.

Oh and I also added some lines of documentation for `lz4_compress`.

### DIFF
--- a/gym/wrappers/frame_stack.py
+++ b/gym/wrappers/frame_stack.py
@@ -19,45 +19,39 @@ class LazyFrames(object):
         lz4_compress (bool): use lz4 to compress the frames internally
 
     """
+    __slots__ = ('frame_shape', 'dtype', 'shape', 'lz4_compress', '_frames')
+
     def __init__(self, frames, lz4_compress=False):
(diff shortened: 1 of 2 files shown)


pzhokhov:
Should we rather re-implement `__getitem__` method to only decompress one frame instead of all?

KristianHolsheimer:
Sorry, I missed the notification on your comment. Yes, you're right. I'll get on that once I have a spare couple of minutes.

KristianHolsheimer:
@pzhokhov I rewrote `LazyFrames.__getitem__` and rebased on the current `master`.

PR #1913, opened 2020-05-13
Lazily seed np rng

jjh42:
In some cases where env are create and torn down rapidly, np seeding takes a significant fraction of the time. This makes seeding only take place when the RNG is actually used.


instance01:
That's a nice change. But check that unit tests work.
Running `pytest` is your key to success here.

PR #1914, opened 2020-05-13
Pole balancing documentation of observation matches code

BSVogler:
The returned angle is returned in the unit of radians and not degrees.
observation number 3 is the angular velocity


christopherhesse:
Thanks for the PR!  I would merge this but it has conflicts.

PR #1916, opened 2020-05-15
pyglet GL context for SimpleImageViewer case

shashikg:
@pzhokhov @christopherhesse

### DIFF
--- a/gym/envs/classic_control/rendering.py
+++ b/gym/envs/classic_control/rendering.py
@@ -59,7 +59,10 @@ def get_window(width, height, display):
     config = screen[0].get_best_config() #selecting the first screen
     context = config.create_context(None) #create GL context
 
-    return pyglet.window.Window(width=width, height=height, display=display, config=config, context=context)
+    if viewer_type == 'Viewer':
+        return pyglet.window.Window(width=width, height=height, display=display, config=config, context=context)
+    elif viewer_type == 'SimpleImageViewer':
+        return pyglet.window.Window(width=width, height=height, display=display, config=config, context=context, vsync=False, resizable=True)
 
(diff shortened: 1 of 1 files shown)


gym/envs/classic_control/rendering.py
```diff
@@ -51,23 +51,26 @@ def get_display(spec):
...
     """
     screen = display.get_screens() #available screens
     config = screen[0].get_best_config() #selecting the first screen
     context = config.create_context(None) #create GL context
 
-    return pyglet.window.Window(width=width, height=height, display=display, config=config, context=context)
+    if viewer_type == 'Viewer':
+        return pyglet.window.Window(width=width, height=height, display=display, config=config, context=context)
+    elif viewer_type == 'SimpleImageViewer':
+        return pyglet.window.Window(width=width, height=height, display=display, config=config, context=context, vsync=False, resizable=True)
```
pzhokhov, line 65: how about window_kwargs argument rather, and call `pyglet.window.Window(width=width, height=height, display=display, config=config, context=context, **window_kwargs)` here?

pzhokhov: OTOH I wonder if there is a strong reason for vsync=False and resizabble=True in SimpleImageViewer in the first place

shashikg: I am not sure but maybe because we are fixing the window size, so we set the vsyns=False.

shashikg: > how about window_kwargs argument rather, and call `pyglet.window.Window(width=width, height=height, display=display, config=config, context=context, **window_kwargs)` here?

done!

PR #1919, opened 2020-05-17
Mountaincar env reward structure

dveb8886:
I implemented this exact reward structure into my agent training code, and was able to successfully train my agent to the point where it scales the mountain successfully every run. the original -1 reward was not very helpful

I decided to share this reward structure since it makes more sense, and a lack of positive reward is a common complaint about this environment

### DIFF
--- a/gym/envs/classic_control/mountain_car.py
+++ b/gym/envs/classic_control/mountain_car.py
@@ -113,8 +113,8 @@ class MountainCarEnv(gym.Env):
 
     def reset(self):
         # for reward, changing slightly so max_pos starts at this random value
-        self.max_pos = self.np_random.uniform(low=-0.6, high=-0.4)
-        self.state = np.array([self.max_pos, 0])
+        self.max_position = self.np_random.uniform(low=-0.6, high=-0.4)
+        self.state = np.array([self.max_position, 0])
         return np.array(self.state)
 
     def _height(self, xs):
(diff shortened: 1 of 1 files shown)


dveb8886:
Just as a note from my runs: I test my agents by running against the env 30 times, and then taking the average of the total reward (all rewards throughout the entire run added up) .

I achieved:
- 26.83 for a run that failed to reach the flag once or twice during the 30 runs
- 22.13 for a run that can get to the flag every single time

Use this as a baseline if you decide to run your own tests to compare. Note my numbers don't include the reward of 10 at the very end (as i added that as an afterthought)

Revanthmk:
doesn't this change the objective of the game. I don't think there should be any punishment for getting away from the end point. Also change the description above if you are going in with this reward system.

dveb8886:
> I don't think there should be any punishment for getting away from the end point

Are you referring to the -1 you get when you are not making progress? (There is no explicit punishment for the car moving in the opposite direction, so I assume you are referring to this)

Assuming so, I saw no reason to change that since the -1 should continue to emulate frustration at a lack of progress. I want the agent to get punished for taking too long

dveb8886:
As for objective, the objective still remains to reach the flag. You get a higher overall score if you do it in the most efficient manner possible

Revanthmk:
Yeah exactly. Previously the objective of the agent was just to reach the flag with least number of steps. Now the agent is punished for moving away from the flag, which restricts the movement of the agent in a different way than intended.

BSVogler:
Mountaincar is a standard control problem. You cannot change the reward function because it better suits your algorithm. This changes the problem definition and is against the principles of the openai gym.

Please consider closing this PR.

pzhokhov:
Roughly seconding what @BSVogler said - this will be a different environment then, and not the one described in the Sutton and Barto's textbook (and would make comparisons of algorithms with published results tricky). I would rather recommend making it a separate package.

PR #1924, opened 2020-05-20
Fix: FrozenLake-v0 | is_valid function | condition

likefrankie:
The following conditional statement appears incorrect:
line 58: if (res[r_new][c_new] not in '#H'):

I expected 'H' not '#H'. My recommendation is:

line 58: if (res[r_new][c_new] != 'H'):


likefrankie:
Thank you! I am glad that I could be helpful. On Jun 19, 2020 5:25 PM, pzhokhov <notifications@github.com> wrote:
@pzhokhov approved this pull request.

Thanks @likefrankie! Good catch, that's an artifact of how the map used to be generated (it was using # as a filler symbol). Your version is perfectly valid now.

PR #1928, opened 2020-05-23
Set default image color to white

pedrofreire:
In gym.envs.classic_control.rendering, Geom has a color attribute, [which is initialized as black](https://github.com/openai/gym/blob/2ec4881c22b129d1f06173d136529477c0d8d975/gym/envs/classic_control/rendering.py#L175). This is a good default for most Geom subclasses, such as lines and circles. However, in the case of Image instances, this causes images to be rendered as black, which seems undesirable; in particular, if the image does not have transparency, the rendered result is a black rectangle.

This has been shown to cause confusion amongst users:
https://stackoverflow.com/questions/59410495/python-pyglet-rgb-image-is-just-black
https://stackoverflow.com/questions/61024893/how-can-i-load-my-own-picture-in-openai-gym

This can be easily fixed by setting the default color for Image instances as white, which causes images to be displayed normally.

Example code and before/after results:

```python

viewer = rendering.Viewer(800, 600)
viewer.set_bounds(-110, 110, -110, 110)
image = rendering.Image('soccer-field.jpg', 200, 200)
viewer.add_geom(image)
viewer.render()

```

Before:
soccer-field-black-square

After:
soccer-field-visible

### DIFF
--- a/gym/envs/classic_control/rendering.py
+++ b/gym/envs/classic_control/rendering.py
@@ -321,6 +321,7 @@ class Line(Geom):
 class Image(Geom):
     def __init__(self, fname, width, height):
         Geom.__init__(self)
+        self.set_color(1.0, 1.0, 1.0)
         self.width = width
         self.height = height
         img = pyglet.image.load(fname)


christopherhesse:
Thanks for the PR!  Does this modify the behavior of any of the gym environments?

pedrofreire:
I don't think so, at least not in master. The only current usage of rendering.Image seems to be on [Pendulum](https://github.com/openai/gym/blob/9e9bb2830fac7845da9c30f892b98d4ee4825400/gym/envs/classic_control/pendulum.py#L84), which uses it for drawing a [rotation arrow](https://github.com/openai/gym/blob/9e9bb2830fac7845da9c30f892b98d4ee4825400/gym/envs/classic_control/assets/clockwise.png), which is a black drawing and thus is unaffected by this change.

PR #1938, opened 2020-06-03
Fix linux dependencies in Readme.rst

floringogianu:
Due to bad formatting the `apt-install ...` command is not readable on github.
Also a couple of dependencies for `atari-py` and `box2d-py` are missing: `cmake zlib1g zlib1g-dev swig`.


christopherhesse:
Whoops, thanks for the PR!

PR #1940, opened 2020-06-04
Rescale observation wrapper

### LINKED ISSUE #1939: Feature request: observation space rescale wrapper
It would be great to have a `RescaleObservation`-wrapper similar to the existing [RescaleAction](https://github.com/openai/gym/blob/d8908cbf101c253ea9b4fb0858f1e90d899b6596/gym/wrappers/rescale_action.py#L7)-wrapper.

The observation spaces can generally be unbounded, so it should verify that the `observation_space.{low,high}` are finite and throw an error if that's not the case.

hartikainen:
Closes #1939

### DIFF
--- a/gym/wrappers/rescale_observation.py
+++ b/gym/wrappers/rescale_observation.py
@@ -11,21 +11,45 @@ def rescale_values(values, old_low, old_high, new_low, new_high):
     return rescaled_values
 
 
+def verify_observation_space_type(observation_space):
+    if not isinstance(observation_space, spaces.Box):
+        raise TypeError("Expected Box observation space. Got: {}"
+                        "".format(type(observation_space)))
+
+
+def verify_observation_space_bounds(observation_space):
+    if np.any(~np.isfinite((
+            observation_space.low, observation_space.high))):
+        raise ValueError(
+            "Observation space 'low' and 'high' need to be finite."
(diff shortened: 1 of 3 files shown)


gym/wrappers/rescale_observation.py
```diff
@@ -0,0 +1,138 @@
...
+    return observation_space
+
+
+class RescaleObservation(gym.ObservationWrapper):
+    def __init__(self, env, low, high):
+        r"""Rescale observation space to a range [`low`, `high`].
+
+        For `Box` spaces, `low` and `high` can be either scalar or vector, and
+        will be broadcasted according to numpy broadcasting rules. For `Tuple`
+        and `Dict` spaces, both `low` and `high` are expected to be scalar.
```
pzhokhov, line 48: could you specify in the docstring that elements of Tuple or Dict space are expected to be Box?

pzhokhov:
Thanks for the PR @hartikainen! Looks good, but I would add 
1) an explanation about are allowed types of sub-spaces in case of Dict or Tuple space
2) (if you think that could be useful) recursive rescaling logic to allow elements of Dict or Tuple to be Dict or Tuple

jkterry1:
I'm closing this per #2259

Issue #1856, opened 2020-03-28
blackjack.py dealer busting not checked ?


fusiller:
Hi there,
from a newbie carefully reading the code to learn how to use your great environment.
It seems to me that the dealer can go bust without the player getting his reward for winning.
line 105
There is no is_bust() test and the cmp() function will return player loosing to a busted dealer.
In the comment though : " If the dealer goes bust the player wins."
Regards
Alexandre

rhalbersma:
There is an `is_bust` check inside `score()` on line 35, which returns 0 for a busted hand. So a player who didn’t bust himself, will always win against a busted dealer.

christopherhesse:
Are you sure there's no `is_bust` test?  `cmp` compares the output of `score` which seems to have that test: https://github.com/openai/gym/blob/master/gym/envs/toy_text/blackjack.py#L35

christopherhesse:
Whoops should have hit refresh, thanks for answering @rhalbersma!

PR #1952, opened 2020-06-11
BlackJack-v0: Allow dealer to lose by going over...

hughdavenport:
Currently the dealer can go over 21 and still win. This patch fixes this by checking whether the dealer is bust after drawing


christopherhesse:
This other issue seems to indicate that this doesn't happen in practice: https://github.com/openai/gym/issues/1856. Are you certain this PR is necessary?

cpnota:
This seems to be handled [here]( https://github.com/openai/gym/blob/8d404e810eb2d4c5bdd208d6023e3e4b0b02203a/gym/envs/toy_text/blackjack.py#L35). I think this can be closed.

PR #1964, opened 2020-06-18
Update hotter_colder.py

turksoyomer:
I'm proposing to add "done=True" if our agent reaches to the correct answer.

### DIFF
--- a/gym/envs/toy_text/hotter_colder.py
+++ b/gym/envs/toy_text/hotter_colder.py
@@ -50,6 +50,7 @@ class HotterColder(gym.Env):
 
         elif action == self.number:
             self.observation = 2
+            done = True
 
         elif action > self.number:
             self.observation = 3


pzhokhov:
Thanks for the idea @turksoyomer! I think only changing done condition (without changing the reward function) may not work particularly well here. Imagine a very smart agent, that somehow knows the correct number and is trying to maximize the episode return. The strategy in the current version would be select that number and keep selecting it till the end of the episode, getting to episode return of 1 * length_of_the_episode. However, in with the changes you are proposing this behavior would only get return of 1 (as the episode would end immediately). Rather, the agent would be forced to guess the number that is closest to the correct, but not correct, until the end of the episode, and only guess correctly at the very last step. 
We could rewrite both done and reward, for instance, only giving the agent reward 1 if it correctly guesses the number, and then terminating the episode (similar to procgen's https://github.com/openai/procgen coinrun). But that may make early learning very hard - as for unskilled agent reaching the reward randomly would be unlikely. 
If I may ask, what is the purpose of this proposed change?

turksoyomer:
Thank you for your comment @pzhokhov. I didn't think about the results of the policy of the agent. What you said makes sense to me now. With my approach, the agent can learn to do different tricks to get the maximum reward.
My proposition was about ending the game when the agent reaches to the number. But with our reward strategy, it's not logical.
I'm closing this branch now. Thank you again.

PR #1974, opened 2020-06-23
include low, high, & dtype in spaces.Box.__repr__

bionicles:
multitask learners need a way to make a sensor for each space, and it's possible for two different Box spaces to have the same shape but different low/high/dtype, so this pull request just adds the minimum low value, maximum high value, and the dtype to `Box.__repr__` so we can use `str(box_space)` as a key in a ModuleDict of sensors (helps normalize inputs)


instance01:
Seems like the Python 3.5 build is failing since f-strings were only introduced in Python 3.6.

PR #1986, opened 2020-07-03
Allow HotterColder and GuessingGame to work with Box class.

TristanBester:
1) The 'contains' method (line 125) of the gym/spaces/box class requires a Numpy array/list to be passed into the 'x' parameter. If the user passes the action as a list both the GuessingGame (line 63) and HotterColder (line 48) environments can not perform the required comparisons. By casting the action to a list the issue is resolved. 
2) The reward value in the HotterColder class (line 62) is a float and thus not subscriptable, subscript was removed.

### DIFF
--- a/gym/envs/toy_text/guessing_game.py
+++ b/gym/envs/toy_text/guessing_game.py
@@ -58,7 +58,7 @@ class GuessingGame(gym.Env):
         return [seed]
 
     def step(self, action):
-        assert self.action_space.contains(action)
+        assert self.action_space.contains([action])
 
         if action < self.number:
             self.observation = 1
@@ -87,3 +87,4 @@ class GuessingGame(gym.Env):
         self.guess_count = 0
         self.observation = 0
         return self.observation
+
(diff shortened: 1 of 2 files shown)


pzhokhov:
Thanks for the PR @TristanBester! 
- contains issue - you are right, currently putting in lists as an action breaks the environment; but I would argue that your proposed change is not quite the way to go, because it breaks the frequently used stanza of sort `env.step(env.action_space.sample())` (and this is what failing travis tests are indicating). Maybe a better way to go would be to check if the action is a list, and if so, convert it to numpy array? Note that with single-value numpy arrays work fine in the comparisons. 
- indexing the reward - if the action is a single-valued numpy array at the line 57 in hotter_colder.py, the line 62 would work fine too :)

pzhokhov:
Another / additional possibility is to check if the action is a float (instead of a list of numpy array), and convert it to the latter.

TristanBester:
Thanks so much for the feedback I really appreciate it, this is my first pull request so I'm still quite new to this.

pzhokhov:
moving the discussion to this PR: https://github.com/openai/gym/pull/1992

PR #1987, opened 2020-07-05
To make toy_text DiscreteEnv's reset() method and step(action) method return consistent type of a state.

jacklinquan:
To make toy_text DiscreteEnv's reset() method and step(action) method return consistent type of a state.
The reset() method of DiscreteEnv returns a state of type numpy.int64, while step(action) returns a state of type int. This inconsistency could cause some problem for some agents.
In reset() method, force the returned state as type int.


pzhokhov:
thanks for the PR @jacklinquan ! It seems that the problem is not quite in the fact that reset returns numpy.int64, but rather that it does not match the type of the states in the transition table. In other words, it is possible that after this change there would be an opposite inconsistency for some environments. Could you provide more details here about the issue you are seeing?

jacklinquan:
Hi @pzhokhov ,
Thanks for the comments!
Recently I played with a basic Q Learning Agent(not DQN, just a sort of tabular one) which inserts state-action Q value dynamically. So initially it doesn't need to know the real size of the Q-table. Before it inserts the state it checks if the state has "flatten" method(that means it is a numpy object and can be flattened) and flattens it then turns it into tuple. In the result Q-table, I found all the states are duplicated, because some of them are of type int, for example 5, while some of them are of type tuple, for example (5,). The root cause is that the env's reset() method returns numpy object, but its step(action) method returns just int.
It is not a very big issue, the agent still works fine. But the complexity is doubled, and the training time needs to be prolonged.

pzhokhov:
Hi @jacklinquan! So are you using the DiscreteEnv directly? Could you post here the code that sets the env up? 
I double checked and all the environments that use it (Taxi, FrozenLake, CliffWalking) produce int observations both in `reset` and `step` methods.

jacklinquan:
> Hi @jacklinquan! So are you using the DiscreteEnv directly? Could you post here the code that sets the env up?
> I double checked and all the environments that use it (Taxi, FrozenLake, CliffWalking) produce int observations both in `reset` and `step` methods.

Hi @pzhokhov ,
I'm using a standard toy_text env that inherits DiscreteEnv, specifically "Taxi-v3". This issue is subtle because when you use print() to display a numpy number, it just shows you exactly the same as type int. You have to use type() to reveal its real type.
Please try this code:
``` Python
import gym

env = gym.make('Taxi-v3')
state = env.reset()
print(state) # 221, or any number between 0~500
print(type(state)) # <class 'numpy.int64'>

new_state, r, d, info = env.step(0)
print(new_state) # 321
print(type(new_state)) # <class 'int'>
```

pzhokhov:
very interesting! Ok, I agree, but let's also add an explicit cast in `step` method then.

pzhokhov:
(I'll add the change)

PR #1992, opened 2020-07-11
Toy text fix

TristanBester:
The 'contains' method (line 125) of the gym/spaces/box class requires a Numpy array/list to be passed into the 'x' parameter. If the user passes the action as a list both the GuessingGame (line 63) and HotterColder (line 48) environments can not perform the required comparisons. By casting the action to a Numpy array the issue is resolved.

### DIFF
--- a/gym/envs/toy_text/hotter_colder.py
+++ b/gym/envs/toy_text/hotter_colder.py
@@ -43,7 +43,16 @@ class HotterColder(gym.Env):
         return [seed]
 
     def step(self, action):
+<<<<<<< HEAD
+        if isinstance(action, (int, float)):
+            action = np.array([action])
+        elif isinstance(action, list):
+            action = np.array(action)
+
         assert self.action_space.contains(action)
+=======
+        assert self.action_space.contains([action])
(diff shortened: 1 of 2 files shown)


gym/envs/toy_text/hotter_colder.py
```diff
@@ -59,10 +64,11 @@ def step(self, action):
         self.guess_count += 1
         done = self.guess_count >= self.guess_max
 
-        return self.observation, reward[0], done, {"number": self.number, "guesses": self.guess_count}
+        return self.observation, reward, done, {"number": self.number, "guesses": self.guess_count}
```
pzhokhov, line 67: I think the change in this line is failing the tests. Specifically, it returns reward as an array now instead of a scalar, and that does not conform to the Env API

pzhokhov:
Thanks @TristanBester! I agree with this change, but please fix the test (I think it is related to `reward[0] -> reward` change).

justinkterry:
@pzhokhov this can be closed given that #2082 was merged

PR #2002, opened 2020-07-24
Populate TimeLimit.truncated on every step

ryanjulian:
Writing samplers which use the `info` return from `Env.step()` is much easier if the keys in `info` are stable.

This PR populates `info['TimeLimit.truncated']` on every step of training, not just on truncated steps. This allows samplers to test for timeouts without having to test for key existence first.

### DIFF
--- a/gym/wrappers/time_limit.py
+++ b/gym/wrappers/time_limit.py
@@ -15,8 +15,9 @@ class TimeLimit(gym.Wrapper):
         assert self._elapsed_steps is not None, "Cannot call env.step() before calling reset()"
         observation, reward, done, info = self.env.step(action)
         self._elapsed_steps += 1
-        if self._elapsed_steps >= self._max_episode_steps:
-            info['TimeLimit.truncated'] = not done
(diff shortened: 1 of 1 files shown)


pzhokhov:
I would argue that this can be fixed pretty easily in user's code via `info.get('TimeLimit.truncated', False)`. TimeLimit is the only wrapper currently included in the gym that populates info dict, so I cannot really make an argument along the lines of "this is not how things are done in the gym" :) ; but our thinking was that wrappers and envs only populate info dict with relevant info, not with all possible keys. This probably only becomes relevant when you have a lot of keys, and also need to serialize and send info dicts (for instance, in some distributed settings); but I would still argue that it is cleaner to not clutter info dicts with default values.

vwxyzjn:
Yeah I agree with @pzhokhov, we shouldn't enforce this. 

> This allows samplers to test for timeouts without having to test for key existence first.

To my untrained eyes this should not be a bottleneck to the sampler. @ryanjulian Are there other benefits of this PR?

PR #2009, opened 2020-08-01
Fix MountainCar reward

jb41:
Fixes bug with the wrong reward from MountainCar env.

### DIFF
--- a/gym/envs/classic_control/mountain_car.py
+++ b/gym/envs/classic_control/mountain_car.py
@@ -101,7 +101,11 @@ class MountainCarEnv(gym.Env):
         done = bool(
             position >= self.goal_position and velocity >= self.goal_velocity
         )
-        reward = -1.0
+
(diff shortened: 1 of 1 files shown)


SkyWriter:
Spent a few days figuring out why my A2C doesn't solve it only to find out that it actually gives no reward ever.

crypt096 (approved):

pzhokhov:
I don't think this change correct, actually - the MountainCar reward is supposed to be -1 for every step of the episode until termination - so that agents that solve the task faster get higher (less negative) reward. I can see an argument for a reward being rather `reward = -1.0 if not done else 0.0`, but that likely does not affect learning at all.

jkterry1:
So per my discussion with @cpnota I do not believe that this change should be made as the "suffering" ends when the environment is over. I'm going to close this and all duplicate PRs and add a comment on the code clarifying this.

PR #2020, opened 2020-08-11
Fix problem with variable space sizes/bounds in KellyCoinflipGeneralizedEnv

AlexKuhnle:
Clip/resample game parameters to be able to fix obs/action space sizes/bounds


AlexKuhnle:
Made it optional with default false, and added a comment line briefly describing what it's for.

pzhokhov (approved):

PR #2022, opened 2020-08-12
Update discrete.py

Freakwill:
Just correct an expression


pzhokhov:
what is the purpose of the FiniteSet class? It seems like it re-names the outcomes of Discrete.sample()? 
Also, we do not include probability distribution code into spaces classes themselves, so I don't think the `sample_with_proba` method should be a part of "Discrete" class

jkterry1:
After further review and discussing it with a few people, it's not clear to me that this is actually a good idea in its current implementation and the PR is stale, so i'm going to close this for now.

PR #2023, opened 2020-08-12
Add support for tuple input on MultiBinary space

lsylusiyao:
When I'm using [ray-project/ray](https://github.com/ray-project/ray), I've found out that when I'm using multi-dimensions space as the input n of `MultiBinary`, problem occurs on init part. After debugging, it appeared that the dealing way of n is not suitable for multi-dimensions input. 
Here's the issue in [ray-project/ray](https://github.com/ray-project/ray): [#10024 on ray](https://github.com/ray-project/ray/issues/10024).

### DIFF
--- a/gym/spaces/multi_binary.py
+++ b/gym/spaces/multi_binary.py
@@ -19,11 +19,13 @@ class MultiBinary(Space):
     '''
     
     def __init__(self, n):
-        if type(n) in [tuple, list] and len(n) != 1:
-            self.n = tuple(n)
+        self.n = n
+        if (type(n) in [tuple, list] and len(n) != 1) \
+            or (type(n) is np.ndarray and len(n.shape) != 1):
+            input_n = n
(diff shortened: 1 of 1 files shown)


gym/spaces/multi_binary.py
```diff
@@ -32,7 +37,7 @@ def contains(self, x):
 
     def to_jsonable(self, sample_n):
         return np.array(sample_n).tolist()
-
+
```
pzhokhov, line 40: please remove the spaces

lsylusiyao:
@pzhokhov Thank you for all your suggestions. 
I've added `length != 1` because I'd like to deal with input like this: `self.obs = MultiBinary([3])`. However, now I believe that this is not essential since the docstrings give out the usage. 
Also I've finish the shape check of `contains`.

pzhokhov (approved):

Issue #1972, opened 2020-06-21
vector.make crashes with NotImplementedError in envs with custom spaces, and Box crashes with None in shape


bionicles:
To use RL for datasets, I made a String space, and unfortunately, this crashes when you attempt to make vector env with your custom env with custom spaces:

```
  File "/home/bion/miniconda3/lib/python3.7/site-packages/gym/vector/__init__.py", line 61, in make
    return AsyncVectorEnv(env_fns) if asynchronous else SyncVectorEnv(env_fns)
  File "/home/bion/miniconda3/lib/python3.7/site-packages/gym/vector/async_vector_env.py", line 88, in __init__
    observation_space=observation_space, action_space=action_space)
  File "/home/bion/miniconda3/lib/python3.7/site-packages/gym/vector/vector_env.py", line 34, in __init__
    self.observation_space = batch_space(observation_space, n=num_envs)
  File "/home/bion/miniconda3/lib/python3.7/site-packages/gym/vector/utils/spaces.py", line 41, in batch_space
    return batch_space_dict(space, n=n)
  File "/home/bion/miniconda3/lib/python3.7/site-packages/gym/vector/utils/spaces.py", line 70, in batch_space_dict
    for (key, subspace) in space.spaces.items()]))
  File "/home/bion/miniconda3/lib/python3.7/site-packages/gym/vector/utils/spaces.py", line 70, in <listcomp>
    for (key, subspace) in space.spaces.items()]))
  File "/home/bion/miniconda3/lib/python3.7/site-packages/gym/vector/utils/spaces.py", line 43, in batch_space
    raise NotImplementedError(
```

The issue is, it's not clear how to make a String space out of the regular spaces in gym, because strings are variable length; we could pad them, but padding is wasteful

Can we add a simple escape hatch to just return a tuple of any-type stuff if the space isn't in the list of supported batch spaces?

Here's the line: https://github.com/openai/gym/blob/809a8ad21ba66b2a17987005a04be3881808bf37/gym/vector/utils/spaces.py#L43

Side note: I'm pretty sure we can drop Binary, MultiBinary, MultiDiscrete, and maybe Discrete, if you guys want to reduce the size of the codebase for maintenance simplification, everything is ultimately just a Box, and Box can model every single one of those other spaces. 

[...]

PR #2038, opened 2020-08-31
Allow custom spaces in VectorEnv

### LINKED ISSUE #1945: custom class or a string as env.observation_space in custom gym environment
can I have a custom class or a string as env.observation_space in my custom environment? right now, it throws an error at env.reset(). 

If I can use a string, how do I define the env.observation_space?

even better, if I can use my custom class, how do I define env.observation_space? 

I think there is no way around without defining it because env.reset() and gym.vector.make() are throwing errors ?

tristandeleu:
- Adds support for custom spaces in `AsyncVectorEnv` and `SyncVectorEnv`. This will not batch the observations, and will return a tuple of individual observations.
 - Fixes #1945 & #1972

### DIFF
--- a/gym/error.py
+++ b/gym/error.py
@@ -165,3 +165,10 @@ class ClosedEnvironmentError(Exception):
     Trying to call `reset`, or `step`, while the environment is closed.
     """
     pass
+
+class CustomSpaceError(Exception):
+    """
+    The space is a custom gym.Space instance, and is not supported by
+    `AsyncVectorEnv` with `shared_memory=True`.
+    """
+    pass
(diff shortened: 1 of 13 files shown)


pzhokhov (approved):
I like this (even though I don't fully agree with the motivation behind strings as custom spaces)! The only thing I would request is adding a comment (maybe in spaces/space.py) along the lines of: 
`Space class can be subclassed to create a custom observation / action space, however, keep in mind that most cases should be covered by the existing space classes (Box, Discrete, ...) and container space classes (Tuple, Dict). Moreover, parametrized probability distributions and batching functions are only well-defined for the existing spaces, so some learning algorithms may not work out of the box.  TLDR think twice before subclassing`

pzhokhov (approved):

saikrishna-1996 (commented):
looks good to me, thanks!

PR #2044, opened 2020-09-05
Add gym-derk to environments

FredrikNoren:
Hi,

I'm the author of [Derk's Gym](https://gym.derkgame.com), which I implemented to be as compatible as possible with OpenAI gym. It's a multi-agent environment though, and as far as I can tell there's no clear consensus on how to handle that. I tried to model the API after the multiagent-particles-env.

Would also love to get feedback on the environment itself!


pzhokhov:
Looks super fun! Yeah, you are right, Gym API is mostly tuned for single player; and I don't think there is a perfect consensus on how to handle the multi-agent environments... One easy hack for multiplayer environments is vectorization - assuming observation and action spaces are the same, all you have to do is dedicate different indices of environments to different agents.
This repo: https://github.com/openai/multi-agent-emergence-environments takes a somewhat different approach, and joins all observations spaces and all action spaces in a tuple. There is a lot of surrounding code to deal with it too.

FredrikNoren:
@pzhokhov Thank you! Yeah for performance reasons I ended up handling observations and actions as single numpy arrays (I guess that's what you mean with vectorization).

PR #2070, opened 2020-10-16
Flattened space/point dtype mismatch

### LINKED ISSUE #2069: flattened point dtype does not match flattened space dtype
Originally found this by using the FlattenObservationWrapper and seeing that my flattened space yielded different precision points than the original space. This is a bug that actually originates from the `spaces.utils.flatten_space` function and is caused by some missing dtype arguments. 

Simplified examples of the bug:
```
from gym.spaces import utils
from gym.spaces import MultiDiscrete, MultiBinary

original_space = MultiDiscrete([2, 2, 100])
flattened_space = utils.flatten_space(original_space)

original_sample = original_space.sample()
flattened_sample = utils.flatten(original_space, original_sample)

assert flattened_space.dtype == flattened_sample.dtype
```

wmmc88:
closes #2069 

I've also added a pytest that demonstrates the error and than my changes fix the error.

### DIFF
--- a/gym/spaces/utils.py
+++ b/gym/spaces/utils.py
@@ -161,12 +161,12 @@ def flatten_space(space):
         return Box(low=0,
                    high=1,
                    shape=(space.n, ),
-                   dtype=np.int8
+                   dtype=space.dtype
                    )
     if isinstance(space, MultiDiscrete):
         return Box(
             low=np.zeros_like(space.nvec),
             high=space.nvec,
-            dtype=np.int64
+            dtype=space.dtype
         )
     raise NotImplementedError
--- a/gym/spaces/utils.py
+++ b/gym/spaces/utils.py
@@ -46,7 +46,7 @@ def flatten(space, x):
         return np.asarray(x, dtype=space.dtype).flatten()
     elif isinstance(space, Discrete):
         onehot = np.zeros(space.n, dtype=space.dtype)
-        onehot[x] = 1.0
+        onehot[x] = 1
         return onehot
     elif isinstance(space, Tuple):
         return np.concatenate(
@@ -73,7 +73,7 @@ def unflatten(space, x):
     defined in ``gym.spaces``.
     """
     if isinstance(space, Box):
-        return np.asarray(x, dtype=np.float32).reshape(space.shape)
+        return np.asarray(x, dtype=space.dtype).reshape(space.shape)
     elif isinstance(space, Discrete):
         return int(np.nonzero(x)[0][0])
     elif isinstance(space, Tuple):
@@ -94,9 +94,9 @@ def unflatten(space, x):
         ]
         return OrderedDict(list_unflattened)
     elif isinstance(space, MultiBinary):
-        return np.asarray(x).reshape(space.shape)
+        return np.asarray(x, dtype=space.dtype).reshape(space.shape)
     elif isinstance(space, MultiDiscrete):
-        return np.asarray(x).reshape(space.shape)
+        return np.asarray(x, dtype=space.dtype).reshape(space.shape)
     else:
         raise NotImplementedError
 
--- a/gym/spaces/utils.py
+++ b/gym/spaces/utils.py
@@ -156,10 +156,15 @@ def flatten_space(space):
             high=np.concatenate([s.high for s in space]),
         )
     if isinstance(space, MultiBinary):
-        return Box(low=0, high=1, shape=(space.n, ))
+        return Box(low=0,
+                   high=1,
+                   shape=(space.n, ),
+                   dtype=np.int8
+                   )
     if isinstance(space, MultiDiscrete):
         return Box(
             low=np.zeros_like(space.nvec),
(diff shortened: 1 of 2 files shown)


gym/spaces/utils.py
```diff
@@ -43,9 +43,9 @@ def flatten(space, x):
     ``gym.spaces``.
     """
     if isinstance(space, Box):
-        return np.asarray(x, dtype=np.float32).flatten()
+        return np.asarray(x, dtype=space.dtype).flatten()
```
pzhokhov, line 46: using space.dtype  here instead of np.int64 seems inconsistent with the clause for MultiDiscrete and MultiBinary

pzhokhov:
changes look good, thanks for adding a test the demonstrates the error! 
I have two minor nitpicks - 1) using space.dtype in some clauses and explicit dtypes in the others - I think we should be consistent
and 2) for Discrete, we are actually changing the dtype (is used to be inconsistent, np.int64 in the Discrete space definition and np.float32 in the flatten function). Given that np.float32 is more float-compatible, I'd rather use np.float32 - maybe change the dtype in discrete.py then?

gym/spaces/utils.py
wmmc88: I originally did this because the underlying dtype of Box can be defined by the user, but the types of discrete, multi_discrete and multi_binary are hardcoded to certain values.

wmmc88: I will change them all to be space.dtype for consistency. this will also accommodate for if anyone chooses to change the underlying types of multi_discrete, discrete,  multi_binary (which i found has actually been done a couple times before).

wmmc88:
> Given that np.float32 is more float-compatible, I'd rather use np.float32 - maybe change the dtype in discrete.py then?

I don't think this is the right approach to store discrete information as floats. In this pr, I've already fixed the discrepency in flatten vs space definition dtype, and its intentional that the dtype doesnt change when flattening. It will only change if there's a conflict in dtypes (ex a dict with a box(default float32) and a discrete(always int64)) and in this case, flatten and flatten space will automatically promote the dtype to the smallest type that fits both. This behaviour is intended and is actually already handled internally by the np.concatenate function.

wmmc88:
I noticed some more space discrepancies in the unflatten function. I added tests to catch them and then made the relevant fixes. Now, you are able to flatten a sample and then unflatten it and the dtypes of the flattened and unflattened value should match. This is on top of the initial fix that ensures the the flattened space dtype and flattened value match (and that dtypes are only changed (ie promoted) if absolutely needed)

pzhokhov:
Nice, thanks! 
As for dtype for flattened Discrete samples being float or int - storing values as one-hot encoded is already sub-optimal, I don't think it matters much which data type is then used for one-hot values, so it makes sense to use the dtype that's most convenient to use. Typically, we'd use flatten and unflatten to deal with outputs (or inputs) of linear layers, and those would be floats - which is why I am arguing for using floats as dtype. 
OTOH, if you convince me that int dtype is more appropriate / conventient for some other use case, I'll merge as is :)

wmmc88:
> Nice, thanks!
> As for dtype for flattened Discrete samples being float or int - storing values as one-hot encoded is already sub-optimal, I don't think it matters much which data type is then used for one-hot values, so it makes sense to use the dtype that's most convenient to use. Typically, we'd use flatten and unflatten to deal with outputs (or inputs) of linear layers, and those would be floats - which is why I am arguing for using floats as dtype.
> OTOH, if you convince me that int dtype is more appropriate / conventient for some other use case, I'll merge as is :)

Sorry I'm not quite sure I understand what you want me to change exactly. Could you elaborate?

Are you suggesting i make changes to the dtype in `discrete.py`?

pzhokhov:
@wmmc88 
> Are you suggesting i make changes to the dtype in discrete.py (as you suggested earlier)?
Yes, or specify a use case in which discrete dtype benefits from being integer (rather than float). Thanks!

wmmc88:
> Are you suggesting i make changes to the dtype in discrete.py (as you suggested earlier)?
> Yes, or specify a use case in which discrete dtype benefits from being integer (rather than float). Thanks!

Well, changing the dtype in `discrete.py` would make it always `np. float32`, including when a discrete space is used on its own. The one-hot encoding inefficiency justification you provided only applies for when the space is flattened, not when its just used as a vanilla discrete space. The whole point of the discrete space is to store discrete values, so I would say it wouldn't make sense if it used a `np.float32` instead of a numpy integer type. If the user wanted to store a single floating type, they would have just used a box space with `shape=()`.

Further, both Discrete and MultiDiscrete Spaces assert that the space has an n>0 (of course). So I think the spaces would make most sense to be `np.uint64` instead of the `np.int64` they are right now. Similarly I think MultiBinary should be using `np.uint8` instead of `np.int8`

pzhokhov:
ok fine, be it your way @wmmc88  :) This probably still merits a version bump, I'll release a new version after this merge.

Issue #28, opened 2016-04-29
Fix all instances of tobytes vs tostring


gdb:
```
File "MsPacman.py", line 4, in <module>
env.render()
File "/Library/Python/2.7/site-packages/gym/core.py", line 146, in render
return self._render(mode=mode, close=close)
File "/Library/Python/2.7/site-packages/gym/envs/atari/atari_env.py", line 95, in _render
self.viewer.imshow(img)
File "/Library/Python/2.7/site-packages/gym/envs/classic_control/rendering.py", line 287, in imshow
image = pyglet.image.ImageData(self.width, self.height, 'RGB', arr.tobytes(), pitch=self.width * -3)
AttributeError: 'numpy.ndarray' object has no attribute 'tobytes'
```

PR #2084, opened 2020-11-04
Remove Python 3.5 support, travis and setup.py maintenance

justinkterry:
In this PR:

-Removed Python 3.5. It's no longer supported, included with any supported Linux distro, or supported by NumPy/SciPy/etc.
-Bumped Ubuntu version used for tests to current LTS
-Bumped Python subversions used for tests to current ones
-Added max version to Pillow. Pyglet depends on it for certain functions, and the range you've chosen isn't compatible with the newest version (8.0.0, released Oct 14)
-All tests run on all Python versions now (this seems to have been accidentally disabled in the Python 2.7 deprecation?)
-Added the minimum version of OpenCV needed to properly function to setup.py
-Removes an unused failover I found chasing down the pillow problem that was supposed to be removed with issue #28

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,13 @@
-dist: xenial
+dist: focal
 language: python
 python:
   - "3.7"
 services:
   - docker
 env:
-  - PY_VER=3.5.6
-  - PY_VER=3.6.8
-  - PY_VER=3.7.3
-  - PY_VER=3.8.1
-
+  - PY_VER=3.6.12
+  - PY_VER=3.7.9
+  - PY_VER=3.8.6
 
 install: "" # so travis doesn't do pip install requirements.txt
 script:
@@ -22,4 +20,4 @@ deploy:
     password: $TWINE_PASSWORD
     on:
         tags: true
-        condition: $PY_VER = 3.5.6
+        condition: $PY_VER = 3.6.12
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: xenial
+dist: focal
 language: python
 python:
   - "3.7"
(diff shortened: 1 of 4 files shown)


justinkterry:
I'll fix the tests in the morning

justinkterry:
Proper Python 3.8 testing was seemingly disabled in the docker file? I re-enabled it and am working through the errors:

-I removed a pillow failover that hasn't been relevant since 2011
-I'm added a minimum version of pillow of 6.2.1 (chosen because it was released about a year ago) since a large library should have such a thing, and would prevent that problem from happening again
-I added a minimum version to opencv while I was at it, since again that's a good thing to have. It's the minimum version that plays nice and has the expected features.

pzhokhov (approved):
LGTM, thanks! The story with the all tests is this - the atari and mujoco tests were failing on 3.8 for some reason... It seems like with your changes (updated ubuntu version + swig + cmake) they are passing again. Long term I'd prefer to add 3.8 and 3.9 wheels to the atari-py build, and remove swig and cmake from gym build, but I think the current version works for now.

justinkterry:
-Ubuntu didn't have anything to do with getting them to work
-I assumed you guys weren't going to touch atari-py since ALE-py should be coming out soon?

pzhokhov:
Hmm I think mujoco is still not playing nicely: https://travis-ci.org/github/openai/gym/jobs/742009047

pzhokhov:
atari-py vs ALE-py - good point, actually

justinkterry:
I'm a contributor to the ALE and one of two maintainers of a large project that depends heavily on Gym (https://github.com/PettingZoo-Team/PettingZoo) so I know some things, but MuJuCo's basically magic to me, especially with those errors. It also has seemed like you guys have completely stopped maintaining MuJuCo-Py and the related Gym environments?

Issue #2062, opened 2020-10-05
car_racing.py memory leaking issue


Jaekyung-Cho:
https://github.com/openai/gym/blob/38a1f630dc9815a567aaf299ae5844c8f8b9a6fa/gym/envs/box2d/car_racing.py#L527

There is memory leaking in car_racing.py code.

pyglet.graphics.vertex_list() function makes the memory almost double,

And it can be fixed with `vl.delete()` under this line
https://github.com/openai/gym/blob/38a1f630dc9815a567aaf299ae5844c8f8b9a6fa/gym/envs/box2d/car_racing.py#L530

praveenVnktsh:
Thanks a lot. This fixed the leak issue.

@Jaekyung-Cho I guess it would be great if you could submit a pull request for this.

JunkyByte:
Hello, I've experienced the same memory leak and applied the solution given by @Jaekyung-Cho.
While it reduces the problem (greatly) it does not eliminate it, I'm running out of ram pretty quickly using multiple CarRacing environments in parallel (on a 64gb ram machine).

Could you share your results for the following code? I see constant increase in ram usage.
(For fair comparison I'm sharing a single environment test code to reproduce so that there's no doubt that it is related to gym / CarRacing env).

```python
import gym
import psutil
import os

def memory_used():
    return psutil.Process(os.getpid()).memory_info().rss * 1e-6  # To megabyte

e = gym.make('CarRacing-v0')
e.unwrapped.verbose = 0

for i in range(10000000):
    e.reset()

    if i % 1000 == 0:
        print('Ram Used: %f' % memory_used())
```

which outputs (numbers are megabyte)

```
Ram Used: 153.264128
Ram Used: 172.523520
Ram Used: 181.149696
Ram Used: 181.149696
Ram Used: 197.632000
Ram Used: 197.632000
Ram Used: 197.632000
Ram Used: 202.498048
Ram Used: 204.931072
Ram Used: 226.336768
Ram Used: 229.040128
Ram Used: 231.743488
Ram Used: 234.176512
Ram Used: 236.609536
Ram Used: 239.312896
Ram Used: 241.745920
Ram Used: 244.449280
Ram Used: 280.604672
Ram Used: 283.037696
(...)
```

looking for a solution, would you be so kind to tell me if you are able to reproduce the issue?

Jaekyung-Cho:
> Hello, I've experienced the same memory leak and applied the solution given by @Jaekyung-Cho.
[...]

PR #2096, opened 2020-11-17
pull request about issue #2062 (car_racing.py memory leaking issue )

Jaekyung-Cho:
Pull request about issue #2062 

There is memory leaking issue caused by pyglet module and solved with simply delete after drawing vertexs 

I confirmed with e-mail from @praveenVnktsh

### DIFF
--- a/gym/envs/box2d/car_racing.py
+++ b/gym/envs/box2d/car_racing.py
@@ -590,6 +590,7 @@ class CarRacing(gym.Env, EzPickle):
             len(polygons) // 3, ("v3f", polygons), ("c4f", colors)  # gl.GL_QUADS,
         )
         vl.draw(gl.GL_QUADS)
+        vl.delete()
         self.score_label.text = "%04i" % self.reward
         self.score_label.draw()
 
--- a/gym/envs/box2d/car_racing.py
+++ b/gym/envs/box2d/car_racing.py
@@ -528,6 +528,7 @@ class CarRacing(gym.Env, EzPickle):
             len(polygons_) // 3, ("v3f", polygons_), ("c4f", colors)  # gl.GL_QUADS,
         )
         vl.draw(gl.GL_QUADS)
+        vl.delete()
 
     def render_indicators(self, W, H):
         s = W / 40.0


SwapnilPande (changes requested):
Hi @Jaekyung-Cho,

I've been debugging the same issue. Looks like we need to add `vl.delete()` after `vl.draw` in function `render_indicators` (line 592). It looks like the memory consumption is still growing after this change, but much much slower.

Jaekyung-Cho:
> Hi @Jaekyung-Cho,
> 
> I've been debugging the same issue. Looks like we need to add `vl.delete()` after `vl.draw` in function `render_indicators` (line 592). It looks like the memory consumption is still growing after this change, but much much slower.

Exactly. Memory leaking still remain, but I cannot find where it occur. Though slower leaking issues doesn't matter in my case.

SwapnilPande:
Looks like this is an issue with Box2D, I was able to isolate the memory leak down to creating the tiles in `_create_track`. Despite calling `DestroyBody` on all of the tiles in `_destroy`, Box2D does not properly destroy the tiles sometimes. This memory leak can be reproduced with this minimum code:

```
import Box2D
from Box2D.b2 import fixtureDef, polygonShape

import psutil
import os

def memory_used():
    return psutil.Process(os.getpid()).memory_info().rss * 1e-6  # To megabyte


world = Box2D.b2World((0,0))


road1_l = (0, 0)
road1_r = (0, 1)
road2_l = (1, 0)
road2_r = (1, 1)
vertices = [road1_l, road1_r, road2_r, road2_l]

fd_tile = fixtureDef(
                shape = polygonShape(vertices=
                    [(0, 0),(1, 0),(1, -1),(0, -1)]))

fd_tile.shape.vertices = vertices

for i in range(1000000):
    t = world.CreateStaticBody(fixtures=fd_tile)
    world.DestroyBody(t)

    if i % 10 == 0:
        print('Ram Used: %f' % memory_used())
```

Looks like it's time to submit an issue to pybox2d. :D

The slow leaking doesn't matter in my case either, though it'd be nice to have this issue resolved. Would you be able to add the second `vl.delete()` statement to make the PR complete for the gym codebase?

Jaekyung-Cho:
> Looks like this is an issue with Box2D, I was able to isolate the memory leak down to creating the tiles in `_create_track`. Despite calling `DestroyBody` on all of the tiles in `_destroy`, Box2D does not properly destroy the tiles sometimes. This memory leak can be reproduced with this minimum code:
> 
> ```
> import Box2D
> from Box2D.b2 import fixtureDef, polygonShape
> 
> import psutil
> import os
> 
> def memory_used():
>     return psutil.Process(os.getpid()).memory_info().rss * 1e-6  # To megabyte
> 
> 
> world = Box2D.b2World((0,0))
> 
> 
> road1_l = (0, 0)
> road1_r = (0, 1)
> road2_l = (1, 0)
> road2_r = (1, 1)
> vertices = [road1_l, road1_r, road2_r, road2_l]
> 
> fd_tile = fixtureDef(
>                 shape = polygonShape(vertices=
>                     [(0, 0),(1, 0),(1, -1),(0, -1)]))
> 
> fd_tile.shape.vertices = vertices
> 
> for i in range(1000000):
>     t = world.CreateStaticBody(fixtures=fd_tile)
>     world.DestroyBody(t)
> 
>     if i % 10 == 0:
>         print('Ram Used: %f' % memory_used())
> ```
> 
> Looks like it's time to submit an issue to pybox2d. :D
> 
> The slow leaking doesn't matter in my case either, though it'd be nice to have this issue resolved. Would you be able to add the second `vl.delete()` statement to make the PR complete for the gym codebase?

Oh! I think I misunderstand your mention. I'm so sorry about that. I add vl.delete() in render_indicators function and make PR now. Thank you for your help.

hirekk:
`vl.delete()` should also be added at the end of the `render_road` method (line 530), since the memory leak still occurs. As far as I can tell, this fixes the issue completely.

yoavalon:
This solution works well for me. Memory remains low,

SlipknotTN:
Me too

xeviknal:
Any updates in this one? Do you think this PR is any close to be merged?

xeviknal (approved):
I've been able to reproduce the issue. Once the PR is applied, the memory leak is gone or at least not exploding.

Jaekyung-Cho:
> I've been able to reproduce the issue. Once the PR is applied, the memory leak is gone or at least not exploding.

Thanks for reproducing instead of me!!

xeviknal:
@pzhokhov @zuoxingdong I see you've been contributing quite a lot last months. Do you know what we can do to move this forward? are you maintainers? I am guessing this issue is making people to start forking. I'd be pretty good for this community to approach this nasty bug. thank you

pzhokhov (approved):

PR #2104, opened 2020-11-20
Bugfix: Allow Nesting of Sync/Async VectorEnvs

lebrice:
Allows for the nesting of Async/Sync VectorEnvs

- 2-level nesting of any combination of SyncVectorEnv/AsyncVectorEnv  (i.e. Sync/Sync, Sync/Async, Async/Sync, Async/Async*)
 - Tested with the "CubeCrash-v0" and "CartPole-v0" environments

*: When nesting Async/Async environments, only the innermost env can have `daemon=True`, since daemonic processes cannot have children.

Somewhat related to #2072 (based on a suggestion from @tristandeleu ). A new Wrapper could also be eventually be introduced in another PR to "unchunk" the observations/actions/rewards/dones, (flatten the first two batch dimensions).

There currently aren't any tests specifically about the handling of Tuple/Dict/non-standard observation/action spaces. I could add some, if those cases aren't already covered by `test_[sync/async]_vector_env.py`.


jkterry1:
Reviewer: @vwxyzjn

vwxyzjn:
@lebrice thanks for contributing the PR. Would you mind fixing the linting error? Also, would you mind giving a short code sample to test it out? Maybe it's just me but I had a hard time understanding the `partial(...)` related code in `test_vector_env.py`.

jkterry1:
@lebrice could you please fix tests?

jkterry1:
@lebrice and fix the merge conflicts?

lebrice:
Yes, will do! Currently in the rush for ICLR.

pseudo-rnd-thoughts:
@lebrice Do you have any plans for this PR?

lebrice:
Hey @pseudo-rnd-thoughts!
Yeah I just rebased + pushed, so I consider this as ready for review again.
If you want to take your revenge and roast my PR, be my guest! :P

pseudo-rnd-thoughts:
Thanks @lebrice could you fix the lint issue. 
The PR looks fine, I can’t see any issues however could you explain an application and use case of this PR? 
Plus why should this be in gym and not in a separate project that provides this extension?

lebrice:
> Could you explain an application and use case of this PR?

Chunking, is the practice of having multiple sequential operations within each asynchronous process. This is useful in reducing the overhead of multiprocessing, and allowing users with a limited number of cores, to have a very large number of vectorized environments.

For example, in my experience, with only a small number of cpus, e.g. 4 cpus,  laptops start struggling really hard when using pure AsyncVectorEnvs with batch sizes >~ 32, even for simple envs like CartPole, since that involves creating 32 python processes.
With this fix, you can instead create just one worker process per CPU, i.e. an AsyncVectorEnv with num_envs = 4, where each worker is stepping through a SyncVectorEnv with a num_envs of 8.

> Plus why should this be in gym and not in a separate project that provides this extension?

Because this is more of a bugfix than an extension, and it's an improvement to the existing components in gym, not a new component.

Hope that clears it up, thanks for taking a look!

lebrice:
Hey @pseudo-rnd-thoughts @vwxyzjn there seems to be a bug in the test setup here: https://github.com/openai/gym/runs/6114261693?check_suite_focus=true#step:4:96

tests/vector/test_vector_env.py
```diff
@@ -58,3 +62,114 @@ def test_custom_space_vector_env():
...
+    assert env.observation_space.dtype == single_observation_space.dtype
+
+    from gym.vector.utils.spaces import iterate
+
+    def batch_size(space: Space) -> int:
+        return len(list(iterate(space, space.sample())))
+
+    assert batch_size(env.action_space) == n_outer_envs
+
+    with env:
```
pseudo-rnd-thoughts, line 149: Why are we using environment as a context manager here?

pseudo-rnd-thoughts:
@lebrice that is a strange bug as I can't replicate it locally, neither from terminal or docker environment. 
Have you merged with the master? This bug has not occurred to anyone else, could you make a minor change to rerun the CI and see if that effects it

tests/vector/test_vector_env.py
lebrice: So it gets closed and the resources are freed.
This isn't absolutely necessary, since the env going out of scope should have the same effect.

lebrice:
I believe you can re-run the workflows that failed from the Actions view @pseudo-rnd-thoughts . This would be better IMO than me making an empty commit.
If you can't see a "re-run failed workflows" button on the top right, then let me know and I'll push a little commit.

tests/vector/test_vector_env.py
```diff
@@ -1,7 +1,11 @@
+from functools import partial
+from typing import Callable, Type
+
 import numpy as np
 import pytest
 
-from gym.spaces import Tuple
+from gym import Space, spaces
```
pseudo-rnd-thoughts, line 7: Could you remove spaces as I dont think you should need to import the module and add to the following line

tests/vector/test_vector_env.py
pseudo-rnd-thoughts: Ok, that makes sense. My issue is that it is a code style that we do not follow anywhere else in the tests so could you just have a `env.close()` at the end of the script to make it simpler for anyone looking at the tests

tests/vector/test_vector_env.py
lebrice: fixed in 49ee20904ac3a4a1dba3020d1ebd11076848f376

lebrice:
Hey @pseudo-rnd-thoughts. Sure, I guess we could mention that this is now possible (or that the bug has been fixed, depending on your perspective)  in the "advanced" portion of the VectorEnv API documentation.

However, I think it's probably best that we hold off until #2072, so that we don't encourage users to use chunking manually, and instead properly document the new VectorEnv subclass in the docs.

Does that sounds reasonable?

pseudo-rnd-thoughts:
Hi @lebrice, that sounds reasonable to me. It is just at some point, we need to update the vector API for these changes

tests/vector/test_vector_env.py
lebrice: Fixed in https://github.com/openai/gym/pull/2104/commits/f47596fc5f28338b41fb5874836e3155bdbde497 .

However, just for the record, I disagree:
I think having to remember to close the env at the env is ugly, error-prone, and also unnecessary, since in all cases the env is closed when it gets out of scope. In my opinion, the only tests where an env should be closed explicitly are 1) the tests related to closing the envs, and 2) tests that require creating a temporary environment just to check spaces (as in the case here).

For example, consider this:
https://github.com/lebrice/gym/blob/49ee20904ac3a4a1dba3020d1ebd11076848f376/tests/vector/test_vector_env.py#L52-L54

lebrice:
Hey @pseudo-rnd-thoughts  can we run the workflows again?

lebrice:
No rush, thanks @pseudo-rnd-thoughts !

RedTachyon:
Can you explain exactly the interaction between this and #2072? As I understand, this PR would enable an underlying mechanism for the env batching, and then we would want the users to directly use the solution implemented in #2072?

gym/vector/async_vector_env.py
```diff
@@ -635,6 +635,16 @@ def _worker(index, env_fn, pipe, parent_pipe, shared_memory, error_queue):
     assert shared_memory is None
     env = env_fn()
     parent_pipe.close()
+
+    def step_fn(actions):
```
RedTachyon, line 639: Why are we defining this function? It doesn't seem to be called anywhere. If it's actually used, it'd need a type hint and some comments/docstring

gym/vector/async_vector_env.py
```diff
@@ -717,7 +737,10 @@ def _worker_shared_memory(index, env_fn, pipe, parent_pipe, shared_memory, error
                     pipe.send((None, True))
             elif command == "step":
                 observation, reward, done, info = env.step(data)
-                if done:
+                if isinstance(env, VectorEnv):
+                    # VectorEnvs take care of resetting the envs that are done.
+                    pass
+                elif done:
```
RedTachyon, line 743: Would it maybe be better to just do a single condition `if done and isinstance(env, Env)` or `if done and not isinstance(env, VectorEnv)`? (not sure which would be clearer, but the `pass` gave me a bit of a pause since it's rarely present in released code, and it felt more like something unfinished)

gym/vector/sync_vector_env.py
```diff
@@ -71,8 +69,11 @@ def __init__(self, env_fns, observation_space=None, action_space=None, copy=True
         self.observations = create_empty_array(
             self.single_observation_space, n=self.num_envs, fn=np.zeros
         )
-        self._rewards = np.zeros((self.num_envs,), dtype=np.float64)
-        self._dones = np.zeros((self.num_envs,), dtype=np.bool_)
+        shape = (self.num_envs,)
+        if isinstance(self.envs[0].unwrapped, VectorEnv):
```
RedTachyon, line 73: Is it ever possible that this condition would be different for `self.envs[0]` and `self.envs[1]`? Also, do we actually need to unwrap it? (I'm not sure off the top of my head how wrappers interact with VectorEnvs)

gym/vector/async_vector_env.py
```diff
@@ -717,7 +737,10 @@ def _worker_shared_memory(index, env_fn, pipe, parent_pipe, shared_memory, error
...
                     info["terminal_observation"] = observation
```
RedTachyon, line 744: Do we actually need this with the AutoReset wrapper introduced recently? And wouldn't it cause a redundant double reset in some cases?

tests/vector/test_vector_env.py
```diff
@@ -58,3 +62,113 @@ def test_custom_space_vector_env():
 
     assert isinstance(env.single_action_space, CustomSpace)
     assert isinstance(env.action_space, Tuple)
+
+
+@pytest.mark.parametrize("base_env", ["Pendulum-v1", "CartPole-v1"])
```
RedTachyon, line 67: Why parametrize only over the two envs instead of all of them?

gym/vector/sync_vector_env.py
lebrice: I dont see how this would only be true for some envs.. I mean, it's certainly outside the reasonable use case here.

Yeah unwrapping it is necessary, since most wrappers can work the same way for both envs and VectorEns.

gym/vector/async_vector_env.py
lebrice: Yeah this is still necessary, since the `if done` check doesn't work with VectorEnvs. This check is there so we don't reset the env when the episode is done (as is currently still done in the case of a single env here).
Not sure if/how the AutoReset wrapper relates to this.
Lmk if that wasn't clear.

tests/vector/test_vector_env.py
lebrice: You mean over all gym envs?
I mean, sure, I'm all for it, but the current vectorenv tests (i.e. the only test above) is only using CartPole-v1

gym/vector/async_vector_env.py
lebrice: Fixed in fcebd76b000980bac0d261b5bff1d6ecc48645ab

gym/vector/async_vector_env.py
lebrice: Fixed in 055087b6ae7711f9565f681b2f6befe563f6ba3e

gym/vector/async_vector_env.py
lebrice: Fixed in 055087b6ae7711f9565f681b2f6befe563f6ba3e

tests/vector/test_vector_env.py
lebrice: Added test parametrized with all classic_control + toy_text envs in 9c0e308cc03aff71ffca12105f9e9f5085451d97

gym/vector/async_vector_env.py
lebrice: Oops. Actually fixed in 15208e3c67c89df95792e2291bd9b62107e838a1

pseudo-rnd-thoughts:
@lebrice Could you fix the CI issues?

lebrice:
Hey there @pseudo-rnd-thoughts ! Yes, will do!

tests/vector/test_vector_env.py
lebrice: Hey @RedTachyon is this good? (testing with all envs where `should_skipp_env_spec_for_test = False`)?
As-is, the parametrization of this test generates 579 tests, which take about 1:20 to run on my end (with pytest xdist and 4 parallel workers). I think this will probably take something like 5 minutes to run on GitHub, depending on the machine's hardware.

(here's a link for that function btw: https://github.com/lebrice/gym/blob/e913bc81b83b43ae8ca9b3a02c981b74d31017ea/tests/envs/spec_list.py#L20)

lebrice:
Hey @pseudo-rnd-thoughts I think I fixed everything on my end, but I'd need your approval so the CI hooks can run.

pseudo-rnd-thoughts:
@lebrice There still seems to be an issue with the PR plus looking at the runtime, the CI has jumped from around 1 to 2 minutes to 10+ minutes. Could you investigate why your PR is causing the time to increase

lebrice:
> @lebrice There still seems to be an issue with the PR plus looking at the runtime, the CI has jumped from around 1 to 2 minutes to 10+ minutes. Could you investigate why your PR is causing the time to increase

Yes that's what I was mentioning here:

https://github.com/openai/gym/pull/2104#discussion_r876380518

I must have misinterpreted what he meant by "using all of them" (gym envs)? In any case, I'd be happy to turn it down, if that's not what he meant.

pseudo-rnd-thoughts:
Sorry, I hadn't seen that comment. Personally, I think that is overkill and we could test with a single env type, just to see if it works. I will message @RedTachyon about it

pseudo-rnd-thoughts:
@lebrice In the meantime, could you fix the CI so it works with the current tests as the number of environments can be modified easily

PR #2122, opened 2020-12-06
Bump pyglet version to support OpenGL in BigSur

liavkoren:
Apple went and did some stuff in BigSur to the way OpenGL is present in the filesystem which BigBroke pyglet which broke some gyms. (Specifically, `classic_control/rendering.py` was falling over when doing `from pyglet.gl import *`. See https://github.com/pyglet/pyglet/issues/274#issuecomment-691702291 and https://github.com/pyglet/pyglet/issues/274#issuecomment-691702291 for details. But the tl;dr is version bumping makes it happy.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ setup(name='gym',
                 if package.startswith('gym')],
       zip_safe=False,
       install_requires=[
-          'scipy', 'numpy>=1.10.4', 'pyglet>=1.4.0,<=1.5.0', 'Pillow<=7.2.0', 'cloudpickle>=1.2.0,<1.7.0',
+          'scipy', 'numpy>=1.10.4', 'pyglet>=1.4.0,<=1.5.11', 'Pillow<=7.2.0', 'cloudpickle>=1.2.0,<1.7.0',
       ],
       extras_require=extras,
       package_data={'gym': [


justinkterry:
What sort of testing did you conduct for this? I looked into the bounds of supported Pyglet support in the last PR that was merged to master, and it would surprise me if this functions without issue.

liavkoren:
@justinkterry What is the testing regime? I verified locally that this fixes the box2d envs for me, but haven't looked at the release notes for the newer versions of Pyglet. If bumping up to 1.5.11 is too aggressive, I could look at what the smallest incremental bump that fixes this issue is; beyond that, if you have any pointers about which parts of the code you're concerned about, I may have some time to look into further things or write some tests.

justinkterry:
Did you verify that everything still looks right locally? That's what I meant.

The other that I believe needs to be resolved is the pillow version. Ik pretty current version cap will have to be removed to properly function with the newest pyglet version, which will in return require increasing the minimum pyglet version.

If you can look into those two things then as far as I'm aware this should be good.

justinkterry:
I fixed the above comment so that its coherent english, sorry about that. If this actually does work with no changes to pyglet version I'd be happy to be proven wrong.

liavkoren:
> I fixed the above comment so that its coherent english, sorry about that. 

Hah, no worries. 🤣 

Cartpole seems to work fine; Lunar lander renders, but I haven't done any other fiddling with it, or checked the other box2d envs yet. I can take a look at the pillow version, altho if requirements haven't been touched in a few weeks, there may be a real quagmire waiting there -- pip just switched over to a new, stricter dependency resolver, which won't tolerate inconsistencies..

jkterry1:
Closed by #2135

PR #2123, opened 2020-12-09
fix reward in mountain_car

tocom242242:
I change reward in mountain car.
I think if code comments is correct, when the car reached top position, the agent should be got reward 0.

The code comments is as below
>         Reward of 0 is awarded if the agent reached the flag (position = 0.5)
>         on top of the mountain.
>         Reward of -1 is awarded if the position of the agent is less than 0.5.

I'm sorry if I'm wrong.

Thank you


zha:
Fix #2194

jkterry1:
Closed per #2009

PR #2125, opened 2020-12-11
Minor fix to ImageEncoder capture_frame

mattvlaw:
ImageEncoder was not writing frames in the capture_frame function.
Unindented the call to write a frame in line 303 of video_recorder.py so that it is reachable.


casio12r (approved):
I found the same issue too. Without this change, the saved mp4 videos are empty.

japonophile (approved):

mattvlaw:
Fixed in #2139

PR #2133, opened 2020-12-18
fix build - only install mujoco for python 3.6, 3.7,

pzhokhov:


justinkterry:
If I understand correctly this just disables mujuco tests on python 3.8 and 3.9? that seems like a less than ideal solution.

pzhokhov:
@justinkterry correct. FWIW to my knowledge mujoco never worked well with 3.8 and 3.9 - so I am down to re-enable them once we make the tests work (and not segfault)

christopherhesse (approved):
LGTM!  I'd double check the operator precedence on that bash command though, if it matters.

pzhokhov:
@christopherhesse thanks! Basically, I checked that 3.6 and 3.7 builds install mujoco, and 3.8 and 3.9 do not... Do you think there are any other gotchas?

justinkterry:
@pzhokhov My knowledge of mujuco is limited, but shouldnt the mujuco-py version just be bumped then?

pzhokhov:
no, I am pretty sure we tried it in the past (as in, mujoco 2.0) and it had more issues of its own. This is not to say it's not possible, but at the time we decided the investment in fixing all the breaks that mujoco 2.0 introduced was not worth it.

pzhokhov:
see https://github.com/openai/gym/pull/1731 for context

pzhokhov:
@justinkterry I scheduled a new pypi release, 0.18.0  - will be done at the end of this build: https://travis-ci.org/github/openai/gym/builds/750492489

PR #2135, opened 2020-12-21
Remove constraints on package versions

ZhiqingXiao:
**Issue:**
Current constraints on versions of pillow preventing Gym 0.18 from being installed in Windows Python 3.9.
Reason: In `setup.py`, Pillow's version <=7.2.0, but Pillow did not release Python 3.9 Windows binary for that version. Pillow released Python 3.9 Windows binary for the latest few versions.

**Why it happens:**
PR #2084 limited the version of Pillow to be <=7.2.0 in order to be compatible with the old version of pyglet. If it did not do so, codes would not be compiled. 

For example, a method in Pillow called `tostring()` has been deprecated in favor of the method `tobytes()`, but old pyglet did not change to adjust it. Now pyglet has updated it. (Cf. https://github.com/pyglet/pyglet/blob/master/pyglet/image/codecs/pil.py#L77)

**Solution:**
Remove the constraints on package versions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ setup(name='gym',
                 if package.startswith('gym')],
       zip_safe=False,
       install_requires=[
-          'scipy', 'numpy>=1.10.4', 'pyglet>=1.4.0,<=1.5.0', 'Pillow<=7.2.0', 'cloudpickle>=1.2.0,<1.7.0',
+          'scipy', 'numpy>=1.10.4', 'pyglet>=1.4.0', 'Pillow', 'cloudpickle>=1.2.0,<1.7.0',
       ],
       extras_require=extras,
       package_data={'gym': [


justinkterry:
The pillow cap is due to the pyglet version cap. The codebase probably shouldn't even pass tests with the newer version of pyglet from what I understand. What validation have you done?

ZhiqingXiao:
@justinkterry The old version of pyglet can not pass Travis CI tests, but the latest version does. Therefore, I remove the version cap for both pyglet and pillow. To my knowledge, this is the best way to make Gym available in Windows Python 3.9.
If you think the current version of Gym is not compatible with the latest version of pyglet, can you please share the reason? Thanks in advance!

Baekalfen:
Seems like it's also causing problems on Mac for the exact same reason.

tthoraldson:
I'm currently experiencing this issue as well. Any ETA on getting this merged?

monrax:
pip install gym throws an error because of this (I have Python 3.9 installed, with zlib 1.2.11):

`RuntimeWarning: Pillow 7.2.0 does not support Python 3.9 and does not provide prebuilt Windows binaries. We do not recommend building from source on Windows. `
`(...)`
`The headers or library files could not be found for zlib, a required dependency when compiling Pillow from source.`
`(...)`
`ERROR: Failed building wheel for Pillow`

Is there another way to install while we wait for this to be merged?

pytorchinfo:
@pzhokhov

mcginnie:
This appears to be an ongoing problem. Is there an eta for the merge?

PR #2139, opened 2020-12-25
Fix video recording.

mahiuchun:
Tested with `python examples/agents/random_agent.py`. The code before the change would create empty video files.

### DIFF
--- a/gym/wrappers/monitoring/video_recorder.py
+++ b/gym/wrappers/monitoring/video_recorder.py
@@ -300,7 +300,7 @@ class ImageEncoder(object):
         if frame.dtype != np.uint8:
             raise error.InvalidFrame("Your frame has data type {}, but we require uint8 (i.e. RGB values from 0-255).".format(frame.dtype))
 
-            self.proc.stdin.write(frame.tobytes())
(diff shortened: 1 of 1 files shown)


khazit:
Thank you ! Fixes it for me too, tested with this :
```python
import gym
env = gym.make('CartPole-v0')
env = gym.wrappers.Monitor(env, "recording", force=True)
env.reset()
while True:
    obs, rew, done, info = env.step(env.action_space.sample())
    if done:
        break
```

japonophile (approved):

feiden (approved):
I stumbled on this problem by chance, and almost submitted the exact same bug fix. Approve!

pzhokhov (approved):

CrazyNicolas:
I already fix the code but still have same problem(262bytes), can I ask why?

CrazyNicolas:
> I already fix the code but still have same problem(262bytes), can I ask why?

Oh! I see, if you change the code, you need reboot your PC to activate the change. I reboot and succeed!

manuel-delverme:
Why is this not released yet?

PR #2147, opened 2021-01-06
Video recorder records empty files because indentationError

acenturyandabit:
I noticed this while working on windows :3


acenturyandabit:
Never mind, it appears I wasn't the first one to find this out... #2139

PR #2149, opened 2021-01-07
Fix reward calculation for MountainCar-v0

zal-jlange:
Previously, the reward was fixed at -1, although described as being 0 if the position was >= 0.5.


jkterry1:
Closed per #2009

Issue #1925, opened 2020-05-21
Cannot record the video by Wrappers.Monitor


JingruiYu:
I am trying to record the performance by the wrappers.Monitor. However, only .json files were generated, not any video files. No error was reported. How can I get the video file? 

**system**
ubuntu 16, version 0.17.2,

**code**
`env = gym.make('CartPole-v0')
    env = gym.wrappers.Monitor(env, "recording",force=True)`

Thank you very much!

christopherhesse:
I ran the following script and it worked fine, are you actually stepping the environment?

```
import gym
env = gym.make('CartPole-v0')
env = gym.wrappers.Monitor(env, "recording",force=True)
env.reset()
while True:
    obs, rew, done, info = env.step(env.action_space.sample())
    if done:
        break
```

JianmingTONG:
Hi, I do get the video under "recording" directory. However, this video cannot be played successfully. Is there any solution to it?
I am using:
Ubuntu 18.04
gym 0.18.0

https://user-images.githubusercontent.com/30002985/103414749-3c252680-4bba-11eb-8ee1-c2cbded08d28.mp4

cxk1998:
I have the same problem. I get a video but it's only 1kb.

pavlexander:
same issue... been trying to investigate it with no success...

First enable logging:
```python
gym.logger.set_level(gym.logger.DEBUG)
```

Logs:

> D:\dev\Python_global_vEnv\Scripts\python.exe "C:\Program Files\JetBrains\PyCharm Community Edition 2020.2.3\plugins\python-ce\helpers\pydev\pydevd.py" --multiproc --qt-support=auto --client 127.0.0.1 --port 49874 --file D:/dev/PycharmProjects/LearningPy/all/gym_testing.py
> pydev debugger: process 13860 is connecting
> 
> Connected to pydev debugger (build 202.7660.27)
> INFO: Making new env: CartPole-v0
> INFO: Clearing 4 monitor files from previous run (because force=True was provided)
> INFO: Starting new video recorder writing to D:\dev\PycharmProjects\LearningPy\all\vid\openaigym.video.0.13860.video000000.mp4
> DEBUG: Capturing video frame: path=D:\dev\PycharmProjects\LearningPy\all\vid\openaigym.video.0.13860.video000000.mp4
[...]

PR #2151, opened 2021-01-08
Fix indent level in video recorder

mattmolinare:
Since it is declared after an exception is raised, the following statement on line 303 is never executed:

    self.proc.stdin.write(frame.tobytes())

The solution is to once unindent the statement. Refer to https://github.com/openai/gym/issues/1925#issuecomment-753465510


vakker:
Is there any progress on this PR? It's a rather trivial fix, it should be merged asap.

vakker:
@mattmolinare this was fixed in #2139 so I assume this PR is not relevant any more.

PR #2158, opened 2021-01-14
Mark abstract methods as abstract

### LINKED ISSUE #2136: Base Env class should be abstract
Hi, thank you for this lib,
The Env class should be abstract but it is not. Instead its methods have a `raise NotImplementedError` body.
There are advantages to making this abstract: for example PyLance can properly analyze gym code.

I think Env should subclass `ABC` and methods should be annotated as `@abstractmethod`.

jakkes:
Not marking abstract methods as abstract causes some code analysers (e.g. PyLance) to not identify the method properly. In this PR, I marked all abstract method of the base environment classes as abstract.

Fixes #2136


jkterry1:
Reviewer: @vwxyzjn

vwxyzjn:
This seems to be a pretty straightforward fix. @jkterry1 LGTM.

jkterry1:
@jakkes can you please make the linting tests pass before I merge this?

PR #2162, opened 2021-01-21
close env after done

### LINKED ISSUE #1978: close the env when finished
https://github.com/openai/gym/blob/345c65973fc7160d8be374745a60c36869d8accc/gym/envs/box2d/lunar_lander.py#L449

Shall we add `env.close()` before returning here?
I've seen error below if it's not closed.
`ImportError: sys.meta_path is None, Python is likely shutting down`.

miluChen:
Fixes #1978 
@christopherhesse


haluptzok:
Yes - looks like that will fix the same issue

PR #2164, opened 2021-01-25
Security Fix for Arbitrary Code Execution - huntr.dev

huntr-helper:
https://huntr.dev/users/B3EF has fixed the Arbitrary Code Execution vulnerability 🔨. Think you could fix a vulnerability like this?

Get involved at https://huntr.dev/

Q | A
Version Affected | ALL
Bug Fix | YES
Original Pull Request | https://github.com/418sec/gym/pull/1
Vulnerability README | https://github.com/418sec/huntr/blob/master/bounties/pip/gym/1/README.md

### 📊 Metadata *
Arbitrary Code Execution
### ⚙️ Description *
OpenAI Gym is a toolkit for developing and comparing reinforcement learning algorithms. This is the gym open-source library, which gives you access to a standardized set of environments.
### 💻 Technical Description *
untrusted loading of data by the pickle.load function leading to Arbitrary code execution.
### 🐛 Proof of Concept (PoC) *

Screenshot from 2021-01-07 18-26-03

### 🔥 Proof of Fix (PoF) *

Screenshot from 2021-01-07 18-29-46

### 👍 User Acceptance Testing (UAT)
The library has been run successfully after the fix being applied.


vwxyzjn:
Hello, thanks for submitting the fax. I'm having a hard time trying to understand the exploit. In your README.md file, I see the source code as follows

```
import os
import pickle
#os.system('pip3 install gym')
from gym.vector.utils import misc
#payload formation
class ArbitraryCode:
    def __reduce__(self):
        cmd = ('xcalc')
        return os.system, (cmd,)
dumps = pickle.dumps(ArbitraryCode())
misc.CloudpickleWrapper.__setstate__(dumps,dumps)
```

Does the exploit work by having the user importing some packages that have this code? If that's the case, couldn't the hacker just let the user run something like the following?

```
import subprocess
subpross.spawn("xcalc")
```

JamieSlome:
@B3EF ^

B3EF:
hi @vwxyzjn , this exploit code just shows how the particular function is exploitable due to unpickling of untrusted data.
its just show how the code is vulnerable , its just  a code review bug as i have just imported your package and tested it.

jkterry1:
Reviewer: @vwxyzjn

vwxyzjn:
@B3EF thanks for the response. Sorry if I am asking obvious questions, but could you tell me under what circumstances would this be a problem? As an end user, would there be situations where I do something like 

```
import gym
env = gym.make(“CartPole-V1”)
```

which the hacker would leverage to just executes arbitrary code somehow?

B3EF:
https://www.youtube.com/watch?v=jwzeJU_62IQ
 hope this helps , and i have just introduced a fix . rather i am not the one who found this bug .
thanks for your response. and first of all my fix isn't a perfect one as the maintainers have to verify which  all fns are allowed during deserialising of data at that point

jkterry1:
So I sat down and looked through this in detail myself and this does not fix any problems for Gym. Specifically, the unpickling logic at issue is only used to clone objects that users have deliberately installed and run. Gym essentially runs third party apps, without sandboxing, like a bash script would call a Python script or something. For this reason, there is no way to prevent arbitrary code execution; that's ultimately already happening.

B3EF:
Sorry For wasting your time on this, and thanks for the info.
<3

PR #2189, opened 2021-03-10
Update Pillow Python Library for CVE-2020-35653 & CVE-2021-27921

dciborow:


dciborow:
Curious to see if any builds pick up a break with this package update.

dciborow:
#2203

danijar:
Thanks! Why are you pinning Pillow to a specific version? Wouldn't it be better going forward to use Pillow without specifying the version number? Versions before 7.2 worked according to the current `setup.py` and versions after 7.2 also work according to Travis.

danijar:
https://github.com/openai/gym/pull/2212

dciborow:
#2199

dciborow:
okay i loosened the pin up.

dciborow:
@joschu how does this PR look?

dciborow:
@joschu thanks for the merge!

Any chance we can get a 0.18.1 release with this update? My Component Governance system is very angry at my current Pillow version...

joschu:
Not sure how to do this. @pzhokhov ?

pzhokhov:
> Not sure how to do this. @pzhokhov ?

typically, this would require simply making a git tag with new version and pushing it, CI would do the rest. However, the build is failing currently, I'll look into fixing it.

pzhokhov:
ok so the story is a bit tangled here - the build is broken because we are using relatively old patch versions of python (3.6.8 and 3.7.3); somehow, builds timeout when trying to pull those. When using newer patch versions that are actually listed on python dockerhub repo page, things seem to work, however, then mujoco segfaults.  We could move forward without mujoco tests (I made a PR: https://github.com/openai/gym/pull/2220/files), which will let us release newer versions. Otherwise, more investigation is needed into how to make mujoco work with newer versions of python. Thoughts @joschu @christopherhesse ?

pzhokhov:
Another approach could be to make our own public mirror for old python images, say, on ACR. OTOH, finicky properties of mujoco wrt python versions are a really sore spot, I'd rather we did not propagate that madness further.

pzhokhov:
For the reference, here's how a build that fails to pull python 3.6.8 image from dockerhub looks like: https://www.travis-ci.org/github/openai/gym/jobs/766940827

dciborow:
@pzhokhov 
Started getting a lot of safety warnings due to Pillow, and the old pin from gym.
If we can not get a new release up onto pypi, I am going to need to roll gym out of my team's project. Any active work or ETA on 18.0.1?

image

PR #2196, opened 2021-03-17
Monitor directory fix

### LINKED ISSUE #2195: Monitor wrapper used 2 different directory paths when it starts
When an environment is started with the `Monitor` wrapper in the `_start` method the directory is set with `directory = os.path.abspath(directory)`. In some cases this leads to 2 different directory paths, so `StatsRecorder` files end up in one and `VideoRecorder` files in the other.

Setting `self.directory = os.path.abspath(directory)` at the very beginning and using `self.directory` everywhere solves the problem.

vakker:
This fixes #2195


jkterry1:
Reviewer: @shreyansjainn

shreyansjainn:
@vakker can you please fix merge conflicts if you see this?

vakker:
@shreyansjainn it should be good now

shreyansjainn (approved):

PR #2212, opened 2021-04-05
Do not depend on a specific Pillow version.

danijar:
Several people have been unable to install Gym because it depends on Pillow 7.2 which does not support Python 3.9.

Python 3.9 is the standard on Ubuntu and Mac now.

The rationale behind removing the version number is that according to `setup.py` versions before 7.2 worked fine and according to Travis CI, current versions also work fine.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ setup(name='gym',
                 if package.startswith('gym')],
       zip_safe=False,
       install_requires=[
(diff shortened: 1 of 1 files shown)


danijar:
@pzhokhov @justinkterry Please take a look

jkterry1:
@danijar Given that the default max pillow version was raised to 8.2, this appears to no longer be relevant so I'm closing this. If problems still exist though, please let me know.

PR #2217, opened 2021-04-08
convert nchain into DiscreteEnv

truongsinh:
- inline doctest included
- bump version to NChain-v1, even though the construct API is exactly the same (but setter/getter is different due to iscreteEnv inheritance)


jkterry1:
@cpnota would you be willing to review this when you have a moment?

jkterry1:
@truongsinh Could you please merge master in and fix the conflicts?

jkterry1:
So I went ahead and went through this myself. This PR makes changes to files entirely beyond the scope for some reason and as best I can tell removes key functionality from the environment class (e.g. the init method) so I'm going to close this.

Issue #2203, opened 2021-03-22
Gym does not support Python 3.9 because of Pillow<=7.2.0


danijar:
Python 3.9 is supported by Pillow 8.0+ only ([source](https://github.com/python-pillow/Pillow/issues/5241#issuecomment-770408990)). Gym currently depends on Pillow<=7.2.0. This means `pip3 install gym` fails for newer Python versions. Depending on Pillow<=7.2.0 also means gym cannot be installed in parallel with projects that require more up-to-date versions ([source](https://github.com/openai/gym/issues/2199#issuecomment-803637005)).

<details>
<summary>Error message</summary>

```
pip3 install --user -U gym
Requirement already satisfied: gym in ./Library/Python/3.9/lib/python/site-packages (0.18.0)
Collecting Pillow<=7.2.0
  Using cached Pillow-7.2.0.tar.gz (39.1 MB)
Requirement already satisfied: cloudpickle<1.7.0,>=1.2.0 in ./Library/Python/3.9/lib/python/site-packages (from gym) (1.6.0)
Requirement already satisfied: numpy>=1.10.4 in ./Library/Python/3.9/lib/python/site-packages (from gym) (1.20.1)
Requirement already satisfied: pyglet<=1.5.0,>=1.4.0 in /usr/local/lib/python3.9/site-packages (from gym) (1.5.0)
Requirement already satisfied: scipy in ./Library/Python/3.9/lib/python/site-packages (from gym) (1.6.1)
Requirement already satisfied: future in ./Library/Python/3.9/lib/python/site-packages (from pyglet<=1.5.0,>=1.4.0->gym) (0.18.2)
Building wheels for collected packages: Pillow
  Building wheel for Pillow (setup.py) ... error
  ERROR: Command errored out with exit status 1:
[...]

PR #2220, opened 2021-04-15
fix build by disabling mujoco and moving to newer python patch versions

pzhokhov:
moving to newer python patch versions is necessary to avoid timeouts when pulling docker images for them (somehow, newer images provide better bandwidth? Or maybe old images are not hosted anymore at all?). At the same time, mujoco segfaults on 3.6.13 and 3.7.10. Hence, this PR disables mujoco tests.


dciborow:
please get this in so a new version of gym can go out before i have to pull it from my project....

dciborow:
@joschu , @christopherhesse , I am trying to get a new release of for this PR #2189 , but believe we need reviews from you for this one before a release can be made?

joschu (approved):

dciborow:
@pzhokhov , we got an ack! Are we able to move forward with release?

vi3itor:
@pzhokhov Thank you for the updates! I very much need this PR to be accepted for [google-research/football](https://github.com/google-research/football) because of the same issue as #2203. Do you need @christopherhesse approval to release a new version?

dciborow:
@joschu , @pzhokhov , 
Understand this repo is mostly in maintenance mode. My team would rather have an internal fork of this that we can keep secure ourselves. But as I am going to be doing the work regardless, it would actually be even easier for me if I could be an admin in the repo and set up dependabot for the package. 
Right now we are still in testing, so this is more of a test case for us then a blocker. But at this point, we aren't sure if we could get real updates out as fast as we need. 

Is there any interest in having some help on the build release side of this package so that next time we can roll an update for this out faster?

pzhokhov:
@dciborow thanks for the ping! While external maintenance could benefit the quality of the maintenance, unfortunately, our security team is not comfortable with giving admin access outside the org.

pzhokhov:
also, released a new version :)

PR #2231, opened 2021-05-13
Update lunar_lander.py

haluptzok:
cleanup before exit to eliminate error message printed out:  "Exception ignored in: <bound method Viewer.__del__ of <gym.envs.classic_control.rendering.Viewer...."
when running:
python lunar_lander.py


jkterry1:
@haluptzok This is closed by merging #2162 right?

PR #2235, opened 2021-05-31
Changed travis-ci.org to travis-ci.com

Anonymous-Ol:
Travis-CI checks have failed because it needs more permissions.

PR #2249, opened 2021-07-01
Make Dict a Mapping

krfricke:
Implementing `__len__` and `__contains__` will make the `Dict` space a proper `Mapping`


cclauss:
No tests run because of #2235

wookayin:
```
isinstance(space, collections.Mapping)
```
will be still False so I don't think is sufficient. Nor is a Sequence.

PR #2254, opened 2021-07-23
Fix EzPickle for Atari envs

Indoril007:
The full_action_space parameter of AtariEnv was not being passed to EzPickle. This would lead to issues when this parameter was set to True and the copies of the environment were used. The copies would not have a full action space whilst the original environment would.


cclauss:
No tests run because of #2235

jkterry1:
Closed per https://github.com/openai/gym/commit/b60a40db6ecbb3d498a908647ebb5fc0431f8799 (I didn't want the changes outside of the fix)

PR #2257, opened 2021-07-26
Upgrade to modern Python syntax

cclauss:
This pull request was created with the command [`pyupgrade --py36-plus **/*.py`](https://github.com/asottile/pyupgrade#readme)

Python f-strings simplify the code and [should speed up execution](https://www.scivision.dev/python-f-string-speed).

### DIFF
--- a/scripts/generate_json.py
+++ b/scripts/generate_json.py
@@ -15,7 +15,7 @@ steps = ROLLOUT_STEPS
 ROLLOUT_FILE = os.path.join(DATA_DIR, 'rollout.json')
 
 if not os.path.isfile(ROLLOUT_FILE):
-    logger.info("No rollout file found. Writing empty json file to {}".format(ROLLOUT_FILE))
+    logger.info(f"No rollout file found. Writing empty json file to {ROLLOUT_FILE}")
     with open(ROLLOUT_FILE, "w") as outfile:
         json.dump({}, outfile, indent=2)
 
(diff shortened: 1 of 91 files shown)


scripts/generate_json.py
```diff
@@ -74,12 +74,12 @@ def add_new_rollouts(spec_ids, overwrite):
...
     for spec in environments:
         if not overwrite and spec.id in rollout_dict:
-            logger.debug("Rollout already exists for {}. Skipping.".format(spec.id))
+            logger.debug(f"Rollout already exists for {spec.id}. Skipping.")
         else:
             modified = update_rollout_dict(spec, rollout_dict) or modified
 
     if modified:
-        logger.info("Writing new rollout file to {}".format(ROLLOUT_FILE))
+        logger.info(f"Writing new rollout file to {ROLLOUT_FILE}")
```
mzjp2, line 82: Admittedly this isn't a new problem, but one of the things you may want to do here (especially since this will turn into a `pylint` error down the line) is use `%` formatting here instead of f-strings, see [here](https://stackoverflow.com/questions/34619790/pylint-message-logging-format-interpolation) for why, and `pylint` (and flake8 -- not sure?) will throw a warning at you down the line

cclauss:
Too many conflicts after the black reformatting in #2265 The lazy evaluation of logging statements should be done in a separate, manual PR(s) prior to another attempt at using pyupgrade.

PR #2258, opened 2021-07-27
GitHub Action to lint Python code

cclauss:
Output: https://github.com/cclauss/gym/actions


jkterry1:
@cclauss

Would you be interesting in making the following contributions in the near-ish future?
1) Making the codebase compliant to a stye standard so that your tests pass
2) Since apparently Travis has permanently gone walk about, replace the all the travis tests with github actions tests

PR #2262, opened 2021-07-28
Fix CI: add tests to GitHub actions

mzjp2:
Hey folks,

Hope I'm not stepping on any toes here, but am a big fan of gym and thought I'd help out a little. This PR mimics the previous travis setup (from my understanding of reading the configuration) on GitHub actions in that:

* I build a `docker` image using the state of the repository
* I run `pytest --forked` as a container running the image I built in the step above

I've made a couple of improvements as I went along, including removing the disabling of `mujuco` and re-enabled the `mujuco` tests on Python 3.8 and 3.9.

I've also temporarily removed the `lint_python` job, since it's currently failing and my next task is to open a new, seperate PR that runs the linting _inside_ the Docker container.

In preparation for that, you'll note that I've removed the `CMD` that specifies `pytest` from the `Dockerfile` and moved it to the responsibility of the runner, so that I can re-use this image with all the dependencies built in to run linting tools inside this docker image.

I wanted to do the linting in a seperate PR and have this build here green, which I think is important. When I introduce the linting PR, I'll also try my best to make the lint job green, rather than having it fail.

Final note, the build times here are pretty quick, but are also slightly slower and hammering pypi and debian mirrors more than we should as good citizens, so I'll also open a new PR to hopefully [fiddle about with the github actions](https://evilmartians.com/chronicles/build-images-on-github-actions-with-docker-layer-caching) and enable docker layer caching for us.


mzjp2:
It looks like I'll need one of you to approve my workflow running. You can see it running sucessfully on my fork over here: https://github.com/mzjp2/gym/actions/runs/1073236815

mzjp2:
I'm also happy to provide a `Makefile` to help people run tests and linting locally inside a Docker container.

jkterry1:
Awesome, thanks a ton! 

-I approved it so it can run
-How on earth did you get MuJuCo to build? It's proprietary software requiring a key
-MuJuCo was disabled on Python 3.8+ because Gym uses an old version of MuJuCo. Fixing the MuJuCo situation is a future problem. I'm comfortable fully disabling tests for it, at least temporarily.
-Adding the relevant makefile would be create, as long as it's very clear to use
-Shouldn't the docker layer caching etc. be included in this PR and not a new one?
-If you could do a PR integrating linting into the tests immediately following this PR that'd be greatly appreciated.

mzjp2:
> -How on earth did you get MuJuCo to build? It's proprietary software requiring a key

Honestly, I'm not actually sure it _is_ being built, but our tests are passing regardless. In either case, I compared the logs in github actions to the old Travis ones, and they are identical, including the "mujuco failed to build" log line 🤔 

> -MuJuCo was disabled on Python 3.8+ because Gym uses an old version of MuJuCo. Fixing the MuJuCo situation is a future problem. I'm comfortable fully disabling tests for it, at least temporarily.

Gotcha, let me poke in a little bit on what's happening here and why things are passing...

> -Adding the relevant makefile would be create, as long as it's very clear to use

I might leave this for a seperate PR, because (see below)...

> -Shouldn't the docker layer caching etc. be included in this PR and not a new one?

Yep, fair point -- let me push to this PR once I have it running (testing GH actions is a pain 😅)

> -If you could do a PR integrating linting into the tests immediately following this PR that'd be greatly appreciated.

It'll most likely be tomorrow night, I'm off to bed soon, but the work here should hopefully unlock CI and using the docker image to setup linting without it failing on the `pip install` stage as it was before this PR

mzjp2:
Okay, so quick progress update:

* I've removed the mujuco stuff and testing without it atm, it's all passing -- so all good
* Implemented caching, turns out it actually took _longer_ given how quick our image build times are since `docker buildx` has a lot of overhead and GH Actions itself isn't great at copying around large caches and compressing/de-compressing them (our images are ~1.5gb)

You can see an example of a cached run [here](https://github.com/mzjp2/gym/actions/runs/1073476987) -- it takes longer than the non-cached run and has wayyy more complexity (uses `docker buildx`) 

so I've pushed the minor fixes here, but reverted the caching changes entirely. I'd suggest we get this in, I can work on the `linting` stuff tomorrow night or happy for someone else to pick it up (should be straightforward from here)

cclauss:
Nice!!  Instead of removing lint_python, please consider picking up the #2263 change so we can keep improving…

mzjp2:
> Nice!! Instead of removing lint_python, please consider picking up the #2263 change so we can keep improving…

Awesome!! I'll undo my deletion, and then you guys can take care of merging `master` back into my branch once your PR is approved and get my PR in too

jkterry1:
I just merged this, I'll leave the further refinements to you for tomorrow night.

pzhokhov:
This is great, thanks a lot @mzjp2 (and @jkterry1 on who's watch it got merged!)

PR #2263, opened 2021-07-28
lint_python: sudo apt-get install python-opengl

cclauss:


jkterry1:
Can you walk me through why this change is nescesary?

cclauss:
Sorry...  I was trying to understand why `pytest .` was failing so I was trying
https://github.com/openai/gym#installing-everything but that does not seem to help much.

PR #2282, opened 2021-08-01
Float pillow version

jkterry1:


dciborow:
@jkterry1 , can we get a new github release with this update soon?

PR #2286, opened 2021-08-02
Delete dead code (reopening)

jkterry1:
(I accidentally merged this a second ago)

This code appears to be very old and unused and like it probably shouldn't be here anymore, so I'm removing it. I'll leave this open for a few days in case anyone has objections. I believe that the new docs (with a list of external tutorials etc.) coming later this month will cover every plausible service these had.


jkterry1:
(flakey test fixed in #2287)

PR #2289, opened 2021-08-02
Add compliance testing for API

### LINKED ISSUE #2278: API Compliance Testing
To help users with creating custom environments, we should have a user test for environments to ensure the Gym API is properly followed, similar to what PettingZoo has. The easiest solution is to upstream the checker in SB3.

Crosspost from https://github.com/DLR-RM/stable-baselines3/issues/525

amtamasi:
Closes #2278


jkterry1:
This is awesome!

-Please make this test be applied to all environments in Gym as part of CI tests
-Please look through the API in PettingZoo and pull the relevant errors and warnings from it that aren't included here. E.g. there's a bunch of warnings/recommendations about observation/action spaces in the PettingZoo tests that I don't believe are here

gym/utils/env_checker.py
```diff
@@ -0,0 +1,345 @@
...
+    if np.any(observation_space.low != 0) or np.any(observation_space.high != 255):
+        warnings.warn(
+            f"It seems that your observation space {key} is an image but the "
+            "upper and lower bounds are not in [0, 255]. "
+            "Generally, CNN policies assume observations are within that range, "
+            "so you may encounter an issue if the observation values are not."
+        )
+
+
+def _check_unsupported_spaces(
```
tristandeleu, line 68: I think Gym by default should support any valid `Space`, including (nested) spaces, `Tuple`, or even custom spaces. The fact that it may not be compatible with specific libraries should be checked on those libraries' side (e.g. SB3 for these particular requirements).

amtamasi: That's fair, I tend to agree. I wasn't sure if you all wanted to keep those library-specific warnings. I can take that one out!

jkterry1:
@antamasi- as long as the code is attributed and you say it was under the MIT license you don't have to copy paste the entire license it into the file

Issue #2279, opened 2021-07-31
Discussion of Vector Environment API Breaking Changes


jkterry1:
There are numerous long standard complaints about the vector environment API, to the point where it is the one thing that most people have seemed to think breaking changes are warranted for (which I agree with). I've created this thread to come to a consensus on the solution.

Because whatever finalized changes we come up with here will be very significant to the community and this can only really be done once, I am specifically insisting that the following people who I've worked with previously agree to whatever breaking changes are made:

Benjamin Black (@benblack769, UMD and PettingZoo)
Antonin Raffin (@araffin, DLR RM and Stable Baselines)
Chris Nota (@cpnota, UMass Amherst and the Autonomous Learning Library)
Costa Huang (@vwxyzjn, Weights and Biases and CleanRL)
Jesse Fareboro (@JesseFarebro, MILA and the Arcade Learning Environment)

joschu:
You might want to check out https://github.com/openai/gym3

tristandeleu:
> There are numerous long standard complaints about the vector environment API, to the point where it is the one thing that most people have seemed to think breaking changes are warranted for (which I agree with).

Can you be more precise about these complaints that warrant breaking the API, in order to get a better understanding of what needs to be done? As far as I know, almost all the [issues involving `VectorEnv`](https://github.com/openai/gym/issues?q=is:issue+VectorEnv+OR+AsyncVectorEnv+OR+SyncVectorEnv) here have been addressed one way or another.

To be specific, there are a number of things that can be improved in my opinion, but very few of them require breaking the API:
 - Add documentation. This should be the number 1 priority, but unfortunately it does not seem to be part of the current plan for the documentation #2276 
 - Refactor the [`_worker`](https://github.com/openai/gym/blob/7c9ae6d14087fe50714d59bc36b1797560961710/gym/vector/async_vector_env.py#L402) and [`_worker_shared_memory`](https://github.com/openai/gym/blob/7c9ae6d14087fe50714d59bc36b1797560961710/gym/vector/async_vector_env.py#L438) in order to avoid code duplication and to simplify custom workers (custom workers should be treated as [an advanced feature](https://github.com/openai/gym/blob/7c9ae6d14087fe50714d59bc36b1797560961710/gym/vector/async_vector_env.py#L71-L77) anyway)
 - Add `call`/`get_attr`/`set_attr` #1600
[...]

PR #2296, opened 2021-08-05
Make RecordEpisodeStatistics work with VectorEnv

vwxyzjn:
Following up with #2279, this PR makes `RecordEpisodeStatistics` work with `VectorEnv` as well. That is

```python
import gym
from gym.vector import SyncVectorEnv

def make_env(gym_id, seed):
    def thunk():
        env = gym.make(gym_id)
        env = gym.wrappers.RecordEpisodeStatistics(env)
        env.seed(seed)
        env.action_space.seed(seed)
        return env
    return thunk

envs = SyncVectorEnv([make_env("CartPole-v1", 1 + i) for i in range(2)])
envs.reset()
for i in range(100):
    for info in infos:
        if "episode" in info.keys():
            break
```

will produce the same results as

```python
import gym
from gym.vector import SyncVectorEnv

def make_env(gym_id, seed):
    def thunk():
        env = gym.make(gym_id)
        env.seed(seed)
        env.action_space.seed(seed)
        return env
    return thunk

envs = SyncVectorEnv([make_env("CartPole-v1", 1 + i) for i in range(2)])
envs.reset()
for i in range(100):
    for info in infos:
        if "episode" in info.keys():
            break
```

The reason why this PR is important is that some envs don't allow you to insert a `RecordEpisodeStatistics` in the `make_env` function. The [procgen environment](https://github.com/openai/procgen) is one such example. So this PR will allow you to do something like this:

```python
import gym
from procgen import ProcgenEnv

envs = ProcgenEnv(num_envs=2, env_name="coinrun", num_levels=0, start_level=0, distribution_mode='hard')
envs.reset()
for i in range(100):
    for info in infos:
        if "episode" in info.keys():
            break
```

### DIFF
--- a/gym/wrappers/record_episode_statistics.py
+++ b/gym/wrappers/record_episode_statistics.py
@@ -11,7 +11,9 @@ class RecordEpisodeStatistics(gym.Wrapper):
         if not isinstance(env, gym.vector.VectorEnv):
             self.num_envs = 1
             self.env_is_vec = False
-        self.t0 = time.time()  # TODO: use perf_counter when gym removes Python 2 support
+        self.t0 = (
+            time.time()
+        )  # TODO: use perf_counter when gym removes Python 2 support
         self.episode_count = 0
         self.episode_returns = None
         self.episode_lengths = None
@@ -25,7 +27,9 @@ class RecordEpisodeStatistics(gym.Wrapper):
         return observations
 
     def step(self, action):
-        observations, rewards, dones, infos = super(RecordEpisodeStatistics, self).step(action)
+        observations, rewards, dones, infos = super(RecordEpisodeStatistics, self).step(
+            action
+        )
         self.episode_returns += rewards
         self.episode_lengths += 1
         if not self.env_is_vec:
@@ -36,11 +40,20 @@ class RecordEpisodeStatistics(gym.Wrapper):
                 infos[i] = infos[i].copy()
                 episode_return = self.episode_returns[i]
                 episode_length = self.episode_lengths[i]
-                episode_info = {"r": episode_return, "l": episode_length, "t": round(time.time() - self.t0, 6)}
+                episode_info = {
+                    "r": episode_return,
+                    "l": episode_length,
+                    "t": round(time.time() - self.t0, 6),
+                }
                 infos[i]["episode"] = episode_info
                 self.return_queue.append(episode_return)
                 self.length_queue.append(episode_length)
                 self.episode_count += 1
                 self.episode_returns[i] = 0
                 self.episode_lengths[i] = 0
-        return observations, rewards, dones if self.env_is_vec else dones[0], infos if self.env_is_vec else infos[0]
+        return (
+            observations,
+            rewards,
+            dones if self.env_is_vec else dones[0],
+            infos if self.env_is_vec else infos[0],
+        )
--- a/gym/wrappers/record_episode_statistics.py
+++ b/gym/wrappers/record_episode_statistics.py
@@ -1,39 +1,46 @@
 import time
 from collections import deque
+import numpy as np
 import gym
 
 
 class RecordEpisodeStatistics(gym.Wrapper):
     def __init__(self, env, deque_size=100):
         super(RecordEpisodeStatistics, self).__init__(env)
-        self.t0 = (
-            time.time()
-        )  # TODO: use perf_counter when gym removes Python 2 support
-        self.episode_return = 0.0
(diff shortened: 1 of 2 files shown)


jkterry1:
Could you please add tests before I merge?

vwxyzjn:
@jkterry1 Could you approve  workflows?

vwxyzjn:
Hey @jkterry1 can you approve the workflows again

gym/wrappers/record_episode_statistics.py
```diff
@@ -1,39 +1,59 @@
...
         self.return_queue = deque(maxlen=deque_size)
         self.length_queue = deque(maxlen=deque_size)
 
     def reset(self, **kwargs):
-        observation = super(RecordEpisodeStatistics, self).reset(**kwargs)
-        self.episode_return = 0.0
-        self.episode_length = 0
-        return observation
+        observations = super(RecordEpisodeStatistics, self).reset(**kwargs)
+        self.episode_returns = np.zeros(self.num_envs, dtype=np.float32)
```
tristandeleu, line 25: `self.num_envs` is not defined here if `env` is a `VectorEnv` instance.

gym/wrappers/record_episode_statistics.py
```diff
@@ -1,39 +1,59 @@
...
 import gym
 
 
 class RecordEpisodeStatistics(gym.Wrapper):
     def __init__(self, env, deque_size=100):
         super(RecordEpisodeStatistics, self).__init__(env)
+        self.env_is_vec = True
+        if not isinstance(env, gym.vector.VectorEnv):
+            self.num_envs = 1
+            self.env_is_vec = False
```
tristandeleu, line 13:
```suggestion
        self.env_is_vec = isinstance(env, gym.vector.VectorEnv)
        self.num_envs = getattr(env, 'num_envs', 1)
```

gym/wrappers/test_record_episode_statistics.py
```diff
@@ -22,3 +22,17 @@ def test_record_episode_statistics(env_id, deque_size):
                 break
     assert len(env.return_queue) == deque_size
     assert len(env.length_queue) == deque_size
+
+
+@pytest.mark.parametrize("env_id", ["CartPole-v0"])
+def test_record_episode_statistics_with_vectorenv(env_id):
+    envs = gym.vector.make(env_id, asynchronous=False)
```
tristandeleu, line 29: With the corresponding imports
```python
from gym.vector import AsyncVectorEnv, SyncVectorEnv
from gym.vector.tests.utils import make_env
```

```suggestion
@pytest.mark.parametrize("klass", [SyncVectorEnv, AsyncVectorEnv])
@pytest.mark.parametrize("num_envs", [1, 4])
def test_record_episode_statistics_with_vectorenv(klass, num_envs):
    env_fns = [make_env("CartPole-v0", i) for i in range(num_envs)]
    envs = klass(env_fns)
```

vwxyzjn: Unfortunately it’s gonna fail with `AsyncVectorEnv` because the `envs.env.envs[0].spec.max_episode_steps` is inaccessible. Maybe I should just hardcore a 201 instead of `envs.env.envs[0].spec.max_episode_steps`?

gym/wrappers/record_episode_statistics.py
vwxyzjn: I don’t follow. `VectorEnv` has a `num_envs` attribute, right?

vwxyzjn (commented):
Thank you @tristandeleu so much for the review. I just have one comment regarding the test cases.

gym/wrappers/record_episode_statistics.py
tristandeleu: Oh right the wrapper inherits the properties from `env`, sorry my bad!

gym/wrappers/test_record_episode_statistics.py
tristandeleu: Oh sorry I didn't see that you were looping over that later. Then you can ignore this (maybe keeping the parametrization for `num_envs`?).

vwxyzjn: Done. Would you mind Allowing the GitHub action workflow runs? I have some weird setup That makes it difficult to run test cases locally….

tristandeleu: @jkterry1

Issue #2297, opened 2021-08-05
Deprecating the `Monitor` in favor of `RecordEpisodeStatistics` and `RecordVideo`


vwxyzjn:
Kicking off a discussion thread here. Currently, the [`Monitor`](https://github.com/openai/gym/blob/master/gym/wrappers/monitor.py) has two features: recording videos via the [video_recorder.py](https://github.com/openai/gym/blob/master/gym/wrappers/monitoring/video_recorder.py) and recording stats via the [stats_recorder.py](https://github.com/openai/gym/blob/master/gym/wrappers/monitoring/stats_recorder.py)

## Issues
* The [stats_recorder.py](https://github.com/openai/gym/blob/master/gym/wrappers/monitoring/stats_recorder.py) does not allow the environment to reset if it's not done (see #599). In my experience, this causes some confusions and can be a pain to deal with, especially when the users want to do something custom. 
* Because the way the `Monitor` is setup, it can't be easily applied to vectorized env directly.
* `Monitor` also has some legacy stuff such as [`mode (['evaluation', 'training'])`](https://github.com/openai/gym/blob/3586d4a3b7605c49e0fd1413afb3624006b9302a/gym/wrappers/monitor.py#L27) and an info to use [`gym.upload`](https://github.com/openai/gym/blob/3586d4a3b7605c49e0fd1413afb3624006b9302a/gym/wrappers/monitor.py#L189-L192). I am also not sure how to use the [`resume` argument](https://github.com/openai/gym/blob/3586d4a3b7605c49e0fd1413afb3624006b9302a/gym/wrappers/monitor.py#L24).

## Potential resolution

I feel the `Monitor` class is doing too many things at once, and it will be a good idea to deprecating the `Monitor` in favor of `RecordEpisodeStatistics` and a newly-created `VideoRecorder`. 

* Note the [stats_recorder.py](https://github.com/openai/gym/blob/master/gym/wrappers/monitoring/stats_recorder.py) has duplicate functionality as `RecordEpisodeStatistics`. We can add the stats saving functionality to `RecordEpisodeStatistics` if desired.

vwxyzjn:
If folks are happy with this proposal, I am happy to prepare PRs.

vwxyzjn:
How do you feel about this PR @tristandeleu? Would love to get your thoughts!

tristandeleu:
[...]

PR #2300, opened 2021-08-06
Add RecordVideo wrapper

vwxyzjn:
This PR adds a lightweight `RecordVideo` wrapper. It allows for both episodic video recording like [gym.wrappers.Monitor](https://github.com/openai/gym/blob/master/gym/wrappers/monitor.py) and fixed interval recording like SB3's [VecVideoRecorder](https://stable-baselines3.readthedocs.io/en/master/guide/vec_envs.html#vecvideorecorder). The motivation behind this PR can be found at #2297. 

## Get started

You can record fixed interval video by doing

```python
import gym
envs = gym.vector.make("CartPole-v1", num_envs=2, asynchronous=False)
envs = gym.wrappers.RecordVideo(envs, "videos", record_video_trigger=lambda x: x % 100 == 0, video_length=200)
envs.reset()
for i in range(1000):
    for info in infos:
        if "episode" in info.keys():
            break
```

By not specifying `video_length`, by default the videos are episodic

```python
import gym
envs = gym.vector.make("CartPole-v1", num_envs=2, asynchronous=False)
envs = gym.wrappers.RecordVideo(envs, "videos", record_video_trigger=lambda x: x % 100 == 0)
envs.reset()
for i in range(1000):
    for info in infos:
        if "episode" in info.keys():
            break
```

It also works with the non-vector gym envs. Overall just fairly flexible.
```python
import gym
env = gym.make("CartPole-v1")
env = gym.wrappers.RecordVideo(env, "videos", record_video_trigger=lambda x: x % 100 == 0)
observation = env.reset()
for _ in range(1000):
  env.render()
  action = env.action_space.sample() # your agent here (this takes random actions)

  if done:
    observation = env.reset()
env.close()
```

```python
import gym
from gym.vector import SyncVectorEnv

def make_env(gym_id, seed):
    def thunk():
        env = gym.make(gym_id)
        env = gym.wrappers.RecordVideo(env, "videos", record_video_trigger=lambda x: x % 100 == 0)
        env.seed(seed)
        env.action_space.seed(seed)
        return env
    return thunk

envs = SyncVectorEnv([make_env("CartPole-v1", 1 + i) for i in range(2)])
envs.reset()
for i in range(1000):
    for info in infos:
        if "episode" in info.keys():
            break
```

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -39,7 +39,6 @@ class Env(object):
     # Set these in ALL subclasses
     action_space = None
     observation_space = None
-    is_vector_env = False
 
     @abstractmethod
     def step(self, action):
@@ -231,7 +230,6 @@ class Wrapper(Env):
         self.observation_space = self.env.observation_space
         self.reward_range = self.env.reward_range
         self.metadata = self.env.metadata
-        self.is_vector_env = self.env.is_vector_env
 
     def __getattr__(self, name):
         if name.startswith("_"):
--- a/gym/core.py
+++ b/gym/core.py
@@ -39,6 +39,7 @@ class Env(object):
     # Set these in ALL subclasses
     action_space = None
     observation_space = None
+    is_vector_env = False
 
     @abstractmethod
     def step(self, action):
@@ -230,6 +231,7 @@ class Wrapper(Env):
         self.observation_space = self.env.observation_space
         self.reward_range = self.env.reward_range
         self.metadata = self.env.metadata
+        self.is_vector_env = self.env.is_vector_env
 
     def __getattr__(self, name):
         if name.startswith("_"):
--- a/gym/vector/sync_vector_env.py
+++ b/gym/vector/sync_vector_env.py
@@ -111,6 +111,3 @@ class SyncVectorEnv(VectorEnv):
             "observation spaces from all environments must be "
             "equal.".format(self.single_observation_space)
         )
-
-    def render(self, mode="human", **kwargs):
-        return self.envs[0].render(mode, **kwargs)
--- a/gym/vector/sync_vector_env.py
+++ b/gym/vector/sync_vector_env.py
@@ -33,6 +33,7 @@ class SyncVectorEnv(VectorEnv):
         self.env_fns = env_fns
         self.envs = [env_fn() for env_fn in env_fns]
         self.copy = copy
+        self.metadata = self.envs[0].metadata
 
         if (observation_space is None) or (action_space is None):
             observation_space = observation_space or self.envs[0].observation_space
@@ -110,3 +111,6 @@ class SyncVectorEnv(VectorEnv):
             "observation spaces from all environments must be "
             "equal.".format(self.single_observation_space)
         )
+
+    def render(self, mode="human", **kwargs):
+        return self.envs[0].render(mode, **kwargs)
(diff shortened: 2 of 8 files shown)


vwxyzjn:
I will add test cases once this change / API design is agreed upon.

jkterry1:
I'm open to the change. I don't want to merge any new featurea or breaking changes until a maintaince release of Gym is cut (hopefully at the beginning of next week).

gym/core.py
```diff
@@ -39,6 +39,7 @@ class Env(object):
     # Set these in ALL subclasses
     action_space = None
     observation_space = None
+    is_vector_env = False
```
tristandeleu, line 42: I'm not a huge fan of overloading the `Env` API to handle `VectorEnv`. Ideally `VectorEnv` should be able to exist without having to touch too much the base API. I understand that this may be useful for wrappers to handle both regular `Env` and `VectorEnv`, but I would be more in favor to use `VectorEnvWrapper` in those cases that would require `VectorEnv`-specific features.

gym/vector/sync_vector_env.py
```diff
@@ -110,3 +111,6 @@ def _check_observation_spaces(self):
             "observation spaces from all environments must be "
             "equal.".format(self.single_observation_space)
         )
+
+    def render(self, mode="human", **kwargs):
+        return self.envs[0].render(mode, **kwargs)
```
tristandeleu, line 116: I think the `render` method of `VectorEnv` should return the rendered images for all the environment, similar to #1624, and not just the first environment. If rendering is only enabled for the first environment, I think it would make more sense to have the wrapper at the level of an individual environment. Something like:

```python
import gym

def make_env(i):
    def _make():
        env = gym.make('CartPole-v1')
        if i == 0:
            env = gym.wrappers.RecordVideo(
                env, 'videos',
                record_video_trigger=lambda x: x % 100 == 0,
                video_length=200
            )
        return env
    return _make

envs = gym.vector.AsyncVectorEnv([make_env(i) for i in range(4)])

observations = envs.reset()
for i in range(1000):
    actions = envs.action_space.sample()
    observations, rewards, dones, infos = envs.step(actions)
```

jkterry1:
@vwxyzjn Can you please add CI testing for your wrapper?

gym/core.py
vwxyzjn: That makes sense. I will find a workaround for this.

gym/vector/sync_vector_env.py
vwxyzjn: That's a good point. I think a reasonable next step from this is to get a resolution for #1624, and then merge this PR.

vwxyzjn:
Thanks @tristandeleu for the review!

vwxyzjn:
@jkterry1 will add the tests after we agree upon the API.

vwxyzjn:
The commits above

1. removes the "overloading the Env API to handle `VectorEnv`".
2. make `RecordVideo` work with regular gym envs but not vector gym envs.

vwxyzjn:
This is really awkward to say, but the way that I set up the PR is a little screwed up (i.e. using the master branch). As a result, it's kind of difficult for me to pull the documentation changes and add more documentation. Would it be possible to Mars is PR first, and then I'll follow up with another PR with documentation?

jkterry1:
Merged, please create the PR :)

PR #2307, opened 2021-08-11
Fix pendulum rendering when sending zero control

Olimoyo:
When sending a control input of zero the rendering was wrong because `if self.last_u:` would evaluate to False.


jkterry1:
@Olimoyo @RyanNavillus will this require a version bump?

Olimoyo:
Afaik, this only changes the rendering and nothing else with the underlying environment. However, I am not too sure what the best practices are for version bumps.

PR #2310, opened 2021-08-11
test_python on macOS, not Linux

cclauss:
Also, `mypy --ignore-missing-imports`


jkterry1:
Can you explain why these changes are needed...?

PR #2316, opened 2021-08-16
Autoclose manually maintained video recorders

XuehaiPan:
When I am using [`VideoRecorder`](https://github.com/openai/gym/blob/master/gym/wrappers/monitoring/video_recorder.py) to recorder my rollout videos, I will get a error if the program is exited unexpectedly (e.g. a keyboard interrupt).

```
[rawvideo @ 0x5646579bf4c0] Invalid buffer size, packet size 327680 < expected frame_size 1920000
Error while decoding stream #0:0: Invalid argument
```

`gym.wrappers.Monitor` has been registered with an closer and can be autoclosed at program exit. However, for low level API users, `VideoRecorder` itself may should have autoclose feature as well.

PR #2318, opened 2021-08-16
Autoclose manually maintained video recorders

XuehaiPan:
When I am using [`gym.wrappers.monitoring.video_recorder.VideoRecorder`](https://github.com/openai/gym/blob/master/gym/wrappers/monitoring/video_recorder.py) to recorder my rollout videos, I will get a error if the program is exited unexpectedly (e.g. a keyboard interrupt).

```
[rawvideo @ 0x5646579bf4c0] Invalid buffer size, packet size 327680 < expected frame_size 1920000
Error while decoding stream #0:0: Invalid argument
```

`gym.wrappers.Monitor` has been registered with a closer and can be autoclosed at program exit. However, for low level API users, `VideoRecorder` itself may should have autoclose feature as well.

I know I can use the higher API `gym.wrappers.Monitor` to handle this. But I just want to record only one episode, so I prefer to use `VideoRecorder` rather than `gym.wrappers.Monitor`. In addition, `gym.wrappers.Monitor` does not allow users to customize the file name.

This PR is reopened for #2316 #2317.

### DIFF
--- a/gym/wrappers/monitoring/video_recorder.py
+++ b/gym/wrappers/monitoring/video_recorder.py
@@ -10,10 +10,14 @@ import distutils.version
 import numpy as np
 
 from gym import error, logger
+from gym.utils import closer
 
 
 def touch(path):
     open(path, "a").close()
+
+
+video_recorder_closer = closer.Closer()
 
 
 class VideoRecorder(object):
@@ -37,6 +41,7 @@ class VideoRecorder(object):
         modes = env.metadata.get("render.modes", [])
         self._async = env.metadata.get("semantics.async")
         self.enabled = enabled
+        self._recorder_id = video_recorder_closer.register(self)
 
         # Don't bother setting anything else if not enabled
         if not self.enabled:
@@ -144,7 +149,7 @@ class VideoRecorder(object):
                 self._encode_image_frame(frame)
 
     def close(self):
-        """Make sure to manually close, or else you'll leak the encoder process"""
+        """Flush all data to disk and close any open frame encoders."""
         if not self.enabled:
             return
 
@@ -178,9 +183,17 @@ class VideoRecorder(object):
 
         self.write_metadata()
 
+        # Stop tracking this for autoclose
+        video_recorder_closer.unregister(self._recorder_id)
+        self.enabled = False
+
     def write_metadata(self):
         with open(self.metadata_path, "w") as f:
             json.dump(self.metadata, f)
+
+    def __del__(self):
+        # Make sure we've closed up shop when garbage collecting
+        self.close()
 
     def _encode_ansi_frame(self, frame):
         if not self.encoder:


jkterry1:
Please make tests pass and then we'll take a look :)

vwxyzjn:
Hey @XuehaiPan thanks for the contribution. Would you mind giving an example of how the end users would use the API?
The one example in the test cases does not look like a pattern that end-users would use.

```python
 def test_autoclose():
     env = gym.make("CartPole-v1")
     rec = VideoRecorder(env)
     env.reset()
     rec.capture_frame()

     rec_path = rec.path
     with video_recorder_closer.lock:
         num_registered = len(video_recorder_closer.closeables)
     del rec

     with video_recorder_closer.lock:
         assert len(video_recorder_closer.closeables) == num_registered - 1
     assert os.path.exists(rec_path)
     f = open(rec_path)
     assert os.fstat(f.fileno()).st_size > 100
```

XuehaiPan:
> The one example in the test cases does not look like a pattern that end-users would use.

Agreed.

I use this `del rec` statement to test the autoclose feature when remove the recorder instance without an explicit `close()` call. A typical code would look like:

```python
def play(env, path):
    rec = VideoRecorder(env, path)
    env.reset()
    rec.capture_frame()
    while True:
        rec.capture_frame()

    # function ends without an explicit rec.close()
    # the Python interpreter will implicitly `del rec` at garbage clean

def main():
    play(gym.make("CartPole-v1"), '.')
```

I'll refactor this.

vwxyzjn:
Thank you the example above is very helpful. I just gave the following code a run using master.

```
import gym
from gym.wrappers.monitoring import video_recorder

env = gym.make("CartPole-v1")
rec = video_recorder.VideoRecorder(env, "video.mp4")
env.reset()
rec.capture_frame()
for i in range(100):
    rec.capture_frame()
    if i == 20:
        raise
```

To my surprise, the video is still logged. I also tried doing a keyboard interrupt and the video is still logged, which doesn't seem to match your description
> When I am using gym.wrappers.monitoring.video_recorder.VideoRecorder to recorder my rollout videos, I will get a error if the program is exited unexpectedly (e.g. a keyboard interrupt).

Maybe this is an operating system thing? Mine is macos.

XuehaiPan:
> To my surprise, the video is still logged.

It seems that the `stdin` of the subprocess (`ImageEncoder`'s backend process) will be closed when the main process exiting.

https://github.com/openai/gym/blob/2853ce4797a4dadb1f8702c8e35143d9fbd43ea6/gym/wrappers/monitoring/video_recorder.py#L436-L440

However, without an explicit `rec.close()`, the subprocess is still running even if the `rec` object is not accessable (local variable defined in function). And the resources will be released only when the main process is exited, which is not suitable for Jupyter Notebook users.

Peek 2021-08-19 18-00

gym/wrappers/monitoring/tests/test_video_recorder.py
```diff
@@ -5,7 +5,7 @@
 import numpy as np
 
 import gym
-from gym.wrappers.monitoring.video_recorder import VideoRecorder
+from gym.wrappers.monitoring.video_recorder import VideoRecorder, video_recorder_closer
```
pzhokhov, line 8: I don't think this new import is used

gym/wrappers/monitoring/video_recorder.py
```diff
@@ -10,12 +10,16 @@
...
 
 from gym import error, logger
+from gym.utils import closer
 
 
 def touch(path):
     open(path, "a").close()
 
 
+video_recorder_closer = closer.Closer()
```
pzhokhov, line 20: I believe using Closer() is not actually necessary at all for the new test to pass - as it is garbage collector that calls `__del__` directly, which then calls close(). My understanding is Closer() is only useful in the cases when you explicitly call Closer().close() before terminating the program.

jkterry1:
@XuehaiPan Could you please take a look at peter zhokov's comments?

gym/wrappers/monitoring/tests/test_video_recorder.py
XuehaiPan: Yes, and this could be removed.

This was imported for the draft version of `test_autoclose` (https://github.com/openai/gym/commit/e8e81d8c441c88df6a1225fe9fcd31b663183fb7), and the test case had been updated as suggested in https://github.com/openai/gym/pull/2318#issuecomment-899923137.

gym/wrappers/monitoring/video_recorder.py
XuehaiPan: I don't think we need the closer either, as we have called `close()` in `__del__()`. I add this closer here to follow the design of `Monitor`:

https://github.com/openai/gym/blob/8da6224b7263f9541a55d541d8268f68ece7d99e/gym/wrappers/monitor.py#L178-L196

And as commented at the top of this PR (https://github.com/openai/gym/pull/2318#issue-713169148):

> gym.wrappers.Monitor has been registered with a closer and can be autoclosed at the program exit. However, for low level API users, VideoRecorder itself may should have autoclose feature as well.

gym/wrappers/monitoring/tests/test_video_recorder.py
jkterry1: Could you please great a small PR adressing this and the other note?

XuehaiPan: I have opened a PR #2404 to test for the number of recorders that registered in the closer.

PR #2321, opened 2021-08-17
More setup.py fixes

jkterry1:


tristandeleu:
This is confusing: cloudpickle is used by `AsyncVectorEnv`, which is not a wrapper. It doesn't make sense to have it as an extra requirement for wrappers.

Also, what is the motivation behind moving cloudpickle to extra requirements? I understand why you'd like to move OpenCV and Scipy as they are large libraries, but cloudpickle is a very lightweight dependency, why not keep it as a general requirement for the core functionality `AsyncVectorEnv`?

jkterry1:
Yep, I was just trying something with CI. It'll stay as a core dependency.

PR #2323, opened 2021-08-17
Change action rescale argument names

jkterry1:
`a` and `b` are not suitable names here. This should only be a breaking change in very rare circumstances and the fix where it is breaking is trivial.

### DIFF
--- a/gym/wrappers/rescale_action.py
+++ b/gym/wrappers/rescale_action.py
@@ -4,11 +4,11 @@ from gym import spaces
 
 
 class RescaleAction(gym.ActionWrapper):
-    r"""Rescales the continuous action space of the environment to a range [a,b].
+    r"""Rescales the continuous action space of the environment to a range [min_action, max_action].
 
     Example::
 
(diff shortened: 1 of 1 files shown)


gym/wrappers/rescale_action.py
```diff
@@ -4,33 +4,33 @@
...
             env.action_space, spaces.Box
         ), "expected Box action space, got {}".format(type(env.action_space))
-        assert np.less_equal(a, b).all(), (a, b)
+        assert np.less_equal(min_action, max_action).all(), (min_action, max_action)
 
         super(RescaleAction, self).__init__(env)
-        self.a = np.zeros(env.action_space.shape, dtype=env.action_space.dtype) + a
-        self.b = np.zeros(env.action_space.shape, dtype=env.action_space.dtype) + b
+        self.min_action = np.zeros(env.action_space.shape, dtype=env.action_space.dtype) + min_action
+        self.b = np.zeros(env.action_space.shape, dtype=env.action_space.dtype) + max_action
```
vwxyzjn, line 24: This should be self.max_action instead of self.b

PR #2324, opened 2021-08-17
Create draft md file for wrappers documentation page

jkterry1:


docs/wrappers.md
```diff
@@ -1,14 +1,78 @@
...
+## Lambda Wrappers
+
+`ActionLambda(env, change_action_fn, change_space_fn)` [text]
+* Bring over from SuperSuit
+
+`ObservationLambda(env, observation_fn, observation_space_fn)` [text]
+* Bring over from SuperSuit, replaces TransformObservation
+
+`RewardLambda(env, reward_fn)` [text]
+* Bring over from SuperSuit, replaces TransformReward
```
tristandeleu, line 63: Any reason why renaming `TransformObservation` and `TransformReward`? This seems like a easily avoidable breaking change.

docs/wrappers.md
```diff
@@ -1,14 +1,78 @@
...
+`RecordEpisodeStatistic(env)` [text]
+* Needs review (including for good assertion messages and test coverage)
+
+`RecordVideo(env, ...)` [text]
+* https://github.com/openai/gym/pull/2300
+
+`TimeLimit(env, max_episode_steps)` [text]
+* Needs review (including for good assertion messages and test coverage)
+
+Some sort of vector environment conversion wrapper needs to be added here, this will be figured out after the API is changed.
```
tristandeleu, line 78: What would be the advantage of having a wrapper for converting an environment to a vector environment compared to using `SyncVectorEnv`/`AsyncVectorEnv` directly?

I don't think that would be appropriate, for example let's say you have a wrapper `Vectorize`, and then you do
```python
env = gym.make('CartPole-v1')
envs = Vectorize(env)
```
Then how do you define [`envs.unwrapped`](https://github.com/openai/gym/blob/master/gym/core.py#L271-L273) for parallel environments? Do you always run a version of the wrapped environment in the main process (which is unnecessary most of the time)? Or do you run that environment in a subprocess (for example, the first process, but that probably rules out `envs.unwrapped.unwrapped`)?

jkterry1:
-The changes to the transform wrappers will meaningfully change the internal function of them. The wrappers aren't versioned; if they were we could just bump versions.
-I have no immediate or fully thought out plans for what I want to do about the vector environment wrapper I alluded to.

tristandeleu:
> -The changes to the transform wrappers will meaningfully change the internal function of them.

Sorry I did not understand, can you precise? Do you mean that `TransformObservation` and `TransformReward` currently have limitations that warrant a change of name?

> -I have no immediate or fully thought out plans for what I want to do about the vector environment wrapper I alluded to.

If this is a WIP, I would suggest to not include it in the documentation to avoid confusion.

PR #2325, opened 2021-08-17
Remove Pixel Observation Wrapper

jkterry1:
For all the other wrappers in Gym, specific uses of them exist. I am unaware of any standard use for this wrapper that's in common place, and if some very obscure use cases exist they can use lambda wrappers or a custom version of this wrapper. I'll leave this up for awhile for people to comment in case I'm missing something.


tristandeleu (changes requested):
Pixel observations is a pretty common use-case (e.g. continuous control tasks from pixel representation). This looks like a very useful wrapper to keep.

PR #2328, opened 2021-08-17
Fix flatten utilities for spaces

### LINKED ISSUE #1915: `flatdim` for `MultiDiscrete` is inconsistent
Related: https://github.com/DLR-RM/stable-baselines3/pull/13#issuecomment-628078922

The implementation in https://github.com/openai/gym/pull/1319 does not treat `Discrete` and `MultiDiscrete` spaces the same way.
- for `Discrete`, one hot encoding is assumed:
https://github.com/openai/gym/blob/b2727d6fd396f7fe2b19a541ef3a6ea47559cdb3/gym/spaces/utils.py#L21-L22
- for `MultiDiscrete`, the shape (here number of sub-spaces) is used:
https://github.com/openai/gym/blob/b2727d6fd396f7fe2b19a541ef3a6ea47559cdb3/gym/spaces/utils.py#L29-L30

### LINKED ISSUE #2319: MultiDiscrete flatten_space inconsistencies
I ran into some seemingly glaring inconsistencies with [flattening](https://github.com/openai/gym/blob/bfd4f6d179c90c56356b7cd3bba323fdabd0d2d2/gym/spaces/utils.py#L103) a [MultiDiscrete space](https://github.com/openai/gym/blob/master/gym/spaces/multi_discrete.py)

1. `flatten_space(MultiDiscrete)` does __NOT__ behave the same as `flatten_space(Tuple([Discrete, Discrete]))`. Flattening a `Discrete` space results in a one-hot vector space. In turn, flattening a `Tuple([Discrete, Discrete])` results in a space of concatenated one-hot vectors. Therefore it would be expected that flattening a `MultiDiscrete` space would also result in a concatenated one-hot vectors space. This is __NOT__ the case. Instead `flatten_space(MultiDiscrete)` results in a space with the same dimensions as the original `MultiDiscrete` space with bounds drawn from the ranges of each variable in the original `MultiDiscrete` space. Therefore the flattening of `MultiDiscrete` has _almost_ no functional purpose; but it does seem to introduce a significant bug (see 2) 
2. `flatten_space(MultiDiscrete)` upper bound is invalid and causes invalid sample draws! The upper bound shouldn't be inclusive. This can be easily demonstrated with a simple `MultiDiscrete([2,2])` space:
```
from gym import spaces
a = spaces.MultiDiscrete([2,2])
valid_samples = [[0,0], [1,0], [0,1], [1,1]]

# this works
for i in range(100):
    sample = a.sample()
    assert list(sample) in valid_samples, "{} is not a valid sample of {}".format(flat_sample, valid_samples)

af = spaces.flatten_space(a)
# this will fail because a sample will be drawn with a value of 2 in one of the elements
for i in range(100):
    flat_sample = af.sample()
    assert list(flat_sample) in valid_samples, "{} is not a valid sample of {}".format(flat_sample, valid_samples)
```

See related thread [here](https://github.com/openai/gym/issues/1830), however inconsistency 2 was not identified nor was inconsistency 1 really addressed.

tristandeleu:
- Fix #1915: `flatdim` now returns concatenated one-hot vectors for `MultiDiscrete` spaces
 - Fix #2319: `flatten_space` now returns a `Box` space corresponding to concatenated one-hot vectors for `MultiDiscrete` spaces
 - Minor fix to the tests

### DIFF
--- a/gym/spaces/tests/test_utils.py
+++ b/gym/spaces/tests/test_utils.py
@@ -23,7 +23,7 @@ from gym.spaces import Box, Dict, Discrete, MultiBinary, MultiDiscrete, Tuple, u
         ),
         (Tuple((Discrete(5), Discrete(2), Discrete(2))), 9),
         (MultiDiscrete([2, 2, 100]), 104),
-        (MultiBinary(10), 20),
+        (MultiBinary(10), 10),
         (
             Dict(
                 {
--- a/gym/spaces/tests/test_utils.py
+++ b/gym/spaces/tests/test_utils.py
@@ -6,67 +6,40 @@ import pytest
 from gym.spaces import Box, Dict, Discrete, MultiBinary, MultiDiscrete, Tuple, utils
 
 
-@pytest.mark.parametrize(
-    ["space", "flatdim"],
-    [
-        (Discrete(3), 3),
-        (Box(low=0.0, high=np.inf, shape=(2, 2)), 4),
-        (Tuple([Discrete(5), Discrete(10)]), 15),
-        (
-            Tuple(
-                [
-                    Discrete(5),
-                    Box(low=np.array([0, 0]), high=np.array([1, 5]), dtype=np.float32),
-                ]
+spaces = [
+    Discrete(3),
+    Box(low=0.0, high=np.inf, shape=(2, 2)),
+    Box(low=0.0, high=np.inf, shape=(2, 2), dtype=np.float16),
+    Tuple([Discrete(5), Discrete(10)]),
+    Tuple(
+        [
+            Discrete(5),
+            Box(low=np.array([0, 0]), high=np.array([1, 5]), dtype=np.float32),
+        ]
+    ),
+    Tuple((Discrete(5), Discrete(2), Discrete(2))),
+    MultiDiscrete([2, 2, 100]),
+    MultiBinary(10),
+    Dict(
+        {
+            "position": Discrete(5),
+            "velocity": Box(
+                low=np.array([0, 0]), high=np.array([1, 5]), dtype=np.float32
             ),
-            7,
-        ),
-        (Tuple((Discrete(5), Discrete(2), Discrete(2))), 9),
-        (MultiDiscrete([2, 2, 100]), 104),
-        (MultiBinary(10), 10),
-        (
-            Dict(
-                {
-                    "position": Discrete(5),
-                    "velocity": Box(
-                        low=np.array([0, 0]), high=np.array([1, 5]), dtype=np.float32
-                    ),
-                }
-            ),
-            7,
-        ),
-    ],
-)
+        }
+    ),
+]
+
+flatdims = [3, 4, 4, 15, 7, 9, 104, 10, 7]
+
+
+@pytest.mark.parametrize(["space", "flatdim"], zip(spaces, flatdims))
 def test_flatdim(space, flatdim):
     dim = utils.flatdim(space)
     assert dim == flatdim, "Expected {} to equal {}".format(dim, flatdim)
 
 
-@pytest.mark.parametrize(
-    "space",
-    [
-        Discrete(3),
-        Box(low=0.0, high=np.inf, shape=(2, 2)),
-        Tuple([Discrete(5), Discrete(10)]),
-        Tuple(
-            [
-                Discrete(5),
-                Box(low=np.array([0, 0]), high=np.array([1, 5]), dtype=np.float32),
-            ]
-        ),
-        Tuple((Discrete(5), Discrete(2), Discrete(2))),
-        MultiDiscrete([2, 2, 100]),
-        MultiBinary(10),
-        Dict(
-            {
-                "position": Discrete(5),
-                "velocity": Box(
-                    low=np.array([0, 0]), high=np.array([1, 5]), dtype=np.float32
-                ),
-            }
-        ),
-    ],
-)
+@pytest.mark.parametrize("space", spaces)
 def test_flatten_space_boxes(space):
     flat_space = utils.flatten_space(space)
     assert isinstance(flat_space, Box), "Expected {} to equal {}".format(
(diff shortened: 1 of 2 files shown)


jkterry1:
@araffin @rallen10 does this work for you guys?

rallen10:
@jkterry1 and @tristandeleu
No, unfortunately. It seems to introduce a new behavior for flattening `MultiDiscrete`, but it is not correct. Using the same test as given [here](https://github.com/openai/gym/issues/2319#issue-972037384), when I draw samples from a `flatten_space(MultiDiscrete([2,2])`, I do not get a 4-dim concatenation of one-hot vectors as expected by the comment in this pull request. Instead I get a 2-dim vector of `float32`s. I'm not even sure how to interpret these floating point values with respect the the unflattened `MultiDiscrete` space.

rallen10:
I noted [here](https://github.com/openai/gym/issues/2319) that I _expected_ `MultiDiscrete` to be functionally equivalent to `Tuple(Discrete,Discrete,....)`, however that was just my own expectation (although I think it is a very reasonable one). However, changing the code to match this expectation is almost certainly a major version change to gym since it would break backward compatibility (unless you include some kind of `legacy` flag for `MultiDiscrete`). Perhaps people actually rely on `MultiDiscrete` being functionally different from `Tuple(Discrete,...)`, but this distinction should be well documented/

The true "bug" in my original issue is the fact that `flatten_space(MultiDiscrete)` results in samples out of the bounds of the original, unflat `MultiDiscrete` space.

rallen10:
Also, there seem to be remarkably few unit tests for gym spaces. For example, assuming that `flatten_space(MultiDiscrete)` _should_ result in a concatenation of one-hot vectors, then there should at least be a unit test in `gym/spaces/tests/test_spaces,py` or `gym/spaces/tests/test_utils,py` that checks the shape and dtype of the samples from the flat space. A test like this would have immediately found the error I mention [here](https://github.com/openai/gym/pull/2328#issuecomment-901267895).

Sorry, I'm not trying to be critical. I know the gym support role you have inherited is a major undertaking and I appreciate that you are doing it. I'm just trying to point out any weakness I see so that gym can be made better.

tristandeleu:
> No, unfortunately. It seems to introduce a new behavior for flattening `MultiDiscrete`, but it is not correct. Using the same test as given [here](https://github.com/openai/gym/issues/2319#issue-972037384), when I draw samples from a `flatten_space(MultiDiscrete([2,2])`, I do not get a 4-dim concatenation of one-hot vectors as expected by the comment in this pull request. Instead I get a 2-dim vector of `float32`s. I'm not even sure how to interpret these floating point values with respect the the unflattened `MultiDiscrete` space.

It does introduce a new behavior for flattening `MultiDiscrete`: elements of `MultiDiscrete` get flattened as a concatenation of one-hot vectors, which if I understand correctly is the behavior your are expecting. For the particular example of `MultiDiscrete([2, 2])`, it gets converted to a `Box` of size 4. Here is an example:

```python
space_1 = Tuple((Discrete(2), Discrete(2)))
space_2 = MultiDiscrete([2, 2])

space_1 = Tuple((Discrete(2), Discrete(2)))
space_2 = MultiDiscrete([2, 2])

assert flatten_space(space_1) == flatten_space(space_2)
# Box([0 0 0 0], [1 1 1 1], (4,), int64)
```

And similarly for the `flatten` function

```python
sample_1 = (1, 0)
sample_2 = np.array([1, 0])

assert sample_1 in space_1
assert sample_2 in space_2

print(flatten(space_1, sample_1))  # array([0, 1, 1, 0])
print(flatten(space_2, sample_2))  # array([0, 1, 1, 0])
```

Running the snippet you gave in #2319 does raise an error, but it is an expected error since the flatten space is of size 4, and `[[0, 0], [1, 0], [0, 1], [1, 1]]` are not the valid samples of that space:
```python
AssertionError                            Traceback (most recent call last)
test_issue2319.py in <module>
     12 for i in range(100):
     13     flat_sample = af.sample()
---> 14     assert list(flat_sample) in valid_samples, "{} is not a valid sample of {}".format(flat_sample, valid_samples)

AssertionError: [1 0 1 1] is not a valid sample of [[0, 0], [1, 0], [0, 1], [1, 1]]
```

rallen10:
@tristandeleu aaahhh...I see what was happening. I was getting completely different results as you for `flatten_space` when running the above code snippet. It was because I cloned the `master` branch of your fork and forgot to checkout the `task/fix-flatten` branch before testing. 

Now that I'm testing the correct branch, the behavior seems correct and my simple tests pass. Thanks for the fix!

As a follow on, I agree that the concatenated one-hot vectors are the expected result, but that they also represent a backward-incompatible change to gym. Anyone who currently uses `flatten_space(MultiDiscrete)` in their code will have errors thrown if they update gym after this merge. Since gym is still technically version 0.x.x, I guess backward-incompatible changes are acceptable/expected.

jkterry1:
Cool. I'm going to wait until @araffin is back from vacation to take a look at this, and if he's happy I'll merge this. I'm comfortable with making the breaking change here.

araffin:
Apart from the missing comments in the code, I just tested it and it seems to work as expected ;)

ahalev:
The behavior of `flatten_space` on a Tuple of Discrete spaces does not seem correct to me. If I take a space to be `Tuple((Discrete(5), Discrete(5))`, I'm looking at a space with 25 unique elements. But calling `flatten_space` on this space gives me a `Box(0, 1, (10,))`, which has 2^10 unique actions. This seems incorrect, unless I am misinterpreting the intention of `Tuple`.

RedTachyon:
@ahalev 

So I'm completely on the fence here. But I think there isn't a single correct solution, considering the intention of that function is to convert everything to a Box. If you're converting a Discrete, that's obviously not going to be perfect.

There are more possible elements sure, but not all of them need to be used. This messes with `space.sample`, but otherwise makes sense.

I could be convinced that a tuple of `Discrete`s should be converted into a product-sized Box, i.e. `Box(0, 1, (25,))` in your example, but that would probably be a nightmare to propely handle in arbitrarily nested spaces.

Overall I *think* it's okay to keep it as is. The misinterpretation isn't at the level of `Tuple`, but at the level of `flatten_space`. The intention of this, as far as I understand, is converting any arbitrary space into a constant(?)-sized vector, i.e. a one-dimensional `Box`, because that's what neural networks like to have. The current approach does that, and just that. 

If this means anything to you, the flattening operation is a monomorphism/injective (i.e. no two inputs are mangled together), but it's not an isomorphism/bijective, because that goal is inherently impossible.

Issue #2336, opened 2021-08-20
Fix the dtypes of observation spaces in Classic Control envs


RedTachyon:
**Problem**: On the example of `MountainCarContinuous-v0` because that's where I noticed the problem. The `env.observation_space` claims that observations are of type `float32`, but the actual observations are `float64`, which is a pain to deal with sometimes, but primarily it is simply internally inconsistent. Note that this issue is also present in other environments from this family.

**Replication**: 
```python
import gym
env = gym.make("MountainCarContinuous-v0")

print(f"Observation space: {env.observation_space.dtype}")  # Out: Observation space: float32
print(f"Real observation dtype: {env.reset().dtype}") # Out: Real observation dtype: float64
```

**Solution (proposed)**: I think the best way is to fix it is to actually use `float32` as that's the standard in DL libraries. The easiest way would be adding a `.astype(np.float32)` cast in `.reset()` and `.step()` methods of each environment. It might be possible to only explicitly add it to `.reset()` and then make sure no new float64's appear during a step, but that would require more changes in the code.

I can whip up a pull request with this fix.

jkterry1:
Please create a PR :). Yes, just make it fp32.

Issue #2298, opened 2021-08-05
Why contains function of gym.spaces.Box does not check dtype of the input?


ozgraslan:

jkterry1:
box can take any input dtype...?

tristandeleu:
This is what @ozgraslan probably means:
```python
import numpy as np
from gym import spaces

space = spaces.Box(low=0., high=1., shape=(2,), dtype=np.float32)
sample = np.array([0., 0.5], dtype=np.float64)

assert space.contains(sample)
assert sample in space  # Alternatively
```

I agree `contains` should check for the dtype as well. If I were to guess, this was probably not included to allow lists to be valid elements of the space: since lists are [implicitly converted to `np.ndarray`](https://github.com/openai/gym/blob/ddd650a3fe700dc0c9ebfd1b78027265fb909eac/gym/spaces/box.py#L141-L142), they would automatically be converted to a `np.float64` dtype if it is a list of floats, and the comparison would fail for the default `dtype=np.float32` of `Box`. If dtype is checked for `Box`, it should probably be also checked for `MultiDiscrete` and `MultiBinary`.

ozgraslan:
Exactly! Thank you for the explanation @tristandeleu.

RedTachyon:
Is it enough to simply make a condition in each `contains` of the datatypes being equal? Could be a simple fix. I don't think we'd ever want a different type to be "implicitly" accepted? 

Possibly also worth considering the distinction between `float`, `np.float`, `torch.float64` etc. Should a `np.float32` space accept a python float? Should a `np.float64` do it?

tristandeleu:
> Take this example:
> 
> ```
> space = spaces.Box(low=0., high=10., dtype=np.float32)
> assert space.contains(np.asarray(np.pi, dtype=np.float64)
> ```
> 
> Why is pi no longer in the interval [0, 10], just because I add more digits of precision to it?

Maybe this example would be more convincing for why we need to check for dtype as well:
```python
space = Box(low=0, high=255, shape=(3,), dtype=np.uint8)
assert [8.13, 21.34, 55.89] in space
assert np.array([8.13, 21.34, 55.89], dtype=np.float64) in space
```
[...]

PR #2340, opened 2021-08-21
Fix dtypes to be consistent with observation_space

RedTachyon:
As per Issue #2336, these environments now return float32 observations.

Environments affected: all Classic Control envs, and BipedalWalker (it got caught by tests)

I added tests to check for agreement between the `observation_space.dtype` and `obs.dtype` after a reset and a step, but only for Box spaces. My reasoning is that it covers probably 90% of observation spaces, and checking for (possibly nested?) Tuple/Dict spaces would be a whole bunch of extra logic within tests. This, however, would be automatically checked by `contains` if #2298 is resolved, but that'd be another PR.

Initially I also changed the dtype for Car Racing, but there `observation_space` indicates `uint8`, so while I'm willing to bet that everyone just casts it to a float anyways, it's consistent

### DIFF
--- a/gym/envs/box2d/bipedal_walker.py
+++ b/gym/envs/box2d/bipedal_walker.py
@@ -487,7 +487,7 @@ class BipedalWalker(gym.Env, EzPickle):
             done = True
         if pos[0] > (TERRAIN_LENGTH - TERRAIN_GRASS) * TERRAIN_STEP:
             done = True
-        return np.array(state), reward, done, {}
+        return np.array(state, dtype=np.float32), reward, done, {}
 
     def render(self, mode="human"):
         from gym.envs.classic_control import rendering
--- a/gym/envs/box2d/car_racing.py
+++ b/gym/envs/box2d/car_racing.py
@@ -392,7 +392,7 @@ class CarRacing(gym.Env, EzPickle):
                 done = True
                 step_reward = -100
 
-        return self.state.astype(np.float32), step_reward, done, {}
+        return self.state, step_reward, done, {}
 
     def render(self, mode="human"):
         assert mode in ["human", "state_pixels", "rgb_array"]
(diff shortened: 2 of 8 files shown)


jkterry1:
My vote is to break car racing here for consistency purposes and make it FP32. As I understand there's no compelling reason for the environment to be uint8 the way there is for graphical environments.

tristandeleu:
The observations of `CarRacing-v0` are images, so it makes more sense to keep the dtype of the observation space as a `np.uint8`.

RedTachyon:
While they are images, I find it hard to come up with a practical situation where its useful to have the observation represented as uint8. So I rather agree with @jkterry1, but without a very strong opinion

tristandeleu:
This kind of post-processing should be handled by the user, not the environment. Leaving the raw images allows users to have full control over the post-processing (e.g. data augmentations). At least this should never be a `np.float32` array of values between `0.0` and `255.0` (usually when images are casted to `np.float32`, they are rescaled between `0.0` and `1.0`). If you really want to get rescaled values, you can add a wrapper to the environment.

```python
class FloatObservation(gym.ObservationWrapper):
    def __init__(self, env):
        super().__init__(env)
        self.observation_space = Box(low=0., high=1.,
            shape=env.observation_space.shape, dtype=np.float32)

        return observation.astype(np.float32) / 255.

env = gym.make('CarRacing-v0')
env = FloatObservation(env)

observation = env.reset()
assert np.all(0. <= observation)
assert np.all(observation <= 1.)
assert observation.dtype == np.float32
```

jkterry1:
I agree with tristain on keeping uint8 if the observations are images.

RedTachyon:
Quick note, I actually reintroduced the float32 thing with the last commit with the intention that it'd be reverted before merging if uint8 is decided, but that doesn't seem to be the case at the moment, sorry about that

Issue #2342, opened 2021-08-21
[Proposal] Remove or Move Algorithmic Environments


jkterry1:
### Proposal

While many lesser used environments in Gym are still important in certain contexts (e.g. toy text), as far as I know the algorithmic ones are not. During the process of generating documentation, it's become clear that there's a not of maintenance for these that's just never really happened. My preference would be to remove this entirely or to move them into a different repo with a maintainer if anyone will volunteer as such. If anyone has strong objections to this, please let me know.

jkterry1:
@tristandeleu

jkterry1:
After thinking about this a bit more, my full proposal is to remove them from Gym and offer to link to anyone who wants to maintain these in a separate repo in the 3rd party envs.

tristandeleu:
As I have previously said in #2259, removing features in general does not sound like a good idea, no matter how unlikely you think people use it. What kind of maintenance do these environments would need precisely? They look fairly well tested.

PR #2344, opened 2021-08-21
Remove Algorithmic Environments

jkterry1:
Per https://github.com/openai/gym/issues/2342


jkterry1:
For future travelers:

In the process of documenting the environments in detail, a fairly long list of concerns about the environments were raised. Based on searching through the number of projects on GitHub that weren't just importing all Gym environments or the like, seeing the number of projects that use it on GitHub (you can search all projects), the best thing for us appears to be moving them to a separate repo, like research code would be.

tristandeleu:
> In the process of documenting the environments in detail, a fairly long list of concerns about the environments were raised.

Would you mind sharing details of this "fairly long list of concerns"?

jkterry1:
Thank you for your input Tristan.

@alfred100p knows more about the bugs than me (he found them). miffyli (from SB3) also raised some concerns in discord. I did not fully dive into understanding them in detail for the following reason:

Regardless of the functionality of the environments, they are very rarely used compared to other environments and moving environment logic outside Gym to promote easier maintenance was a core goal of OpenAI's with Gym before the repo was transferred to me. It is difficult to argue that these environments play a "core" role the other types do. This spirit is also why the Gym logic for Atari environments is being moved to the ALE (a process that began years ago), which a PR was just finally opened for. These environments are also not dying, they are going to a separate maintained repo that users will still be able to install them from, that will be linked in the release notes and in 3rd party environments. Currently it's under my username but it'll be fixed and transferred to the new maintainer before I merge this PR.

alfred100p:
The issues I found for Algorithmic Environments are:

- Issue based on inconsistency with gym website of the <a href="https://gym.openai.com/envs/RepeatCopy-v0/">Repeat Copy</a> Task and <a href="https://github.com/openai/gym/blob/master/gym/envs/algorithmic/repeat_copy.py">Implementation</a>- Here there are two issues:
     * based on the existing code the output should be [x1 to xk xk to x1 x1 to xk].
     *  there is an unspecified character m which isn't used in the code, it may have been a parameter used to specify how many times the term x1 to xk xk to x1 should have been repeated, but there is no task mentioned in the paper(<a href="http://arxiv.org/abs/1511.07275">Learning Simple Algorithms from Examples</a>) from which all other tasks were taken from, so its not clear what was intended for m.
- Issue related to ReversedAddition- I propose we can add a new task like ReversedAdditionN instead of just ReversedAddition and ReversedAddition3 as the current implementation can support any number of rows to be added.

jkterry1:
Another reason I support removing these an environment being a default environments in Gym is reflective of things that are good for new researchers in RL to look at and/or are good general benchmarks. Things like mountain car or space invaders or the MuJuCo envs are great for that, these aren't.

PR #2348, opened 2021-08-22
Remove AtariEnv in favour of official ALE Python package

JesseFarebro:
This PR removes `atari-py` in favour of the official ALE package `ale-py`. More information can be found in https://github.com/mgbellemare/Arcade-Learning-Environment/pull/382.

The major changes to Gym will be as follows:

* ROMs are no longer packaged as part of `atari-py`. The current solution is to use `python -m atari_py.import_roms`. The new `ale-py` method of importing ROMs will be the shell command `ale-import-roms`.

* `clone_state()`, `clone_full_state()` will return the actual `ALEState` object now instead of an object meant for serialization. This should have no impact on users as `ALEState` can be Pickled and serialized without issues. 

* It is now recommended to use `gym.make('AtariID', render_mode='human')` instead of `env.render(mode='human')` for both render modes. `ale-py` shipps now with SDL2 and supports full 60FPS viewing, audio, scaling, etc. You don't get this by using `env.render()`

* Environment defaults have changed: `frameskip = 5, repeat_action_probability = 0.25, full_action_space = True`. This follows the most recent methodology set out in Machado et al. 2018. This has no impact on the existing Gym IDs which remain unchanged.

It's worth noting that as far as users are concerned everything is backwards compatible. `ale-py` will still search for ROMs that were imported via `atari-py` and the environment parameters for `v0` and `v4` haven't been changed.


jkterry1:
@JesseFarebro I believe the version of ALE-Py needs to be pegged instead of just specifying a minimum version since this impacts environment logic?

jkterry1:
For additional context, the problem with Atari-Py is that it used an incredibly out of date version of the ALE with many now fixed inconsistencies, bugs and performance issues. This has been a planned upgrade to Gym for something like 2 years now.

setup.py
```diff
@@ -9,7 +9,7 @@
 
 # Environment-specific dependencies.
 extras = {
-    "atari": ["atari-py==0.2.6", "opencv-python>=3."],
+    "atari": ["ale-py~=0.7", "opencv-python>=3."],
```
JesseFarebro, line 12: @jkterry1 made it more strict, there won't be any breaking changes on patch releases.

gym/envs/__init__.py
```diff
@@ -758,7 +757,7 @@ def _merge(a, b):
     "yars_revenge",
     "zaxxon",
 ]:
-    for obs_type in ["image", "ram"]:
+    for obs_type in ["rgb", "ram"]:
```
JesseFarebro, line 760: `ALGymEnv` still accepts `image` as an obs_type but will display a deprecation warning asking the user to switch to one of: `rgb` or `grayscale`.

gym/envs/__init__.py
```diff
@@ -774,9 +773,9 @@ def _merge(a, b):
 
         register(
             id="{}-v0".format(name),
-            entry_point="gym.envs.atari:AtariEnv",
+            entry_point="ale_py.gym:ALGymEnv",
             kwargs={
-                "game": game,
+                "game": name,
```
JesseFarebro, line 778: The environment now takes the game name in camel case instead of the ID in snake case. This kwarg was only used by Gym as `atari-py` was what interpreted the game ID. From what I can tell no one else uses this kwarg.

setup.py
```diff
@@ -9,7 +9,7 @@
 
 # Environment-specific dependencies.
 extras = {
-    "atari": ["atari-py==0.2.6", "opencv-python>=3."],
+    "atari": ["ale-py~=0.7rc", "opencv-python>=3."],
```
JesseFarebro, line 12: Gotta make sure to remove `rc` after final release.

gym/envs/__init__.py
JesseFarebro: This has now been reverted to `game`.

PR #2354, opened 2021-08-23
Fix a RecordVideo bug

vwxyzjn:
Currently, the `RecordVideo`'s `record_video_trigger` won't work in the episodic case. In effect, it will record videos at every episode and your video folder would look like this:
image

This is because `self.start_video_recorder()` is called at every reset. This PR fixes this issue. By running either of the code below, you would get

image

```python
import gym
env = gym.make("CartPole-v1")
env = gym.wrappers.RecordVideo(env, "videos", record_video_trigger=lambda x: x % 100 == 0)
observation = env.reset()
for _ in range(1000):
  env.render()
  action = env.action_space.sample() # your agent here (this takes random actions)

  if done:
    observation = env.reset()
env.close()
```

```python
import gym
from gym.vector import SyncVectorEnv as SyncVectorEnvBase
class SyncVectorEnv(SyncVectorEnvBase):
    def render(self, mode='human', close=False):
        return self.envs[0].render(mode)
envs = SyncVectorEnv([lambda: gym.make("CartPole-v1")])
envs = gym.wrappers.RecordVideo(envs, "videos", record_video_trigger=lambda x: x % 100 == 0)
envs.reset()
for i in range(1000):
    for info in infos:
        if "episode" in info.keys():
            break
```

### DIFF
--- a/gym/wrappers/record_video.py
+++ b/gym/wrappers/record_video.py
@@ -32,7 +32,8 @@ class RecordVideo(gym.Wrapper):
 
     def reset(self, **kwargs):
         observations = super(RecordVideo, self).reset(**kwargs)
-        self.start_video_recorder()
+        if self._video_enabled():
+            self.start_video_recorder()
         return observations
 
     def start_video_recorder(self):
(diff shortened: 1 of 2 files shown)


jkterry1:
Can you add tests surrounding problems like this for the future?

vwxyzjn:
Tests added :)

PR #2365, opened 2021-08-27
Fix `seed` method for `spaces.Tuple` and `spaces.Dict`

### LINKED ISSUE #1864: Seed and sample behaviour in Tuple space
If I create a Tuple space containing Discrete spaces, fix the seed, and then generate a sample, all elements of the sampled tuple are the same.  For example, (0, 0) or (4, 4).  Conversely, if I create a MultiDiscrete space, fix the seed, and then generate a sample, elements of the sampled tuple are not necessarily the same.  For example, (0, 4) or (1, 3).

This is especially problematic when a Tuple space contains Tuple spaces (i.e. where MultiDiscrete is not an adequate replacement):

```python
from gym.spaces import Tuple, MultiDiscrete, Discrete

my_seed = 0

# Test 1

print("Test 1:")

a1 = Tuple([Tuple([Discrete(5) for _ in range(2)]) for _ in range(2)])
b1 = Tuple([MultiDiscrete([5 for _ in range(2)]) for _ in range(2)])

a1.seed(my_seed)
b1.seed(my_seed)

for _ in range(5):
    print(a1.sample(), b1.sample())

# Test 2

print("Test 2:")

a2 = Tuple([Tuple([Discrete(5) for _ in range(2)]) for _ in range(2)])
b2 = Tuple([MultiDiscrete([5 for _ in range(2)]) for _ in range(2)])

for a2_space, b2_space in zip(a2, b2):
    my_seed += 1
    a2_space.seed(my_seed)
    b2_space.seed(my_seed)

for _ in range(5):
    print(a2.sample(), b2.sample())
```
```
Test 1:
((0, 0), (0, 0)) (array([0, 4]), array([0, 4]))
((3, 3), (3, 3)) (array([3, 1]), array([3, 1]))
((4, 4), (4, 4)) (array([0, 3]), array([0, 3]))
((3, 3), (3, 3)) (array([1, 3]), array([1, 3]))
((3, 3), (3, 3)) (array([0, 2]), array([0, 2]))
Test 2:
((0, 0), (4, 4)) (array([4, 2]), array([4, 3]))
((4, 4), (0, 0)) (array([0, 0]), array([3, 3]))
((3, 3), (0, 0)) (array([3, 0]), array([4, 2]))
((2, 2), (4, 4)) (array([0, 2]), array([1, 2]))
((4, 4), (1, 1)) (array([3, 2]), array([2, 2]))
```

Is this expected/desirable behaviour for Tuple or am I using it incorrectly?  I would have expected Tuple to behave for arbitrary tuples as MultiDiscrete behaves for a single tuple.

### LINKED ISSUE #2041: Dict space seeding makes child spaces correlated
```
    def seed(self, seed=None):
        [space.seed(seed) for space in self.spaces.values()]
```
[This line](https://github.com/openai/gym/blob/master/gym/spaces/dict.py#L48) means that all child spaces get the exact same seed. As a consequence, when sampling from the dict space later, the results of those child spaces may seem correlated.

Example:
```
>>> from gym.spaces import Dict, Discrete
>>> d = Dict(a=Discrete(10), b=Discrete(10))
>>> # Initially, the space is seeded with None, which makes the child spaces random.
>>> d.sample()
OrderedDict([('a', 0), ('b', 2)])
>>> d.sample()
OrderedDict([('a', 9), ('b', 1)])
>>> d.seed(42)
>>> # Now, sampling will make the child spaces all produce the same values.
>>> d.sample()
OrderedDict([('a', 9), ('b', 9)])
>>> d.sample()
OrderedDict([('a', 7), ('b', 7)])
>>> d.sample()
OrderedDict([('a', 7), ('b', 7)])
>>> d.sample()
OrderedDict([('a', 2), ('b', 2)])
```

The fix for this is pretty trivial, and I have it written already. Is this a PR that you would accept?

### LINKED ISSUE #2166: sampling from spaces.Tuple yields non-random behavior
I noticed a glitch when setting the seed for a Tuple action space. For example if I:
- create a Tuple with two identical elements (both Boxes)
- set the seed for the Tuple space

then every time I sample, both elements have the same values.
code:
```
from gym import spaces

action_space = spaces.Tuple(
    [spaces.Box(low=0.0, high=1.0, shape=(1,)) for _ in range(2)]
)
action_space.seed(0)
for i in range(4):
    print(action_space.sample())
```
output
```
(array([0.05436005], dtype=float32), array([0.05436005], dtype=float32))
(array([0.9653909], dtype=float32), array([0.9653909], dtype=float32))
(array([0.63269097], dtype=float32), array([0.63269097], dtype=float32))
(array([0.29001734], dtype=float32), array([0.29001734], dtype=float32))
```
It seems to be broadcasting the same seed to every element of the Tuple. (If I don't set the seed I don't get identical values). 

One workaround is to directly set a different seed for each element of the Tuple.
```
for i, space in enumerate(action_space):
    space.seed(0 + i)
```
It seems like this would be desirable as the default behavior.

XuehaiPan:
`Tuple.seed` and `Dict.seed` will return `None` when a list of integers is expected.

Besides, the `seed` function sets the same seed for all subspaces if the argument `seed` is not `None`. This will cause a lack of stochasticity.

```python
>>> t = Tuple((Discete(100), Discete(100), Discete(100)))
>>> print(t.seed(0))
None  # a list of integers is expected

>>> t.sample()
(78, 78, 78)  # always the same
>>> t.sample()
(70, 70, 70)
>>> t.sample()
(59, 59, 59)
```

Closes #2166, Closes  #2041, Closes #1864.


tristandeleu:
There is this issue: #2166

jkterry1:
@tristandeleu Are you happy with this? Also this doesn't actually close any of those issues if I'm reading correctly
@XuehaiPan This needs tests

tristandeleu:
> @tristandeleu Are you happy with this? Also this doesn't actually close any of those issues if I'm reading correctly

You're not reading correctly, this PR should close those three issues. The PR looks good to me.

jkterry1:
@tristandeleu Thanks a ton

@XuehaiPan can you please add CI tests for this to make sure similar regressions don't happen in the future?

XuehaiPan:
I have tried to implement 3 test cases for `seed` method: `test_seed_returns_list`, `test_seed_reproducibility`, `test_seed_subspace_incorrelated`.

PR #2373, opened 2021-08-28
Forward environment properties to the wrapper

### LINKED ISSUE #2175: Base class Wrapper(env) does not allow action_space updates from within original environment
Hi,

I am making a gym wrapper for PyBullet. I extract the action limits from PyBullet to create the `action_space`. However, these limits can only be retrieved from the environment after PyBullet is initialized, which happens after calling `env.reset()`. Therefore, at the moment of calling `gym.make('env-v0')`, the action space is not yet initialized. This should be no problem, as I will set them eventually in the `env.reset()` routine. However, after updating the `self.action_space` in the `env.reset()` routine, I find that `env.action_space=None`. 

I suspect that my issue lies with the `gym.Wrapper(gym.Env)` wrapper.

Because I registered my custom environment `env-v0` with `max_episode_steps=200`, my environment is wrapped in a TimeLimit wrapper, which is a subclass of the `gym.Wrapper(gym.Env)` wrapper (see wrapper code below). This wrapper assigns `self.action_space` in `__init__(self, env)`, which does not make it straightforward to updated `action_space` later on from within the original environment (e.g. by calling `env.reset()`). What is the reason for this implementation?

Why is there no method defined for `action_space` in `gym.Wrapper(gym.Env)`, as given below?
```
    @property
    def action_space(self):
        return self.env.action_space
```

Indeed, when I remove the  `max_episode_steps=200` argument when registering my environment, `self.action_space` is correctly updated after calling `env.reset()` as it is not wrapped in a `gym.Wrapper(gym.Env)` wrapper.

The same problem occurs for `self.observation_space`, `self.reward_range`, and `self.metadata`.
```
class Wrapper(Env):
    """Wraps the environment to allow a modular transformation.

    This class is the base class for all wrappers. The subclass could override
    some methods to change the behavior of the original environment without touching the
    original code.

    .. note::

        Don't forget to call ``super().__init__(env)`` if the subclass overrides :meth:`__init__`.

    """
    def __init__(self, env):
        self.env = env
        self.action_space = self.env.action_space
        self.observation_space = self.env.observation_space

    def __getattr__(self, name):
        if name.startswith('_'):

    @property
    def spec(self):
        return self.env.spec

    @classmethod
    def class_name(cls):
        return cls.__name__

    def step(self, action):
        return self.env.step(action)

    def reset(self, **kwargs):

    def render(self, mode='human', **kwargs):
        return self.env.render(mode, **kwargs)

    def close(self):
        return self.env.close()

    def seed(self, seed=None):
        return self.env.seed(seed)

    def compute_reward(self, achieved_goal, desired_goal, info):
        return self.env.compute_reward(achieved_goal, desired_goal, info)

    def __str__(self):
        return '<{}{}>'.format(type(self).__name__, self.env)

    def __repr__(self):
        return str(self)

    @property
    def unwrapped(self):
        return self.env.unwrapped
```

tristandeleu:
- Do not set the wrapper properties in `__init__`, but forward the environment properties by default.
 - Fixes #2175 

```python
class CustomEnv(gym.Env):
    observation_space = None
    action_space = None

    def reset(self):
        self.observation_space = gym.spaces.Box(low=0., high=1., shape=(2,))
        self.action_space = gym.spaces.Discrete(2)
        return self.observation_space.sample()

    def step(self, action):
        return (self.observation_space.sample(), 0., False, {})

class CustomWrapper(gym.Wrapper):
    pass

env = CustomWrapper(CustomEnv())
env.reset()
print(f'Action space: {env.action_space}')
# Before: Action space: None
# After: Action space: Discrete(2)
```

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -224,10 +224,11 @@ class Wrapper(Env):
 
     def __init__(self, env):
         self.env = env
-        self.action_space = self.env.action_space
-        self.observation_space = self.env.observation_space
-        self.reward_range = self.env.reward_range
-        self.metadata = self.env.metadata
+
+        self._action_space = None
+        self._observation_space = None
+        self._reward_range = None
+        self._metadata = None
 
     def __getattr__(self, name):
         if name.startswith("_"):
@@ -243,6 +244,46 @@ class Wrapper(Env):
     @classmethod
     def class_name(cls):
         return cls.__name__
+
+    @property
+    def action_space(self):
+        if self._action_space is None:
+            return self.env.action_space
+        return self._action_space
+
+    @action_space.setter
+    def action_space(self, space):
+        self._action_space = space
+
+    @property
+    def observation_space(self):
+        if self._observation_space is None:
+            return self.env.observation_space
+        return self._observation_space
+
+    @observation_space.setter
+    def observation_space(self, space):
+        self._observation_space = space
+
+    @property
+    def reward_range(self):
+        if self._reward_range is None:
+            return self.env.reward_range
+        return self._reward_range
+
+    @reward_range.setter
+    def reward_range(self, value):
+        self._reward_range = value
+
+    @property
+    def metadata(self):
+        if self._metadata is None:
+            return self.env.metadata
+        return self._metadata
+
+    @metadata.setter
+    def metadata(self, value):
+        self._metadata = value
 
     def step(self, action):
         return self.env.step(action)
--- a/gym/tests/test_core.py
+++ b/gym/tests/test_core.py
@@ -90,10 +90,10 @@ properties = [
 ]
 
 
-@pytest.mark.parametrize("klass", [UnittestEnv, UnknownSpacesEnv])
+@pytest.mark.parametrize("class_", [UnittestEnv, UnknownSpacesEnv])
 @pytest.mark.parametrize("props", properties)
-def test_wrapper_property_forwarding(klass, props):
-    env = klass()
+def test_wrapper_property_forwarding(class_, props):
+    env = class_()
     env = NewPropertyWrapper(env, **props)
 
     # If UnknownSpacesEnv, then call reset to define the spaces
--- a/gym/tests/test_core.py
+++ b/gym/tests/test_core.py
@@ -1,4 +1,7 @@
-from gym import core
+import pytest
+import numpy as np
+
+from gym import core, spaces
 
 
 class ArgumentEnv(core.Env):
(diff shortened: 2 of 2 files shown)


jkterry1:
@tristandeleu can you add tests? Otherwise I believe this looks good.

tristandeleu:
What is the status of this PR?

gym/core.py
```diff
@@ -244,6 +245,46 @@ def spec(self):
...
         return cls.__name__
 
+    @property
+    def action_space(self):
+        if self._action_space is None:
+            return self.env.action_space
+        return self._action_space
+
+    @action_space.setter
+    def action_space(self, space):
```
RedTachyon, line 255: I'm assuming it's intended that the setter only affects the wrapper, and does not get propagated to the inner env. I'm not sure if this makes any difference in the end, but I'm also not fully immediately convinced which option is better? Is there any use case in which either option is better?

gym/tests/test_core.py
```diff
@@ -9,9 +12,99 @@ def __init__(self, arg):
...
+    {
+        "observation_space": spaces.Box(
+            low=0.0, high=1.0, shape=(64, 64, 3), dtype=np.float32
+        ),
+        "action_space": spaces.Discrete(2),
+    },
+]
+
+
+@pytest.mark.parametrize("klass", [UnittestEnv, UnknownSpacesEnv])
```
RedTachyon, line 93: Minor nitpick - is `klass` often used instead of `class`? I'd intuitively go rather for `cls`

RedTachyon (approved):
Overall this looks solid and solves what it's supposed to, just one minor doubt about a variable name (see review)

One thing I'm thinking about now is whether it would be useful for wrappers to forward all properties' getters by default (which would also solve this issue as a special case). This way if someone has e.g. `env.current_state` defined, it would be forwarded through all the wrappers. But it'd probably require too much effort, with the only benefit being sometimes not having to write `.unwrapped`

gym/tests/test_core.py
tristandeleu: By convention, [`cls` is reserved for classmethods](https://www.python.org/dev/peps/pep-0008/#function-and-method-arguments). PEP8 suggests using `class_` instead, but `klass` is also a fairly common name, even [CPython uses it](https://github.com/python/cpython/search?q=klass).

gym/core.py
tristandeleu: That's already the case: #1347

```python
class CustomEnv(gym.Env):
    g = 9.81

class CustomWrapper(gym.Wrapper):
    pass

env = CustomWrapper(CustomEnv())
print(env.g)  # 9.81
env.g = 1.62
print(env.g)  # 1.62
```

The reason why we can't just rely on `__getattr___` for `observation_space`, `action_space`, `reward_range` and `metadata` is that since [`Wrapper` inherits from `Env`](https://github.com/openai/gym/blob/5a4709b158009ba8786c5df2490d8a601bc80ce4/gym/core.py#L212), [those properties will get set](https://github.com/openai/gym/blob/5a4709b158009ba8786c5df2490d8a601bc80ce4/gym/core.py#L32-L39) in every wrapper, and so forwarding will not work properly: for example, if you don't set `observation_space` in your wrapper, because you'd like it to be the same as the wrapped environment, then you'd always get `wrapped_env.observation_space == None`.

tristandeleu:
> One thing I'm thinking about now is whether it would be useful for wrappers to forward all properties' getters by default (which would also solve this issue as a special case). This way if someone has e.g. `env.current_state` defined, it would be forwarded through all the wrappers. But it'd probably require too much effort, with the only benefit being sometimes not having to write `.unwrapped`

That's already the case: #1347

```python
class CustomEnv(gym.Env):
    g = 9.81

class CustomWrapper(gym.Wrapper):
    pass

env = CustomWrapper(CustomEnv())
print(env.g)  # 9.81
env.g = 1.62
print(env.g)  # 1.62
```

The reason why we can't just rely on `__getattr___` for `observation_space`, `action_space`, `reward_range` and `metadata` is that since [`Wrapper` inherits from `Env`](https://github.com/openai/gym/blob/5a4709b158009ba8786c5df2490d8a601bc80ce4/gym/core.py#L212), [those properties will get set](https://github.com/openai/gym/blob/5a4709b158009ba8786c5df2490d8a601bc80ce4/gym/core.py#L32-L39) in every wrapper, and so forwarding will not work properly: for example, if you don't set `observation_space` in your wrapper, because you'd like it to be the same as the wrapped environment, then you'd always get `wrapped_env.observation_space == None`.

gym/core.py
tristandeleu: I don't think mutating `observation_space`/`action_space` in the inner envs is a good idea. There may be cases where you'd like to have access to the space of the original environment (e.g. through `env.unwrapped.action_space`). If the setter was propagated to the inner env, then you'd always have `env.action_space == env.unwrapped.action_space`.

Issue #2357, opened 2021-08-26
Box.contains fails on tuple input


FirefoxMetzger:
```python
from gym.spaces import Box
import numpy as np

space = Box(low=np.array((-1, -1, -1)), high=np.array((1, 1, 1)))
space.contains((0,0,0))

```

Produces the following stack-trace

```
Traceback (most recent call last):
  File "box_bug.py", line 5, in <module>
    space.contains((0,0,0))
  File "/home/firefoxmetzger/.local/lib/python3.8/site-packages/gym/spaces/box.py", line 144, in contains
    x.shape == self.shape and np.all(x >= self.low) and np.all(x <= self.high)
AttributeError: 'tuple' object has no attribute 'shape'
```

A quick look at the code reveals that only lists are promoted to arrays.

https://github.com/openai/gym/blob/edc233288e4336825540ca2c0dc2e6b8a9291dfc/gym/spaces/box.py#L140-L145

I would lobby for using a more generic call to `np.asarray` to convert anything ArrayLike; perhaps combined with a type hint update `Any -> np.typing.ArrayLike`:

```python
def contains(self, x):
    x = np.asarray(x)

    if x.shape != self.shape:
        return False

    if np.any( (x<self.low) | (x>self.high) ):
        return False

    return True
```

tristandeleu:
> @tristandeleu Oh? Am I missing something? What situation do you have in mind where this conversion would be detrimental?

Not detrimental, but less intuitive from a UX perspective because this conversion happens internally to Gym, and is not something that the user requested. But really at the end of the day, I am not strongly against the internal conversion to Numpy arrays.

FirefoxMetzger:
> this conversion happens internally to Gym, and is not something that the user requested.

@tristandeleu I'm confused. `np.asarray` would create a copy of its input if it isn't already a ndarray (with correct type and order), which will then be handed to gc once the function returns, so the user would keep the original object that was passed into `space.contains` and (performance aside) wouldn't even notice that gym creates a temporary array internally in some cases. I can see how this can be troublesome if `x` was a tensor located on the GPU since this can have significant performance implications. Then again `space.contains` doesn't seem like a function one would expect to see in the critical loop of HPC code, so I'm not sure if my argument here counts.

Which part is unintuitive from your perspective?

tristandeleu:
[...]

Issue #2369, opened 2021-08-27
[Proposal] Moving uncommonly used toy text environments from Gym into different repo


jkterry1:
Per the rational in https://github.com/openai/gym/pull/2363, I'd like to not have environments that aren't widely used and don't represent good baseline challenges in the RL community in Gym by default. In toy text, there is a mix. E.g. frozen lake and taxi are clearly very widely used in classes and otherwise, and blackjack is important to include due to its role in barto and sutton. This leaves guessing game, hotter colder, nchain and roulette. They are by far the least used environments in Gym according to gitHub statistics, and I am unaware of a compelling reason for them to exist the way that blackjack has. If a compelling reason for keeping any of those four does exist that I'm just missing, I would be happy to keep them. If not, I'm planning to have them moved to a new repo (maybe gym-legacy-toytext).

jkterry1:
@Rohan138 the kellycoin env presumably should be removed as well

Issue #2275, opened 2021-07-30
Inclusion of Changelog


jkterry1:
In #2259, @araffin suggested including release notes in a changelog instead of just in github releases like SB3. This has two ostensible benefits:

-It can make creating full GitHub release notes earlier since they release exist. On a personal level for other libraries I've maintained though, I've found dealing with this in PRs more annoying than just going through commits, and ultimately this matter only impacts me personally.
-It helps people based off master, which is a fair point. In libraries like SB3, there's a lot of people using master for good reason. I don't believe this is something that would ever be done under normal circumstances for Gym though? If people disagree with me on this I can go ahead and include a proper changelog.

araffin:
>In #2259, @araffin suggested including release notes in a changelog instead of just in github releases like SB3. This has two ostensible benefits:

As an additional remark, just putting release notes in Github releases is fine as long as the releases don't take took long to be released (something we do in SB3, where each release may take some months before we consider it reliable enough) and that nothing is forgotten.

PR #2374, opened 2021-08-28
box.contains check dtype and promote non-ndarrays

FirefoxMetzger:
Closes: https://github.com/openai/gym/issues/2357 and #2298

Instead of only casting list to ndarray, cast any class to ndarray (if possible) and emit a warning when casting. Also, check if the dtype of the input matches the dtype of the space.

### DIFF
--- a/gym/spaces/box.py
+++ b/gym/spaces/box.py
@@ -140,14 +140,14 @@ class Box(Space):
 
     def contains(self, x):
         if not isinstance(x, np.ndarray):
-            warnings.warn('Casting input x to numpy array.')
+            warnings.warn("Casting input x to numpy array.")
             x = np.asarray(x, dtype=self.dtype)
-        
+
         return (
-            np.can_cast(x.dtype, self.dtype) and
-            x.shape == self.shape and
-            np.any(x >= self.low) and
-            np.any(x <= self.high)
+            np.can_cast(x.dtype, self.dtype)
+            and x.shape == self.shape
+            and np.any(x >= self.low)
+            and np.any(x <= self.high)
         )
 
     def to_jsonable(self, sample_n):
--- a/gym/spaces/box.py
+++ b/gym/spaces/box.py
@@ -1,5 +1,5 @@
 import numpy as np
-from warnings import warn
+import warnings
 
 from .space import Space
 from gym import logger
--- a/gym/spaces/box.py
+++ b/gym/spaces/box.py
@@ -1,4 +1,5 @@
 import numpy as np
+from warnings import warn
 
 from .space import Space
 from gym import logger
@@ -138,10 +139,15 @@ class Box(Space):
         return sample.astype(self.dtype)
 
     def contains(self, x):
-        if isinstance(x, list):
-            x = np.array(x)  # Promote list to array for contains check
+        if not isinstance(x, np.ndarray):
+            warnings.warn('Casting input x to numpy array.')
+            x = np.asarray(x, dtype=self.dtype)
+        
         return (
-            x.shape == self.shape and np.all(x >= self.low) and np.all(x <= self.high)
+            np.can_cast(x.dtype, self.dtype) and
+            x.shape == self.shape and
+            np.any(x >= self.low) and
+            np.any(x <= self.high)
         )
 
     def to_jsonable(self, sample_n):
--- a/gym/wrappers/test_flatten_observation.py
+++ b/gym/wrappers/test_flatten_observation.py
@@ -19,7 +19,7 @@ def test_flatten_observation(env_id):
         space = spaces.Tuple(
             (spaces.Discrete(32), spaces.Discrete(11), spaces.Discrete(2))
         )
-        wrapped_space = spaces.Box(-np.inf, np.inf, [32 + 11 + 2], dtype=np.float64)
+        wrapped_space = spaces.Box(0, 1, [32 + 11 + 2], dtype=np.int64)
     elif env_id == "KellyCoinflip-v0":
         space = spaces.Tuple(
             (spaces.Box(0, 250.0, [1], dtype=np.float32), spaces.Discrete(300 + 1))
--- a/gym/wrappers/test_flatten_observation.py
+++ b/gym/wrappers/test_flatten_observation.py
@@ -25,7 +25,7 @@ def test_flatten_observation(env_id):
             (spaces.Box(0, 250.0, [1], dtype=np.float32), spaces.Discrete(300 + 1))
         )
         low = np.zeros((302,), dtype=np.float64)
-        high = np.array([250.] + [1.] * 301, dtype=np.float64)
+        high = np.array([250.0] + [1.0] * 301, dtype=np.float64)
         wrapped_space = spaces.Box(low, high, [1 + (300 + 1)], dtype=np.float64)
 
     assert space.contains(obs)
--- a/gym/wrappers/test_flatten_observation.py
+++ b/gym/wrappers/test_flatten_observation.py
@@ -19,12 +19,12 @@ def test_flatten_observation(env_id):
         space = spaces.Tuple(
             (spaces.Discrete(32), spaces.Discrete(11), spaces.Discrete(2))
         )
-        wrapped_space = spaces.Box(-np.inf, np.inf, [32 + 11 + 2], dtype=np.float32)
+        wrapped_space = spaces.Box(-np.inf, np.inf, [32 + 11 + 2], dtype=np.float64)
     elif env_id == "KellyCoinflip-v0":
         space = spaces.Tuple(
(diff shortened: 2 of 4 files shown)


gym/spaces/box.py
```diff
@@ -1,4 +1,5 @@
 import numpy as np
+from warnings import warn
```
FirefoxMetzger, line 2:
```suggestion
import warnings
```

It is better to match the style of the rest of the code-base.

FirefoxMetzger:
@tristandeleu Looks like gym's tests depend on Box.contains old behavior of not checking types (https://github.com/openai/gym/pull/2374/checks?check_run_id=3450441531#step:4:252). Potentially downstream code will do this, too, making the current PR a breaking change. What is gym's approach to such scenarios? Change the function to make the dtype check optional and backward-compatible (i.e. add a `enforce_dtype_match` kwarg or similar) or modify the existing tests since this is a bugfix?

@jkterry1 Since the current version of the PR would contain a breaking change (due to the additional check suggested by #2298 ) do you have an opinion/preference on this matter?

tristandeleu:
There are two issues coming out of those tests:
 - `KellyCoinflipEnv` is returning an observation without specifying the dtype, so it falls back to `np.float64` by default even though the `observation_space` has a dtype of `np.float32`. I think this fix would be reasonable in that case (and it might not even matter too much since this environment might get moved out of Gym #2369):
<details>

```diff
diff --git a/gym/envs/toy_text/kellycoinflip.py b/gym/envs/toy_text/kellycoinflip.py
index 1a47c0a..4f305ab 100644
--- a/gym/envs/toy_text/kellycoinflip.py
+++ b/gym/envs/toy_text/kellycoinflip.py
@@ -79,7 +79,7 @@ class KellyCoinflipEnv(gym.Env):
         return self._get_obs(), reward, done, {}

     def _get_obs(self):
-        return np.array([self.wealth]), self.rounds
+        return np.array([self.wealth], dtype=np.float32), self.rounds

     def reset(self):
         self.rounds = self.max_rounds
@@ -236,11 +236,11 @@ class KellyCoinflipGeneralizedEnv(gym.Env):

     def _get_obs(self):
         return (
-            np.array([float(self.wealth)]),
+            np.array([float(self.wealth)], dtype=np.float32),
             self.rounds_elapsed,
             self.wins,
             self.losses,
-            np.array([float(self.max_ever_wealth)]),
+            np.array([float(self.max_ever_wealth)], dtype=np.float32),
         )

     def reset(self):
```
</details>

 - The tests for the `FlattenObservation` wrapper are failing because of an invalid dtype for the space. This is something I missed in #2328. Flattening a `Tuple` of `Discrete` spaces leads to a `Box` spaces corresponding to multi-hot encoding (with dtype `np.int64`), and flattening a `Tuple` of `Box` and `Discrete` has to have a dtype of `np.float64` ([dtype resolved with `np.result_type`](https://github.com/openai/gym/blob/150404104afc57ea618acca34d2d1f797d779869/gym/spaces/utils.py#L213)). I would suggest to fix those tests to reflect that (this is not a problem with this PR, on the contrary it highlighted an incorrect test!).
<details>

```diff
diff --git a/gym/wrappers/test_flatten_observation.py b/gym/wrappers/test_flatten_observation.py
index f190081..2db176c 100644
--- a/gym/wrappers/test_flatten_observation.py
+++ b/gym/wrappers/test_flatten_observation.py
@@ -19,12 +19,12 @@ def test_flatten_observation(env_id):
         space = spaces.Tuple(
             (spaces.Discrete(32), spaces.Discrete(11), spaces.Discrete(2))
         )
-        wrapped_space = spaces.Box(-np.inf, np.inf, [32 + 11 + 2], dtype=np.float32)
+        wrapped_space = spaces.Box(0, 1, [32 + 11 + 2], dtype=np.int64)
     elif env_id == "KellyCoinflip-v0":
         space = spaces.Tuple(
             (spaces.Box(0, 250.0, [1], dtype=np.float32), spaces.Discrete(300 + 1))
         )
-        wrapped_space = spaces.Box(-np.inf, np.inf, [1 + (300 + 1)], dtype=np.float32)
+        wrapped_space = spaces.Box(-np.inf, np.inf, [1 + (300 + 1)], dtype=np.float64)

     assert space.contains(obs)
     assert wrapped_space.contains(wrapped_obs)
```
</details>

FirefoxMetzger:
@tristandeleu I updated the tests; are we worried about informing downstream about this change? I.e. should we bump a version, update a changelog, or leave a comment somewhere saying "Breaking change: Box.contains now also matches dtype when checking for membership" or something?

gym/wrappers/test_flatten_observation.py
```diff
@@ -19,12 +19,12 @@ def test_flatten_observation(env_id):
         space = spaces.Tuple(
             (spaces.Discrete(32), spaces.Discrete(11), spaces.Discrete(2))
         )
-        wrapped_space = spaces.Box(-np.inf, np.inf, [32 + 11 + 2], dtype=np.float32)
+        wrapped_space = spaces.Box(-np.inf, np.inf, [32 + 11 + 2], dtype=np.float64)
```
tristandeleu, line 22:
```suggestion
        wrapped_space = spaces.Box(0, 1, [32 + 11 + 2], dtype=np.int64)
```

This doesn't have too much impact on the test, but this should be `dtype=np.int64` (the flattened space takes the dtype of `Discrete`, [which is `np.int64`](https://github.com/openai/gym/blob/150404104afc57ea618acca34d2d1f797d779869/gym/spaces/discrete.py#L17)). Also the bounds should be 0 and 1 ([according to the rules for flattening the space](https://github.com/openai/gym/blob/150404104afc57ea618acca34d2d1f797d779869/gym/spaces/utils.py#L200-L214), since flattening a `Tuple` of `Dsicrete` spaces leads to multi-hot vectors).

gym/wrappers/test_flatten_observation.py
```diff
@@ -19,12 +19,12 @@ def test_flatten_observation(env_id):
...
     elif env_id == "KellyCoinflip-v0":
         space = spaces.Tuple(
             (spaces.Box(0, 250.0, [1], dtype=np.float32), spaces.Discrete(300 + 1))
         )
-        wrapped_space = spaces.Box(-np.inf, np.inf, [1 + (300 + 1)], dtype=np.float32)
+        wrapped_space = spaces.Box(-np.inf, np.inf, [1 + (300 + 1)], dtype=np.float64)
```
tristandeleu, line 27:
```suggestion
        low = np.zeros((302,), dtype=np.float64)
        high = np.array([250.] + [1.] * 301, dtype=np.float64)
        wrapped_space = spaces.Box(low, high, [1 + (300 + 1)], dtype=np.float64)
```

The dtype is correct, but the bounds should be `np.zeros((302,), dtype=np.float64)` and `np.array([250.] + [1.] * 301, dtype=np.float64)`.

tristandeleu:
@FirefoxMetzger Yes this will probably require bumping the version, to avoid any surprises. I don't know what is the status on the changelog though (there was a discussion here #2275, but I don't know what was the outcome), but it would definitely be a good idea to inform users of this change through the changelog.

jkterry1:
This was stated elsewhere, but the changelogs are going in GitHub release notes, e.g. here is the changelog for the last release: https://github.com/openai/gym/releases/tag/0.19.0

jkterry1:
@FirefoxMetzger this needs tests

FirefoxMetzger:
@jkterry1 I've added a unit test to avoid regression regarding the two issues addressed by this PR. Also note that this is also covered by existing tests (the two that were adjusted as part of this PR).

jkterry1:
Just to confirm, the only version bump that would be required would be for KellyCoinflip right?

FirefoxMetzger:
@jkterry1 Good catch with the environment version. I'm not an active user of KellyCoinflip, so I can't estimate if the reduction in precision (float64 -> float32) will affect performance or existing algorithms. If so, then the version of the environment should indeed be bumped.

The version bump I was talking about with @tristandeleu was related to the overall gym version. The function `Box.contains` behaves differently/more strictly now (it now checks if the provided object has the same dtype as the space it represents). As such, it may break existing code outside of gym that (explicitly or implicitly) relies on `Box.contains` not enforcing the dtype. Hence the question about bumping versions or how to communicate this best.

rohanb2018:
I have sort of an unrelated comment, but I noticed in `Box.contains` that the `np.all` clauses were changed to `np.any` clauses. Was there a particular reason for this change? I feel like `np.all` is the correct logic for `Box.contains`, because we need all of the coordinates of the test point `x` to satisfy the lower/upper bounds, not just any of them.

FirefoxMetzger:
@rohanb2018 Good catch; this is not how it should be. It changed from `np.all(x >= self.low)` to `not np.any(x < self.low)` and then back to `np.any(x >= self.low)` which is of course not correct. I'll flip the signs in a new PR later today.

PR #2396, opened 2021-09-06
Pass **kwargs in step() of env wrappers

johannespitz:
I find myself inserting this key-word passing all the time into the step function of different gym.Wrapper's.
One particular use case: When using the FrameStack Wrapper for observation stacking on a custom PyBullet or MuJoCo env, it is often useful to pass arguments in the step function that enable specific logging code that should otherwise not be executed because it is very expensive (i.e. logging each simulation step, logging contact information...).

### DIFF
--- a/gym/wrappers/atari_preprocessing.py
+++ b/gym/wrappers/atari_preprocessing.py
@@ -100,11 +100,11 @@ class AtariPreprocessing(gym.Wrapper):
             low=_low, high=_high, shape=_shape, dtype=_obs_dtype
         )
 
-    def step(self, action):
+    def step(self, action, **kwargs):
         R = 0.0
 
         for t in range(self.frame_skip):
-            _, reward, done, info = self.env.step(action)
+            _, reward, done, info = self.env.step(action, **kwargs)
             R += reward
             self.game_over = done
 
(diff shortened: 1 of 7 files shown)


jkterry1:
I'm closing this for now per the discussion in #2399. One work around for this within the confines of the API would be dictionary action spaces where such controls are treated as actions.

johannespitz:
Since you now closed the discussion I was wondering if you would consider to merge or open this PR again. 
I don't see any reason not to pass key word arguments in a wrapper class. And most importantly the current implementation (i.e. passing them in `reset` but not in `step`) is inconsistent because the base class does not accept them in either function (as discussed in #2399).

johannespitz:
@jkterry1 
Hi, I understand that you don't want to change the signature of the core class but as I wrote above that shouldn't be a reason for not merging this simple modification here.
I would really appreciate if you could reopen the PR or alternatively explain why it is a bad idea.
Best, Johannes.

jkterry1:
Could you please explain the inconsistency again?

johannespitz:
Thanks for the reply.
As far as I can see all `reset` functions pass the key word arguments but the `step` functions do not.
That's inconsistent because the core `env` does not have key word arguments in its signature for neither of them.
Merging this PR would fix the inconsistency. 
(The other alternative, of course, would be to remove the key word arguments from `reset` but that would certainly break something.)

Issue #1880, opened 2020-04-14
gym.spaces.Box allows arbitrary shape assignment, but the true value of shape does not change.


AnanthHari:
**Description:** After creation of a Box space, the shape variable _can_ be assigned a new value. For instance,
```python
>>> from gym.spaces import Box
>>> a = Box(low=0, high=1, shape=(2,3))
>>> a; a.low
Box(2, 3)
array([[0., 0., 0.],
       [0., 0., 0.]], dtype=float32)
>>> a.shape = (4,5)
>>> a; a.low
Box(4, 5)
array([[0., 0., 0.],
       [0., 0., 0.]], dtype=float32)
```

We can see that the representation of the Box object acknowledges the change in shape, but `box.low` is unchanged. The actual behavior is that the shape of the Box object remains unchanged even after simply assigning new value to `shape` attribute. (Also, it is now in an awkward state as both `a.contains(np.zeros((2,3)))` and `a.contains(np.zeros((4,5)))` do not return `True`).

This creates confusion to new users who might think that the change in `shape` attribute changes the shape of the underlying `Box` object. The misleading notion that changing the `shape` attribute changes the `Box`'s shape also comes from the fact that `low` and `high` input parameters might be scalars, and they could potentially be arranged in different shapes by assigning different `box.shape` values.

**Solution:** I propose the following solution. Store the shape of Box in an internal variable named `__shape__`, and use `shape` as a method, instead of an attribute. Use `@property ` decorator to use `box.shape` instead of `box.shape()` like this:
```python
@property
def shape(self):
    return self.__shape__
```
This ensure backward compatibility. Now, trying to assign a new shape value to `shape` attribute of Box fails, because `shape` is not a variable.
```python
>>> a = Box(low=0, high=1, shape=(2,3))
>>> a.shape = (5,1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: can't set attribute
```

We also need to patch up the base class in `space.py`. Change 
```python
self.shape = None if shape is None else tuple(shape)
```
as 
```python
if not hasattr(self, "shape") or self.shape is None:
    self.shape = None if shape is None else tuple(shape)
```
I can make a pull request if the bug/feature description warrants changing the current structure of `Box` space.

jkterry1:
@AnanthHari I'm the maintainer of Gym now; mind making that PR?

RedTachyon:
[...]

PR #2397, opened 2021-09-06
Changed .shape to be a property

RedTachyon:
In relation to issue #1880, it's similar (but not exactly the same) as what the OP there suggested. Basically the shape is now stored as `self._shape`, and has a `@property`-based getter. There is no setter, as changing the shape might also require changing `low` and `high`, and at that point it's probably better to just create a new object.

It shouldn't change the behavior at all, except that there's an exception thrown if someone tries to manually change `self.shape`. 

I opted for a `_name` naming because dunder methods (as suggested by OP) are typically reserved for other stuff.

It'd be good if there's another set of eyes to see if it for sure doesn't break anything. It's simple enough, and the tests (should) pass, but

### DIFF
--- a/gym/spaces/box.py
+++ b/gym/spaces/box.py
@@ -57,7 +57,7 @@ class Box(Space):
         if np.isscalar(high):
             high = np.full(shape, high, dtype=dtype)
 
-        self.shape = shape
+        self._shape = shape
         self.low = low
         self.high = high
 
--- a/gym/spaces/space.py
+++ b/gym/spaces/space.py
@@ -19,7 +19,7 @@ class Space(object):
     def __init__(self, shape=None, dtype=None):
         import numpy as np  # takes about 300-400ms to import, so we load lazily
 
-        self.shape = None if shape is None else tuple(shape)
+        self._shape = None if shape is None else tuple(shape)
         self.dtype = None if dtype is None else np.dtype(dtype)
         self._np_random = None
 
@@ -32,6 +32,11 @@ class Space(object):
             self.seed()
 
         return self._np_random
+
+    @property
+    def shape(self):
+        """Return the shape of the space as a nonchangeable property"""
+        return self._shape
 
     def sample(self):
         """Randomly sample an element of this space. Can be
--- a/gym/spaces/space.py
+++ b/gym/spaces/space.py
@@ -35,7 +35,7 @@ class Space(object):
 
     @property
     def shape(self):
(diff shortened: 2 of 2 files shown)


RedTachyon:
I'm slightly confused by the tests failing, it worked on my fork, and the error seems to happen in some seemingly unrelated video recording tests.

jkterry1:
Yep, @vwxyzjn needs to fix a flakey test he accidentally added

jkterry1:
I'm probably missing something, but how does your PR resolve this from Ananth's issue?

```
We also need to patch up the base class in space.py. Change

self.shape = None if shape is None else tuple(shape)
as

if not hasattr(self, "shape") or self.shape is None:
    self.shape = None if shape is None else tuple(shape)
```

RedTachyon:
So what Ananth originally proposed was changing `shape` to become a `@property` only for `Box` spaces. In that case, the mentioned condition is necessary to check if the concrete space (e.g. Box) has that property, and otherwise, it will be stored as usual. 

So in particular, with that approach the behavior of a Discrete space doesn't change, i.e. you have the same problem there (you can do `space.shape = (5,)`, not get an error, but not get any expected effect)

Instead, I handled it for all spaces by turning `shape` into a `@property` in the `Space` class. 

As far as I can tell, the only differences this makes are:
- The problem is removed from all spaces instead of just one (although arguably, in the other ones it's less important)
- If someone made a custom space where they manually do `self.shape = ...`, it would need to be changed. Note that this would be pretty bad practice, since the shape is handled by `Space` itself via inheritance.

My reasons:
- While Box uses `shape` the most, it is also defined for all other spaces, so the behavior should stay as consistent as possible between them
- It's (imo) cleaner than checking whether the subclass already defined `shape` before calling `super().__init__` (and it'd get messy if the subclass first calls that, and then wants to define the shape by itself)

---

tl;dr Ananth's version was specific, and needed a general guard to make sure it doesn't break/affect other spaces. My version is general, so other spaces are automatically affected (fixed?), so that guard is not necessary.

ahmedo42:
@RedTachyon I think the tests will pass once #2394 is merged , test is fixed there

araffin:
This PR (and gym 0.20 release) breaks all SB3 pre-trained models :confused: :face_with_head_bandage:  (and all pickled Spaces), to be backward compatible you have to patch `__setstate__`:
https://github.com/DLR-RM/stable-baselines3/issues/573#issuecomment-921627566

Issue #1879, opened 2020-04-13
Blackjack environment does not handle naturals correctly


rhalbersma:
**TL;DR**: the current blackjack code does not correctly implement the Sutton and Barto example, which in turn also does not correctly implement the usual casino rules as described by Thorp, "Beat the Dealer".

Lines 105-107 of `blackjack.py` do not correctly implement the scoring of naturals (21 with 2 cards). A natural by either the player or the dealer beats a 21 from more than 2 cards from the other player. The current code rewards this with 0. 

A fix is to select the score from a small dictionary, keyed on the combination of `is_natural` on the player's and dealer's hands. Since the case of the player busting has already been handled, it is also convenient to directly test for the dealer busting, which simplifies the scoring of regular hands to `sum_hand`, obviating the need for `score()`.

```
if is_bust(self.dealer):
    reward = 1.
else:
    reward = {
        (True, False)   : 1.5 if self.natural else 1.,
        (True, True)    : 0.,
        (False, True)   : -1.,
        (False, False)  : cmp(sum_hand(self.player), sum_hand(self.dealer))
    }[(is_natural(self.player), is_natural(self.dealer))]
```

Note that Sutton and Barto correctly describe the positive reward for a player natural versus a dealer 21 (albeit that they use +1 instead of +1.5 as described by their cited Thorp). However, also Sutton and Barto neglect to mention the case of a player 21 against a dealer natural, scoring it as 0 instead of -1. This might also explain their discrepancy (mentioned on p.99) with Thorp: S & B recommend hitting on a soft 18 with a dealer's ace showing. Since this ace gives the dealer a 4/13 probability of a natural, even a player's 21 is will not help him some of the time. This will make it better to stand on such a soft 18.

I can submit a PR for a 'Blackjack-v1' (since fixing this scoring is a breaking change). If such a PR is welcomed, I'd also want to fix some other small things, such as the order of the returned observation. It's more convenient to return this a `(usable_ace, player count, dealer card)` and to convert `usable_ace` to an `int` instead of a `Boolean`. The integer conversion makes it possible to directly index NumPy arrays with the returned observation tuple, which simplifies table-based algorithms. The permuted order of `usable_ace` makes displaying NumPy arrays (for Q-values / policies e.g.) naturally appear in the format of S & B, i.e. two matrices, one for soft and hard hands each.

christopherhesse:
[...]

PR #2398, opened 2021-09-06
Fix the natural handling in Blackjack

RedTachyon:
See #1879 and the complications I described there.

Blackjack-v0 is kept as is, including the `natural` argument that isn't actually activated in Blackjack-v0

I added a Blackjack-v1 variant which implements the game described in S&B as accurately as it gets, to the best of my understanding (which is limited by only a vague knowledge of the original game - so I had to rely on the book).

### DIFF
--- a/docs/toy_text/taxi.md
+++ b/docs/toy_text/taxi.md
@@ -86,5 +86,7 @@ gym.make('Taxi-v3')
 
 ### Version History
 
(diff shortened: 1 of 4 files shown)


jkterry1:
Can you please merge in master so tests can properly function?

jkterry1:
Please remove v0 per discussion on the issue

PR #2404, opened 2021-09-11
Update test case for `VideoRecorder`

XuehaiPan:
Refer to #2318 for more details.

### DIFF
--- a/gym/wrappers/monitoring/tests/test_video_recorder.py
+++ b/gym/wrappers/monitoring/tests/test_video_recorder.py
@@ -1,13 +1,9 @@
 import gc
-import json
 import os
-import shutil
-import tempfile
 import time
-import numpy as np
 
 import gym
-from gym.wrappers.monitoring.video_recorder import VideoRecorder, video_recorder_closer
+from gym.wrappers.monitoring.video_recorder import VideoRecorder
 
 
 class BrokenRecordableEnv(object):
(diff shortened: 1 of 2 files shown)


jkterry1:
@XuehaiPan this fixes both of those notes right?

XuehaiPan:
This PR takes the tests for `closer` back for the first comment. https://github.com/openai/gym/pull/2318#discussion_r704789524

As commented in https://github.com/openai/gym/pull/2318#discussion_r704791882 (the second comment) and https://github.com/openai/gym/pull/2318#discussion_r706634689, we could remove all `closer`s in `gym` entirely (not only for `VideoRecoder`, but also for `Monitor`, etc.), and call `close()` in the `__del__()` method.

vwxyzjn:
I like the idea of having "`close()` in the `__del__()` method", which seems like a much cleaner solution than a customized `Closer`. If we move forward with this idea, would it make the test cases cleaner?

XuehaiPan:
> I like the idea of having "`close()` in the `__del__()` method", which seems like a much cleaner solution than a customized `Closer`. If we move forward with this idea, would it make the test cases cleaner?

Definitely. This will make both the implementation and the tests simpler for classes with a `close()` method (e.g., `Env`, `Monitor`, `VideoRecorder`, etc.).

However, if we want to test the side effect (e.g. create a video file) for a `close()` method that called implicitly by `__del__()` (the `test_autoclose` in this PR), we should do explicit garbage collection and wait for a few seconds (as we cannot know when will the Python interpreter do a garbage cleaning):

```python
import time
import gc

rec_path, proc, num_registered = record()

gc.collect()  # do explicit garbage collection for test
time.sleep(5)  # wait for subprocess exiting

assert proc.poll() is not None  # subprocess is terminated
```

jkterry1:
@XuehaiPan can you please make the remaining requested changes (by costa and the second comment) so this can be merged?

XuehaiPan:
I think the first commit (https://github.com/openai/gym/pull/2404/commits/a888dfa168a5881b6590d9541e1c33c95216a20f) is enough to reply to both of the comments. I have removed `closer` for `VideoRecorder` in the new commit (https://github.com/openai/gym/pull/2404/commits/78e1edbaeaa5d86a60712c2010c65e50ea5168a4) to simplify the tests. (https://github.com/openai/gym/pull/2404#issuecomment-917709758 and https://github.com/openai/gym/pull/2404#issuecomment-917865225)

Issue #2210, opened 2021-04-02
Seed Returned by gym.seeeding.np_random not usable with numpy Random Generators


maximilianreimer:
Hi everyone,
I just stumble upon a minor bug (?): The seed returned by `np_random, seed = gym.seeding.np_random(seed)` can not be used to seed another `numpy.random.RandomGenerator(seed)` or reseed and exiting one (e.g. `numpy.random.seed(seed)`) since those except numbers between between 0 and 2**32 - 1.  I think that should be possible, shouldn't it? 

The concrete failure case I have is: 
```python
import numpy as np
from gym.utils import seeding

user_given_seed = None
np_random, seed = seeding.np_random(user_given_seed)

# since in some places np.random.randn or similar is used we want to seed the global numpy random generator as well
np.random.seed(seed)
```
Results in: 
```
...
   np.random.seed(seed)
  File "mtrand.pyx", line 244, in numpy.random.mtrand.RandomState.seed
  File "_mt19937.pyx", line 166, in numpy.random._mt19937.MT19937._legacy_seeding
  File "_mt19937.pyx", line 180, in numpy.random._mt19937.MT19937._legacy_seeding
ValueError: Seed must be between 0 and 2**32 - 1
```

And I think the user might expect that a generator seeded by returned seed it the same as the one that has been returned by the methods. 

```
import numpy as np
from gym.utils import seeding

# setting this to 1 resulted in a seed in the valid range. 
user_given_seed = 1
np_random, seed = seeding.np_random(user_given_seed)

np_random_2 = np.randm.RandomState(seed)

asser np_random_2.random() == np_random() # fails
```
As far as I can see that's not the case currently. 

I am using:
* Python 3.9
* gym 0.15.3

RedTachyon:
So to the best of my understanding, the issue here is that the gym seeding uses 8-bit bigint (uint64), whereas numpy only accepts 4-bit ints (uint32, as stated in the docs).

I think this can be fixed by setting create_seed (seeding.py#L49) to only use 4 bits. Or ditching the seeding tools entirely as per #1663

jkterry1:
Closing in favor of the PR

Issue #1663, opened 2019-08-28
Remove seeding utils as NumPy 1.17 supports better random number generators


zuoxingdong:
In the [release of Numpy 1.17](https://numpy.org/devdocs/release.html#new-features), it seems that it has improvement in seeding designed for parallel processes.

I am curious about if it could be a good idea to remove the [seeding utils](https://github.com/openai/gym/blob/master/gym/utils/seeding.py) which was originally designed for reducing correlation due to weak seeding (Python 2 in particular) ? 

What do you think ? @pzhokhov @christopherhesse

christopherhesse:
I'm not sure we want to require numpy 1.17 just yet, maybe wait until it has been out for a while.

jkterry1:
Numpy 1.17 can safely be required now

jkterry1:
Closing in favor of the PR

Issue #2407, opened 2021-09-14
Add deprecation notices for v0/v4 versions of the Atari environments and eventually remove them


jkterry1:
In the next release we need to add deprecation notices when importing all the v0/v4 versions of the Atari environments, in the near future (say 3 months) we should remove the v0 environments outright, and in about a year we should remove the v4 environments outright.

@JesseFarebro

lebrice:
Hey @jkterry1 , did the atari environments get removed?
```python3
from gym.envs.atari import AtariEnv  # Doesn't work
```
And even after `pip install gym[atari]`, and using the recommended syntax from the `gym/envs/README.md`:
```python3
from gym.envs import atari  # also doesn't work
AtariEnv = atari.AtariEnv
```

JesseFarebro:
Hi @lebrice, they have been removed from Gym. Once you install `gym[atari]` it installs the new Arcade Learning Environment python wrapper `ale-py`. You can essentially achieve the same `AtariEnv` by importing `ale_py.gym.ALGymEnv`, i.e.,

```py
from ale_py.gym import ALGymEnv
```

Just note that some of the default parameters to `ALGymEnv` have changed. See this blog post for additional info: [https://brosa.ca/blog/ale-release-v0.7](https://brosa.ca/blog/ale-release-v0.7). Furthermore, the class is now located here: https://github.com/mgbellemare/Arcade-Learning-Environment/blob/master/src/python/gym/environment.py.

lebrice:
Thanks for the explanation @JesseFarebro. 
However, I can't help but wonder why this change was made in such a backward-incompatible way! Would you consider re-introducing the package in the `gym.envs` namespace using something like a try-except clause? I feel like this might have broken literally all the RL research code out there!

JesseFarebro:
I don't understand what your use case is? It's not common to import the actual `AtariEnv` class. This wasn't even documented anywhere and is what I'd consider off-label. The documented method for instantiating Atari environments is to call `gym.make`, i.e.,

```py
import gym

env = gym.make('Breakout-v0')
```

This still works as intended, it just hooks into the aforementioned `ALGymEnv` class. Why would you be directly importing `AtariEnv`? All arguments that you'd pass to `AtariEnv` can be passed to `gym.make`. From what I can see on Github search this isn't commonplace and definitely won't break all RL research code, the `gym.make` paradigm is what's ubiquitous. Am I missing something here?

lebrice:
[...]

PR #2422, opened 2021-09-25
Seeding update

RedTachyon:
See Issue #1663

This is a bit of a ride. The base change is getting rid of the custom seeding utils of gym, and instead using `np.random.default_rng()` as is recommended with modern versions of NumPy. I kept the `gym.utils.seeding.np_random`  interface and changed it to basically being a synonym for `default_rng` (with some API difference, consistent with the old `np_random`)

Because the API (then `RandomState`, now `Generator`) changed a bit, `np_random.randint` calls were replaced with `np_random.integers`, `np_random.rand` -> `np_random.random`, `np_random.randn` -> `np_random.standard_normal`. This is all in accordance with the recommended NumPy conversion.

My doubts in order of (subjective) importance

### Doubt number 1: 
In `gym/utils/seeding.py#L18` I'm accessing a "protected" variable `seed_seq`. This serves accessing the random seed that was automatically generated under the hood when the passed seed is None. (it also gives the correct value if the passed seed is an integer) 
An alternative solution would be restoring the whole `create_seed` machinery which generates a random initial seed from `os.urandom`. I was unable to find another way to get the initial seed of a default Generator (`default_rng(None)`) instance.

### Doubt number 2:
In `gym/spaces/multi_discrete.py#L64`. Turns out that a `Generator` doesn't directly support `get_state` and `set_state`. The same functionality seems to be achievable by accesing the bit generator and modifying its state directly (without getters/setters). 

### Doubt number 3:
As mentioned earlier, this version maintains the `gym.utils.seeding` file with just a single function. Functionally, I don't think that's a problem, but might feel a bit redundant from a stylistic point of view. This could be replaced by changing something like 17 calls of this function that occur in the codebase, but at the moment I think it'd be a bad move. The reason is that the function passes through the seed that's generated if the passed seed is None (see Doubt 1), which has to be obtained through mildly sketchy means, so it's better to keep it contained within the function. I don't think passing the seed extremely necessary, but that would somewhat significantly change the actual external API, I tend to be hesitant about changes like this if there's no good reason.
Overall I think it's good to keep the `seeding.np_random` function to keep some consistency with previous versions. The alternative is just completely removing the concept of "gym seeding", and using NumPy. (right now "gym seeding" is basically an interface for NumPy seeding)

### Doubt number 4:
Pinging @araffin as there's a possibility this will (again) break some old pickled spaces in certain cases, and I know this was an issue with SB3 and the model zoo. Specifically, if you create a `Space` with the current master branch, run `space.sample()` at least once, and then pickle it, it will be pickled with a `RandomState` instance, which is now considered a legacy generator in NumPy. If you unpickle it using new gym code (i.e. this PR), `space.np_random` will still point to a `RandomState`, but the rest of the code expects `space.np_random` to be a `Generator` instance, which has a few API changes (see the beginning of this post).

Overall I don't know how important it is for the internal details of gym objects to remain the same - which is more or less necessary for old objects to be unpicklable in new gym versions. There's probably a way of a custom unpickling protocol as a compatibility layer - I'm not sufficiently familiar with this to do it, but I imagine it should be doable on the user side? (i.e. not in gym) 

### Doubt number 2137: (very low importance)
This doesn't technically solve #2210. IMHO this is absolutely acceptable, because `np.random.seed` is part of the legacy seeding mechanism, and is officially discouraged by NumPy. Proper usage yields expected results:
```python
import numpy as np
from gym.utils import seeding

user_given_seed = None
np_random, seed = seeding.np_random(user_given_seed)

# since in some places np.random.randn or similar is used we want to seed the global numpy random generator as well
rng = np.random.default_rng(seed)
```

## tl;dr
This definitely needs another set of eyes on it because I'm not confident enough about the nitty-gritty details of NumPy RNG. There are a few things I'm not 100% happy with from a stylistic point of view, but as far as my understanding goes, the functionality is what it should be. There's also the question of supporting old pickled objects, which I think is a whole different topic that needs discussing now that gym is maintained again.


vwxyzjn:
Thanks for contributing this PR. Could you add a few test cases on some environments like `CartPole-v1` and assert the observations and actions are the same given the same random seed?

RedTachyon:
I think this is already present in `gym.envs.tests.test_determinism.test_env` (https://github.com/openai/gym/blob/master/gym/envs/tests/test_determinism.py), can you check if that's what you mean?

XuehaiPan:
> Because the API (then RandomState, now Generator) changed a bit, `np_random.randint` calls were replaced with `np_random.integers`, `np_random.rand` -> `np_random.random`, `np_random.randn` -> `np_random.standard_normal`. This is all in accordance with the recommended NumPy conversion.

As @RedTachyon already noted here, such changes drop backward compatibility for users who using `self.np_random` to keep reproducibility in their custom environments. Some legacy code would suddenly be unfunctional after upgrading their `gym` backend version. We should be really careful about that.

In the NumPy API reference ([What’s New or Different](https://numpy.org/doc/stable/reference/random/new-or-different.html#new-or-different)), the only two benefits I can see are that `Generator` provides more distribution functions than `RandomState` and is a little bit faster. However, these new functionalities we have never used in `gym` by far. Such changes will lead to ugly implementations that is shown in [Random sampling (`numpy.random`)](https://numpy.org/doc/stable/reference/random/index.html#quick-start) almost everywhere when users use `self.np_random`:

```python
try:
    rng_integers = rng.integers
except AttributeError:
    rng_integers = rng.randint
a = rng_integers(1000)
```

to ensure their code is compatible with different version of `gym`.

------

In [NEP 19 — Random number generator policy](https://numpy.org/neps/nep-0019-rng-policy.html):

> ***All current usages of `RandomState` will continue to work in perpetuity***, though some may be discouraged through documentation. Unit tests can continue to use the full complement of `RandomState` methods. Mixed `RandomState/Generator` code can safely share the common `BitGenerator` state. Unmodified `RandomState` code can make use of the new features of alternative BitGenerator-like settable streams.

As `gym` is a widely used framework, I think we should provide stable interfaces.

araffin:
>As gym is a widely used framework, I think we should provide our interfaces as stable as we can. I don't think is a good idea to introduce these breaking changes.

I second @XuehaiPan on that.
What is the rationale behind that breaking change? (as the feature is not buggy nor deprecated)

XuehaiPan:
> What is the rationale behind that breaking change? (as the feature is not buggy nor deprecated)

Here is my worry about I mentioned above:

> Some legacy code would suddenly be unfunctional after upgrading their gym backend version. We should be really careful about that.

---

I'm one of the `gym` users that writing custom RL environments. And it's very nice to use `self.np_random` rather than `np.random` to write reproducible environments. And I mentioned that:


> Such changes will lead to ugly implementations that are shown in Random sampling (`numpy.random`) almost everywhere when users use self.np_random:
>
> ```python
> try:
>     rng_integers = rng.integers
> except AttributeError:
>     rng_integers = rng.randint
> a = rng_integers(1000)
> ```
>
> to ensure their code is compatible with different versions of `gym`.

RedTachyon:
My main rationale is essentially avoiding technical debt. The RandomState interface is considered legacy now and will not be receiving updates, and instead the Generator is the "correct" way of doing RNG with NumPy since 1.17. I'm not entirely sure what is the level of adoption of this change in the general community, but that's the direction things are moving. 

The necessary changes in downstream code are also not particularly difficult, and are also essentially an update in the direction of "best practices". 

From what I understand, Generator also behaves better with parallel processes in some way, although I'm unable to find exactly how that works. 


Ultimately, there's again the overarching issue of how much backwards-compatibility we want to maintain now that gym is actually maintained again. In a way, breaking changes were always a part of gym due to the environment versioning, and old code that depends e.g. on LunarLander-v0 or v1 had to be updated. Conversely, striving to keep everything compatible with all previous versions will make it more difficult to make actual updates and avoid racking up technical debt.

XuehaiPan:
> The necessary changes in downstream code are also not particularly difficult, and are also essentially an update in the direction of "best practices".

Users will need to refactor their code and specify something like `gym >= 0.21.0` (or any other version). Nowadays, users only write only a piece of code build on top of lots of dependency packages. Although such changes in downstream code are not difficult, it needs the whole community to update their code almost simultaneously and push a new version to PyPI and/or conda as quickly as possible. I don't think it is possible to achieve.

araffin:
> My main rationale is essentially avoiding technical debt. 

Is it really technical debt? Is Gym currently limited by the current implementation? What features are missing?
(from what I see, gym only uses very basic functions from the random generator)

> how much backwards-compatibility we want to maintain now that gym is actually maintained again

Maintained does not mean we need to break code at every release.
See also my comment in https://github.com/openai/gym/issues/2407#issuecomment-919918855 (quoting @jkterry1  intentions in the "Plans for future maintenance").

> In a way, breaking changes were always a part of gym due to the environment versioning, and old code that depends e.g. on LunarLander-v0 or v1 had to be updated. 

Breaking changes in environment should occur only when a bug is fixed, only when it is *necessary*.

Too many things rely on gym and doing breaking changes at every release is the best thing you can do to annoy every user. That would be ok if gym was in an early stage of development but that's not the case any more.

The only result you will get is people capping gym version to avoid breaking changes: https://github.com/DLR-RM/stable-baselines3/issues/573 and https://github.com/cyprienc/stable-baselines3/commit/e7a48d29df5b1f519d8e037ac6ba91f9608cef42

> From what I understand, Generator also behaves better with parallel processes in some way, although I'm unable to find exactly how that works.

If this piece of code was a bottleneck in current gym env, I would agree to change it.
But currently no benchmark has been done and the ratio benefits/cost seems pretty low.

jkterry1:
-Correct me if I'm wrong here, but the breaking changes here should be for a very minor use case, and I don't believe that it should break SB3 or any other major repo?

-@RedTachyon is there anyway to retain backwards compatibility with a warning instead of breaking things?

RedTachyon:
Hm, so the way I see it there are two issues to deal with, which are somewhat orthogonal.

Number 1: existing codebases (particularly custom environments) can use `self.np_random.randn` and other legacy methods like this.

Number 2: spaces (and envs? does anyone ever pickle and store envs?) that were pickled in old versions can get wonky if unpickled with a new version.

---

Re 1: my best idea for a solution is using a very thin wrapper class like this:

```python
class NGenerator:        
    def __init__(self, generator: np.random.Generator):
        self.generator = generator
        
    def __getattr__(self, name):
        return getattr(self.generator, name)
    
    def __repr__(self):
        return repr(self.generator)
    
    def __str__(self):
        return str(self.generator)
    
    def rand(self, *dims):
        # some warning
        return self.generator.random(dims)
    
    def randn(self, *dims):
        return self.generator.standard_normal(dims)
    
    def randint(self, low, high=None, size=None, dtype=int):
        return self.generator.integers(low=low, high=high, size=size, dtype=dtype)
    
    def seed(self, seed: int):
        self.generator = np.random.default_rng(seed)
```

Add some docstrings, add some warnings, and in principle it should act exactly like the original Generator, except it has these three additional methods, which optionally pass a warning. This would still use the new "correct" RNG, but would allow using it with the legacy interface. 

Then the `seeding.np_random` would return a generator wrapped in this class, and people would still be able to happily use `randint` etc.

The one thing this can't do is `setattr`, but doing `setattr` on a `Generator` or `RandomState` object would be incredibly cursed and most likely - not possible. (naively forwarding `__setattr__` causes some problems in the init, there's probably a way to do it, but it's probably also not worth it)


---

Re 2: Do we care about the pickled spaces sampling with the `exact same` behavior as before pickling? If so, we could extend the wrapper class described above to accept either a `Generator` or a `RandomState` and expose a shared API with some `isinstance` checks. This would be significantly more spaghettified.

That being said, I don't think this is actually important, it feels like an *extremely* niche use case (i.e. pickle a space or an environment, then unpickle it in the future and require the RNG sequence to be exactly the same).

If we're content with just gracefully unpickling a space and giving it some reasonable `self.np_random`, maybe extending the `__setstate__` so that it just creates a new `Generator` wrapped in the `NGenerator` wrapper would be enough? Or even just set it to `None` and let the lazy seeding do its job.

Note: the `NGenerator` and `self.generator` names are completely arbitrary and I'm not attached to them whatsoever.

jkterry1:
Your solution to 1 seems fine to me provided no one else has further objections

Regarding 2, unless there's an application I'm aware of we don't need to worry about that kind of pickling across different Gym versions. All the ones I'm aware of that this would impact only cache these objects instead of permanently saving them.

araffin:
> Number 2: spaces (and envs? does anyone ever pickle and store envs?) that were pickled in old versions can get wonky if unpickled with a new version.

Most env cannot be pickled anyway. But spaces are (to check if the new env matches the one used for training).
However, is `np_random` used in spaces?

XuehaiPan:
I think adding the legacy function alias to the "new" `Generator` is a good approach for backward compatibility.

------

For the solution provided in https://github.com/openai/gym/pull/2422#issuecomment-927396154, I have some advice here:

1. Warned in [numpy.random.Generator](https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.Generator):

> The function `numpy.random.default_rng` will instantiate a `Generator` with numpy’s default `BitGenerator`.
>
> **No Compatibility Guarantee**
>
> `Generator` does not provide a version compatibility guarantee. In particular, as better algorithms evolve the bitstream may change.

We should use an explicit `BitGenerator` rather than the default of NumPy (which may change between versions).

Currently, the default `BitGenerator` is `numpy.random.PGC64`. And noted in [Upgrading `PCG64` with `PCG64DXSM`](https://numpy.org/doc/stable/reference/random/upgrading-pcg64.html#upgrading-pcg64-with-pcg64dxsm):

> We have introduced a new PCG64DXSM BitGenerator that will eventually become the new default BitGenerator implementation used by default_rng in future releases. PCG64DXSM solves the statistical weakness while preserving the performance and the features of PCG64.

Although `PCG64DXSM` is a better version of `PCG64`, it is only available on `numpy >= 1.21.0`. I suggest to explicitly use the `Generator(PCG64(seed))` rather than `np.random.default_rng(seed)`.

2. It is unnecessary to write a new wrapper class for the `Generator`, and that introduces new technical debt here. Just using inheritance is enough:

```python
from pkg_resources import parse_version

import numpy as np
import gym
from gym.logger import warn


# TODO: Remove this class and make it alias to `Generator` in Gym release 0.25.0
#       RandomNumberGenerator = np.random.Generator
    REMOVE_SINCE = '0.25.0'

    assert parse_version(gym.__version__) < parse_version(REMOVE_SINCE)

    def rand(self, *size):
        warn('Function `rng.rand(*size)` is marked as deprecated '
             f'and will be removed in gym {self.REMOVE_SINCE} in the future. '
             'Please use `Generator.random(size)` instead.')

        return self.random(size)

    def randn(self, *size):
        warn('Function `rng.randn(*size)` is marked as deprecated '
             f'and will be removed in gym {self.REMOVE_SINCE} in the future. '
             'Please use `rng.standard_normal(size)` instead.')

        return self.standard_normal(size)

    def randint(self, low, high=None, size=None, dtype=None):
        warn('Function `rng.randint(low, [high, size, dtype])` is marked as deprecated '
             f'and will be removed in gym {self.REMOVE_SINCE} in the future. '
             'Please use `rng.integers(low, [high, size, dtype])` instead.')

        return self.integers(low, high, size, dtype)

    def get_state(self):
        warn('Function `rng.get_state()` is marked as deprecated '
             f'and will be removed in gym {self.REMOVE_SINCE} in the future. '
             'Please use `rng.bit_generator.state` instead.')

        return self.bit_generator.state

    def set_state(self, state):
        warn('Function `rng.set_state(state)` is marked as deprecated '
             f'and will be removed in gym {self.REMOVE_SINCE} in the future. '
             'Please use `rng.bit_generator.state = state` instead.')

        self.bit_generator.state = state


# Shortcut for RandomNumberGenerator
RNG = RandomNumberGenerator


def get_rng(seed=None):
    if isinstance(seed, np.random.BitGenerator):
        return RNG(seed)
    if isinstance(seed, np.random.Generator):
        return seed
    return RNG(np.random.PCG64(seed))
```

I could create a PR for this if needed.

RedTachyon:
> However, is np_random used in spaces?

Very rarely, but yes, I describe it in the original PR post in the part where I tagged you. Tl;dr under some conditions, it's used in calling `space.sample()`

RedTachyon:
@XuehaiPan I'm fine with this solution too, this should probably just be added directly in this PR, I can do it if there's agreement. 

My thinking about my approach was that `default_rng` is the officially recommended constructor as per numpy/numpy#15391.

I don't necessarily agree that the two solutions are very different from a technical debt point of view, in the end both simply introduce a new class which will be deleted later.

So for the difference in functionality: with a wrapper class, we use `default_rng` which might not have stable behavior, but from what I understand will have stable API. So we'd automatically get new RNG mechanisms.

With an inherited class, we manage the RNG mechanism manually (for better or for worse), but a very big (subjectively) benefit is that we'd still have `isinstance(self.np_random, np.random.Generator)`


As a minor note regarding the exact code snippet you posted - making a new  `get_rng` function is unnecessary imo, this can all be handled within `gym.utils.seeding.np_random`

XuehaiPan:
> this should probably just be added directly in this PR, I can do it if there's agreement.

I'm fine with that.

> As a minor note regarding the exact code snippet you posted - making a new `get_rng` function is unnecessary imo, this can all be handled within `gym.utils.seeding.np_random`.

Agreed. The snippet here is just an example.

gym/utils/seeding.py
```diff
@@ -14,84 +9,6 @@ def np_random(seed=None):
             "Seed must be a non-negative integer or omitted, not {}".format(seed)
         )
 
-    seed = create_seed(seed)
-
-    rng = np.random.RandomState()
-    rng.seed(_int_list_from_bigint(hash_seed(seed)))
+    rng = np.random.default_rng(seed)
+    seed = rng.bit_generator._seed_seq.entropy
```
XuehaiPan, line 13:
```suggestion
    sq = np.random.SeedSequence(seed)
    seed = sq.entropy
    rng = get_rng(sq)
```

XuehaiPan:
Continue to https://github.com/openai/gym/pull/2422#issuecomment-927628102

> Or we can keep this backward compatibility in perpetuity.

Should we keep `self.np_random` to provide the exact same functionality as `np.random`, i.e., always have `np_random.rand` and `np_random.randint` etc.? As:

> NumPy will continue to keep `RandomState` as a stable API in the future.

------

Otherwise, the name `np_random` could be misleading after removing the helper class in the future.

```python
>>> self.np_random.randint
AttributeError: 'numpy.random._generator.Generator' object has no attribute 'randint'

>>> np.random.randint
<built-in method randint of numpy.random.mtrand.RandomState object at 0x7f6e4d1ec740>
```

jkterry1:
I would like to keep the waring in place. This is functionality that should be removed in the future, and will _have_ to be as numpy plans to deprecate this eventually. We'll just leave it as a warning for maybe a year (or longer, depending) so there can be a smooth transition period without unnecessary breaks.

jkterry1:
Also I'm going to wait and merge this PR until the release after the next one

RedTachyon:
So this new commit is basically what @XuehaiPan wrote above, the changes include:
- Removed an explicit mention of the version when it'd be removed, I don't know if we want to commit to a specific version right now.
- Removed the possibility to pass a BitGenerator as an argument to `np_random` - it was always the case that you can pass `Optional[int]` and I think that's absolutely good enough
- As mentioned earlier, the whole initialization logic is kept within `np_random` to keep it consistent.

I'm not 100% sure I'm happy with the `seed` method, but then again, it shouldn't be used anyways, and I don't see better alternatives

gym/utils/seeding.py
```diff
@@ -14,84 +13,82 @@ def np_random(seed=None):
...
-    while bigint > 0:
-        bigint, mod = divmod(bigint, 2 ** 32)
-        ints.append(mod)
-    return ints
+# TODO: Remove this class and make it alias to `Generator` in Gym release 0.25.0
+# RandomNumberGenerator = np.random.Generator
+class RandomNumberGenerator(np.random.Generator):
+
+    # assert parse_version(gym.__version__) < parse_version(REMOVE_SINCE)
+
```
XuehaiPan, line 27:
```suggestion
# TODO: Remove this class and make it alias to `Generator` in the future after a smooth migration
# RandomNumberGenerator = np.random.Generator
class RandomNumberGenerator(np.random.Generator):

```

And remove the unnecessary import of `Generator` and `parse_version`.

gym/utils/seeding.py
```diff
@@ -14,84 +13,82 @@ def np_random(seed=None):
...
+        )
+
+        self.bit_generator.state = state
+
+    def seed(self, seed=None):
+        warn(
+            "Function `rng.seed(seed)` is marked as deprecated "
+            "and will be removed in the future. "
+            "Please use `rng = get_rng(seed)` to create a separated generator instead."
+        )
```
XuehaiPan, line 82:
```suggestion
        warn(
            "Function `rng.seed(seed)` is marked as deprecated "
            "and will be removed in the future. "
            "Please use `rng, seed = gym.utils.seeding.np_random(seed)` to create a separated generator instead."
        )
```

tristandeleu:
> Furthermore, wouldn't this improve the situation with vector environments? I've never used vector environments but I can't imagine the RNG usage is very robust. This way you could take advantage of `SeedSequence` spawning where the user actually has control, not Gym forcing a decision upon us.

The way this is currently handled in `VectorEnv` is to have [sequential seeds](https://github.com/openai/gym/blob/000a2a0d513729bd560bced898967f7cc90d5530/gym/vector/async_vector_env.py#L162-L180), which is clearly not ideal; you always have the option to set the seeds of sub-environments manually though. A much better solution, as was discussed above in https://github.com/openai/gym/pull/2422#issuecomment-927318444, would be to fork the `Generator`.

JesseFarebro:
@RedTachyon Sorry, I should have been more specific. I'd argue that for Gym's use case we'd have something like:

```py
import gym
import numpy as np

env = gym.make('EnvID', seed=np.random.get_default_rng())
```

Environments should handle it as follows:

```py
import numpy as np

class Env:
    def __init__(self, /, seed=None):
        self._rng = np.random.get_default_rng(seed)
        self._action_space = spaces.Discrete(..., rng=self._rng)

# Initialize with a `SeedSequence`
seed = np.random.SeedSequence(12583)
env = Env(seed=seed)

# Initialize with the default Generator
rng = np.random.get_default_rng()
env = Env(seed=rng)

# Initialize with only an integer
seed = 12583
env = Env(seed=seed)

# Initialize with custom Generator
rng = np.random.Generator(np.random.MT19937(12583))
env = Env(seed=seed)
```

I still have no idea why `seed()` is a function on the environment class. For the ALE alone this results in a really weird paradigm where we have to load the ROM and set the game mode + difficulty in `seed`. This clearly isn't the intended purpose of `seed()` and it makes no sense why I should be able to create an environment and then seed it multiple times. Even Gym3 did away with `seed()` in favour of supplying a keyword argument.
 
Regardless of the benefits, I mentioned in my previous comment, just from a practical perspective, who likes managing Python `random.seed`, `numpy.random` seed, TF / Torch / Jax random seed, Gym random seed, etc. Out of all of these, the only one that makes any sense is managing another seed for libraries that may run on an accelerator like TF / Torch / Jax. These may have different needs that can't be met by `np.random`. Gym though? This seems out of scope for Gym.

> I'm still a fan of the self.np_random pattern (even though the name is mildly unfortunate now that the API is different). While it might not give the full flexibility, I don't believe that is necessary. Instead, it gives very high convenience in managing the RNG for the use cases that are actually necessary.

Could you elaborate on this because I disagree. Give the environment a seed upon construction and it can provide the RNG to `gym.spaces` and either use the RNG to initialize the environment or keep it around if it needs the RNG in, say, `reset` or `step`.

---

@tristandeleu 
> The way this is currently handled in VectorEnv is to have sequential seeds, which is clearly not ideal; you always have the option to set the seeds of sub-environments manually though. A much better solution, as was discussed above in #2422 (comment), would be to fork the Generator.

What do you mean by fork the generator? I think the intended use is to initialize a `SeedSequence` in which you can call `spawn` (see [numpy docs](https://numpy.org/doc/stable/reference/random/parallel.html)). Given the current implementation in this PR, we manage the `SeedSequence` for you which wouldn't be ideal.  The numpy API doesn't allow you to pass a `SeedSequence` to initialize a `SeedSequence`. Thus you can't even use the numpy sanctioned API to seed Gym as things stand (Note: `spawn()` returns a list of child `SeedSequence`s).

My proposal would make this trivial as you could do something like:

```py
import numpy as np

class AsyncVectorEnv:
    def __init__(self, env_fns, /, seeds = None):
        if seeds is None:
            seeds = [None] * len(env_fns)
        assert isinstance(seeds, collections.abc.Sequence)
        
        self.rngs = map(np.random.get_default_rng, seeds)
        
        ...
        
        for ..., rng in zip(..., self.rngs):
            process = ctx.Process(
                    target=...,
                    args=(
                        ...,
                        rng
                    ),
                )


my_seed = 12858391

seed_seq = np.random.SeedSequence(my_seed)
seeds = seed_seq.spawn(10)

env = AsyncVectorEnv(..., seeds=seeds)
```

XuehaiPan:
> I still have no idea why `seed()` is a function on the environment class.

I think the answer is quite simple. If the only way to seed an environment is specify the `seed` or `generator` at initialization `__init__`, how can we support this:

```python
import copy

env = MyEnv(*very_long_argument_list)
env.seed(0)

clone = copy.deepcopy(env)
clone.seed(1)
```

> The concept of "passing around a generator" discussed in the Generator changes in np.random should be applied here as well. Why can't I provide a Generator to Gym where I have complete control over seeding it? This removes a lot of the complexities discussed here and brings multiple advantages (e.g., see https://github.com/google/jax/blob/main/design_notes/prng.md).

Do we really care about the flexibility? For me, the only thing I concern about is making my custom environment and spaces are reproducable. I think seed the RNGs with an integer or `None` is good enough in practice. And we only accept `Optional[int]` can alleviate a lot of maintenance work.

> Thus you can't even use the numpy sanctioned API to seed Gym as things stand (Note: spawn() returns a list of child SeedSequences).

By now, we use `subseeds = self.np_random.choice(INT_MAX, size=n_child)` to generate (or spawn) a sequence of seeds correlated with the main seed (see `spaces.Tuple.seed` and `spaces.Dict.seed`). We could use `SeedSequence.spawn` in the future.

RedTachyon:
@JesseFarebro From what I understand then, the only major difference you're proposing is that the `env.seed` method would be removed, and creating the RNG would be moved to environment initialization, and that's it? I can't really see the benefit of this, it would be removing a (somewhat widely used?) method, with the only reason I see being that it's annoying to implement in some specific (i.e. ALE) environments. 

I wouldn't be opposed to including a `seed` argument in the env initialization/`env.make`, but at the moment this is (imo sufficiently well) emulated by
```python
env = Env(...) # env = gym.make(...)
env.seed(0)
```

If it's *really* important, we could make the `seed` argument accept different types after all, e.g. `Union[None, int, SeedSequence, Generator, BitGenerator]` and as far as I can tell, it's almost exactly as what you're proposing, except the end user has an option to do `env.seed` after initialization. At the moment I kept `seed: Optional[int]` because simplicity is good.

I do agree that there probably isn't much need to call `env.seed` multiple times, at least to the best of my knowledge. But the pattern seems relatively entrenched, in my opinion is harmless and intuitive.

Keep in mind that our two version aren't *that* different in how the RNG is handled for the end-user and for env developers. Ultimately you still have a field within the environment, called either `self.np_random` or `self._rng` which is managed by the environment and available within other methods.

--- 
>Could you elaborate on this because I disagree.

I still feel like we're talking about almost the same thing. What I meant by `self.np_random` is what you wrote as `self._rng`. The impact of controlling a `Space` seed is minimal (if some code actually relies on it, I'd be extremely concerned), and other than that, the Env still maintains an inner RNG and passes it around implicitly.

In other words, as far as I understandyour pattern is:
```python
env = gym.make(env_id, seed=0)
# RNG managed by the env from now on
env.reset()
for t in range(100):
    env.step(some_action)
```

and the current pattern is
```python
env = gym.make(env_id)
env.seed(0)
# RNG managed by the env from now on
env.reset()
for t in range(100):
    env.step(some_action)
```

---
>Do we really care about the flexibility? For me, the only thing I concern about is making my custom environments and spaces reproducable. We don't (or shouldn't) completely control how these random numbers are generated. That's what the "random" stands for.

I strongly agree with this. Let's remember that gym is *not* a general purpose numerical computation library (as opposed to numpy, scipy, pytorch, jax, which are the libraries that were linked as references earlier). It's a tool built for a specific purpose, and controlling the PRNG is very far from the essence of that purpose (i.e. RL)

---
I'll skip the VecEnv discussions because I never actually used the built-in ones. I feel like that's a problem for a separate issue, and as long as what we do here makes it possible for VecEnv to work properly, it's good enough.

tristandeleu:
> What do you mean by fork the generator? I think the intended use is to initialize a `SeedSequence` in which you can call `spawn` (see [numpy docs](https://numpy.org/doc/stable/reference/random/parallel.html)). Given the current implementation in this PR, we manage the `SeedSequence` for you which wouldn't be ideal. The numpy API doesn't allow you to pass a `SeedSequence` to initialize a `SeedSequence`. Thus you can't even use the numpy sanctioned API to seed Gym as things stand (Note: `spawn()` returns a list of child `SeedSequence`s).
> 
> My proposal would make this trivial as you could do something like:

Sorry for the confusion, I did meant spawning multiple `SeedSequence`s from a given `SeedSequence`.

araffin:
>   Do you know if this pattern is ever actually in use? 
> In fact, since right now manually passing the seed in a space.sample is not possible, it can't be the case that someone uses it like that while passing the seed around.

https://github.com/DLR-RM/stable-baselines3/blob/3efab0d267e74cb03264411d4500ddde0c163404/stable_baselines3/common/base_class.py#L552

and 
https://github.com/DLR-RM/stable-baselines3/blob/3efab0d267e74cb03264411d4500ddde0c163404/stable_baselines3/common/off_policy_algorithm.py#L402

?

tristandeleu:
> Do you know if this pattern is ever actually in use? I'd honestly expect that a random agent like this would have its own RNG managing, and if it doesn't, then they likely don't care about it enough to pass the RNG around "properly".

I don't know if this pattern is used, and if I were to do it I'd most likely handle the RNG on the agent's side as well. I just wanted to raise awareness that this could be an issue.

> In fact, since right now manually passing the seed in a `space.sample` is not possible, it can't be the case that someone uses it like that while passing the seed around. So the question is whether we want to enable a pattern like this. In my opinion, the answer is no, because a better pattern anyway would be having a separate sampler.

You can't manually set the seed in `space.sample`, but you can in `Space`, so you could possibly have a situation where the user did `env.action_space.seed(0)` to fix the seed for reproducibility. Again, I don't think this is a good practice, and I agree that we shouldn't enable this pattern, but we have to keep that in mind if any changes are made on the RNG.

RedTachyon:
@araffin This isn't really the same though. The first link is basically a wrapper for `space.seed()` (among other things) which is still possible the exact same way. The latter does use `space.sample()`, but without paying extra attention to the RNG (because it's implicitly handled by the space). With the current state of the PR, both of these things will work the exact same way w.r.t. the interface. The discussion is about making it more explicit.

The conditions for it to be "that pattern" are to me the following:

0. Use the built-in `space.sample` for action selection
0. Actually use those actions for some training logic
0. Pay attention to how the RNG is passed through these samples

I should have been clearer in my last message, but for me, the example you gave falls into the 'if it doesn't, then they likely don't care about it enough to pass the RNG around "properly"' clause.

In fact I'm pretty sure having code with the three conditions is impossible at the moment, because `space.sample` doesn't accept any arguments, so if you want manual control over your generator, you must use an external sampler. 

---

Also one note
>Very specific to Space though, how about having space.sample(rng: Generator = np.random.default_rng(None))?

If this were to be introduced (imo redundant and therefore a bad idea), it would have to be `space.sample(rng: Optional[Generator] = None`, and if it's None, it uses the internally managed RNG. Because without that, then you have a whole mess with backwards compatibility.

JesseFarebro:
@XuehaiPan
> I think the answer is quite simple. If the only way to seed an environment is specify the seed or generator at initialization __init__, how can we support this:

```py
import copy

env = MyEnv(*very_long_argument_list)
env.seed(0)

clone = copy.deepcopy(env)
clone.seed(1)
```

> If the answer is make a new method set_rng(), it is almost the same with seed(). And the latter is much simpler and backward compatible. Besides, you can call self.seed() in your __init__(self, seed=None), rather than a separated call.

This perfectly encapsulates everything that's wrong with `.seed()`. What even is this? Why would you want to do this? What would you expect `seed()` to do here? I guarantee it's not doing what you're expecting. In the ALE for example this would completely reset the environment and this would be equivalent to just instantiating the environment from scratch. The contract for `seed()` just makes no sense IMO. Even newer environments completely ignore `seed()`, e.g., Procgen won't do anything with `.seed()`, it requires a keyword argument instead.

> Using a global RNG is thread unsafe. Different result may produce by the system process management. So we assign a separated RNG for each environment and space controlled by a main seed. The only purpose of the seeding control is reproducablility.

Obviously, hence my argument for "passing a generator". There's no momentum for wanting to implement breaking changes in Gym, if there was I think it's clear spaces should never have implemented a `sample` method. I'd like to think if it was to be implemented today it would take a generator as input which removes this unneeded complexity of managing child RNGs inside of Gym.

> Do we really care about the flexibility? For me, the only thing I concern about is making my custom environments and spaces reproducable. We don't (or shouldn't) completely control how these random numbers are generated. That's what the "random" stands for.

My concerns prioritize reproducibility over convenience. Given the current state of our community, I think this is the correct choice to make. I don't see how what I proposed makes things any more difficult from the perspective of an environment maintainer. You initialize an RNG from the seed, you use it. Simple.

> I think seed the RNGs with an integer or None is good enough in practice. And we only accept Optional[int] can alleviate a lot of maintenance work. We provide utils.seeding.np_random here, users can write their own utilities for RNG control if they don't like it.

This is a non-issue. There's zero maintenance work, `np.random.get_default_rng` handles all of these cases for you. It accepts `seed: Union[None, int, Sequence[int], SeedSequence, BitSequence, Generator]`.

---

@RedTachyon

> @JesseFarebro From what I understand then, the only major difference you're proposing is that the env.seed method would be removed, and creating the RNG would be moved to environment initialization, and that's it? I can't really see the benefit of this, it would be removing a (somewhat widely used?) method, with the only reason I see being that it's annoying to implement in some specific (i.e. ALE) environments.

I mean this is one way of doing it, I'm just trying to get my point across that first off, `seed()` is terrible. As I mentioned previously, `seed()` just isn't flexible enough and its contract isn't well defined. Like I said, popular environments like the ALE have ill-intended side-effects, Procgen just completely ignores `.seed()`, Brax (which Gym wants to support natively) accepts a seed as a keyword argument. I wouldn't say this widely used method is very coherent and definitely not something anyone could rely on. I know in my own work I always assume `seed()` is somehow broken.

Can we agree that we can and should do better? If we can't then there's no point in continuing to press the issue.

> If it's really important, we could make the seed argument accept different types after all, e.g. Union[None, int, SeedSequence, Generator, BitGenerator] and as far as I can tell, it's almost exactly as what you're proposing, except the end user has an option to do env.seed after initialization. At the moment I kept seed: Optional[int] because simplicity is good.

Regardless of our disagreement, I think this is a good change. Simply calling `get_default_rng(seed)` would solve this.

>I do agree that there probably isn't much need to call env.seed multiple times, at least to the best of my knowledge. But the pattern seems relatively entrenched, in my opinion is harmless and intuitive.

How is this intuitive? This is the farthest thing from intuitive... What would you expect `seed()` to do? I certainly have no idea? In more complex environments where say RNG is involved only when the environment first starts (like the ALE, or say VizDoom, Procgen, etc.) should `seed()` reset the environment completely? It's not clear from the documentation and this is certainly going on right now. This use case is pretty fundamental to many of the most popular environments in Gym and we all just completely break `seed()` right now.

>Keep in mind that our two version aren't that different in how the RNG is handled for the end-user and for env developers. Ultimately you still have a field within the environment, called either self.np_random or self._rng which is managed by the environment and available within other methods.

In my ideal world, you can't even `sample` from a space and the RNG is only used by the environment how the environment sees fit. The generator can be serialized independently of the environment and this also allows you to control for stochasticity if you were doing search.

XuehaiPan:
@JesseFarebro

> > I think the answer is quite simple. If the only way to seed an environment is specify the seed or generator at initialization **init**, how can we support this:
> 
> ```python
> import copy
> 
> env = MyEnv(*very_long_argument_list)
> env.seed(0)
> 
> clone = copy.deepcopy(env)
> clone.seed(1)
> ```
> 
> > If the answer is make a new method `set_rng()`, it is almost the same with `seed()`. And the latter is much simpler and backward compatible. Besides, you can call `self.seed()` in your `__init__(self, seed=None)`, rather than a separated call.
> 
> This perfectly encapsulates everything that's wrong with `.seed()`. What even is this? Why would you want to do this? What would you expect `seed()` to do here? I guarantee it's not doing what you're expecting. In the ALE for example this would completely reset the environment and this would be equivalent to just instantiating the environment from scratch. The contract for `seed()` just makes no sense IMO. Even newer environments completely ignore `seed()`, e.g., Procgen won't do anything with `.seed()`, it requires a keyword argument instead.

------

For clarity:

> Why would you want to do this? What would you expect `seed()` to do here? I guarantee it's not doing what you're expecting.

For the snippet shown above, it is easy to make multiple uncorrelated environments using `copy.deepcopy` rather than pass the long arguments here. For example:

```python
import copy
import multiprocessing as mp


def sample(worker_id, env, policy, buffer):
    local_env = copy.deepcopy(env)
    local_env.seed(worker_id)

    # some code to sample rollout


env = MyEnv(*very_long_argument_list)
env = Wrapper1(env, *very_long_argument_list1)
env = Wrapper2(env, *very_long_argument_list2)
# ...
env = WrapperN(env, *very_long_argument_listN)

manager = mp.Manager()
buffer = manager.list()
workers = []
for worker_id in range(16):
    worker = mp.Process(target=sample, args=(worker_id, env, policy, buffer))
    worker.start()
    workers.append(worker)
```

Otherwise, you should pass all the arguments for the environment and wrapper classes and the arguments of the wrappers to the worker function, in order to create the environment in the local process separately.

> In the ALE for example this would completely reset the environment and this would be equivalent to just instantiating the environment from scratch.

Indeed, seed the environment with a new seed is equivalent to creating a new instance. Similarly, it is suggested to create a new `Generator` instance rather than seed the existing `RNG` in NumPy API reference. But it is really inconvenient (e.g. the example above). Besides, the only thing we do in `seed()` is set the states of the RNGs in the environment. However, it could be a lot of work in `__init__`, such as value range check for the arguments. With the `seed()` method, we can bypass these and just `copy.deepcopy(env)`.

JesseFarebro:
@XuehaiPan

This still makes no sense to me? How doesn't this solve your problem:

```py
import multiprocessing as mp

def make_env(rng):
    env = MyEnv(*very_long_argument_list, seed=rng)
    env = Wrapper1(env, *very_long_argument_list1)
    env = Wrapper2(env, *very_long_argument_list2)
    # ...
    env = WrapperN(env, *very_long_argument_listN)

    return env

def sample(rng, policy, buffer):
    local_env = make_env(rng)

    # some code to sample rollout


my_initial_seed = 128413
seed_seq = np.random.SeedSequence(my_initial_seed)

manager = mp.Manager()
buffer = manager.list()
workers = []

# Generate your child SeedSequences
seeds = seed_seq.spawn(16)
for seed in zip(seeds):
    worker = mp.Process(target=sample, args=(seed, policy, buffer))
    worker.start()
    workers.append(worker)
```

Add more keyword arguments to `make_env` and make it configurable with Gin or Hydra or something. The code above is reproducible and doesn't rely on `copy` which can be unreliable. As you mentioned in your previous comment there's an equal amount of overhead with `copy + seed` as just initializing the environment from scratch, so this isn't an issue.

If on the other hand, you want to begin a rollout from a specific state in the environment this wouldn't even be possible with your code as `seed()` might reset the environment altogether. With the ALE you could do something like: 

```py
import multiprocessing as mp

def make_env(rng):
    env = MyEnv(*very_long_argument_list, seed=rng)
    env = Wrapper1(env, *very_long_argument_list1)
    env = Wrapper2(env, *very_long_argument_list2)
    # ...
    env = WrapperN(env, *very_long_argument_listN)

    return env

def sample(rng, state, policy, buffer):
    local_env = make_env(rng)
    local_env.restore_state(state)

    # some code to sample rollout


my_initial_seed = 128413
seed_seq = np.random.SeedSequence(my_initial_seed)

root_env = make_env(seed_seq)
# ... rollout to some state S
state = root_env.clone_state(include_rng=False)

manager = mp.Manager()
buffer = manager.list()
workers = []

# Generate your child SeedSequences
seeds = seed_seq.spawn(16)
for seed in zip(seeds):
    worker = mp.Process(target=sample, args=(seed, state, policy, buffer))
    worker.start()
    workers.append(worker)
```

---

> I agree that the seed() method should be called only once for an environment, which leads to "just remove seed method and add a new argument to __init__". But it is really inconvenient (e.g. the example above). Besides, the only thing we do in seed() is set the states of the RNGs in the environment. However, it could be a lot of work in __init__(), such as value range check for the arguments, especially we usually wrap the environment with multiple wrappers (same cost in wrapper.__init__()). With the seed() method, we can bypass these, just copy.deepcopy(env) and seed it with another seed.

I don't understand this argument either. Sure, there will be work done to seed the environment, but in most cases you're only initializing the environment once. How is there any cost in `wraper.__init__()`? This calls `__init__` on the wrapper class which is a subclass of `Env` but this won't be initializing the RNG, this should only be handled by the actual environment implementation? `__init__` on the root environment is only ever called once, see your example: `MyEnv(...)`.

XuehaiPan:
```python
def make_env(rng):
    env = MyEnv(*very_long_argument_list, seed=rng)
    env = Wrapper1(env, *very_long_argument_list1)
    env = Wrapper2(env, *very_long_argument_list2)
    # ...
    env = WrapperN(env, *very_long_argument_listN)

    return env

def sample(rng, state, policy, buffer):
    local_env = make_env(rng)
    local_env.restore_state(state)

    # some code to sample rollout
```

In my opinion, hard code the `make_env` in a generic function `sample` is not a good idea. Besides, all `very_long_argument_list` to `very_long_argument_listN` should be part of the arguments for `make_env` not only `rng` here. In some use cases, it could be hard to write a short `make_env` function, such as a multi-agent environment with some user-defined agents (sometimes needs to create models, load checkpoints, etc.) as part of the environment.

In addition, I can access all intermediate variables (e.g. `wrapper1`) in my example but the function `make_env` only returns `env`.

> How is there any cost in `wrapper.__init__()`?

I meant the similar cost in `env.__init__`, such as validation of the arguments. The validation is only need once and `copy.deepcopy` can bypass this.

Here I given an other use case: make tabular in the `__init__` method:

```python
class MyEnv(gym.Env):
    def __init__(self, size, *args):
        vaildate_args(size, *args)  # computational expensive

        self.Q = self.make_tabular(size, *args)

    @staticmethod
    def make_tabubar(*args):
        Q = np.zeros((size, size))
        # some computational expensive code
        return Q
```

In the above example, the function `vaildate_args(size, *args)` should be only called once, as we instantiating multiple environments with almost the same arguments. For the computational expensive `make_tabubar`, once the NumPy array `Q` is calculated, just `copy.deepcopy(Q)` (very fast) will bypass the duplicate computing. I know we can decorate the function with something like `functools.lru_cache`. But the Python built-in caches only support multi-threading, and will not work in multi-processing at all. You will need to compute the tabular again in the other process.

JesseFarebro:
@XuehaiPan

> In my opinion, hard code the make_env in a generic function sample is not a good idea. Besides, all very_long_argument_list to very_long_argument_listN should be part of the arguments for make_env not only rng here.
> In some use cases, it could be hard to write a short make_env function, such as a multi-agent environment with some user-defined agents (sometimes needs to create models, load checkpoints, etc.) as part of the environment.

Yeah, that's why I said you should configure it by using something like Gin / Hydra or just supplying the kwargs to `make_env`. Your other issues shouldn't be issues, let the environment persist and just send weight deltas to your workers. As for this `validate_args` thing I don't see how any environment would do validation that would add any meaningful overhead. There's also a non-zero overhead with doing a `copy.deepcopy`.

I think the main thing here is that in my opinion keeping `.seed()` has downsides that outweigh the *design choices* you mention. Since everything is so inconsistent and fragmented I guess no one has to abide by `.seed()` anyways, e.g., https://github.com/openai/procgen/issues/21#issuecomment-575005387.

RedTachyon:
@JesseFarebro So if I understand correctly, your main objection is to the existence of the `seed` method, is that correct? In general I do see your point and can agree that in an ideal world, maybe this wouldn't be present, but then again, there's a whole ton of things I'd like to change about gym that are at this point so entrenched that it'd be necessary to just create a new library.

Other than that, as far as I understand, you agree that after creating the environment, it should hold an attribute `self.np_random: Generator` or `self._rng: Generator` or whatever you call it, and use it for whatever reset/step logic?

RedTachyon:
Oops, closed by accident. In any case, this is pending a decision what to do about the `seed` method

araffin:
> Oops, closed by accident. In any case, this is pending a decision what to do about the seed method

>This perfectly encapsulates everything that's wrong with .seed(). What even is this? Why would you want to do this? What would you expect seed() to do here? I guarantee it's not doing what you're expecting. In the ALE for example this would completely reset the environment and this would be equivalent to just instantiating the environment from scratch. The contract for seed() just makes no sense IMO. Even newer environments completely ignore seed(), e.g., Procgen won't do anything with .seed(), it requires a keyword argument instead.

@JesseFarebro the `env.seed()` method is very useful when you want to ensure reproducibility or put two environments in the same state.
An implicit constrain is that you need to call `env.reset()` afterward (we could make that constrain explicit?), as I agree `env.seed()` without reset does not make sense.

From an RL algorithm point of view, when the user passes a `env` object, you have no idea how it was created, how many wrappers it has around, and sometimes don't even have direct access to it (e.g., env launched in a subprocess, you want to avoid launching new process every time you want to be back in a certain env state), so `env.seed()` seems to be the only solution here (if you have better one, please share ;)).
And in case creating an environment is a costly operation, `env.seed()` avoid such overhead.

The only other solution I see is to force the user to provide an environment in a certain state to the algorithm (and force the user to have a `seed` argument in the env creation as you propose) but that does not solve the overhead of re-creating env.

RedTachyon:
FWIW in the meantime I'm just about convinced by @JesseFarebro's points about the `env.seed` method. There might be details we disagree about, but I'm honestly not sure, the method thing is the biggest thing atm.

@araffin The way I see it being changed at the moment is changing `env.__init__` and `env.reset` to accept a `seed` argument, and then the (re)seeding logic can be processed there. Since, as you say, `env.seed()` should require a `env.reset()` call afterwards, we can combine them within `env.reset()` and all the functionality is preserved. I suppose if an explicit seed is passed in `env.reset`, that's a signal that the env might have to rebuild whatever internals it has. Otherwise, you can just define it once on initialization and keep the seed at the default `None` for `env.reset`, and the behavior will be the same as currently. Does that make sense?

Of course there would be some deprecation message and a temporary compatibility layer. Maybe making `seed` basically call `reset` with the right argument?

jkterry1:
So after several offline discussions with people in this thread, I believe the prudent move is to make seed an argument to reset and deprecate the seed method with warning. The seed method can be kept in a deprecated state to ensure backwards compatibility for the foreseeable future and to ensure a smooth transition. I believe making this design decision makes the other concerns here moot, though please correct me if I'm wrong.

XuehaiPan:
> I believe the prudent move is to make `seed` an argument to `reset` and deprecate the `seed` method with warning.

https://github.com/openai/gym/blob/45902be4d082e4f2c289a449a30efbf76a0616a2/gym/core.py#L60-L74

As documented in `gym.Env.reset`, we should not set the RNG states at the beginning of every episode. The RNG states will change on the interaction with agents and the states are completely different between episodes (not need to reseed). We only need to seed the environment either in `__init__` or an explicit `seed()` call (only once at the beginning of training). If the default value of `seed` is `None` in the new version of `reset()`, the users should manage all seeds for every episode, otherwise, it is completely unreproducible.

------

It is easy to add functionality. If the users want to seed the environment at initialization or every `reset()`:

```python
class SeedAfterInit(MyEnv):
    def __init__(self, *args, seed=None, **kwargs):
        super().__init__(*args, **kwargs)
        self.seed(seed)


class SeedOnReset(gym.Wrapper):
    def reset(*args, seed=None, **kwargs):
        self.seed(seed)
        return super(*args, **kwargs)
```

However, it's hard to remove it if the new functionality has been added:

```python
class DoNotSeedOnReset(gym.Wrapper):
    def reset(*args, seed=None, **kwargs):
        seed = self.np_random.bit_generator._seed_seq.entropy
        return super(*args, seed=seed, **kwargs)  # if we force to set seed at reset
```

Besides, if the `seed` is an optional argument in `reset()`, how can we determine whether the users explicitly pass `reset(seed=None)`, or the users omitted it as `reset()` but they do not want to reseed it.

jkterry1:
In this proposal, users would have to repeatedly submit the same seed to reset the needed to run it again and the doc strings for reset would be changed. This change to reset and seed would represent rethinking how to do things a bit; Gym was hacked together years ago before it (or DRL) was widely used and then more or less abandoned so there is room for change and improvement.

XuehaiPan (commented):
Such new changes do not make sense to me. It looks like we removed `env.seed` method and make the base class's method `gym.Env.reset` behave like `seed`. This is just a rename. If we cannot leave the seeding functionality, why we should remove it?

1. Why `gym.Env.reset()` returns a `seed`? It should return an observation.
2. We only need to seed the environment once. When will we need to seed the environment on every episode? If we decide to remove the `seed` method, it should be put into `__init__` rather than `reset`.
3. How can we know:
    - The user omits the optional argument `seed` (call `env.reset()`): the user doesn't want to reseed the environment).
    - The user explicitly passes `None` as seed (call `env.reset(seed=None)`): the user wants to seed the environment as random as possible).

    Solution:

    ```python
    class Env:

        __marker = object()

        def reset(self, seed=__marker):
            if seed is not self.__marker:
                self.np_random, seed = seeding.np_random(seed)
    ```

    https://github.com/python/cpython/blob/3875a6954741065b136650db67ac533bc70a3eac/Lib/_collections_abc.py#L947-L961

4. There are many widely-used RL frameworks that rely on the `seed` method. Such as [RLlib](https://github.com/ray-project/ray) and [stable-baseline3](https://github.com/DLR-RM/stable-baselines3):
    - https://github.com/ray-project/ray/search?q=env.seed
    - https://github.com/DLR-RM/stable-baselines3/search?q=env.seed

    Since we removed the `seed` method in this PR (which is not backward compatible), it is meaningless to make our new RNGs (`np.random.Generator`) to be compatible with `np.random.RandomState`. It is not backward compatible anyway.

I can imagine the only result for these changes will be users in the community marking `gym <= 0.25.0` in their project requirements.

gym/utils/seeding.py
```diff
@@ -1,45 +1,120 @@
...
-import random as _random
 import struct
-import sys
+from typing import Optional
+
+import numpy as np
+from numpy.random import Generator
 
 from gym import error
+from gym.logger import warn
```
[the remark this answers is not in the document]
jkterry1: @RedTachyon

vwxyzjn:
Hey @RedTachyon, this PR looks good to me. One thing is that this PR contains a lot of discussions. Would you mind summarizing these changes for gym's users? Eventually this summary would be helpful for the release purposes.

Also, a quick question. As a user, it seems I would only to make the following changes, right?

```diff
 env = gym.make(gym_id)
-env.seed(1)
+env.reset(seed=1)
```

And once I set a seed with `env.reset(1)`, it's going to keep resetting with that seed until I specifically call `env.reset(seed=2)`, right?

XuehaiPan:
> And once I set a seed with env.reset(1), it's going to keep resetting with that seed until I specifically call env.reset(seed=2), right?

I'm afraid not. See point 2 and 3 in https://github.com/openai/gym/pull/2422#pullrequestreview-787800359.

------

In the old version, I can ensure reproducibility by seeding the environment at the beginning (only once):

```python
seed = args.seed  # seed from command line option

env = gym.make(<id>)
env = env.seed()  # only seed once

observation = env.reset()
done = False
while not done:
    action = policy(observation)

observation = env.reset()  # another episode (no need for reseeding)
done = False
while not done:
    action = policy(observation)
```

In the new version, I cannot ensure my reproducibility with the same code because the environment will be reset as random as possible:

```python
seed = args.seed  # seed from command line option

env = gym.make(<id>)

observation = env.reset()  # reset with `seed=None` as random as possible
done = False
while not done:
    action = policy(observation)

observation = env.reset()  # another episode, reset with `seed=None` as random as possible again
done = False
while not done:
    action = policy(observation)
```

gym/core.py
```diff
@@ -58,7 +63,7 @@ def step(self, action):
         raise NotImplementedError
 
     @abstractmethod
-    def reset(self):
+    def reset(self, seed: Optional[int] = None):
```
XuehaiPan, line 66:
```suggestion
    def reset(self, *, seed: Optional[int] = None):
```

Mark `seed` as a keyword-only argument. Same modification for all subclasses.

vwxyzjn:
@XuehaiPan that might not be true. You could reproduce using the new API with

```
seed = args.seed  # seed from command line option

env = gym.make(<id>)
env.reset(1)

observation = env.reset()  # reset with `seed=None` as random as possible
done = False
while not done:
    action = policy(observation)

observation = env.reset()  # another episode, reset with `seed=None` as random as possible again
done = False
while not done:
    action = policy(observation)
```


The reset function is

```
    @abstractmethod
    def reset(self, seed: Optional[int] = None):
        """Resets the environment to an initial state and returns an initial
        observation.
        Note that this function should not reset the environment's random
        number generator(s); random variables in the environment's state should
        be sampled independently between multiple calls to `reset()`. In other
        words, each call of `reset()` should yield an environment suitable for
        a new episode, independent of previous episodes.
        Returns:
        """
        # Initialize the RNG if it's the first reset, or if the seed is manually passed
        if seed is not None or self.np_random is None:
            self.np_random, seed = seeding.np_random(seed)
```

So since the `self.np_random` is set when you call the first reset, the first `env.reset(1)` effectively acts as `env.seed(1)`.

XuehaiPan:
> So since the self.np_random is set when you call the first reset, the first env.reset(1) effectively acts as env.seed(1).

Then, another question: what if I want to reseed the environment again as randomly as possible? I will need:

```python
del env.np_random
env.reset(seed=None)
```

RedTachyon:
> reseed the environment again as randomly as possible

What exactly do you mean by that? And specifically, how is it different from not reseeding it explicitly at all? 

(I'll try to address the other things soon)

XuehaiPan:
> What exactly do you mean by that?

@RedTachyon For example, I what to test my policy on randomly initialized starting state for multiple episodes (reseed the environment at the beginning of each episode with `None`). See point 3 in https://github.com/openai/gym/pull/2422#pullrequestreview-787800359.

Reseed the environment with `None`:

```python
for epi in range(<num_episode>):
    env.np_random = None
    observation = env.reset(seed=None)
    done = False
    while not done:
        action = policy(observation)
```

Otherwise, generate multiple seed first:

```python
seeds = [...]

for seed in seeds:
    env.np_random = None
    observation = env.reset(seed=seed)
    done = False
    while not done:
        action = policy(observation)
```

RedTachyon:
@XuehaiPan I still don't understand. What's the difference between reseeding "as randomly as possible" (which means you don't control the random seed and the PRNG), versus not reseeding at all (which means you use whatever the state of the PRNG is at the moment, i.e. you don't control the random seed and the PRNG).

I agree that there would be a difference in the specific numbers generated, but in either case they will be not reproducible since you're not controlling the seed.

XuehaiPan:
> What's the difference between reseeding "as randomly as possible" (which means you don't control the random seed and the PRNG), versus not reseeding at all (which means you use whatever the state of the PRNG is at the moment, i.e. you don't control the random seed and the PRNG).

I agree with that. What I'm concerned about is that the seeding update takes away the rights to seed the environment again with `None`.

araffin:
with this PR, what would be the way to seed multiple envs that are in different processes?

before:
```python
seed = 1
# at the beginning of training
# `env.seed(seed + rank)` is called in each process so next call to `reset()` will provide expected obs
vec_env.seed(seed) 

# later in the code, e.g., evaluating the agent
obs = vec_env.reset()
```

after?
```python
seed = 1
vec_env.reset(seed) # calling reset(seed + rank) without retrieving the obs?

# later in the code, e.g., evaluating the agent
obs = vec_env.reset()
```

XuehaiPan:
I'd suggest keeping the `seed()` method. If the user want to seed the environment on `reset`, just provide a built-in wrapper:

```python
class Env:
    def seed(seed=None):
        # keep this function

    def reset(self, *, **kwargs):  # only accept keyword arguments
        # do not modify this too


class SeedOnReset(gym.Wrapper):
    __marker = object()

    def reset(self, *, seed=__marker, **kwargs):
        if seed is not self.__marker:
            self.seed(seed)
        return super().reset(**kwargs)
```

RedTachyon:
General note: I won't be discussing *why* `env.seed` is being removed/integrated into `env.reset`, since I'm personally mostly indifferent to this and it's not my decision anyways. I think the comment that brought me towards the current solution is this: https://github.com/openai/gym/pull/2422#issuecomment-929318970 
Not quite sure if Jesse still stands by this solution (after offline discussions), but it's about the content of the message.

@vwxyzjn Some of the doubts might have been cleared up in the meantime, but as a summary of the changes:

1. The internals of seeding now use np.random.Generator as opposed to a np.random.RandomState, which follows the recommendations on how to use numpy RNG.
2. If you never used `env.seed`, and will not use `env.reset(seed=...)`, the behavior should be exactly the same - you just get a bunch of random numbers you have no control over.
3. If you used `env.seed(...); env.reset()`, now you write `env.reset(seed=...)` and should get the same behavior
4. If you used `env.seed(...)` in the middle of an episode, you're out of luck because this doesn't make any sense in a lot of environments. Then again, if you use it like this, you should probably stop anyways. (i.e. this usage is explicitly deprecated; see the rest of the discussion)
5. How the resetting works now (with this PR):
 - On the very first reset, you either give an explicit seed, which is used to initialize the RNG; or you don't, in which case RNG is initialized "however"
 - On any subsequent reset, you can either give an explicit seed, which restarts the RNG; or you don't, in which case RNG just goes on. This is equivalent to calling `env.reset()` before the PR without explicitly calling `env.seed()` before it.
 - In custom environments, you should call `super().__init__(seed=...)` to let this mechanism happen. Or you can just handle it another way, because some environments use an external process (e.g. Atari, Unity) and not the python-side `self.np_random`.


@XuehaiPan I still don't see why someone would need to explicitly reseed with `None`. What's the difference between doing this and doing nothing? Either way you get an uncontrolled PRNG.


@araffin Yea, your `after` seems about right. I don't see why you'd need to discard those observations, since you'll need them in whatever training procedure anyways, but you can. What's wrong with that version?

XuehaiPan:
> 3. If you used `env.seed(...); env.reset()`, now you write `env.reset(seed=...)` and should get the same behavior
> 4. If you used `env.seed(...)` in the middle of an episode, you're out of luck because this doesn't make any sense in a lot of environments. Then again, if you use it like this, you should probably stop anyways. (i.e. this usage is explicitly deprecated; see the rest of the discussion)
> 5. How the resetting works now (with this PR):
> 
> * On the very first reset, you either give an explicit seed, which is used to initialize the RNG; or you don't, in which case RNG is initialized "however"
> * On any subsequent reset, you can either give an explicit seed, which restarts the RNG; or you don't, in which case RNG just goes on. This is equivalent to calling `env.reset()` before the PR without explicitly calling `env.seed()` before it.
> * In custom environments, you should call `super().__init__(seed=...)` to let this mechanism happen. Or you can just handle it another way, because some environments use an external process (e.g. Atari, Unity) and not the python-side `self.np_random`.

Oppose for these which are backward incompatible. These need the downstream code base to be updated. For example, if I'm a package maintainer, I cannot arbitrarily set `gym>=0.19.0` (or what ever version before this PR being merged). I will need to add `if gym.__version__ ....` or `if hasattr(env, 'seed')` conditions in my code (same for other packages depend on `gym`).

------

> @XuehaiPan I still don't see why someone would need to explicitly reseed with `None`. What's the difference between doing this and doing nothing? Either way you get an uncontrolled PRNG.

When the environment have been seeded, the user cannot seed the environment again with `None`. Some third-party environments call `self.reset()` in `__init__` to make sure the environments have been reseted.

```python
class MyEnv(gym.Env):
    def __init__(self):
        super().__init__()
        self.reset(0)

env = MyEnv()
env.reset(seed=None)  # no effect for the seed
```

JesseFarebro:
My preference would be to keep `.seed()` over `.reset(*, seed)`. I don't think moving it to reset solves any of the issues I complained about earlier in this thread and it's just as confusing IMO. I think I underestimated all the edge cases people are using `seed` for (thanks to @XuehaiPan, @araffin, @RedTachyon for the discussions). I do have another proposal:

I feel like this same argument has been going on far longer than Gym. For example, this is an issue in the ALE and we have these two functions `cloneState` and `cloneSystemState`. Not very intuitive, but `cloneState` essentially serializes the state __WITHOUT__ the RNG state (i.e., you'll get a random seed upon restoring the state). `cloneSystemState` serializes the state __WITH__ the RNG state (i.e., upon restoring you'll get the exact same RNG sequence going forward).

Apart from the questionable naming of the system versus whatever non-system is, perhaps mandating proper serialization and copy support could help with some of these use cases. For example, something like,
```py
env = gym.make('...')

# Makes a copy of the environment where RNG is seeded with 1
env_copy_with_new_rng = env.copy(seed=1)

# Makes a copy but keeps the current RNG state
env_copy = env.copy()
```

I think this does a couple of things:

1) This solves my admittedly pedantic annoyances with the `seed` function and its purpose within Gym. I think it's reasonable to want to make a copy of an environment and potentially seed it. I would probably argue that the default behaviour for pickling etc. should be to serialize with RNG. If you need to do anything special you can always call `copy(seed)`.

2) Provides "native" support for copying environments post-initialization. As we've discussed in this thread this is something that's important for Gym as environment initialization can be really expensive. Supporting some type of native copy solidifies this is something environments should support. Environment authors can add further features, e.g., multi-agent case if it's really expensive to initialize the environment but we want a different policy for our opponent: `copy(opponent=...)`. 

Furthermore, I actually think this clears things up with the case where `seed=None` is supplied. `seed(seed=None)` is kind of weird but I understand how this would initialize with an arbitrary seed. I think it's clear that `copy(seed=None)` means that you should not re-seed the environment when you copy it. If you really need to be calling `.copy(seed)` repetitively then perhaps do something like:

```py
from numpy.random import SeedSequence

env = gym.make('...')
ss = SeedSequence()

envs = map(lambda seed: env.copy(seed=seed), ss.spawn(10))
``` 

This solution could also be somewhat backwards compatible. You can provide a default implementation for `copy` as:
```py
def copy(seed=None):
    _self = _copy.deepcopy(self)
    if seed is not None:
        _self.seed(seed)
    return _self
```
and slowly deprecate `seed` in favour of passing `seed` as an argument to the Environment constructor. In the case of the ALE we could now supply `seed` in `__init__`, keep `.seed()` for backwards compatibility, and support a custom `.copy()` which is more efficient than doing `.seed()` in our case.

RedTachyon:
I feel like there's a fundamental problem of how much `gym` should be handling, what assumptions it should make (and enforce), and where it should just users do whatever they want. Obviously any extreme will be bad, since too many assumptions can break some reasonable environments, and if there are no assumptions, then this already exists, and it's called Python.

I don't know to what extent this design work was done by OpenAI at the beginning, but probably not enough. And now we just kinda have to make it up as we go, to figure out the right balance for what actually exists. The only problem is that whenever some design change in either direction seems to make sense, it will likely be a breaking change, and that will inevitably make some people angry. But at the same time, a situation where we're stuck with an outdated architecture just because that's what ended up being made a few years ago is... not ideal.

XuehaiPan:
> I feel like there's a fundamental problem of how much `gym` should be handling, what assumptions it should make (and enforce), and where it should just users do whatever they want.

**tl;dr**: keep the interface as stable as we can. Minimize the impact of downstream code bases.

- new features: **welcome :)**.
- new methods / interfaces: **fine :)**.
- remove old methods / interfaces: **no :(**. (especially widely used by downstream code)

As an interface maintainer / designer, the interface definition should and must be stable (new: fine; remove/change: no). Suppose I create a new PR: **remove `observation_space` and `action_space`, rename it to `obs_space` and `act_space`**, I can imaging everyone will against this.

------

By now, there are 55 third-party environments listed in [Third Party Environments](https://github.com/openai/gym/blob/master/docs/third_party_environments.md) and 23.5k repositories explicitly mark `gym` as their dependency ([Used by `23.5k`](https://github.com/openai/gym/network/dependents?package_id=UGFja2FnZS01MjE5NTAzMA%3D%3D)). The number many be larger since many third-party envionmnets are not open sourced or haven't add to the document yet. And some repositories do not add `gym` as explicit dependency but depend on packages that using `gym`.

This PR contains two breaking changes: 1) replace `RandomState` with `Generator`; 2) remove `seed` method from `Env`. These will affect all users above (23.5k+). In my opinion, non of these changes is necessary.

### Change 1: replace `RandomState` with `Generator` (not necessary but fine with compatibility layers)

Commented in https://github.com/openai/gym/pull/2422#issuecomment-927305081:

> In [NEP 19 — Random number generator policy](https://numpy.org/neps/nep-0019-rng-policy.html):
> 
> > ***All current usages of `RandomState` will continue to work in perpetuity***, though some may be discouraged through documentation. Unit tests can continue to use the full complement of `RandomState` methods. Mixed `RandomState/Generator` code can safely share the common `BitGenerator` state. Unmodified `RandomState` code can make use of the new features of alternative BitGenerator-like settable streams.
> 
> NumPy will continue to keep `RandomState` as a stable API in the future. As `gym` is a widely used framework, I think we should provide our interfaces as stable as we can. I don't think it's a good idea to introduce these breaking changes.

This change is not necessary. If we change the behavior of `self.np_random` cause NumPy suggest us to do so. It's fine but please include compatibility layers (already included in this PR).

### Change 2: remove `seed` method from `Env` (not necessary for any reason)

As an interface definition, many RL environments is a subclass of `gym.Env`. That means: given a `env` instance, it always have:

```
properties:
    action_space: gym.space.Space
    observation_space: gym.space.Space
methods:
    reset: () -> observation
    step: (action) -> observation, reward, done, info
    close: () -> NoReturn
    seed: (seed) -> List[int]
    render: (mode) -> Any
```

It is a paradigm of environment designing and almost everyone follow that. It makes the downstream RL learning framework and thiry-party 

We are welcome to add new methods such as `Env.copy(seed=None)`, but do not remove old methods.

As an interface maintainer / designer, the interface definition should and must be stable (new: fine; remove/change: no). In this PR it is `Env.seed`. Suppose I create a new PR: **remove `observation_space` and `action_space`, rename it to `obs_space` and `act_space`**, I can imaging everyone will against this.

We have seen the painful transition from Python 2 to Python 3. Although there is a migration tool called `2to3`, why it took so long to migrate. The reason is Python 2 has a large code base and there is a large community rely on it. Here is the same situation for `gym.Env` (used by 23.5k+).

If we still want to make changes to `gym.Env`, please add a separated class with a suffix like `gym.EnvV2`. **Do not touch the old interface.** Otherwise, add a compatibility layer like `six`:

```python
if PY3:
    def iterkeys(d, **kw):
        return iter(d.keys(**kw))

    def itervalues(d, **kw):
        return iter(d.values(**kw))

    def iteritems(d, **kw):
        return iter(d.items(**kw))

    def iterlists(d, **kw):
        return iter(d.lists(**kw))

    viewkeys = operator.methodcaller("keys")

    viewvalues = operator.methodcaller("values")

    viewitems = operator.methodcaller("items")
else:
    def iterkeys(d, **kw):
        return d.iterkeys(**kw)

    def itervalues(d, **kw):
        return d.itervalues(**kw)

    def iteritems(d, **kw):
        return d.iteritems(**kw)

    def iterlists(d, **kw):
        return d.iterlists(**kw)

    viewkeys = operator.methodcaller("viewkeys")

    viewvalues = operator.methodcaller("viewvalues")

    viewitems = operator.methodcaller("viewitems")
```

Please add some migration variables like `gym.GYM025` etc.

jkterry1:
@JesseFarebro 

I do not believe that that should be added to Gym. Regarding your new comment about Atari not actually being incompatible with .seed(), I'm going to sit down at some point and think about if we still want to to include the breaking change component of the PR given this.

@XuehaiPan (and others) 

I disagree that making changes like this is an extraordinarily big deal. Every major Python library deprecates and removes/changes features over time, including ones infinitely bigger than us like NumPy or SciPy. This PR is deprecates a feature that will still continue to function for awhile (e.g. roughly 6 months in our case) before using a new function is required, this happens all the time. Additionally, changing existing code to use the new API here would require incredibly small changes that could be made in minutes, which is different from what happened with Python 3.6. Python makes many changes of this scope in sub versions, e.g. 3.6->3.7. I had a Zoom call with the RLlib developers last week (RLlib is the biggest RL library in the world by PyPI installs and as best I can tell has more Gym code than anyone else in the world) and mentioned this to them, and their response to this PR was "ya sure that's fine." Regarding changing the object name, this isn't really standard for marking these sorts of changes, the standard to mark this would be a major version change in Gym. The problem with this is that semantic versioning is super weird as a construct in Gym.

XuehaiPan:
> I disagree that making changes like this is an extraordinarily big deal.

@jkterry1 If the decision is made, it is fine for me. I'm not against any changes to the library. I'm arguing whether these changes are necessary? Especially for the core function.

I think the implementation of the seeding update is this PR could be better:

1. move `seed` to `__init__` rather than `reset` (point 2. in https://github.com/openai/gym/pull/2422#pullrequestreview-787800359).
2. came with point 1. above (point 3. in https://github.com/openai/gym/pull/2422#pullrequestreview-787800359), potential issues for https://github.com/openai/gym/pull/2422#issuecomment-968330221:

> > @XuehaiPan I still don't see why someone would need to explicitly reseed with `None`. What's the difference between doing this and doing nothing? Either way you get an uncontrolled PRNG.
> 
> When the environment have been seeded, the user cannot seed the environment again with `None`. Some third-party environments call `self.reset()` in `__init__` to make sure the environments have been reset.
> 
> ```python
> class MyEnv(gym.Env):
>     def __init__(self):
>         super().__init__()
>         self.reset(seed=0)
> 
> env = MyEnv()
> env.reset(seed=None)  # no effect for the seed
> ```

------

> I disagree that making changes like this is an extraordinarily big deal. Every major Python library deprecates and removes/changes features over time, including ones infinitely bigger than us like NumPy or SciPy. This PR is deprecates a feature that will still continue to function for awhile (e.g. roughly 6 months in our case) before using a new function is required, this happens all the time.

@jkterry1 Indeed, some libraries bigger than us deprecate features over time, but they still support the old functionality and just prompt a warning or nothing. How long will the old functionality supports? The answer for Python and/or NumPy maybe forever. Especially for the core functions.

For example, `threadings.Event.isSet` was deprecated in favor of name `is_set` since Python 2.6 ([Python 2.6 docs `threading.Event`](https://docs.python.org/2.6/library/threading.html#threading.Event.is_set)). But it is still supported by now in Python 3.10:

```pycon
Python 3.10.0 (default, Oct  4 2021, 17:55:55) [GCC 5.4.0 20160609] on linux
>>> import threading
>>> threading.Event.is_set
<function Event.is_set at 0x7fbf1c632050>
>>> threading.Event.isSet
<function Event.isSet at 0x7fbf1c6320e0>
>>>
```

Another example is the `numpy.random.RandomState`:

In [NEP 19 — Random number generator policy](https://numpy.org/neps/nep-0019-rng-policy.html):

> ***All current usages of `RandomState` will continue to work in perpetuity***, though some may be discouraged through documentation. Unit tests can continue to use the full complement of `RandomState` methods. Mixed `RandomState/Generator` code can safely share the common `BitGenerator` state. Unmodified `RandomState` code can make use of the new features of alternative BitGenerator-like settable streams.

All current usages of `RandomState` will continue to work in perpetuity.

XuehaiPan:
BTW, a **special reminder for breaking changes without backward compatibilities**: the source codes of many academic papers are open-source with warnings:

```diff
- "the source code is released as it is without any warranties"
```

jkterry1:
I briefly spoke to people at OpenAI about this and others like Jesse offline and I still believe this is the best course of action going forward

jkterry1:
The old seed method will continue to work until the Gym 1.0 release to support legacy code though. And again, updating code to use this change is absolutely tiny. I've spoken to the largest users of the Gym API in the world about this and other discussed API changes, and essentially their positions universally were "ya sure, this isn't a big deal"

araffin:
> The old seed method will continue to work until the Gym 1.0 release to support legacy code though.

Can we at least make the method forever deprecated as [numpy does it](https://numpy.org/doc/stable/reference/random/legacy.html)?

That would be nice for all the 24k projects that already exist... https://github.com/openai/gym/network/dependents?package_id=UGFja2FnZS01MjE5NTAzMA%3D%3D

efokschaner:
Hi I recently started working with gym, and have been having some fun getting fooled by what's in master with what's released while trying to integrate.

Take this all with a pinch of salt:

It seems that changing `reset()` from not implemented to having an implementation could be a little incompatible for classes that derive, as they will definitely not be calling the base class implementation currently.

If I could offer a suggestion, it would be to split a pure, implementation-free Env interface out from the common implementation stuff shared amongst envs.

Issue #2424, opened 2021-09-27
[Proposal] Relocate wrapper tests


RedTachyon:
### Proposal 

This is basically my inner pedant speaking. The tests for wrappers are all sitting in `gym/wrappers`, when there's a perfectly fine (and empty) `gym/wrappers/tests` directory sitting there, so they should be moved there.

### Motivation

Despaghettification of the file/directory structure

### Pitch

Put wrappers tests into `wrappers/tests`

### Alternatives

Live with it.

### Additional context

The one case where I see this being breaking in any way if someone imports the tests from gym with hardcoded paths and uses them in downstream code, which feels cursed and I really hope nobody does that.

Also in theory it's possible that one of the tests is written in a way that requires it to be in `gym/wrappers` (e.g. with relative imports), but from a quick glance I don't see anything like that, and if there were, it'd be worth fixing (and would trivially appear when running tests)

I can make a quick PR later today and pray that nothing breaks.

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/openai/gym/issues) in the repo (**required**)

jkterry1:
There's also a tests/gym/wrappers. My preference would to move _all_ tests into the format of tests/xxx, e.g. tests/wrappers, tests/envs, etc.

jkterry1:
Closing in favor of PR

PR #2425, opened 2021-09-27
Move all tests to gym/tests/...

RedTachyon:
See Issue #2424

This is kinda WIP, we have to make sure that all the tests actually trigger and pass.

Also from a quick local test it seems that some of the tests have internal imports. PyCharm did its best to convert them, but it might need some more attention.


jkterry1:
@RedTachyon this tentatively looks good to me, however the first 4 file renames shown in the files changed tab don't make a lot of sense to me?

RedTachyon:
@jkterry1 Do you mean `gym/envs/tests/__init__.py → gym/tests/__init__.py` etc?

It's marked as renames, but my guess is that it's the `diff` failing to understand moved files. There was a bunch of inits in the previous version under `gym/envs/tests`, `gym/wrappers/tests`, and now there's a bunch of inits under `gym/tests/envs`, `gym/tests/wrappers` and so on. So it tried to map the former to the latter, but it got a bit wonky in the process because in the end it's just all empty `__init__`s

jkterry1:
I mean things like "gym/spaces/tests/__init__.py → gym/tests/envs/__init__.py"

jkterry1:
Oh, no that should be a bad diff. You're right.

Issue #2412, opened 2021-09-17
[Bug Report] Recorded videos cannot be played on browsers anymore


eleurent:
Hi!

**Describe the bug**
Since the 0.20.0 release, whenever I generate a video through a Monitor and its [VideoRecorder](https://github.com/openai/gym/blob/master/gym/wrappers/monitoring/video_recorder.py#L23), they cannot be played on either Firefox or Chrome.

**Code example**
Here is a minimal example to reproduce the bug:

```python
import gym
from gym.wrappers import Monitor

env = gym.make("CartPole-v0")
env = Monitor(env, directory="test", video_callable=lambda e: True, force=True)
obs = env.reset()
done = False
while not done:
    action = 0 if obs[3] < 0 else 1  # a simple policy just to have longer videos
    obs, reward, done, info = env.step(action)
    env.render()
env.close()
```

**Video obtained with gym v0.20.0**

https://user-images.githubusercontent.com/1706935/133759181-bb60867b-1254-428a-b8fc-b43ae4f61778.mp4

**Video metadata**

```json
{
    "episode_id": 0,
    "content_type": "video/mp4",
    "encoder_version": {
        "backend": "ffmpeg",
[...]

PR #1893, opened 2020-04-24
Fixes video recorder ffmpeg on Centos7 and RHEL7

zlig:
* Conditionally select different arguments for ffmpeg
* Adds meaningful error from the third party application in charge of encoding

PR #2435, opened 2021-10-01
Fixes video recording formatting issues

zlig:
* Address issues reported in https://github.com/openai/gym/issues/2412 by:
  * Reverts some changes from #1893
  * Propose to install `imageio-ffmpeg` instead


vwxyzjn:
I can confirm this works for me in macOS! Thanks for the fix.

vwxyzjn:
@jkterry1 This LGTM.

PR #2440, opened 2021-10-07
Start/Offset Functionality for Discrete space

### LINKED ISSUE #2008: [Feature Request] Add range option to Discrete class
Hi,

Currently, `Discrete` spaces class can be quite restrictive and only allow the creation of positive values starting from zero values. Thus, to define an `action_space` with negative values, we must use `Box` spaces instead:

```
def __init__(self):
        self.action_space = spaces.Box(low=-1, high=1, shape=(1, ), dtype=np.int64)
```

and then, to get a scalar, we must select values by the index, as such:

```
def step(self, action):

        action = action[0]
```

Another option would be to extend the `Space` class and design it as desired (which is what I am doing). However, the alternative to also pass a range of values (e.g., `range(-1,2)`) to a `Discrete` space class would be much appreciated.

Thanks in advance

IshanManchanda:
Adds an optional start parameter to the Discrete space class which allows shifting the captured range from ``{0, 1, ..., n - 1}`` to ``{a, a + 1, ..., a + n - 1}``, closes #2008.

Added it as the last kwarg to not break existing functionality relying on the position of the seed kwarg. Also special-cased the repr function to omit the start value if it is 0 to retain previous behavior.

Progress:
- [X] Basic implementation, current tests passing
- [X] Add new parameters to existing tests in ``tests/spaces/test_spaces.py``
- [X] Add parameter description and example to ``docs/api.md``
- [X] Fix build (numpy int64 serialization) and pre-commit (black reformatting) errors

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ for episode in range(10):
 * [Stable Baselines 3](https://github.com/DLR-RM/stable-baselines3) is a learning library based on the Gym API. It is our recommendation for beginners who want to start learning things quickly.
 * [RL Baselines3 Zoo](https://github.com/DLR-RM/rl-baselines3-zoo) builds upon SB3, containing optimal hyperparameters for Gym environments as well as code to easily find new ones. Such tuning is almost always required.
 * The [Autonomous Learning Library](https://github.com/cpnota/autonomous-learning-library) and [Tianshou](https://github.com/thu-ml/tianshou) are two reinforcement learning libraries I like that are generally geared towards more experienced users.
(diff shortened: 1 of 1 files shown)


gym/spaces/discrete.py
```diff
@@ -5,19 +5,24 @@
...
         >>> Discrete(2)
+        >>> Discrete(3, start=-1)  # {-1, 0, 1}
 
     """
 
-    def __init__(self, n, seed=None):
+    def __init__(self, n, seed=None, start=0):
         assert n >= 0
         self.n = n
+        self.start = start
```
XuehaiPan, line 21:
```suggestion
        self.start = np.int64(start)
```

PR #2444, opened 2021-10-12
Update pendulum.py

zeeshanalipanhwar:
I have added the description of the class as much as I could, it needs to be completed.

### DIFF
--- a/gym/envs/classic_control/pendulum.py
+++ b/gym/envs/classic_control/pendulum.py
@@ -21,9 +21,9 @@ class PendulumEnv(gym.Env):
         2       Speed                     -8                       8
     Actions:
         Type: Box(2)
-        Num   Action (Torque)
-        0     -2
-        1     +2
+        Num   Action
+        0     Apply torque -2
+        1     Apply torque +2
     Reward:
     Starting State:
     Episode Termination:
--- a/gym/envs/classic_control/pendulum.py
+++ b/gym/envs/classic_control/pendulum.py
@@ -20,10 +20,9 @@ class PendulumEnv(gym.Env):
         1                                 -1                       1
         2       Speed                     -8                       8
     Actions:
-        Type: Box(2)
-        Num   Action
-        0     Apply torque -2
-        1     Apply torque +2
+        Type: Box(1)
+        Num     Action               Min                     Max
+        0       Torque               -2                       2
     Reward:
     Starting State:
     Episode Termination:
--- a/gym/envs/classic_control/pendulum.py
+++ b/gym/envs/classic_control/pendulum.py
@@ -14,11 +14,9 @@ class PendulumEnv(gym.Env):
         goal is to swing it up so it stays upright.
     Source:
     Observation:
-        Type: Box(3)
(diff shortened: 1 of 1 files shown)


jkterry1:
CAn you please make tests pass?

gym/envs/classic_control/pendulum.py
```diff
@@ -6,6 +6,28 @@
...
+        The inverted pendulum swingup problem is a classic
+        problem in the control literature. In this version of the
+        problem, the pendulum starts in a random position, and the
+        goal is to swing it up so it stays upright.
+    Source:
+    Observation:
+        Type: Box(3)
+        Num     Observation               Min                     Max
+        0                                 -1                       1
+        1                                 -1                       1
```
RedTachyon, line 20: The first two observations are missing, what is their interpretation?

gym/envs/classic_control/pendulum.py
```diff
@@ -6,6 +6,28 @@
...
+        2       Speed                     -8                       8
+    Actions:
+        Type: Box(2)
+        Num   Action
+        0     Apply torque -2
+        1     Apply torque +2
+    Reward:
+    Starting State:
+    Episode Termination:
```
RedTachyon, line 29: Reward, Starting State and Episode Termination are also missing, I'm pretty sure all of that can be easily found in the code.

gym/envs/classic_control/pendulum.py
```diff
@@ -6,6 +6,28 @@
...
+        goal is to swing it up so it stays upright.
+    Source:
+    Observation:
+        Type: Box(3)
+        Num     Observation               Min                     Max
+        0                                 -1                       1
+        1                                 -1                       1
+        2       Speed                     -8                       8
+    Actions:
+        Type: Box(2)
```
RedTachyon, line 23: This is incorrect, the action space is a `Box(1)`, with the sole action applying a continuous torque in the range [-2, 2]

gym/envs/classic_control/pendulum.py
zeeshanalipanhwar: Those are the Mass and the Length of the Pendulum.

RedTachyon:
>Those are the Mass and the Length of the Pendulum.

This doesn't seem right, why would the observation be mass and length? Those are constants of the environment.

RedTachyon:
@jkterry1 I'd suggest reverting this merge, or doing another PR which would do this correctly. My probelms with the current version:
- Observation space is specified as Box(1) while in reality it's Box(3)
- It's not well formatted, which means that tests will be failing for every PR branching from current master
- Reward isn't specified
- Starting State isn't specified
- Episode Termination isn't specified

Imo it's better to have no docstring compared to a docstring that's actively wrong. I can try to get a proper docstring together at some point, but especially for the time you're gone (if you're still around for now), I'd really prefer if this was reverted

jkterry1:
@RedTachyon just created and merged a PR reverting. If you could create a PR that does this correctly while I'm on vacation that'd be great.

PR #2449, opened 2021-10-16
Add Python 3.10, remove unittest envs

jkterry1:


jkterry1:
Apparently numpy isn't shipping Python 3.10 wheels yet, so this is on hold until that happens.

PR #2450, opened 2021-10-17
Update README.md

treyshotz:


jkterry1:
This isn't the standard syntax for pip install, exactly. Where this comes from is that macos switched from bash to zsh, which resulted in requiring those codes when pip installing things with brackets on macos and no other system.

PR #2454, opened 2021-10-20
Make AtariPreprocessing frameskip setting compatible with v5 envs.

heiner:
This makes the `frame_skip` (= action repeat) setting of the `AtariPreprocessing` wrapper compatible with the `ALE/*-v5` versions of the game.

The reason one might want to do that is that the wrapper applies random no-ops, which should happen at a low-level, before frame skipping.

The access to the private `_frameskip` is unfortunate but necessary with the current version of `atari-py`. If required, I can create a PR for the ALE repository that adds a `frameskip` `@property` to [AtariEnv](https://github.com/mgbellemare/Arcade-Learning-Environment/blob/8705b917e7bdadac133efe5069221ca060c49b31/src/gym/envs/atari/environment.py#L108)

### DIFF
--- a/gym/wrappers/atari_preprocessing.py
+++ b/gym/wrappers/atari_preprocessing.py
@@ -60,10 +60,14 @@ class AtariPreprocessing(gym.Wrapper):
         assert screen_size > 0
         assert noop_max >= 0
         if frame_skip > 1:
-            assert "NoFrameskip" in env.spec.id, (
-                "disable frame-skipping in the original env. for more than one"
-                " frame-skip as it will be done by the wrapper"
-            )
+            if (
+                "NoFrameskip" not in env.spec.id
+                and getattr(env.unwrapped, "_frameskip", None) != 1
+            ):
+                raise ValueError(
+                    "disable frame-skipping in the original env. Otherwise, more than one"
+                    " frame-skip will happen as through this wrapper"
+                )
         self.noop_max = noop_max
         assert env.unwrapped.get_action_meanings()[0] == "NOOP"
 


heiner:
> This is fine for the time being. Once `v0` and `v4` environments are fully deprecated and completely removed I plan on rewriting this wrapper and moving it over to the official ALE repository.

Thanks! And thanks for your work on this important, if understaffed, project!

The lint_python failure looks unrelated (fails at step `Run pip install -e .[nomujoco]` with error message `No matching distribution found for opencv-python>=3.`).

As a somewhat unrelated but still important note: Anecdotally, for training on Atari the _algorithm for downscaling frames_ appears to matter for performance (!?). This is why e.g. in [SeedRL](https://arxiv.org/pdf/1910.06591.pdf#page=17) they point out that they use "Resizing method: bilinear" which maps to [`cv2.INTER_LINEAR`](https://github.com/google-research/seed_rl/blob/f53c5be4ea083783fb10bdf26f11c3a80974fa03/atari/atari_preprocessing.py#L230) as opposed to [`cv2.INTER_AREA`](https://github.com/openai/gym/blob/3eb699228024f600e1e5e98218b15381f065abff/gym/wrappers/atari_preprocessing.py#L157). I expect this might be a reason for people to not use this "official" wrapper in its current form.

heiner:
> > The lint_python failure looks unrelated (fails at step Run pip install -e .[nomujoco] with error message No matching distribution found for opencv-python>=3.).
> 
> Ahh, seems the linting stage is running with Python v3.10.

Any chance we can fix that unrelated check failure?
 
> > As a somewhat unrelated but still important note: Anecdotally, for training on Atari the algorithm for downscaling frames appears to matter for performance (!?). This is why e.g. in SeedRL they point out that they use "Resizing method: bilinear" which maps to cv2.INTER_LINEAR as opposed to cv2.INTER_AREA. I expect this might be a reason for people to not use this "official" wrapper in its current form.
> 
> I would push back against this. Area resizing is the proper method to use, it maintains the on-screen proportions without adding fringing effects you'll see with bilinear. These details should be standardized but we've maintained that image post-processing like resizing should be left in "agent-space". I'm not sure why SeedRL took Dopamine's wrapper and changed the method of interpolation but I'm certain this shouldn't affect performance on average. If you have any more insight into this I'd love to hear why this decision was made.

Anecdotally, getting SOTA on Atari requires tweaking these details. I agree this isn't ideal, and it does show how brittle current RL approaches are. I believe Raphaël Marinier did that work in the case of SeedRL, I'll see if I can get him to comment here.

PR #2467, opened 2021-10-27
Make gym.Env an abstract class

jakkes:
`gym.Env` now inherits from `abc.ABC`, implying it (or any children) cannot be instanced unless all abstractmethods are implemented.

Also removed an unused import in `core.py`.


XuehaiPan:
Not sure if we need to mark some optional methods as `abstractmethod`, such as `gym.Env.render` and `gym.ActionWrapper.reverse_action`. Some simple env/wrapper does not implement these methods but works fine.

 We have too many:

```python
def render(self, *args, **kwargs):
    raise NotImplementedError

def reverse_action(self, *args, **kwargs):
    raise NotImplementedError
```

in tests.

vwxyzjn:
Although this is probably the more correct class design, I feel it has limited practical benefits. And Like @XuehaiPan suggested, marking the optional methods abstract would be a breaking change. Is there a particular use case for this change @jakkes?

jkterry1:
I agree with this, I'll reopen if a compelling arguemtn for this is made

PR #2469, opened 2021-10-29
Fix action dimension check bugs

Ericonaldo:
- **Gym bugs**: Mujoco and Robotic envs do not check the scalar action dimension at all.

- **Description**: I find that Mujoco and Robotic envs accept scalar action without any warning or errors. For example, in ''FetchReach-v1", the action dimension is 4, however a scalar input "0.1" or "[0.1]" is also legal inputs. This can lead to difficult-to-find dimension bugs if people set a wrong action (because I've made one).

- **Ways to reproduce**: 
```
import gym
env = gym.make(''FetchReach-v1") # Try any Mujoco and Robotic envs
env.reset()
env.step(0.1) # This is a legal input
env.step([0.1]) # This is also a legal input
```

- **Reason**: Some numpy operations broadcast the dimension automatically. 
For example,
in [mujoco_env.py](https://github.com/openai/gym/blob/a5c0608efbadfd0a1097aaff284885be18129427/gym/envs/mujoco/mujoco_env.py#L130), the code `self.sim.data.ctrl[:] = ctrl` broadcasts the dimension;
in [robot_env.py](https://github.com/openai/gym/blob/a5c0608efbadfd0a1097aaff284885be18129427/gym/envs/robotics/robot_env.py#L73), the code `action = np.clip(action, self.action_space.low, self.action_space.high)` broadcasts the dimension.

- **My commit**: check the dimension before the above codes as: 
`assert np.shape(action)[-1] == self.action_space.shape[-1], "Action dimension mismatch"`

### DIFF
--- a/gym/envs/robotics/robot_env.py
+++ b/gym/envs/robotics/robot_env.py
@@ -70,9 +70,8 @@ class RobotEnv(gym.GoalEnv):
         return [seed]
 
     def step(self, action):
-        assert (
-            np.shape(action)[-1] == self.action_space.shape[-1]
-        ), "Action dimension mismatch"
+        if np.array(action).shape != self.action_space.shape:
+            raise ValueError("Action dimension mismatch")
 
         action = np.clip(action, self.action_space.low, self.action_space.high)
         self._set_action(action)
--- a/gym/envs/robotics/robot_env.py
+++ b/gym/envs/robotics/robot_env.py
@@ -70,8 +70,10 @@ class RobotEnv(gym.GoalEnv):
         return [seed]
 
     def step(self, action):
-        assert np.shape(action)[-1] == self.action_space.shape[-1], "Action dimension mismatch"
-        
+        assert (
+            np.shape(action)[-1] == self.action_space.shape[-1]
+        ), "Action dimension mismatch"
+
         action = np.clip(action, self.action_space.low, self.action_space.high)
         self._set_action(action)
         self.sim.step()
--- a/gym/envs/robotics/robot_env.py
+++ b/gym/envs/robotics/robot_env.py
@@ -70,6 +70,8 @@ class RobotEnv(gym.GoalEnv):
         return [seed]
 
     def step(self, action):
+        assert np.shape(action)[-1] == self.action_space.shape[-1], "Action dimension mismatch"
+        
         action = np.clip(action, self.action_space.low, self.action_space.high)
         self._set_action(action)
         self.sim.step()
--- a/tests/envs/test_action_dim_check.py
+++ b/tests/envs/test_action_dim_check.py
@@ -1,0 +1,68 @@
+import pickle
+
+import pytest
+
+from gym import envs
+from tests.envs.spec_list import skip_mujoco, SKIP_MUJOCO_WARNING_MESSAGE
+
+
+ENVIRONMENT_IDS = (
+    "FetchPickAndPlace-v1",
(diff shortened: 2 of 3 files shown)


jkterry1:
@Ericonaldo please add tests for this functionality and make CI tests pass

Ericonaldo:
Black modified codes and add a test function in `tests/envs/test_action_dim_check.py`.

tests/envs/test_action_dim_check.py
```diff
@@ -0,0 +1,68 @@
...
+# @pytest.mark.skipif(skip_mujoco, reason=SKIP_MUJOCO_WARNING_MESSAGE)
+@pytest.mark.parametrize("environment_id", ENVIRONMENT_IDS)
+def test_serialize_deserialize(environment_id):
+    env = envs.make(environment_id)
+    print(
+        "Testing {}".format(
+            environment_id,
+        )
+    )
+    if env.action_space.shape[-1] <= 1:
```
vwxyzjn, line 53: What is this used for? It sounds like you are returning if the action space dimension is <= 1, which seems unnecessary as all environments should have at least 1 action space dimension.

gym/envs/robotics/robot_env.py
```diff
@@ -70,6 +70,10 @@ def seed(self, seed=None):
         return [seed]
 
     def step(self, action):
+        assert (
+            np.shape(action)[-1] == self.action_space.shape[-1]
```
vwxyzjn, line 74: Would it be better if we do `assert np.array(action).shape == self.action_space.shape`? I would also try to be more explicit about this, raising a `ValueError` instead of an assertion error.

tests/envs/test_action_dim_check.py
```diff
@@ -0,0 +1,68 @@
...
+    except:
+        pass
+    try:
+        obs = env.step([0.1])
+    except:
+        pass
+
+    assert not obs, "env {}: scalar input dimension does not check".format(
+        environment_id
+    )
```
vwxyzjn, line 68: I see what you are trying to do here. It seems a bit convoluted. I would just check this with a single env, like

```
env = gym.make("HalfCheetah-v2")
assert env.action_space.shape == (9,)

with pytest.raises(ValueError, match="Action dimension mismatch"):
    env.step([0.1]

```

vwxyzjn (commented):
Thanks for the change. I have left a few comments.

tests/envs/test_action_dim_check.py
Ericonaldo: Yeah, the bug only exists for those continuous envs whose action dim >1; for envs with action dim =1, there is no such a bug and therefore the check is unnecessary, so I add a judge.

gym/envs/robotics/robot_env.py
Ericonaldo: Great, I think raising a `ValueError` may be a better choice.

tests/envs/test_action_dim_check.py
Ericonaldo: Cool, lets' get it.

Ericonaldo:
Make the codes clear after the discussion.

vwxyzjn (approved):
LGTM @jkterry1

jkterry1:
@Ericonaldo can you please fix tests so I can merge?

Ericonaldo:
I add a mujoco check following existing test files.

Issue #2452, opened 2021-10-18
[Proposal] Adding type hints


jkterry1:
Per [this](https://sethmlarson.dev/blog/2021-10-18/tests-arent-enough-case-study-after-adding-types-to-urllib3) recently article, we really should work to add type hints to Gym over time. This is something that can be worried about after the major upcoming changes are planned, but none-the-less this is important for a package of Gym's scale. Incremental contributions towards this end are very welcome.

ikamensh:
@lebrice https://docs.python.org/3/library/typing.html#typing.runtime_checkable says: `Note runtime_checkable() will check only the presence of the required methods, not their type signatures.` - this seems like a really weak guarantee for compatibility, therefore maybe not something that should be recommended to gym users. I would be for the first option you have described.

> whether it would be best to make Envs generic in terms of the types of their observations/actions, or in terms of the type of their spaces

I think they should be generic in terms of Obs / Actions types, and specify that spaces share this generic typing as in your first message. Essentially spaces like `TupleSpace` should further restrict the type they can be generic (if type system allows this), therefore one should get type error for `Env[ndarray, ndarray]` with `TupleSpace`.

XuehaiPan:
> The `Tuple[float, float, float, float]` seems extremely unscalable, no? There are plenty of environments with significantly larger state spaces.

We can try:

```python
>>> from typing import Tuple

>>> Tuple[float, float, float, float]
typing.Tuple[float, float, float, float]

>>> Tuple[(float,) * 4]
typing.Tuple[float, float, float, float]

>>> Tuple[(float,) * 4] == Tuple[float, float, float, float]
True

>>> Tuple[(float,) * 10]
typing.Tuple[float, float, float, float, float, float, float, float, float, float]
```

I'd prefer using [`numpy.typing.NDArray`](https://numpy.org/devdocs/reference/typing.html#numpy.typing.NDArray) rather than `typing.Tuple`. Cause `gym.space.Space` usually returns a `numpy.ndarray` instead of `bultins.tuple`.

However [`numpy.typing.NDArray`](https://numpy.org/devdocs/reference/typing.html#numpy.typing.NDArray) requires `numpy >= 1.21.0`, and it do not support specifying the `shape` of the array only the `dtype`. [PEP 646 -- Variadic Generics (Python >= 3.10)](https://www.python.org/dev/peps/pep-0646).

PR #2477, opened 2021-10-30
Adds type checks via `pyright` to CI

ikamensh:
Contribution towards #2452
Adds type checks via `pyright` - first step for having typing in gym, and two first typechecked files (version.py and logging.py).

While `mypy` can be considered official / reference type checker, in practice it has difficulties supporting all accepted typing PEPs without workarounds in user code. As mentioned in the discussion in #2452, I have much better experience in production with `pyright` - it's signal to noise ratio is much higher. Most reported issues are actual deficiencies / missing typing. 

Advantages:
1) Pyright is well maintained, few open issues.
2) Developers in Visual Studio Code can have SAME type checking right in their IDE.

Known drawbacks:
1) I couldn't find a working way of including `pyright` in pre-commit.


gym/logger.py
```diff
@@ -9,42 +10,47 @@
 ERROR = 40
 DISABLED = 50
 
-MIN_LEVEL = 30
```
ikamensh, line 12: type checker: modifying a constant (because fo SNAKE_CASE)

jkterry1:
This looks good, but can you please merge the new .md file you propose into CONTRIBUTING.md?

ikamensh:
done

jkterry1:
Thanks a ton for all your PRs!

Could you please fix the merge conflict? It looks like merging your other PRs resulted in one

ikamensh:
You're welcome! :) Conflict + CI should be fixed now

PR #2482, opened 2021-11-06
Fix the broken link to Creating Environments

JojiJoseph:
Changed *Creating Environments* link from https://github.com/openai/gym/blob/master/docs/creating-environments.md to https://github.com/openai/gym/blob/master/docs/creating_environments.md, as the change in file name isn't reflected in CONTRIBUTING.md.


jkterry1:
Thanks!

PR #2488, opened 2021-11-16
Typing in gym.core.Env

ikamensh:
Adds type hints to `gym.core.Env` class. Type checked with pyright.
Part of #2452 . Implementation inspired by comment from @lebrice

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -1,11 +1,17 @@
+from __future__ import annotations
+
 from abc import abstractmethod
+from typing import TypeVar, Generic, Tuple
 
 import gym
-from gym import error
-from gym.utils import closer
+from gym import error, spaces
 
 
-class Env:
+ObsType = TypeVar("ObsType")
+ActType = TypeVar("ActType")
+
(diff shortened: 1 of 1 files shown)


RedTachyon:
@ikamensh regarding #2486, #2487, #2488, #2489, could you please put them all together into a single PR? It would be much easier to process and review, especially since some of these PRs don't actually work in isolation. Also it would be easier to avoid a merge conflict hell with other work going on in the meantime.

jkterry1:
@ikamensh seconded, that would be very appreciated. Sorry it took as so long to get to look at these

ikamensh:
Can do, but I actually thought it would be easier with split PRs - if some of them need rework, parts can already be merged and therefore less work in progress needs rebasing. I'll have time in a week or so but if some of it can be merged as is would also be appreciated.

jkterry1:
All of them but the space PR (#2487) have merge conflicts; Ariel could you please take a look at that one in the mean time?

PR #2491, opened 2021-11-17
Type cast in `spaces` families

### LINKED ISSUE #2140: `Tuple` __eq__ fails when `spaces` aren't of the same type
When creating a `gym.spaces.Tuple`, if a list of spaces is passed as the `spaces` argument, I believe it should be converted to a tuple, as otherwise the space could be modified, which doesn't seem like an intended behaviour, and comparisons with other equivalent `spaces.Tuple` objects, but created with tuple arguments, will fail:

```python
inner_spaces = [
        Box(0, 1, (2,2)),
        Discrete(2),
        Box(0, 1, (2,2)),
]
space_1 = spaces.Tuple(list(inner_spaces))
space_2 = spaces.Tuple(tuple(inner_spaces))
assert space_1 == space_2 # This fails
assert space_2 == space_1 # This also
```

The way I see it, this can be easily fixed in one of two ways:

1.  Convert the `spaces` argument to a tuple in the constructor:
    ```python
    def __init__(self, spaces):
        self.spaces = tuple(spaces)
        for space in spaces:
            assert isinstance(space, Space), "Elements of the tuple must be instances of gym.Space"
        super(Tuple, self).__init__(None, None)
    ```
    However this might change the behaviour of some programs which depend on being able to modify `Tuple` spaces.

2.  Convert the `spaces` argument to a tuple just when comparing:
    ```python
    def __eq__(self, other):
        return isinstance(other, Tuple) and tuple(self.spaces) == tuple(other.spaces)
    ```
    This won't copy anything if `self.spaces` is already a tuple, however for lists its a small copy.
    
Let me know what you think!

### LINKED ISSUE #2479: [Bug Report] Equals Method of gym.spaces.Tuple does not Work Correclty
**Describe the bug**
Equals methods of `gym.spaces.Tuple` does state that two Space instances do not equal if they contain the same Space but instance a was initialized using a list of subspaces and b using a tuple of instances. 

Suggestion:
1. Convert lists to a tuples in` __init__`
2. Raise error if list is passed to `__init__`
3. Change `__eq__`

**Code example**
```python
from gym.spaces import Tuple
t1 = Tuple(tuple())
t2 = Tuple(list())

t1 == t2
```

**System Info**
 * `pip install gym==0.15.3`
 *  openSUSE Leap 15.3
 * Python version: 3.6.13

XuehaiPan:
## 1. `spaces.Dict`
  
in `__init__`:

  - convert `tuple` to `OrderedDict` as well (same as `list`).
  - raise `AssertionError` when the user does not pass a `dict` or `list` / `tuple` of keyvalue pairs.

------

## 2. `spaces.Tuple`:

in `__init__`:

  - convert input spaces to `tuple`.

in `contains`:

  - convert the input sample to `tuple`.

Fixes #2140.
Fixes #2479.

------

## 3. `spaces.MultiBinary`

in `__init__`:

  - convert input shapes to tuple of `int`s.
  - raise `AssertionError` when any entry of the input shapes is not positive.

------

## 4. `spaces.MultiDiscrete`

in `__init__`:

  - make a copy of the input `nvec`.

    ```python
    >>> import numpy as np
    >>> a = np.array([1, 2, 3], np.int64)
    >>> np.asarray(a) is a
    True

    >>> np.asarray(a)[0] = 4
    >>> a
    array([4, 2, 3])
    ```

  - raise `AssertionError` when any entry of the input shapes is not positive (convert to `int`s first).

in `contains`:

  - convert `tuple` to `np.ndarray` as well (same as `list`).

------

## 5. `spaces.Discrete`

in `__init__`:

  - convert the input shape `int`.
  - raise `AssertionError` when the input shape is not positive (convert to `int`s first).

### DIFF
--- a/gym/spaces/dict.py
+++ b/gym/spaces/dict.py
@@ -1,5 +1,5 @@
 from collections import OrderedDict
-from collections.abc import Mapping
+from collections.abc import Mapping, Sequence
 import numpy as np
 from .space import Space
 
@@ -42,9 +42,15 @@ class Dict(Space, Mapping):
         if spaces is None:
             spaces = spaces_kwargs
         if isinstance(spaces, dict) and not isinstance(spaces, OrderedDict):
-            spaces = OrderedDict(sorted(list(spaces.items())))
-        if isinstance(spaces, list):
+            try:
+                spaces = OrderedDict(sorted(spaces.items()))
+            except TypeError:  # raise when sort by different types of keys
+                spaces = OrderedDict(spaces.items())
+        if isinstance(spaces, Sequence):
             spaces = OrderedDict(spaces)
+
+        assert isinstance(spaces, OrderedDict), "spaces must be a dictionary"
+
         self.spaces = spaces
         for space in spaces.values():
             assert isinstance(
--- a/gym/spaces/discrete.py
+++ b/gym/spaces/discrete.py
@@ -16,8 +16,9 @@ class Discrete(Space):
     """
 
     def __init__(self, n, seed=None, start=0):
-        assert n >= 0 and isinstance(start, (int, np.integer))
-        self.n = n
+        assert n > 0, "n (counts) have to be positive"
+        assert isinstance(start, (int, np.integer))
+        self.n = int(n)
         self.start = int(start)
         super().__init__((), np.int64, seed)
 
--- a/gym/spaces/multi_binary.py
+++ b/gym/spaces/multi_binary.py
@@ -1,3 +1,4 @@
+from collections.abc import Sequence
 import numpy as np
 from .space import Space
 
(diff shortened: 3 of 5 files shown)


RedTachyon (approved):
Overall looks good, one point around the code in this PR - do we need the `OrderedDict`s in `spaces.Dict`? Since Python 3.7 (i.e. the earliest version that's actually supported by gym), dicts are guaranteed to be ordered, so I'd probably prefer just making everything into a `dict`. Not sure if we need to do that in this PR though.

One other minor comment left as review

jkterry1:
@RedTachyon are you happy with it now?

RedTachyon:
@jkterry1 I think can be merged now, there's still a conversation to be had about dict vs OrderedDict, but that can be done in another PR

jkterry1:
@RedTachyon can you please create an issue regarding the OrderedDict thing so it doesn't get forgotten?

Issue #2494, opened 2021-11-20
Unnecessary pyright tests in CI?


jkterry1:
@ikamensh 

Hey, I really appreciate your PR adding pyright and type hint checking in general, but I have a concern. Right now, you have type hint tests running on Linux, macOS and Windows VMs. Could you please create a PR changing this to only Linux? Gym doesn't even officially support Windows and different operating systems won't effect type hints. Also they add a bunch of extra github actions costs. Additionally, is even running type hint checks on different versions of Python necessary?

ikamensh:
> different operating systems won't effect type hints.
I picked up running on different OS from official pyright documentation; I think there are minor differences in the typing of python interpreter itself / some libraries per each OS. But I agree, likely none of them matter for gym.

Re: python version - This sometimes helps catch lack of support for some python version, e.g. here https://github.com/openai/gym/runs/4249041761?check_suite_focus=true . I'm not sure how important it is to save these gh action costs. Maybe a compromise is to run on latest and oldest python versions, e.g. 3.7 and 3.10 now.

PR #2495, opened 2021-11-21
Reduce number of pyright checks

software-samurai:
Hello, I have implemented this per my understanding of issue #2494 .


jkterry1:
Thanks a ton for this, this like like it should be valid, I'll merge if/when CI passes

PR #2502, opened 2021-12-02
[Space] Permutation

zuoxingdong:
The motivation of this PR is to add a new space for permutation operations that cannot be done with `MultiDiscrete`. 

It is useful for some environments in which the actions are to permutate a set of objects.

### DIFF
--- a/gym/spaces/permutation.py
+++ b/gym/spaces/permutation.py
@@ -7,7 +7,7 @@ class Permutation(Space):
         assert n >= k, "`n` has to be no less than `k`"
         self.n = n
         self.k = k
-        self.nvec = np.array([n]*k, dtype=dtype)
+        self.nvec = np.array([n] * k, dtype=dtype)
         super().__init__(shape=self.nvec.shape, dtype=dtype, seed=seed)
 
     def sample(self):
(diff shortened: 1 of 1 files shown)


jkterry1:
So in general I very much support adding new spaces that enable new forms of functionality. However, I have questions:

1. What does this allow, in simple English, that multidiscrete does not?
2. Why can't multidiscrete be reasonably extended to support this?
3. Given the tons of new methods added to many spaces since I took over Gym, does anything need to be added or changed here to be in line with other methods?
4. What's up with the jsonable methods?
5. Are you currently aware of active use cases in RL that would benefit from this space? This isn't a requirement, but I would like to know and I'm not aware of any.
6. This would require writing fairly specialized learning code, yes?

RedTachyon:
My understanding is that this is essentially multidiscrete, but with the constraint that each element is unique. I'm honestly not sure if this is necessary on top of the regular multidiscrete. Any valid permutation would also be a valid multidiscrete, so the only additional functionality it would give is (a) checking if a thing is a permutation (does it need to be done inside gym?), and (b) sampling a permutation directly from the space (which might be nice in some applications, but because sampling from a space in actual learning code is lowkey cursed, isn't a big argument for me)

*If* it is decided that it should be added, it desperately needs more comments and documentation as to what it actually does.

jkterry1:
Hey, I haven't heard anything in a week so I'm going to close this, if you want to comment about this further I'd be happy to listen

PR #2507, opened 2021-12-07
Fix close_extras not implemented by default in VectorEnv

tristandeleu:
- Fix `close_extras` not implemented by default in `VectorEnv`. For example:
```python
class CustomVectorEnv(gym.vector.VectorEnv):
    def __init__(self, num_envs):
        observation_space = gym.spaces.Box(0., 1., shape=(2,))
        action_space = gym.spaces.Discrete(3)
        super().__init__(num_envs, observation_space, action_space)

env = CustomVectorEnv(num_envs=5)
# Before:
# Traceback (most recent call last):
#   File "gym/vector/vector_env.py", line 139, in __del__
#     self.close(terminate=True)
#   File "gym/vector/vector_env.py", line 120, in close
#     self.close_extras(**kwargs)
#   File "gym/vector/vector_env.py", line 97, in close_extras
#     raise NotImplementedError()
# NotImplementedError:
```

### DIFF
--- a/gym/vector/async_vector_env.py
+++ b/gym/vector/async_vector_env.py
@@ -477,6 +477,10 @@ class AsyncVectorEnv(VectorEnv):
         logger.error("Raising the last exception back to the main process.")
         raise exctype(value)
 
+    def __del__(self):
+        if not getattr(self, "closed", True):
+            self.close(terminate=True)
+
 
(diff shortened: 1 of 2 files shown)


vwxyzjn:
This LGTM. @jkterry1

PR #2509, opened 2021-12-09
Adding documentation for CartPole-v1 in docs/classic_control

kingb12:
Hi! I've completed a first document, let me know how it looks and I'll apply any feedback you have to this one and other classic control problem documents I'm writing up.

Changes:
- new document for CartPole-v1 in `docs/classic_control`
- minor tweaks to docstrings in `cartpole.py`
- minor tweaks to doc for Blackjack (v0 throws a deprecation error).
I was able to source some of this from the [prior wiki](https://github.com/openai/gym/wiki/CartPole-v0)

I assumed a value that that is beyond the allowed thresholds for pole angle, x-position, etc and causes the episode to terminate is outside the observation space. e.g. the pole can be at +/- 12 degrees, but a prior doc indicated a range for observed pole angle of +/- 24 degrees. My assumption was this reflects changes to the problem difficulty, but let me know if I'm actually reporting the observation values incorrectly.

### DIFF
--- a/docs/classic_control/cartpole.md
+++ b/docs/classic_control/cartpole.md
@@ -1,0 +1,65 @@
+CartPole-v1
+---
+|Title|Action Type|Action Shape|Action Values|Observation Type| Observation Shape|Observation Values|Average Total Reward|Import|
+| ----------- | -----------| ----------- | -----------|-----------| ----------- | -----------| ----------- | -----------|
+|CartPole-v1|Discrete|(1,)|(0,1)| Box |(4,)|[(-4.8,4.8),(-inf,inf), (~ -0.2095, ~ 0.2095), (-inf, inf)]| |`from gym.envs.classic_control import cartpole`|
+---
+
+### Description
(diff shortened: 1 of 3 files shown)


jkterry1:
Formatting looks good, I'll have someone else review the angle stuff. Please message me when you're all done or I have to do something, I don't regularly check my email that github notifications are sent to because there's so many of them

kingb12:
Thanks for the input all! I added a section noting the discrepancy:


> **Note:** above denotes the ranges of possible observations for each element, but in two cases this range exceeds the
range of possible values in an un-terminated episode:
> - `*`: the cart x-position can be observed between `(-4.8, 4.8)`, but an episode terminates if the cart leaves the
`(-2.4, 2.4)` range.
> - `**`: Similarly, the pole angle can be observed between  `(-.418, .418)` radians or precisely **±24°**, but an episode is 
terminated if the pole angle is outside the `(-.2095, .2095)` range or precisely **±12°**

I also emphasized the exact range is defined by the degree values (vs. radians) in case any methods need to normalize over the range, etc. Let me know if there are any other steps, I'll open a separate PR for the other two docs unless you all prefer otherwise

jkterry1:
Your version history description doens't make sense? It doesn't describe the change that resulted in v1

kingb12:
Sorry it took me a bit to come back to this, just fixed the version history

jkterry1:
This looks great, thanks a ton. Could you please do the other environments I messaged you about in new PRs?

PR #2511, opened 2021-12-09
Mark `seed` a keyword-only argument in `reset()`

XuehaiPan:
Since PR [Seeding update](https://github.com/openai/gym/pull/2422) got merged, the seeding functionality has been moved to `Env.reset` method.

In the old implementation, the user could write:

```python
env.seed(42)
observation = env.reset()
```

with no ambiguity. Besides, the old signature of `Wrapper` is `Wrapper.reset(**kwargs)`, no positional arguments are supported.

Now they could have:

```python
observation = env.reset(42)
```

------

In this PR, the `seed` argument is marked as keyword-only, the users should know what they are exactly doing:

```python
observation = env.reset(seed=42)
```


jkterry1:
I agree that it should be a keyword only argument; @RedTachyon will have to integrate this into his kwargs work with reset

jkterry1:
Closed in favor of Ariel's PR

Issue #2455, opened 2021-10-21
Better error messages for capitalization variance in Gym


jkterry1:
I just spent 15 minutes debugging, only to realize the issue was that I was typing `pendulum-v1` instead of `Pendulum-v1`. These names really ought to be capitalization invariant.

jkterry1:
error messages that responsibly handle this is probably better, yes

kir0ul:
https://github.com/openai/gym/pull/2535 should take care of this issue now.

PR #2513, opened 2021-12-09
Suggest matching environment independent of capitalization

kir0ul:
This PR fixes https://github.com/openai/gym/issues/2455.

### DIFF
--- a/gym/envs/registration.py
+++ b/gym/envs/registration.py
@@ -1,11 +1,11 @@
-import contextlib
+import re
+import sys
 import copy
 import difflib
 import importlib
 import importlib.util
-import re
-import sys
-from typing import Callable, Dict, Generator, Optional, Set, Tuple, Type, Union
+import contextlib
+from typing import Callable, Type, Optional, Union, Dict, Set, Tuple, Generator
 
 if sys.version_info < (3, 10):
     import importlib_metadata as metadata  # type: ignore
@@ -16,7 +16,7 @@ from collections import defaultdict
 from collections.abc import MutableMapping
 from operator import getitem
 
-from gym import Env, error, logger
+from gym import error, logger, Env
 from gym.envs.__relocated__ import internal_env_relocation_map
 
 # This format is true today, but it's *not* an official spec.
--- a/gym/envs/registration.py
+++ b/gym/envs/registration.py
@@ -5,39 +5,28 @@ import difflib
 import importlib
 import importlib.util
 import contextlib
-from typing import Callable, Type, Optional, Union, Dict, Set, Tuple, Generator
+from typing import (
+    Callable,
+    Type,
+    Optional,
+    Union,
+    Tuple,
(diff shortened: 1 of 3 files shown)


jkterry1:
Please use classic control, toy text or Atari environments for the tests (others will have some changes in the near future)

jkterry1:
(also tests are failing)

JesseFarebro (changes requested):
We can get something more robust by using the standard library `difflib`. In particular, you can do,

```py
import difflib

matches = difflib.get_close_matches(id, self.env_specs.keys())
```

This will apply to more than just lower-case correction as is implemented in this PR. If the list of matches is non-zero you could just format the list and provide this to the user.

kir0ul:
> We can get something more robust by using the standard library `difflib`. 

Thanks for the suggestion, I didn't know about `difflib`.

tests/envs/test_registration.py
```diff
@@ -24,11 +26,30 @@ def __init__(self, arg1, arg2, arg3):
 
 
 def test_make():
-    env = envs.make("CartPole-v0")
-    assert env.spec.id == "CartPole-v0"
+    env = envs.make("CartPole-v1")
```
JesseFarebro, line 29: Could you revert these changes to `-v0`?

gym/envs/registration.py
```diff
@@ -1,10 +1,11 @@
-import re
-import sys
+import contextlib
```
JesseFarebro, line 1: Could you revert the `isort` changes here?

tests/envs/test_registration.py
```diff
@@ -89,7 +110,7 @@ def test_missing_lookup():
 def test_malformed_lookup():
     registry = registration.EnvRegistry()
     try:
-        registry.spec(u"“Breakout-v0”")
+        registry.spec("“Breakout-v0”")
```
JesseFarebro, line 113: Could you revert this change? We can include this in a different PR where we clean up these issues.

tests/envs/test_registration.py
```diff
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
+import pytest
```
JesseFarebro, line 2: Could you revert this `isort` change?

JesseFarebro (changes requested):

tests/envs/test_registration.py
kir0ul: The thing is, if I put back the `u` before the string, the `black` pipeline fails, so I'd leave it this way.

kir0ul:
Oops, was working on the wrong branch and pushed your changes here instead of in https://github.com/openai/gym/pull/2535. Will redo it in the right branch.

kir0ul:
I think we can close this since all the code and tests have been copied in https://github.com/openai/gym/pull/2535.

PR #2514, opened 2021-12-09
Remove DiscreteEnv class

### LINKED ISSUE #2381: [Question] What should be done with the GoalEnv and DiscreteEnv classes?
Right now, Gym has a GoalEnv class and Env class as base classes in core.py. The GoalEnv class was added as part of the robotics environments, and impose special requirements on the observation space. From what I can tell, this class has not been used outside of Gym's robotics environments and is largely unnecessary. Unless I'm missing something here, removing this class sounds like the way to go.

carlosluis:
Remove the `DiscreteEnv` class and modify the following environments:

- `CliffWalking`
- `FrozenLake`
- `Taxi`

to now use the regular `Env` class.

Testing: ran small driver script taking one environment step in each case and validated the env.render() output

Fixes #2381


RedTachyon:
Seems good, the test suite covers these environments so the interface should still work. The change is simple enough that the actual behavior should be the same as well

ZhiqingXiao:
@carlosluis  @RedTachyon @jkterry1 

I do not think that it is a good revision. I suggest reverting this merge.

Reasons:
- The class DiscreteEnv is a good abstract of finite MDP.
- We can check the suitability of a unified agent for finite MDP by checking whether an environment is derived from the class DiscreteEnv.
- Removing the class DiscreteEnv introduces obvious code duplications.

carlosluis:
You bring important points about the usefulness of DiscreteEnv.

My initial feeling when reading #2381 was that the small amount of code duplication outweights having to maintain a separate env class, also considering only 3 environments would benefit from it.

I wouldn't mind reverting the change as you bring some fair points, but I'll let @jkterry1 or @RedTachyon  decide on this since they may have additional context I'm missing.

jkterry1:
So I read this, and I still think that the small amount of code duplication is worth it. My main motive for this was the have a standardized class structure going forward. I'm not convinced that having separate classes to model every type of MDP is a good idea, and I have a strong suspicion that this was ever an intended feature of this class having spoke to folks at OpenAI about this.

RedTachyon:
So I missed the initial conversation on this, and in my review I mostly stuck with a purely technical analysis, but tbh I agree with @ZhiqingXiao 

In my opinion, an enormous part of gym's appeal and usefulness is providing a common API and a pattern for RL environments, that is `gym.Env`. Discrete MDPs are not that popular for DRL, but still an important step in RL as a whole, and the DiscreteEnv pretty nicely captures their logic, handling all the boilerplate, again providing a common pattern

As long as DiscreteEnv is a subclass of Env (and it is/was), this is still perfectly compatible with everything else and doesn't add any extra maintenance - the external interface should be identical, and the internal code is still there, just duplicated.

That being said, I figured that this ultimately isn't a big deal either way, so I was willing to just go with the conclusion from #2381. But if there are other voices for bringing it back, I can support it

PR #2515, opened 2021-12-09
Add options to the signature of `env.reset`

### LINKED ISSUE #2399: [Proposal] Custom arguments in step and reset methods
### Proposal 

Add *args and **kwargs to the signatures of `step` and `reset` (or just **kwargs)

### Motivation

Lowkey inspired by #2396 and taking it further, but also by my previous work and thoughts.

Essentially, it is sometimes helpful to pass additional information to the environment that is non-agent-specific, either during the step, or when starting a new episode.

I'd argue this doesn't add any complexity, but rather opens the door to better flexibility when defining custom environments. It can be entirely ignored, also in already existing environments.

### Alternatives

In principle, a solution is just ignoring the problem. Python doesn't care that much about signatures of functions in subclasses, so if you can ignore the IDE screaming at you when working with a custom environment, it should be fine.

Other than that, there are a few hacks possible like changing the action space, but they're exactly that, hacks.

The main way is probably creating a new environment whenever one needs to have new kwargs passed to the environments. Which is fine for Cartpole, but a massive nuisance when creating a new environment is a costly process (e.g. running a game executable)

### Additional context

There was a discussion of a similar nature in issue #337, but it's been 5 years and a pandemic, so I'd like to revisit this.

### Checklist

- [x] I have checked that there is no (recent) similar [issue](https://github.com/openai/gym/issues) in the repo (**required**)

RedTachyon:
Pretty self-explanatory. 

The implementation is straight-forward for regular environments, since it only adds logic that is not used in the built-in environments by default.

A slightly tricky thing is in vector environments, where kwargs get passed both into `reset_async` and `reset_wait`, because the two different classes process them differently. 

As with any "big" change, this definitely needs some more eyes on it.

This closes #2399 (at least to the extent that was agreed in the meantime), and subsumes #2511 for simplicity (easier to just include it here than deal with merge conflicts, or wait for synchronization between merges)


araffin:
Hello,
i'm curious to know the argument that changed this api change status from refused to approved.
A similar discussion: https://github.com/DLR-RM/stable-baselines3/issues/650

In short: 
- you don't need kwargs if you have access to env methods
- using kwargs obfuscates the code (you can litteraly pass anything) and prevents you from doing any type checking

RedTachyon:
@araffin The main point is doing things like curriculum learning, basically adjusting parameters of the environment between episodes, particularly in situations where creating the environment again from scratch is not viable (e.g. it needs to launch some big executable like Unity envs)

As for the points you're bringing up:

>you don't need kwargs if you have access to env methods

You don't need reset if you have access to env methods, or anything else really - you can replace it with a custom env method. The point of gym is a shared interface

>using kwargs obfuscates the code (you can litteraly pass anything)
I don't see how it obfuscates the code since the alternative is either to not do it at all, or use a custom method (which can do literally anything as well).

>and prevents you from doing any type checking

It prevents type checking in the kwargs passed to reset because it's kept flexible on purpose, and can be handled appropriately in each environment. I don't think that's an issue, because in the default case, there is also no type checking, because there are no types to check. As far as I can tell, this does not prevent any type checking outside of this method's signature.

araffin:
> The main point is doing things like curriculum learning, basically adjusting parameters of the environment between episodes,

for curriculum learning, I assume you use that type of code:
```python
# collect data, train the agent for n steps or n epsiodes, ....
run_episode(agent, env)
# optional: periodically evaluate to choose the right level difficulty
agent_performance = evaluate(agent, eval_env)
# compute the level difficulty using different indicators (number of timesteps/episodes, agent performance, ...)
difficulty = curriculum_schedule(agent_performance, current_timesteps, ...)
# update level difficulty
env.set_difficulty(difficulty)
```
I don't really see the need of keyword arguments for the reset method here (and it also looks cleaner to me as the code that runs the episode and collect data is separated from the one choosing the level difficulty).

> The point of gym is a shared interface

`env.reset(**kwargs)` allows anything as input, I wouldn't consider that a shared interface...

> As far as I can tell, this does not prevent any type checking outside of this method's signature.

the thing is if you use wrappers, each wrapper will be passed `**kwargs` and could potentially consume one or more of them, the issue is if you want to know what each wrapper expect, you will have to dig into the logic of each `reset()` code, you cannot just look at the method signature.
This is what I meant by obfuscating, as you allow anything as input, you need to check each wrapper to know what is really happening (and it is error prone as a typo in the keyword argument won't trigger any error).

jkterry1:
One thing to note is that the intention is to have the **kwargs to reset be analogous to infos. If it makes more sense to have an info argument to reset, that's at least a possibility.

ivannz:
First things first, thank you, @RedTachyon, for the work you did on straightening the seed-reset interplay in the envs.

@araffin I would like to offer my perspective that inclines in favour of `**kwargs` in the `.reset` method in wrappers. Basically, `**kwargs` is more akin to obliviousness, rather than obfuscation.

As a running example consider the `seed` keyword arg introduced to `.reset` by PR #2422. At the moment the `.reset` code of many the wrappers in `gym.core` and `gym.wrappers` force-calls `self.env.reset(seed=seed, **kwargs)` of the wrapped env. Take [OrderEnforcing](https://github.com/openai/gym/blob/180d8ddd5cce0d23cdc4cada1faa52f3ae45313c/gym/wrappers/order_enforcing.py), for instance, which currently says
```python
class OrderEnforcing(gym.Wrapper):
    ...

    def reset(self, seed: Optional[int] = None, **kwargs):
        self._has_reset = True
        return self.env.reset(seed=seed, **kwargs)
```
and forces the optional `seed` arg onto the wrapped env. Although this call is consistent with the new `.reset`/`.seed` api in `gym>=0.21` and does not obfuscate, in my opinion, it being overly explicit and specific in its arguments makes it less friendly to many existing downstream third-party environments, including [nle](https://github.com/facebookresearch/nle/blob/0676f37c46e9ae78dca363525f78df300f881c59/nle/env/base.py#L447), [minigrid](https://github.com/maximecb/gym-minigrid/blob/03cf21f61bce58ab13a1de450f6269edd636183a/gym_minigrid/minigrid.py#L706), [obstacle-tower-env](https://github.com/Unity-Technologies/obstacle-tower-env/blob/d2450cd5fea92fff0a881e858a9a44f53a440d19/obstacle_tower_env.py#L182). Admittedly, these envs might not be entirely compatible with the newest `gym.Env.reset` api, and the `gym`, being the upstream library, is not responsible for potential misuse.

Many other wrappers, including [gym.Wrapper](https://github.com/openai/gym/blob/180d8ddd5cce0d23cdc4cada1faa52f3ae45313c/gym/core.py#L302), [gym.ObservationWrapper](https://github.com/openai/gym/blob/180d8ddd5cce0d23cdc4cada1faa52f3ae45313c/gym/core.py#L330), [gym.RewardWrapper](https://github.com/openai/gym/blob/180d8ddd5cce0d23cdc4cada1faa52f3ae45313c/gym/core.py#L344), [gym.ActionWrapper](https://github.com/openai/gym/blob/180d8ddd5cce0d23cdc4cada1faa52f3ae45313c/gym/core.py#L357), [NormalizeObservation](https://github.com/openai/gym/blob/180d8ddd5cce0d23cdc4cada1faa52f3ae45313c/gym/wrappers/normalize.py#L67), and [FrameStack](https://github.com/openai/gym/blob/180d8ddd5cce0d23cdc4cada1faa52f3ae45313c/gym/wrappers/frame_stack.py#L122), have nothing to do with the PRNG and thus *do not need* to know whether a `seed` has been provided, in contrast to what their post-PR `.reset` signatures suggests. Furthermore, many of them, essentially, pass the `.reset` args along anyway.

If we were to generalize from this example, a generic wrapper functionality cloud be more versatile, in my opinion, if it were more `forgiving` towards the arguments of the wrapped env's methods. For example, in wrappers, which do not care about the PRNG seeds, but still need to store some info upon `.reset` calls, it might be possible to write their custom logic something along the lines of:
```python
class OrderEnforcing(gym.Wrapper):
    ...

    def reset(self, **kwargs):
        self._has_reset = True

class FrameStack(ObservationWrapper):
    ...

    def reset(self, **kwargs):
        observation = self.env.reset(**kwargs)
        [self.frames.append(observation) for _ in range(self.num_stack)]
        return self.observation()
```
These `.reset` definitions *force* that their arguments be provided in keyword form, yet *suggest* that they are *oblivious* to the particular keywords provided. The inner logic passes the keywords along to the wrapped env, without needlessly enforcing any arguments, and thus may be considered more *friendly* towards the wrapped envs developed elsewhere.

Less generic wrappers are a tougher design issue. In the case when a wrapper has to rely of some specific particular, even optional, kwarg, it should *request* it explicitly in the `.reset` def. The problem then becomes whether that used kwarg is also expected by the wrapped env and what specifically that env's expected kwarg's type is.

For example, returnig to `seeds`, one might conceive of a wrapper that takes a in `seed` to init its own PRNG from it and fork a new seed via `SeedSequence.spawn` to pass along downstream. Such a wrapper would have to employ introspection capabilities of the `inspect` module to figure out if the wrapped env expects a `seed` kwarg, perhaps using
```python
import gym
import numpy as np

from typing import Optional
from inspect import signature
from sys import maxsize

def expected(fn):
    """Fetch the explicit;y decalred parameters in the signature."""
    return frozenset(
        p.name for p in signature(fn).parameters.values()
        if p.kind != p.VAR_KEYWORD
    )

class WrapperWithOwnPRNG(gym.Wrapper):
    prng = None

    @property
    def ready(self):
        return isinstance(self.prng, np.random.Generator)

    # explicitly declare that `.reset` can take in a seed and do something with
    #   it, even if it is optional.
    def reset(self, seed: Optional[int] = None, **kwargs):
        # get a seed sequence from the provided entropy
        ss = np.random.SeedSequence(seed)

        # the default numpy's rng, i.e. `PCG64`, is our source of randomness
        self.prng = np.random.default_rng(ss)

        # nothing to do if the wrapped env does not accept `seed`
        if 'seed' not in expected(self.env.reset):

        # fork the SeedSequence, produce a random int from it and pass along
        # XXX the `.entropy` of the spawned sequences is identical to
        #  the original SeedSequence, hence we create an aux PRNG to sample
        #  a single large integer
        aux = np.random.default_rng(*ss.spawn(1))  # XXX .spawn returns a list
        return self.env.reset(seed=aux.integers(maxsize), **kwargs)
```

To summarize, i think, intentional use of `**kwargs` in the signatures in the wrappers suggests that their functions act as an oblivious *passthrough* for the undeclared args.

XuehaiPan:
@ivannz

`**kwargs` in `Wrapper.reset()` is necessary for wrapper design, since the wrapper should pass used arguments to the inner side. But it may cause errors if either the developers or the users do not take the arguments to `reset` carefully. The users will need to look into the source code of wrappers and environments.

------

> https://github.com/openai/gym/pull/2515#issuecomment-990927936 `env.reset(**kwargs)` allows anything as input, I wouldn't consider that a shared interface...

`**kwargs` in `Wrapper.reset()` is necessary in wrapper design, but not in `Env.reset()`.

```python
class MyEnv(gym.Env):
    def reset(self, *, arg1, arg2, seed=None):  # no **kwargs in the end
        ...
```

For the unwrapped environment instance, it should not collect remaining `**kwargs` to the inside (inter unwrapped environment).

If the signature of `MyEnv.reset` has `**kwargs`, the user can write:

```python
env = MyEnv()
env = Wrapper1(env)
...
env = WrapperN(env)

env.reset(misspelled_name=xxx)
```

with no exception.

------

> https://github.com/openai/gym/pull/2515#issuecomment-990927936 the thing is if you use wrappers, each wrapper will be passed `**kwargs` and could potentially consume one or more of them, the issue is if you want to know what each wrapper expect, you will have to dig into the logic of each reset() code, you cannot just look at the method signature.
This is what I meant by obfuscating, as you allow anything as input, you need to check each wrapper to know what is really happening (and it is error prone as a typo in the keyword argument won't trigger any error).

Potential issues for wrappers consume arguments:

```python
class Wrapper1(gym.Wrapper):
    def reset(self, *, argX, seed=None, **kwargs):  # `argX` without default value
        # reset something with argX
        # there is no key 'argX' in kwargs
        # 'argX' in kwargs -> False
        return self.env.reset(seed=seed, **kwargs)

class Wrapper2(gym.Wrapper):
    def reset(self, *, argX, seed=None, **kwargs):  # same argument name `argX` without default value
        # reset something with argX
        # there is no key 'argX' in kwargs
        # 'argX' in kwargs -> False
        return self.env.reset(seed=seed, **kwargs)

env = Wrapper2(Wrapper1(MyEnv()))
env.reset(argX=xxx)  # TypeError: Wrapper1.reset() missing 1 required keyword-only argument: 'argX'
```

The outer wrapper (`Wrapper2`) catches the argument `argX` but does not pass it to the inside (in most use-cases). The inner wrapper (or inner environment instance) will not receive the expected argument in `**kwargs`.

For all wrappers and environment `reset` signatures, keyword argument names should be unique.

jkterry1:
I'd like to offer to a few notes:

1) At some point in the near future, Gym will have a comprehensive and completely redone set of wrappers. These will all handle reset correctly, and significantly remove the need for people to write their own wrappers.
2) Passing **kwargs should not be done in normal circumstances, including for any environments built into Gym. It just allows this option for edge cases. If we think it's necessary, I'd be comfortable raising some sort of warning saying that you have to know what you're doing and that it could cause issues when **kwargs are passed.

jkterry1:
After thinking about this more, I really like the idea of issuing a warning when passing **kwargs:

"You're passing a nonstandard argument to reset via **kwargs. This is only allowed for certain edge cases and should not be done under normal circumstances, and may result in unexpected behavior with wrappers, environments, and/or learning code. This should be used cautiously by experienced developers."

Is everyone happy with that warning and including this with the warning?

XuehaiPan:
@jkterry1

> 1. At some point in the near future, Gym will have a comprehensive and completely redone set of wrappers. These will all handle reset correctly, and significantly remove the need for people to write their own wrappers.

It's hard to handle `reset` with `**kwargs` in general wrappers. For example, in wrapper `RescaleAction`:

https://github.com/openai/gym/blob/180d8ddd5cce0d23cdc4cada1faa52f3ae45313c/gym/wrappers/rescale_action.py#L6-L34

it sets its own `action_space` computed from the `env`'s `action_space`. Once the `env.unwrapped.reset(**kwargs)` can change the environment's `action_space`. The general wrapper `RescaleAction` should always recompute the `action_space` to handle this.

> 2. Passing `**kwargs` should not be done in normal circumstances, including for any environments built into Gym. It just allows this option for edge cases. If we think it's necessary, I'd be comfortable raising some sort of warning saying that you have to know what you're doing and that it could cause issues when `**kwargs` are passed.

I can't see any easy approach to raise warnings in the subclasses. We usually use inheritance to create new environment classes and wrapper classes. The warning will not arise if the subclass doesn't call `super().method()`.

```python
class Env:
    def reset(self, *, seed=None, **kwargs):
        if kwargs:
            logger.warn('message')
        # set seed ....

class MyEnv(Env):
    def reset(self, *, seed=None, **kwargs):
        # do not call `super().reset(seed=seed, **kwargs)`
        # no warnings will arise
```

We will need [metaclasses](https://docs.python.org/3/reference/datamodel.html#metaclasses) to achieve this, but it's much more complicated.

XuehaiPan:
I would prefer to do not accept any custom arguments in `reset` at all. Only some sort of common arguments (e.g. `seed`) are supported.

To support some use cases https://github.com/openai/gym/pull/2515#issue-1076064412 and #2399, changing the environment settings (e.g. the number of agents), we can add new API like:

```python
class Env:
    def reset(self, *, seed=None):  # no **kwargs
        # set seed ....

    reset_episode = reset  # alias

    def reset_environment(self, **kwargs):
        pass  # do nothing as a prototype (override in subclasses)
```

Split the `reset` into two methods `reset_episode` and `reset_environment` (or other name `reset_settings`). However, it's hard to write general wrappers to support `reset_environment` too.

jkterry1:
I sat down and thought about this some more, and spoke to @RedTachyon offline as well. @XuehaiPan I take your point about the concerns regarding **kwargs with wrappers. I also do still think that modifying reset to allow for things like curriculum learning, instead of requiring custom methods, is the lesser of the evils.

What if we just made `info` an argument to reset? This would create a clear and simple contract for wrappers to pass things through to environments and preserve name errors in the environment if you do things wrong. This would also satisfy the functionality that we're seeking to add here, and arguably be more gym-like.

@XuehaiPan would this address your concerns?

XuehaiPan:
> What if we just made `info` an argument to reset?

> @XuehaiPan would this address your concerns?

Yes. We should only support a subset of arguments that satisfy the following constraints:

1. it's optional (have a default value): for backward compatibility
2. common for all environments: `info` √   `num_agents` ×
3. will not change the core attributes (e.g. `action_space`) of the environment: for writing general wrappers (`seed` √   `info` ×)

`seed` argument satisfies all constraints above, but I'm afraid `info` (a dict) does not. It looks like we use `info` (a dict) rather than `**kwargs` (`kwargs` is also a dict).

If all official wrappers do not use `info` and pass it into the inner unwrapped environment, I think it's fine. The third-party environments that use `info` have the duty to provide their own wrappers to handle `info` well. For example, `num_agent` in `info` will change `env.unwrapped.action_space`, then the developers should provide their own version of action wrappers (e.g. wrapper `RescaleAction`) to users.

RedTachyon:
So I'm starting to do the change where it won't be **kwargs and just a dictionary of whatever instead, and one question comes up that might simplify things a lot - is it fine if Wrappers maintain the **kwargs signature? That was the case before anything was done about this (e.g. already in 0.21), and it might allow me to remove the ugliest hacks, but maintain backwards-compatibility (though I'll need to make sure of that)

jkterry1:
Unless there's a reason for this I'm unaware of, I have no problem using **kwargs in wrappers

XuehaiPan (approved):
LGTM. We may need another PR to update the documentation.

In the documentation, we should note that:

1. **guideline for developers:** the new argument `options` should be passed into the unwrapped `env` instance, wrappers should not consume any values from `options`.

   ```python
   class MyWrapper(gym.Wrapper):
       def reset(self, **kwargs):
           results = self.env.reset(**kwargs)
           # Wrapper logics
           return results
   ```

2. **warning for developers and end-users:** the `options` argument in `reset` may change some attributes of the unwrapped `env` instance (e.g., `action_space`). These environments should be specially treated by the developers. The user should aware that it may produce unexpected results when they use these environments with wrappers (e.g., `RescaleAction`).

   ```python
   class MyEnv(gym.Env):
       def __init__(self, num_agents=4):
           self.num_agents = num_agents
           self.action_space = spaces.Box(low=-5.0, high=+5.0, shape=(self.num_agents,))

       def reset(self, *, seed=None, options=None):
           super().reset(seed=seed)
           options = options or {}
           num_agents = options.get('num_agents')
           if num_agents is not None:
               self.num_agents = num_agents
               self.action_space = spaces.Box(low=-5.0, high=+5.0, shape=(self.num_agents,))
   
   env = gym.wrappers.RecaleAction(MyEnv(num_agents=4), min_action=-1.0, max_action=+1.0)
   env.unwrapped.action_space  # -> Box([-5. -5. -5. -5.], [5. 5. 5. 5.], (4,), float32)
   env.action_space            # -> Box([-1. -1. -1. -1.], [1. 1. 1. 1.], (4,), float32)
   env.reset(options=dict(num_agents=6))
   env.unwrapped.action_space  # -> Box([-5. -5. -5. -5. -5. -5.], [5. 5. 5. 5. 5. 5.], (6,), float32)
   env.action_space            # -> Box([-1. -1. -1. -1.], [1. 1. 1. 1.], (4,), float32) (expected Box([-1. -1. -1. -1. -1. -1.], [1. 1. 1. 1. 1. 1.], (6,), float32))
   ```

gym/utils/env_checker.py
```diff
@@ -274,6 +274,35 @@ def _check_render(
         env.close()
 
 
+def _check_reset_seed(env: gym.Env, seed: Optional[int] = None) -> None:
+    """
+    Check that the environment can be reset with a random seed.
+    """
+    try:
+        env.reset(seed=seed)
```
JesseFarebro, line 282: These tests should probably be more robust, something like:

```py
import inspect

...

sig = inspect.signature(env.reset)
# Check signature has a seed parameter
assert 'seed' in sig.parameters

seed_param = sig.parameters.get('seed')
# Check the seed parameter is positional or keyword argument
assert seed_param.kind is inspect.Parameter.POSITIONAL_OR_KEYWORD
# Check the default value is None
assert seed_param.default is None
```

gym/utils/env_checker.py
```diff
@@ -274,6 +274,35 @@ def _check_render(
...
+            "This is likely due to not calling `super().reset(seed=seed)` in the `reset` method."
+        )
+
+
+def _check_reset_options(env: gym.Env) -> None:
+    """
+    Check that the environment can be reset with options.
+    """
+    try:
+        env.reset(options={})
```
JesseFarebro, line 299: Same as above, might want to use `inspect.signature` for better conformance.

RedTachyon:
So the more I think about it, the less I'm certain if `inspect.signature` is a good option here. The code you proposed won't quite work if we want wrapped envs to pass tests (which I imagine we do), because wrappers accept **kwargs into their reset signature, so they don't/may not have `seed` explicitly listed. We also don't necessarily require downstream envs to have None as the default argument.

I think what's important is the functionality, and that's what the current tests check directly

RedTachyon:
So in the end I went for both approaches basically, so that the env checker will tell you what's going on. Granted, if the `inspect` method passes and yet passing a seed causes a crash, that's pretty cursed and should never happen, so that's noted in the error messages. 

A few notes on the choices I made here:
- I skipped checking the kind of the seed/options arguments because complaining about writing `def reset(self, seed: ..., options: ...)` as opposed to `def reset(self, *, seed: ..., options: ...)` in a downstream env would be too pedantic and not helpful in my opinion
- I had to include a possibility for `**kwargs` in the signature to prevent wrappers from failing the checker
- If the reset doesn't accept a `seed`, that's an assertion error in the checker. If it accepts it but doesn't actually seed the RNG, that's only a warning, because not all environments need that. If the default value isn't None, that's also a warning, because that's kinda weird but eh, you do you.


The tests seem to be passing, but the fail is due to the thing on master. Which I still don't understand, but hopefully that'll be cleared up Soon ™

Issue #1929, opened 2020-05-23
Integer overflows in unbounded integer Box spaces


iamlucaswolf:
Hi,

I believe I noticed some peculiar bugs in `gym.spaces.Box` around the representation of integer box spaces with unbounded upper ends (e.g. [0, inf)).

First of all, instantiating a Box space with an upper bound of `np.inf` will immediately result in an integer overflow:

```python
>>> space = Box(low=0, high=np.inf, shape=(1,), dtype=np.int)
>>> space.high
array([-9223372036854775808])
```

This is due to line 45f in `gym/spaces/box.py`, which broadcasts the upper bound to all dimensions in the constructed space if the input argument `high` is scalar. The implementation uses `np.full`, which will cause the overflow. 

```python
>>> np.full((1,), np.inf, dtype=np.int)
array([-9223372036854775808])
```

While it's certainly debatable whether these are the intended semantics of float to int conversion (after all, `type(np.inf) == float`), there exists an easy workaround for this issue...

```python
>>> space = Box(low=0, high=np.iinfo(np.int).max, shape=(1,), dtype=np.int)
>>> space.high
array([9223372036854775807])
```

... which will however result in more problems down the line. If we now try to sample from this space, we get:

```python
>>> space = Box(low=0, high=np.iinfo(np.int).max, shape=(1,), dtype=np.int)
>>> space.sample()
array([-8746956029393751040])
```

In this case, the overflow happens in line 95 (in `Box.sample()`), which sets `high = self.high.astype('int64') + 1`; `high` is then used as an (exclusive) upper bound for uniform sampling in 117f.

**TL;DR**

Currently, it seems that integer Box spaces with unbounded upper endpoints can not be represented in `gym`. Trying to do so will result in integer overflows which happen silently along the way.

While I agree that this is a somewhat pathological case, I'd argue that it does have practical implications. I stumbled over it while implementing an `env` class for chess with the board encoding used by AlphaZero.

I'd be happy to provide a PR for this.

christopherhesse:
Thanks for the bug report!   I don't see the use case for unbounded upper endpoints for integers, which have bounded upper end points.  I agree that the high + 1 stuff is an issue, but I think it might work with a special case for int64/uint64, since all smaller integers would fit in those.

What sort of PR are you envisioning there?

iamlucaswolf:
Thanks for your reply!

[...]

PR #2517, opened 2021-12-10
Potential fix for integer overflow in box spaces

jjshoots:
This is a potential fix for [#1929](https://github.com/openai/gym/issues/1929). Essentially, it checks if a np.inf has be used as a bound, then basically does np.inf-2 for the datatype.


RedTachyon:
Can you add tests that show that the specific problem from that issue is solved?

jjshoots:
Sorry for being a total newb in this, how would I go about doing that? Would it mean putting a test function under `gym/tests/spaces/test_spaces.py`?

RedTachyon:
@jjshoots No worries, everyone has to start somewhere. But yes, you can put it in that file, another function that would create the space with the right properties (i.e. int dtype and infinite bounds); and then some assertions on the behavior of that space (e.g. that sampling from a space that should go from 0 to +infinity will be in fact positive - which doesn't seem to be the case with the version on master at the moment from what I gathered)

jjshoots:
@RedTachyon I think it's ready now, it passes all unit tests on my machine.

jkterry1:
@jjshoots its still failing tests

jjshoots:
@jkterry1 From what I can see, the tests that fail are ones related to black formatting and codespell (both of which I've just fixed). I've run all unit tests on my machine (Ubuntu 20.04, Python 3.8.10) and they all seem to pass. Are the tests performed in CI different from the ones when I use `pytest`, or is `pytest` not the way to go about running the tests?

jkterry1:
take a look at the github actions workflow to see how the tests that are run in CI are defined

gym/spaces/box.py
```diff
@@ -50,11 +50,30 @@ def __init__(self, low, high, shape=None, dtype=np.float32, seed=None):
                 "shape must be provided or inferred from the shapes of low or high"
             )
 
+        # return custom infinite if is integer infinite
+        def _get_inf(dtype):
+            if np.dtype(dtype).kind == "f":
+                return np.inf
+            elif np.dtype(dtype).kind == "i":
+                temp = np.ones((1,)).astype(dtype)
```
RedTachyon, line 58: Rename `temp` to something more descriptive please

gym/spaces/box.py
```diff
@@ -50,11 +50,30 @@ def __init__(self, low, high, shape=None, dtype=np.float32, seed=None):
...
+                return 2 ** (temp[0].nbytes * 8 - 1) - 2
```
RedTachyon, line 59: This looks a bit iffy. The idea is to get the maximum/minimum integer, yes? Afaik the "right" way to do it is with `np.iinfo`

gym/spaces/box.py
```diff
@@ -50,11 +50,30 @@ def __init__(self, low, high, shape=None, dtype=np.float32, seed=None):
...
+            else:
+                raise ValueError("unknown dtype for infinite bounds")
```
RedTachyon, line 61: Capitalize the first letter, and maybe give some more information, e.g. 
`f"Unknown dtype {dtype} for infinite bounds"`

(but please check if this actually prints something readable if you test it with e.g. `complex` as type)

gym/spaces/box.py
```diff
@@ -50,11 +50,30 @@ def __init__(self, low, high, shape=None, dtype=np.float32, seed=None):
                 "shape must be provided or inferred from the shapes of low or high"
             )
 
+        # return custom infinite if is integer infinite
+        def _get_inf(dtype):
```
RedTachyon, line 54: As far as I can tell, this function doesn't use any extra information that requires putting it here, right? I'd move it e.g. to the `.../spaces/utils.py` file, nested functions are typically not that great of a thing for readability

tests/spaces/test_spaces.py
```diff
@@ -448,3 +448,49 @@ def test_space_legacy_state_pickling():
...
+        Box(low=-np.inf, high=0, shape=(2,), dtype=np.float32),
+    ],
+)
+def test_infinite_space(space):
+    # for this test, make sure that spaces that are passed in have only 0 or infinite bounds
+    # because space.high and space.low are both modified within the init
+    # so the only way to test that infinite bounded space are truly infinite during test
+    # is to ensure that we only pass in either infinite or 0
+    assert np.all(space.high > space.low), "High bound not higher than low bound"
+    assert space.contains(space.sample())
```
RedTachyon, line 472: Add `space.seed(0)` before this, at the moment this is technically a stochastic test, so it sometimes might work and sometimes fail. Also I'd like to see some more assertions on what this is expected to be, e.g. that the dtype is right, and that the value is right (e.g. for the first two spaces tested, it should be positive - in principle `contains` should test it, but we're messing with the bounds, so a simple sanity check would be valuable.

It'd be good to also vary the shapes a bit, I'd like to see at least one test where it's a box larger than 1D (so e.g. `shape=(2,3)`)
From what I can see in the code, it's handled properly so it should still pass, but I can easily see a naive implementation of this that could pass your current tests, but fail on the example I gave, so it's good to cover our bases

RedTachyon (commented):
So I wrote a bunch of comments which might explain my thought process as I was reading this code, overall I think I'd suggest extracting `_get_inf` into a separate function, and rewriting it in terms of `np.iinfo`. You could also have it handle the sign of the infinity, since it's not necessarily the case that `min_int == -max_int`

And keep in mind that overall this does its job, and I'm being a bit nitpicky right now, so don't get discouraged! But it's better to make sure that everything is done well the first time, than make another PR again later.

gym/spaces/box.py
RedTachyon: (update: I think this will be irrelevant if function is rewritten as suggested)

jjshoots:
@RedTachyon Thanks for the feedback! I really appreciate the help. I've implemented most of the things, except for the one about moving `_get_inf` into `.../spaces/utils.py`. You're right that `_get_inf` does not depend on anything else, I tried putting it into `.../space/utils.py`, but `.../space/utils.py` imports `.../spaces/box.py`, so if I put it in there, we end up with a circular import. At this moment I'm not sure if I should try to resolve the circular import as I don't think that is good design (but do correct me if I'm wrong)? So for now I've left it in `.../spaces/box.py` as a separate static function instead.

RedTachyon:
I'm still not entirely sold on the structure/style. I'd put both get_inf and _get_precision into separate functions, not within the class so that they don't pollute the namespace (i.e. even as static methods, these aren't things that end users are expected to ever care about)

I see the problem you mentioned with utils, and it actually seems a bit odd the way it's structured at the moment, but that's a whole another issue.

jjshoots:
@RedTachyon Is there anything that I am able to do to help resolve this for now? Or are there some refactoring plans in the works and that I should just leave this in place for now.

RedTachyon:
@jjshoots Yes, sorry if that wasn't clear. For now please put the two functions I mentioned (`get_inf` and `_get_precision`) into stand-alone functions, which can even be at the end of that same file. So not static methods, and not functions defined within functions. And then we'll see about some refactoring later, but it would just be moving around functions if anything

RedTachyon:
@jjshoots I meant the functions should be in the outermost scope of that file, right now they're methods, and I don't think that's necessary.  Basically the file should look roughly like this:
```python
import ...

class Box(space):
    ...

def get_inf(...):
    ...

def get_precision(...):
    ...

jjshoots:
@RedTachyon gotcha, made the relevant changes. :)

RedTachyon:
Great, thanks!

@jkterry1 Can be merged

PR #2519, opened 2021-12-10
Fix a typo in requirements

### LINKED ISSUE #2518: [Bug Report] InvalidVersionSpec: Invalid version '3.': empty version component
If you are submitting a bug report, please fill in the following details and use the tag [bug].

**Describe the bug**
The requirements.txt contains the line

```
opencv-python>=3.
```

Which causes the following bug, when gym is pip installed in a `conda` environment with `pip_interop_enabled`.

```
Collecting package metadata (current_repodata.json): \ WARNING conda.core.index:push_record(267): Skipping pypi/pypi::gym-0.21.0-pypi_0 due to InvalidSpec: 3.
done
Solving environment: failed with repodata from current_repodata.json, will retry with next repodata source.

InvalidVersionSpec: Invalid version '3.': empty version component
```

**Code example**
Please try to provide a minimal example to reproduce the bug. Error messages and stack traces are also helpful.

```
conda create --name demo python
conda config --env --set pip_interop_enabled True
pip install gym
conda update --all
```

This can be resolved by changing the version string to 

```
opencv-python>=3.0*
```

RedTachyon:
Closes #2518

I am impressed how nobody noticed this for several months (including myself, because I recall looking at the requirements), but it is what it is.

There's a larger discussion to be had as to what should actually be put in these requirements, since a common "default" workflow is doing

```bash
pip install -r requirements.txt
pip install -e .
```

And this will install a whole bunch of stuff that's very optional to many users. That being said, an even more common workflow is `pip install gym` which seems to avoid this issue. In any case, the typo fix should be an obvious thing in the meantime.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 ale-py~=0.7
-opencv-python>=3.
+opencv-python>=3.0
 box2d-py==2.3.5
 mujoco_py>=1.50, <2.0
 scipy>=1.4.1


seungjaeryanlee:
Considering that it was like that for more than a year and we noticed now, I agree about having a discussion about it.

Also, `setup.py` seems to have the same typo, which we probably want to fix as well.

https://github.com/openai/gym/blob/e5d1ba9a4c532603a4b08ba6dae624188287e69b/setup.py#L20

jkterry1:
The long term fix to this is we're still planning on using lycon2 as a replacement, people are currently working on this as we speak. I'll fix the setup.py entry on master.

PR #2520, opened 2021-12-10
Removed mypy from CI tests

Ryan-Amaral:
Removed the installation and running of mypy from `.github/workflows/lint_python.yml`. 

This was done as `pyright` is now being used for type checking (as per `CONTRIBUTING.md`) and already has a functional test which is limited to specified files. The mypy test is then redundant, and incredibly inefficient due to checking EVERYTHING. Removing mypy from CI testing shaves ~11 minutes off.

Unless I'm mistaken and mypy is used in testing for some other reason that I'm not seeing. At the very least, if mypy must be used in test, I think it should use a more proper configuration to limit what it searches to a reasonable time (just as the pyright test does).

### DIFF
--- a/.github/workflows/lint_python.yml
+++ b/.github/workflows/lint_python.yml
@@ -13,21 +13,21 @@ jobs:
       - run: pytest --doctest-modules . || true
       - run: shopt -s globstar && pyupgrade --py36-plus **/*.py || true
 
-pyright:
-  name: Check types with pyright
-  runs-on: ubuntu-latest
-  strategy:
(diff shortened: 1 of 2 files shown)


jkterry1:
@ikamensh is this good to merge?

ikamensh:
Yes, IMO this is good to merge. According to https://github.com/microsoft/pyright/issues/679 Pyright already has type stubs bundled, so there is no need to do that kind of installing stubs like mypy did in the new setup.

Issue #2099, opened 2020-11-18
CarRacing-v0 - Weird Camera Jerk Issue


praveenVnktsh:
I noticed something weird when training my model on carracing-v0. Whenever a jerky braking situation is given, the camera flips vertically occasionally. Here is a picture of the same:

![flipped car](https://user-images.githubusercontent.com/30774147/99491692-340a8200-2992-11eb-997a-05a3f15a487a.png)

The left window shows the flipped image, and the one on the right is the frame previous to the current frame. The frame after the flipped frame is back to normal.

I noticed that there are a few odd lines in the environment's source file:

https://github.com/openai/gym/blob/c4d0af393ef9fba641bd3ebbbf1f60d291c8475d/gym/envs/box2d/car_racing.py#L421-L423

After some prodding, I noticed that occasionally the velocity of the car becomes negative as a result of which the atan function returns a negative number. The issue seems to be solved by commenting out those lines.

I do not understand the advantage of making the camera move with higher velocities since this can sometimes allow an agent to learn a lot more long range information by just having rapid movements. I observed this in some of my runs!

@olegklimov - Would be great if you could comment on this.

carlosluis:
I conducted a little experiment that confirms the sign switching problem reported by @praveenVnktsh 

Using this code 
```python
        if np.linalg.norm(vel) > 0.5:
            old_angle = ((angle + np.pi) % (2 * np.pi)) - np.pi  # normalize angle
            angle = math.atan2(vel[0], vel[1])
            if np.sign(old_angle) != np.sign(angle):
                print(f"old_angle = {old_angle}, new_angle = {angle}")
```
I then proceeded to steer the car recklessly until I see some angle flipping. Of course, near 0, pi and -pi, I see some angle flipping inherent from numerical approximations, but there's other instances where there's clearly a "wrong" angle flip

```
old_angle = 1.0416961908340454, new_angle = -1.2111190759830692
old_angle = 1.0416961908340454, new_angle = -1.2111190759830692
old_angle = 1.2610578536987305, new_angle = -1.1509918860429935
old_angle = 1.2610578536987305, new_angle = -1.1509918860429935
old_angle = 1.4788283109664917, new_angle = -1.1005596690570452
old_angle = 1.4788283109664917, new_angle = -1.1005596690570452
old_angle = 1.694751262664795, new_angle = -1.0476754974926226
old_angle = 1.694751262664795, new_angle = -1.0476754974926226
[...]

PR #2521, opened 2021-12-10
Remove CarRacing environment from Box 2D

carlosluis:
### Description

- Removes `CarRacing` and `Car` from Box 2D
- If someone tries to import `CarRacing` via `from gym.envs.box2d.car_racing import CarRacing`, they will see a warning.
- If they try to import implicitly from the box2d package, they will see an `ImportError` due to [this missing Box2D class](https://github.com/openai/gym/blob/e5d1ba9a4c532603a4b08ba6dae624188287e69b/gym/envs/box2d/__init__.py#L2)

### Testing
- Verified behaviour when trying to import `CarRacing`.

### Question
1.  I'm probably missing some context on why the other environments in box2d still exist when there's no `Box2D` class anymore. Is the plan to remove them all eventually?
2. Since I couldn't find examples of other removed envs, not sure if there's a minimum of functionality that should be left in the class. Currently it only raises a warning as soon as it's imported, but any call to the env will fail of course


RedTachyon:
Is this related to some issue/discussion that I'm not seeing? Why is this environment being removed?

As far as I remember, the plan is to replace Box2D with something more sane eventually, but I don't see why one of the environments should be removed ahead of time. And simply removing Box2D without a replacement would be a bad move since LunarLander and BipedalWalker both are pretty useful to have.

jkterry1:
Hey, apparently something got lost in translation somewhere. Basically the plan is to remove CarRacing (you can check stats for imports on github, its virtually used), conduct a bunch of maintenance/refactoring to lunar lander and bipedal walker and then use them as the basis of new phys2d environments. Essentially the motivation for removal is that I'd rather remove it than expend the time/energy on rewriting it in brax and the large list of maintenance needed before that could be done.

jkterry1:
@carlosluis can you please fix CI?

jkterry1:
One important piece of historical context behind these environments for those who read this is that, according to OpenAI, the box2d environments came from "some random dude who we'd never met before who created a PR and we thought they were kinda of cool so we merged them." There is no real compelling ideological reason behind including these like some other environments we have in Gym.

araffin:
> Hey, apparently something got lost in translation somewhere. Basically the plan is to remove CarRacing (you can check stats for imports on github, its virtually used), 

virtually used?
https://github.com/search?q=carracing-v0&type=repositories
58 repos and 3k files that refer to "Carracing-v0"...

jkterry1:
Antonin pointed out this work in discussion over discord:

https://arxiv.org/pdf/1809.01999.pdf and the importance of this for LSTMs work on RL, both of which I was unaware of. We'll keep this environment in and conduct the same maintenance on it as the other box2d environments.

@carlosluis Would you mind taking a look at this issue? https://github.com/openai/gym/issues/2099

PR #2522, opened 2021-12-11
Consolidate LunarLander and BipedalWalker envs

andrewtanJS:
This merges the different environment types for LunarLander (continuous) and BipedalWalker (hardcore), and adds kwargs to specify the specific environment.

This maintains backwards compatibility as the id for these environments remain the same.

Specifically, the `LunarLanderContinuous` and `BipedalWalkerHardcore` classes were removed, and their functionality was merged into their parent class. They can be instantiated by specifying the relevant keywords in `env.make` (using `"continuous"=True` and `"hardcore"=True`, respectively).

### DIFF
--- a/gym/envs/box2d/bipedal_walker.py
+++ b/gym/envs/box2d/bipedal_walker.py
@@ -119,9 +119,7 @@ class ContactDetector(contactListener):
 class BipedalWalker(gym.Env, EzPickle):
     metadata = {"render.modes": ["human", "rgb_array"], "video.frames_per_second": FPS}
 
-    hardcore = False
-
-    def __init__(self):
+    def __init__(self, hardcore=False):
         EzPickle.__init__(self)
         self.viewer = None
 
@@ -130,6 +128,8 @@ class BipedalWalker(gym.Env, EzPickle):
         self.hull = None
 
         self.prev_shaping = None
+
+        self.hardcore = hardcore
 
         self.fd_polygon = fixtureDef(
             shape=polygonShape(vertices=[(0, 0), (1, 0), (1, -1), (0, -1)]),
@@ -564,10 +564,6 @@ class BipedalWalker(gym.Env, EzPickle):
         if self.viewer is not None:
             self.viewer.close()
             self.viewer = None
-
-
-class BipedalWalkerHardcore(BipedalWalker):
-    hardcore = True
 
 
 if __name__ == "__main__":
--- a/gym/envs/box2d/bipedal_walker.py
+++ b/gym/envs/box2d/bipedal_walker.py
@@ -14,7 +14,7 @@ from Box2D.b2 import (
 )
 
 import gym
-from gym import spaces
+from gym import error, spaces
 from gym.utils import colorize, seeding, EzPickle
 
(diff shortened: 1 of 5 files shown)


gym/envs/box2d/bipedal_walker.py
```diff
@@ -119,9 +119,7 @@ def EndContact(self, contact):
 class BipedalWalker(gym.Env, EzPickle):
     metadata = {"render.modes": ["human", "rgb_array"], "video.frames_per_second": FPS}
 
-    hardcore = False
-
-    def __init__(self):
+    def __init__(self, hardcore=False):
```
RedTachyon, line 122: Minor nitpick - could you add a type hint? 
`def __init__(self, hardcore: bool = False):`

Analogously for LunarLander. Gym doesn't have full type checks yet, but we're moving in that direction, and it's not much effort to add them here.

RedTachyon (commented):
Looks good overall, I'd be happier if the constructors had type hints for this extra bit of clarity (see comment in code). But in general a really nice change, I remember being surprised at how these envs are handled.

gym/envs/box2d/bipedal_walker.py
andrewtanJS: Sure thing, I've added them to the corresponding environments

RedTachyon:
Actually after some further thought, I think it might be worth keeping the previous classes for the sake of compatibility. While the main approach is `env = gym.make(...)`, it isn't necessarily that uncommon to do `env = EnvClass(...)`.

I'm thinking maybe about adding back that class with a deprecation warning, and a note on how to update code that directly imports e.g. `BipedalWalkerHardcore`. @jkterry1 what do you think?

While I agree that the way this class is implemented at the moment is annoying, it is what it is. Adding an explicit argument to the initializer, and deprecating the redundant classes might be the way to go.

jkterry1:
We're realistically going to have to make this breaking change in the near future anyways for the transition to Brax. Maybe it's the sleep deprivation fairies talking to me here, but I'm unclear on why what your proposing is a better option that what the PR does now given this fact.

jkterry1:
@andrewtanJS could you please add informative errors if people try to create the environments the old way before we merge?

andrewtanJS:
Will do

andrewtanJS:
I've added the old environment classes back, but this time upon calling the classes directly, it'll raise an error alongside some refactoring information.

To make sure that importing is backwards compatible, I've added these classes back to the module `__init__.py`. This should allow for existing code to successfully import these classes, but to then encounter the informative errors if they attempt to initialize it by calling the class directly (i.e. `LunarLanderContinuous()`) instead of calling `gym.make()`.

As per specification, calling `gym.make(LunarLanderContinuous-v2)` still works, and it simply initializes `LunarLander` with the corresponding `continuous` argument.

jkterry1:
@andrewtanJS can you please fix tests?

andrewtanJS:
@jkterry1 gotcha, I've just fixed the pre-commit tests

RedTachyon (approved):
Looks good to merge now

Issue #2510, opened 2021-12-09
[Proposal] Formal API handling of truncation vs termination


jkterry1:
Right now, one of the biggest weaknesses of the Gym API is that Done is used for both truncation and termination. The problem is that algorithms in Q learning family (and I assume others), depend on the differentiation between a terminal state and a truncated state to function as intended. Right now, Gym solves this for some (but not all) environments with infos for built in environments via the time limiting wrapper inserting this information into infos. No third party environments I'm aware of use this, and most learning code uses this. This has resulted in a few problems:

1. The DQN people are deeply unamused because proper reimplementation of their algorithms are very frequently done wrong
2. Many people who are very qualified in RL now have complained about being wildly confused by this while first learning RL
3. Many large institutions (e.g. some large groups at Google brain) refuse to use Gym almost entirely over this design issue, which is bad

This sort of thing in the opinion of myself and those I've spoken to at OpenAI warrants a breaking change in the pursuit of a 1.0 release. There are three options for making the breaking change:

1. Add have step return an extra boolean value in addition to done, e.g. something like `observation, reward, done/terminated, truncated, info = env.step(action)`
2. That option, but return the discount value like dm_env does instead of a boolean
3. Turn `done` into a custom three state variable, or give it a method like `done.is_truncated`

My preference is option 1. I think that option 3 is too confusing and wouldn't generally be used because it's not explicit. I also think that option 2 adds additional confusion and complexity over option 1 without really adding any additional utility. I'm creating this issue so that people can discuss it before a PR is made. This feature is intended to be released in the release of gym after the current.

Regarding breaking changes and backwards compatibility:

- Old style environments can just be wrapped to add this, and we'll include a function in gym.wrappers to do this
- Learning code can be compliant with the new API simply via an extra `_,`

In other words, it's my view that the level of breaking this update will be is a minor annoyance, and I think that this is a very important feature. This change will be the second of three planned small changes to the base API before the 1.0 release (the first was to the seed method, the last will be to the render method). I am obviously aware of the profound role Gym plays in the community, however Gym will almost certainly be used a lot more over the next 5 years than the past 5, and I think that a series of minor breaking changes to clean things up now in the pursuit of a proper 1.0 release will be very worth it.

Please offer your thoughts on which of the three options here is best in the comments if you would like (or new options if I'm missing something) and remain civil and inviting to others.

Rohan138:
[...]

PR #2525, opened 2021-12-12
Fix jerky camera angle in car racing env

### LINKED ISSUE #2099: CarRacing-v0 - Weird Camera Jerk Issue

carlosluis:
Fixes #2099 

You can find context and investigations in the issue.


araffin:
Please wait a little before merging, there might be two additional fixes to include. I will try to detail them tomorrow, but in short:
- Markov assumption is broken when reaching the end of the track (related to #2510 )
- reward on the last tile is weird: only negative (breaks Markov assumption again as no info about the last tile is given)

araffin:
> reward on the last tile is weird: only negative (breaks Markov assumption again as no info about the last tile is given)

In fact, the bug seems to come from the completion of a lap wrongly detected. See below:

https://user-images.githubusercontent.com/1973948/146530604-c0d83979-8f24-45d3-b710-26512b5d3202.mp4

carlosluis:
No worries, we will wait a bit before merging because we want to migrate to pygame from pyglet as well alongside the version bump (context in https://github.com/openai/gym/issues/2099). I won't have time to get around to do that before I go on vacation for the rest of the year, so there's no rush on my side.

araffin:
I found the issue, unfortunately, it is not clear what to do...
My agent was actually cutting corner... so it managed to complete the track without visiting every tile, as a result, the episode is not terminated when the car finishes one full loop.
There are different options:
1. do not change anything, that's the agent fault
2. consider that the track is completed if we pass through the first tile again and that 95% of all tiles are visited. (detecting a full lap is not trivial as going in the wrong direction is allowed)
3. change the way visit is computed but this may introduce other reward hack/bugs

cutting_corner

jkterry1:
It seems to me like #2 is the most natural. Is there any big issue with this approach that I'm not seeing? We'd obviously need to try learning with the changes before merging to ensure no unexpected behaviors occur given though.

carlosluis:
@araffin I implemented the workaround #2 you suggested. I tested manually by controlling myself the car but since you seem to have a good learning setup for this env maybe it's worth it for you to train again with this changes to verify there are no new exploits

RedTachyon:
I don't have any strong opinions on the functionality, but I'd like to see some comments in the code explaining what's going on in lines 391-394 since it's a pretty unusual pattern. Can this be simplified/made more pythonic? Also, couldn't this operation be performed once per initialization or once per reset, and the coordinates of the tile can then be cached?

araffin:
> @araffin I implemented the workaround #2 you suggested. I tested manually by controlling myself the car but since you seem to have a good learning setup for this env maybe it's worth it for you to train again with this changes to verify there are no new exploits

sorry, I've got notifications off on gym, I implemented the fixes in my fork:
https://github.com/araffin/gym/tree/fix/carracing

For detecting the first tile, you should probably use the idx of the tile instead (see my fork where I set an id for each tile).

I think we should provide an argument to choose how much should be visited to consider the lap over.

jkterry1:
@araffin are you happy with carlos' changes?

gym/envs/box2d/car_racing.py
```diff
@@ -81,6 +82,14 @@ def _contact(self, contact, begin):
...
                 self.env.reward += 1000.0 / len(self.env.track)
                 self.env.tile_visited_count += 1
+
+                # Lap is considered completed if enough % of the track was covered
+                if (
+                    tile.idx == 0
+                    and self.env.tile_visited_count / len(self.env.track)
+                    > self.lap_complete_percent
+                ):
+                    self.env_new_lap = True
```
araffin, line 92: I just realized this should be `self.env.new_lap` no?
and in that case, termination is due to reaching lap end (should not be treated as normal termination)

pseudo-rnd-thoughts:
Would you want to make a PR for this?

araffin:
> Would you want to make a PR for this?

well, current implementation is buggy. We also need to fix `TimeLimit` wrapper (it should check if the `TimeLimit.truncated` key already exist or not)

Issue #2526, opened 2021-12-13
[Question] Does gym provide type stubs?


syheliel:
### Question
It seems that `Pylance` can't find the type stubs and there are no package like `gym-types` to fix it.
![image](https://user-images.githubusercontent.com/45957390/145765731-2938b8c1-e53d-4696-bf7b-6354f08a24c6.png)

ilyabosch:
This might be because the library doesn't have `py.typed` file. According to https://www.python.org/dev/peps/pep-0561/ a library should have this in it's file tree to signal that sources should be used for type checking.

jkterry1:
So, Gym does not provide type stubs. However, we're working on adding type hints, which there are issues tracking. Is there any reason to deal with type hints as well given this?

PR #2529, opened 2021-12-14
Typing/basics

ikamensh:
As requested, this PR merges the changes of #2486, #2487, #2488, #2489 and rebases on top of master


ikamensh:
I'm somewhat puzzled by pyright not finding numpy stubs in the venv, any hints are welcome. Also asked the author of GH action: https://github.com/jakebailey/pyright-action/issues/3 . Will follow-up

gym/spaces/space.py
```diff
@@ -1,7 +1,23 @@
...
+import numpy as np
+
 from gym.utils import seeding
 
 
-class Space:
+T_cov = TypeVar("T_cov", covariant=True)
+
+
+class Space(Generic[T_cov]):
```
RedTachyon, line 20: From what I see, the only thing that the generic does is inform the signature of `sample`, correct? 

In general it'd be tempting to connect this generic with Box, Discrete and other specific spaces, which is where this type annotation would actually be useful. But I'm not sure to what extent that works in python. Maybe just defining eg. `class Box(Space[np.ndarray])`?

gym/spaces/space.py
```diff
@@ -1,7 +1,23 @@
...
+    Tuple,
+)
+
+import numpy as np
+
 from gym.utils import seeding
 
 
-class Space:
+T_cov = TypeVar("T_cov", covariant=True)
```
RedTachyon, line 17: Does it have to be covariant? (I'm not sure either way)

gym/spaces/space.py
```diff
@@ -26,41 +42,41 @@ def __init__(self, shape=None, dtype=None, seed=None):
...
-    def np_random(self):
+    def np_random(self) -> np.random.RandomState:
         """Lazily seed the rng since this is expensive and only needed if
         sampling from this space.
         """
         if self._np_random is None:
             self.seed()
 
-        return self._np_random
+        return self._np_random  # type: ignore  ## self.seed() call guarantees right type.
```
RedTachyon, line 52: Wouldn't it pass type checks if `_np_random` gets added as a field to the class? This feels like something that shouldn't require disabling the checker (although it is a little bit unusual)

gym/spaces/space.py
```diff
@@ -26,41 +42,41 @@ def __init__(self, shape=None, dtype=None, seed=None):
...
         return self._shape
 
-    def sample(self):
+    def sample(self) -> T_cov:
         """Randomly sample an element of this space. Can be
         uniform or non-uniform sampling based on boundedness of space."""
         raise NotImplementedError
 
-    def seed(self, seed=None):
+    def seed(self, seed: Optional[int] = None):
```
RedTachyon, line 64: -> list[int]

RedTachyon (commented):
Overall pretty good step, the only thing that gets me a bit doubtful is the generics for spaces since right now they don't really do anything. Would it be enough to just add the proper types into the respective concrete spaces? Or should we add the types to each environment, and encourage users to define them appropriately when they do `class SomeEnvironment(gym.Env[np.ndarray, int]):`

Imo this should be resolved before merging, since at this moment, the type hints for spaces don't really do anything, and I think it's better to avoid changing it back if we find a different, better way. 

It's also worth keeping in mind that the pattern of `gym.make` will just about invalidate all of this, but the object oriented API is still available, and that should work.

gym/spaces/space.py
ikamensh: Yes, `class Box(Space[np.ndarray])` is precisely how I imagined it to work. That's valid python. Making subclasses of the Space more specific, and coherent with the Env object, is the intention behind generic Env and Space:
```python
    action_space: spaces.Space[ActType]
    observation_space: spaces.Space[ObsType]
```

`gym.make` can be helped by a typing stub later on, by having overload on many literal strings we are passing. At least that can help with the most used environments. (roughly how __pow__ is handled in typeshed https://github.com/python/typeshed/pull/6287/files )

gym/spaces/space.py
ikamensh: The problem is to narrow down the type after self.seed() is called - as it is self._np_random is Optional[np.random.RandomState], and pyright doesn't understand that if it's None, than it's made not-None by self.seed() call.

ikamensh:
Thanks for the review.

> only thing that gets me a bit doubtful is the generics for spaces since right now they don't really do anything.

As mentioned in review comments, it's intended for 1) checking compatibility between Env and it's Spaces; 2) usage in subclassed Spaces (currently only sample, I wonder if it can be possible to specify that `contains` output is related to types passed).

ikamensh:
Note: I've also added py.typed placeholder file and verified this should remove warning about stubs not installed ( #2526 )

gym/spaces/space.py
ikamensh: A possible solution to this would be to assert it's an `int` out of all options. But I don't know enough to tell if this will bite us back.

RedTachyon:
I think it's good to merge now. It's not necessarily "complete", but it can already do something, and with the signature of seed, it's probably better to be lenient for now until the type system is complete

Issue #353, opened 2016-09-20
Acrobot-v0 shouldn't be removed


rafaelcp:
A lot of experiments were run on the Acrobot-v0 env and, in my case, I even have it in my thesis proposal. Now I can't compare my new methods because this env doesn't exist anymore. I think it entirely beats the purpose of environment versioning if old versions get removed.

PR #2533, opened 2021-12-20
Adding documentation for Acrobot-v1 in docs/classic_control

kingb12:
Added some docs! 

I noticed `Acrobot-v0` was removed rather than deprecated, I believe in [this commit](https://github.com/openai/gym/commit/5ba5cb7e2dd7fb06412b0ba26fdba14127e87bbf). In the version history I mentioned the functional changes but did not mention the change from direct angle readings to sin/cos pairings provided by the v1 observation space. Let me know if you think there's a better way to handle (there's also a stale issue on [bringing `Acrobot-v0` back](https://github.com/openai/gym/issues/353))


jkterry1:
Hey, this looks good, but could you please note the sin/cos thing in the version history as well?

Issue #2459, opened 2021-10-23
The regex for environment names should not require a version in the name


jkterry1:
In many cases outside of Gym proper, environment versioning doesn't make a lot of sense, so people just peg it at v0 indefinitely. This is mildly annoying and should not be a general requirement.

PR #2535, opened 2021-12-20
Remove mandatory version in environment name

kir0ul:
This PR is an attempt to fix https://github.com/openai/gym/issues/2459.


jkterry1:
@kir0ul tests are failing

kir0ul:
- [ ] Also needs to fix types in `registration.py`

kir0ul:
> @kir0ul tests are failing

It seems there's a problem with the Docker registry:
```
Head "https://registry-1.docker.io/v2/library/python/manifests/3.7": received unexpected HTTP status: 502 Bad Gateway
```

jkterry1:
I forced tests to rerun, it looks fine. 

Regarding "Also needs to fix types in registration.py", what do you mean?

gym/envs/registration.py
```diff
@@ -17,22 +17,49 @@
...
 
-def env_id_from_parts(namespace: str, name: str, version: Union[int, str]) -> str:
+    This format is true today, but it's *not* an official spec.
+    [username/](env-name)-v(version)    env-name is group 1, version is group 2
+
+    2016-10-31: We're experimentally expanding the environment ID format
+    to include an optional username.
+    """
+    # First we try the strict pattern with the version number
+    pattern = re.compile(
```
JesseFarebro, line 31: Could you use the following regular expression:
```
^(?:(?P<namespace>[\w:-]+)\/)?(?:(?P<name>[\w:.-]+?))(?:-v(?P<version>\d+))?$
```
so we can have this logic all in one expression. This way we can just check if the capture group `version` is None or not.

All I changed was to make the name match group non-greedy and make the version group optional (greedy).

gym/envs/registration.py
```diff
@@ -17,22 +17,49 @@
...
+        r"^(?:(?P<namespace>[\w:-]+)\/)?(?P<name>[\w:.-]+)-v(?P<version>\d+)$"
+    )
+
+    match = pattern.fullmatch(env_str)
+    if match:
+        namespace, name, version = match.group("namespace", "name", "version")
+    else:
```
JesseFarebro, line 38: This logic can be removed with the new regular expression but do keep the error checking.

gym/envs/registration.py
```diff
@@ -17,22 +17,49 @@
...
+        pattern = re.compile(r"^(?:(?P<namespace>[\w:-]+)\/)?(?P<name>[\w:.-]+)$")
+        match = pattern.fullmatch(env_str)
+        if not match:
+            raise error.Error(
+                f"Malformed environment ID: {env_str}."
+                f"(Currently all IDs must be of the form {pattern}.)"
+            )
+        namespace, name = match.group("namespace", "name")
+        version = None
+    res = {"matched": match, "pattern": pattern, "id_parts": (namespace, name, version)}
```
JesseFarebro, line 49: Is there a particular reason you were returning `pattern` (it doesn't seem to be used) and `id_parts`? It seems we could just return `match` and call `group` where needed. We're guaranteed to get a valid match out of this function so it's valid to just call `match.group('...')` for whatever we need.

gym/envs/registration.py
```diff
@@ -17,22 +17,49 @@
...
-# This format is true today, but it's *not* an official spec.
-# [username/](env-name)-v(version)    env-name is group 1, version is group 2
-#
-# 2016-10-31: We're experimentally expanding the environment ID format
-# to include an optional username.
-env_id_re: re.Pattern = re.compile(
-    r"^(?:(?P<namespace>[\w:-]+)\/)?(?P<name>[\w:.-]+)-v(?P<version>\d+)$"
-)
 
+def check_env_id(env_str: str) -> dict:
```
JesseFarebro, line 21: Could you rename this to `parse_env_id`?

gym/envs/registration.py
```diff
@@ -706,7 +718,9 @@ def _versions(self, namespace: str, name: str) -> Set[Tuple[int, str]]:
                 lambda version: (
                     int(version),
                     namespace,
-                ),
+                )
+                if isinstance(version, str)
+                else (0, namespace),
```
JesseFarebro, line 723: Should probably return `(None, namespace)` here when there isn't a version. Could you also change the return type to `Set[Tuple[Optional[int], str]]`.

gym/envs/registration.py
```diff
@@ -726,7 +740,9 @@ def _versions(self, namespace: str, name: str) -> Set[Tuple[int, str]]:
                         lambda version: (
                             int(version),
                             relocated_namespace,
-                        ),
+                        )
+                        if isinstance(version, str)
+                        else (0, namespace),
```
JesseFarebro, line 745: Change to `(None, namespace)`.

tests/envs/test_registration.py
```diff
@@ -24,11 +26,30 @@ def __init__(self, arg1, arg2, arg3):
...
 
 def test_make():
-    env = envs.make("CartPole-v0")
-    assert env.spec.id == "CartPole-v0"
+    env = envs.make("CartPole-v1")
+    assert env.spec.id == "CartPole-v1"
     assert isinstance(env.unwrapped, cartpole.CartPoleEnv)
 
 
+@pytest.mark.parametrize(
```
JesseFarebro, line 34: We should probably test some edge case environment IDs:
```
# Valid
MyAwesomeEnv-vfinal-v0
MyAwesomeEnv-vfinal
MyAwesomeEnv--
MyAwesomeEnv-v
```
These edge cases should all be valid.

gym/envs/registration.py
JesseFarebro: Could you also update the return type to be `re.Match`.

JesseFarebro (changes requested):

kir0ul:
> Regarding "Also needs to fix types in registration.py", what do you mean?

Not sure why I don't see them in CI but I get the following errors on my machine when I run Pyright on `registration.py`:
```
$ pyright gym/envs/registration.py
gym/gym/envs/registration.py:678:33 - error: Argument of type "Tuple[int, str]" cannot be assigned to parameter "__a" of type "Mapping[_K@getitem, _V@getitem]" in function "getitem"
    "Tuple[int, str]" is incompatible with "Mapping[_K@getitem, _V@getitem]" (reportGeneralTypeIssues)
gym/gym/envs/registration.py:688:16 - error: Operator "<" not supported for types "int" and "int | str"
    Operator "<" not supported for types "int" and "str" (reportGeneralTypeIssues)
gym/gym/envs/registration.py:787:35 - error: Cannot access member "module" for type "EntryPoint"
    Member "module" is unknown (reportGeneralTypeIssues)
gym/gym/envs/registration.py:787:50 - error: Cannot access member "attr" for type "EntryPoint"
    Member "attr" is unknown (reportGeneralTypeIssues)
gym/gym/envs/registration.py:794:16 - error: "attr" is possibly unbound (reportUnboundVariable)  gym/gym/envs/registration.py:796:48 - error: "module" is possibly unbound (reportUnboundVariable)
6 errors, 0 warnings, 0 infos 
Completed in 1.499sec
```
Also I get the same errors on `master` so I guess that's worth trying to fix?

jkterry1:
Yes, please fix that

gym/envs/registration.py
kir0ul: Right, way better this way, thanks for the hint! :pray:

gym/envs/registration.py
kir0ul: For the `pattern` there was a reason before as it was used in error messages elsewhere, but then I moved the error messages inside the function, so indeed there's no reason anymore, good catch! Agree with the rest.

tests/envs/test_registration.py
kir0ul: Speaking of edge cases, would you consider adding a dev dependency? [Hypothesis](https://pypi.org/project/hypothesis/) is very nice to generate all the edge cases you didn't think of and to catch hidden bugs. For example we could use the [hypothesis.strategies.from_regex()](https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.from_regex) strategy and use the same regex as in the code to automatically generate tests.

JesseFarebro: CC @jkterry1

jkterry1: ya thats fine for tests for me

jkterry1:
@kir0ul can you please fix the merge conflicts

JesseFarebro:
@kir0ul let me know when this is ready to review again. Thanks for your work on this 👍 .

jkterry1:
@kir0ul just checkin in on this

jkterry1:
(you'll need to fix the merge conflict too)

kir0ul:
I kind of lost myself in the typing rabbit hole... but I was able to remove most of the `# type: ignore` statements.
Tests are failing on my fork so this still isn't ready yet.

kir0ul:
Tests are green again! :heavy_check_mark:
I guess this is ready for review @JesseFarebro :slightly_smiling_face:

gym/envs/registration.py
```diff
@@ -441,14 +447,16 @@ def register(self, id: str, **kwargs) -> None:
...
-    def _versions(self, namespace: str, name: str) -> Set[Tuple[int, str]]:
+    def _versions(self, namespace: str, name: str) -> Set[Tuple[Optional[int], str]]:
         # Get the set of versions under the requested namespace
         versions = set(
             map(
                 lambda version: (
                     int(version),
                     namespace,
-                ),
+                )
```
JesseFarebro, line 457: Here we can do,
```py
lambda version: (
    version and int(version),
    namespace
)
```
and get rid of the conditional in the map. Note: `x and f(x)` will evaluate to None if `x is None`, otherwise it will evaluate `f(x)`.

gym/envs/registration.py
```diff
@@ -316,8 +329,10 @@ def make(self, path: str, **kwargs) -> Env:
 
         # We check what the latest version of the environment is and display a warning
         # if the user is attempting to initialize an older version.
-        latest_version, latest_ns = max(versions)
-        if int(version) < latest_version:
+        is_lesser_version, latest_version, latest_ns = self._is_lesser_version(
```
JesseFarebro, line 332: Instead of `_is_lesser_version` can we do something like,

```py
from operator import itemgetter

...

latest_version, latest_ns = max(filter(itemgetter(0), versions), default=(None, None,))
if version and latest_version and int(version) < latest_version:
    ...
```

I think it's clear what's going on with the one-liner solution, we filter by the first element, if it's None it won't be included. If we perform `max([])` the default kicks in and will return `(None, None)`.

gym/envs/registration.py
```diff
@@ -400,9 +409,7 @@ def spec(self, path: str) -> EnvSpec:
             versions = self._versions(namespace, name)
             version_not_found_error_msg += ", ".join(
                 map(
-                    lambda version: env_id_from_parts(
-                        getitem(version, 1), name, getitem(version, 0)  # type: ignore
-                    ),
+                    lambda version: env_id_from_parts(version[1], name, version[0]),
```
JesseFarebro, line 412: It would be even nicer if we returned a `namedtuple` from `_versions` so we could do (`id.namespace, name, id.version`) instead of indexing into the tuple. This isn't necessary but just something I thought of.

tests/envs/test_registration.py
```diff
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
+import pytest
+
 import gym
-from gym import error, envs
+from gym import envs, error
```
JesseFarebro, line 5: Are these formatting changes necessary to pass CI?

tests/envs/test_registration.py
```diff
@@ -24,11 +26,47 @@ def __init__(self, arg1, arg2, arg3):
 
 
 def test_make():
-    env = envs.make("CartPole-v0")
-    assert env.spec.id == "CartPole-v0"
+    env = envs.make("CartPole-v1")
```
JesseFarebro, line 29: Is this also necessary to pass CI all of a sudden?

.gitignore
```diff
@@ -11,6 +11,7 @@
 
 # Virtualenv
 /env
+.direnv/
```
JesseFarebro, line 14: What is `.direnv/`? Does this belong in this commit?

gym/envs/registration.py
```diff
@@ -1,10 +1,10 @@
-import re
-import sys
+import contextlib
```
JesseFarebro, line 1: Are these import formatting changes necessary to pass CI now?

JesseFarebro (changes requested):

tests/envs/test_registration.py
kir0ul: No no it's not, it just removes the warning that the environment is deprecated. Should I keep version 0?

tests/envs/test_registration.py
kir0ul: No no, they're not, I just used [isort](https://pycqa.github.io/isort/) so that imports are standardized (kind of `black` but for imports statements), but I can revert that if you prefer.

.gitignore
kir0ul: It's a virtual env that auto activates when you `cd` at the root of the repo or below: https://direnv.net/
Not sure if it belongs to this PR, it's just something I use, but I can remove it if you prefer.

gym/envs/registration.py
kir0ul: Same as https://github.com/openai/gym/pull/2535#discussion_r776417773

gym/envs/registration.py
kir0ul: Smart!

tests/envs/test_registration.py
JesseFarebro: Yeah, if you don't mind just leaving it at `-v0`, we can fix things like this in another PR. Same thing with the `isort` changes. These are definitely nice to have but we should keep this PR as the minimal changes needed.

.gitignore
JesseFarebro: Yeah, it probably doesn't belong in this commit. You could keep this in your global `.gitignore` file in `~/.config/git/ignore`. I do this with quite a few entries.

gym/envs/registration.py
kir0ul: Right, that was probably overkill.

gym/envs/registration.py
kir0ul: I like this idea!

.gitignore
kir0ul: You're right, it's probably the best way to handle my own setup globally.

JesseFarebro:
Hey @kir0ul, thanks for the revisions. I was starting to see some cracks in a couple of places with the registration system (no fault of your own, it was some debt from when I originally wrote it) so I decided to rework some things. I built off your latest commit, could you take a look here and cherry-pick this commit into this PR then we can continue reviewing from there: https://github.com/JesseFarebro/gym/commit/bdda3500cf07adc718c22ac89f93042d74f296c7

I also combined the `difflib` suggestions into this commit. The other notable changes are:

I created new error types: `NamespaceNotFound`, `NameNotFound`, and `VersionNotFound` so we can construct the error messages in `EnvSpecTree`. Furthermore, I converted `EnvSpec` to a dataclass and I take advantage of `InitVar` and `__post_init__` to do the parsing of `namespace, name, version`. This simplifies the code a lot in other places.

Let me know what you think.

kir0ul:
> I built off your latest commit, could you take a look here and cherry-pick this commit into this PR then we can continue reviewing from there: [JesseFarebro@6cde417](https://github.com/JesseFarebro/gym/commit/6cde41702e503b66525322ca7f4b1776209ffd48)

Done!

> I created new error types: `NamespaceNotFound`, `NameNotFound`, and `VersionNotFound` so we can construct the error messages in `EnvSpecTree`. Furthermore, I converted `EnvSpec` to a dataclass and I take advantage of `InitVar` and `__post_init__` to do the parsing of `namespace, name, version`. This simplifies the code when having to deal with the full environment id.

I think that makes a lot of sense and the code feels cleaner, especially with the new error types.
Maybe a general comment: are there any _cracks_ you were seeing that you think could be converted to new test cases?

As a side note, coverage for `gym/envs/registration.py` is only of 78% and could be improved as many edge cases are not tested.

gym/envs/registration.py
```diff
@@ -169,48 +174,166 @@ def __init__(self):
...
-            return True
+            message += "is deprecated. "
+            message += f"Please use the latest version `v{latest_spec.version}`."
+            raise error.DeprecatedEnv(message)
+        # Otherwise we've asked for a version that doesn't exist.
         else:
-            return False
+            message += f"{name} could not be found. "
+            if all_versions:
+                all_versions_sorted = sorted(
```
JesseFarebro, line 311: Also dangerous and is solved with the above changes to `_assert_version_exists`.

gym/envs/registration.py
```diff
@@ -305,24 +435,26 @@ def make(self, path: str, **kwargs) -> Env:
...
-        versions = self._versions(namespace, name)
+        # Get all versions of this spec.
+        versions = self.env_specs.versions(spec.namespace, spec.name)
 
         # We check what the latest version of the environment is and display a warning
         # if the user is attempting to initialize an older version.
-        latest_version, latest_ns = max(versions)
-        if int(version) < latest_version:
-            latest_id = env_id_from_parts(latest_ns, name, latest_version)
+        latest_spec = max(
```
JesseFarebro, line 443: Maybe rename this to `latest_versioned_spec` so it's clear we're filtering the default out.

gym/envs/registration.py
```diff
@@ -305,24 +435,26 @@ def make(self, path: str, **kwargs) -> Env:
...
+            filter(lambda spec: spec.version, versions),
+            key=lambda spec: cast(int, spec.version),
+            default=None,
+        )
+        if (
+            latest_spec
+            and latest_spec.version is not None
```
JesseFarebro, line 450: This won't be required as we're filtering this case out. This will be more clear if we rename it to `latest_versioned_spec`.

JesseFarebro (changes requested):

JesseFarebro:
Thanks for the changes again @kir0ul. I think we're almost there. I'm a little concerned about the coverage as well. For example, I think we need to cover the case where:

- `TestEnv` has versions: `[None, v5]` and we verify `spec("TestEnv"), spec("TestEnv-v0"), spec("TestEnv-v10")`. (This would have caught the `max` and `sorted` errors when mixing default and non-default envs.

tests/envs/test_registration.py
```diff
@@ -23,12 +25,111 @@ def __init__(self, arg1, arg2, arg3):
...
+        ("MyAwesomeEnv-vfinal", None, "MyAwesomeEnv-vfinal", None),
+        ("MyAwesomeEnv--", None, "MyAwesomeEnv--", None),
+        ("MyAwesomeEnv-v", None, "MyAwesomeEnv-v", None),
+    ],
+)
+def test_register(env_id, namespace, name, version):
+    envs.register(env_id)
+    assert gym.envs.spec(env_id).id == env_id
+    assert version in gym.envs.registry.env_specs.tree[namespace][name].keys()
+    del gym.envs.registry.env_specs.tree[namespace][name]
```
JesseFarebro, line 81: Should probably `del gym.envs.registry.env_specs[env_id]` so we properly run `__delitem__` and we cleanup any empty leftovers.

gym/envs/registration.py
```diff
@@ -62,31 +68,34 @@ class EnvSpec:
...
-
-        match = env_id_re.fullmatch(id)
-        if not match:
-            raise error.Error(
-                f"Attempted to register malformed environment ID: {id}."
-                f"(Currently all IDs must be of the form {env_id_re.pattern}.)"
-            )
-        self._env_name = match.group("name")
+    id_requested: InitVar[str]
+    entry_point: Optional[Union[Callable, str]] = None
```
JesseFarebro, line 72: Should convert all the rest of these to fields. For example,
```py
    entry_point: Optional[Union[Callable, str]] = field(default=None)
    reward_threshold: Optional[int] = field(default=None)
    nondeterministic: bool = field(default=False)
    ...
```

JesseFarebro:
@kir0ul there's no need to worry about hashing `EnvSpec` as I outlined before, I updated the comment here: https://github.com/openai/gym/pull/2535/files#r777573839 to reflect this.

gym/envs/registration.py
```diff
@@ -47,12 +27,38 @@ def load(name: str) -> Type:
...
+def parse_env_id(id: str) -> Tuple[Optional[str], str, Optional[int]]:
+    """Parse environment ID string format.
+
+    This format is true today, but it's *not* an official spec.
+    [username/](env-name)-v(version)    env-name is group 1, version is group 2
+
+    2016-10-31: We're experimentally expanding the environment ID format
+    to include an optional username.
+    """
+    pattern = re.compile(
```
JesseFarebro, line 39: We should probably lift `pattern` outside of this function so it gets compiled once and cached between calls to `parse_env_id`. We do call this function quite a bit. Perhaps again naming it `env_id_re`?

kir0ul: Sounds like a valid point!

tests/envs/test_registration.py
kir0ul: Yes, good catch!

jkterry1:
@JesseFarebro does this look good now?

JesseFarebro:
@jkterry1 not yet, we still need some changes to `_assert_version_exists` and a couple of test cases. 

@kir0ul as I'm not a collaborator on the Gym repo I can't push changes to your PR. I can help you work on this if you add me as a collaborator to your fork.

gym/envs/registration.py
```diff
@@ -62,31 +70,33 @@ class EnvSpec:
...
-            raise error.Error(
-                f"Attempted to register malformed environment ID: {id}."
-                f"(Currently all IDs must be of the form {env_id_re.pattern}.)"
-            )
-        self._env_name = match.group("name")
+    id_requested: InitVar[str]
+    entry_point: Optional[Union[Callable, str]] = field(default=None)
+    reward_threshold: Optional[int] = field(default=None)
+    nondeterministic: bool = field(default=False)
+    max_episode_steps: Optional[int] = field(default=False)
```
JesseFarebro, line 77: The default here should be `None`.

gym/envs/registration.py
```diff
@@ -62,31 +70,33 @@ class EnvSpec:
...
+    order_enforce: Optional[bool] = field(default=True)
```
JesseFarebro, line 78: Is this optional? It was before this PR but I don't believe it is. Can you remove `Optional`.

JesseFarebro (changes requested):

kir0ul:
> @kir0ul as I'm not a collaborator on the Gym repo I can't push changes to your PR. I can help you work on this if you add me as a collaborator to your fork.

Just added you!

I think this is starting to look good now, but I think we still need to check for some edge cases to test.

JesseFarebro:
I agree, the code is looking good.

I'll add that edge case I was talking about above. If you can think of any other test cases feel free to add them.

JesseFarebro (approved):

JesseFarebro:
To summarize these changes:

Environment ID Definition: [$namespace/]$name[-v$version] <- version is now optional

1. Users don't have to provide a version when registering an environment
2. Both the default version (unversioned) and versioned environments can be registered at the same time
3. Users will now get smart suggestions when making typos w.r.t. the namespace or name of the environment
4. Gym will now suggest supported versions whether that is versioned or unversioned.

The only point of contention for me is (2). I don't think versioned and unversioned environments should be able to be registered at the same time. I think it should work as follows:

If you register a versioned environment (e.g., `v0`) you shouldn't be able to register an unversioned environment. By default, if you request the unversioned environment Gym should return the latest versioned environment. This is equivalent to just saying: "give me the latest environment".

Otherwise, you can register the unversioned environment if versions make no sense for your environment.

jkterry1:
I agree with Jesse that support both an unversioned and versioned version at the same name makes sense. Otherwise I'm happy with all this. What changes are required per that one point?

kir0ul:
> What changes are required per that one point?

I'd say something like this:
- [ ] Check if there's already a versioned environment when registering an unversioned one, and raise an error if that's the case.
- [ ] Return the latest versioned environment with a warning if the unversioned environment has been requested.
- [ ] Test that an error is raised when registering an unversioned environment if a versioned environment of the same name already exists.
- [ ] Test that everything works fine when registering a versioned environment if an unversioned environment of the same name already exists.
- [ ] Test that the latest versioned environment is returned if an unversioned environment of the same name has been requested, and test that a that a warning is raised.

Also coverage for `gym/envs/registration.py` has slightly improved up to 79%, and most of the uncovered code is related to `internal_env_relocation_map` and `load_env_plugins()`.

kir0ul:
Hey @JesseFarebro, the last pushed commit should address the last contention point.
Also coverage for `gym/envs/registration.py` is now at 80%.

JesseFarebro:
@kir0ul I'm a little confused with this point:

>  Test that everything works fine when registering a versioned environment if an unversioned environment of the same name already exists.

Did we not agree that this was disallowed? We shouldn't allow a user to register a versioned environment if the unversioned env already exists? Perhaps we weren't in agreement on this? @jkterry1?

If we are disallowing this I have some local changes I'll push to your branch. If we are allowing this there still needs to be some additional changes.

kir0ul:
Ah maybe I misunderstood your last comment, sorry about that.

jkterry1:
Hey guys, what's the status of this PR?

JesseFarebro:
@jkterry1 @kir0ul I should be finishing this up tonight.

gym/envs/registration.py
```diff
@@ -456,10 +456,13 @@ def make(self, path: str, **kwargs) -> Env:
             logger.info("Making new env: %s (%s)", path, kwargs)
         else:
             logger.info("Making new env: %s", path)
-        spec = self.spec(path)
```
JesseFarebro, line 306: This could have failed before if you requested the unversioned environment. `spec()` would try to lookup that version and immediately fail.

gym/envs/registration.py
```diff
@@ -513,36 +517,58 @@ def spec(self, path: str) -> EnvSpec:
     def register(self, id: str, **kwargs) -> None:
         spec = EnvSpec(id, **kwargs)
 
-        # Get all versions of this spec.
-        versions = self.env_specs.tree[spec.namespace][spec.name].values()
```
JesseFarebro: Because we are using `defaultdict` we should never use `.tree[][]` as this will create dicts in places we probably don't want them. You could see this if you call `repr()` on `EnvSpecTree` you'll get blank entries in places you don't expect.

gym/envs/registration.py
```diff
@@ -513,36 +517,58 @@ def spec(self, path: str) -> EnvSpec:
...
+                    f"Can't register the unversioned environment `{spec.id}` "
+                    f"when version `{latest_versioned_spec.version}` "
+                    "of the same name already exists. Note: the default "
+                    "behavior is that the `gym.make` with the unversioned "
+                    "environment will return the latest versioned environment."
+                )
+                raise error.RegistrationError(message)
+        # We might not find this namespace or name in which case
+        # we should continue to register the environment.
+        except (error.NamespaceNotFound, error.NameNotFound):
```
JesseFarebro, line 566: Not sure how I feel about this bare except but I think it's fine for our purposes. If we can't find a namespace or name then we most definitely want to continue to register the environment.

kir0ul:
> @kir0ul do you want to take a look at the most recent commit?

Sure I will have a look, hopefully today if I can or over the weekend, at least for the sake of having another pair of eyes on the code.
 
> I've added a new error type: `RegistrationError` for the case where you try to register a versioned env when unversioned exists or vice versa.
> 
> There were also some other small changes from your last commit.

Sounds good!

gym/envs/registration.py
kir0ul: Well it works, and I don't have a better suggestion either :sweat_smile:

kir0ul:
@JesseFarebro I just made some slight modifications to the tests, otherwise it looks pretty good to me!

JesseFarebro (approved):

PR #2536, opened 2021-12-20
Added singledispatch utility to vector.utils & changed order of space argument.

RushivArora:
After speaking to @jkterry1, I have created this PR for:

1) Adding the singledispatch utility to all the vector.utils to make the functions more generic
2) Changing the order of the **space** argument to be the first in all function definitions in '''vector.utils''' and changing corresponding function calls in **VectorEnv**, **SyncVectorEnv** and **AsyncVectorEnv**

Let me know if anything else needs to be changed!

~ Rushiv


jkterry1:
@lebrice if you could review this I'd greatly appreciate it

jkterry1:
@RushivArora can you please fix tests?

RushivArora:
@jkterry1 I believe the tests are now fixed: just a syntax error by calling on _BaseGymSpaces

RushivArora:
@jkterry1 Sorry for the failed tests! It missed my eye but some of the tests will need reworking since **space** was shifted to be the first argument in the code for the PR.

I'll rework the tests to match the new syntax, test it thoroughly, and push the changes again in a day or two!

RushivArora:
@jkterry1 This latest commit has the reworking of the following tests (in addition to the main PR changes):

- tests/vector/test_numpy_utils.py
- tests/vector/test_shared_memory.py

It passes all the tests on my local system, but it will not pass the tests on your workflow if you use the older ones.

jkterry1:
@RushivArora can you please fix black?

RushivArora:
@jkterry1 will fix it! II'll let you know as soon as I'm done

RushivArora:
@jkterry1 the PR is good to merge!

gym/vector/utils/numpy_utils.py
```diff
@@ -108,29 +110,26 @@ def create_empty_array(space, n=1, fn=np.zeros):
...
+    raise ValueError(
+        f"Space of type `{type(space)}` is not a valid `gym.Space` instance."
+    )
 
 
+@create_empty_array.register(Box)
+@create_empty_array.register(Discrete)
+@create_empty_array.register(MultiDiscrete)
+@create_empty_array.register(MultiBinary)
 def create_empty_array_base(space, n=1, fn=np.zeros):
```
lebrice, line 122: Opinion: All the concrete implementations for each type should have a name starting with an underscore, to indicate that they shouldn't be called directly, and also not to "pollute" the namespace (to make it easy to understand which function is to be used).

lebrice (commented):
Hey there @RushivArora , 

This PR does something similar to one that I opened before (and haven't maintained in a little bit): #2093 

Here are some of the differences between the two PRs:
- Other PR also changes the functions from `gym.spaces.utils`, such as `flatten_space` and `flatten`, (**edit**: but these have already been introduced in #2328 earlier).
- The argument ordering change is backward-compatible: if the arguments are passed in the wrong order, then they are reordered and I think a warning gets raised.
- One subtle thing that `singledispatch`changes is that any calls to the function must pass the space as the first positional argument. This is backward-incompatible also. I added a little wrapper function around the new singledispatch functions, called `_space_as_first_pos_arg` to fix this.

The changes here look good to me.
@jkterry1 are these backward-incompatible changes fine? If so, then I'd be inclined to approve this PR instead of #2093.  

Thanks!

lebrice (approved):

gym/vector/utils/numpy_utils.py
RushivArora: Hi @lebrice, thank you for the review appreciate it! And happy new year :)

I like this suggestion. Looking at the changed in #2328, they haven't renamed their concrete implementations to start with an underscore. Therefore, in the interest of avoiding confusion to future readers of the code I can do one of these two:
1) Either we leave it without an underscore
2) I can make the changes to the concrete implementations of flatten and flatten_space functions in #2328 to start with an underscore, and then do the same for the functions being modified in this PR.

@jkterry1 let me know what works best!

jkterry1: @RedTachyon what do you want done here?

RedTachyon: I'd also prefer underscored names with the same reasoning as @lebrice, overall I don't have much to add beyond that

lebrice: Sounds like option 2 it is then @RushivArora

RushivArora: Changes are made, should be good now @jkterry1 @lebrice

jkterry1:
@lebrice @RedTachyon should I merge this now?

PR #2538, opened 2021-12-21
Update importlib_metadata entry_point calls

JesseFarebro:
On Python 3.10+ there are deprecation warnings for using `.get()` on `EntryPoints`. If we use an updated version of the backport library `importlib_metadata` and the native `importlib.metadata` for Py310+ this shouldn't be an issue.


jkterry1:
@JesseFarebro this has merge conflicts now

PR #2541, opened 2021-12-24
typing in gym.spaces

ikamensh:
This adds typing to spaces.Box and spaces.Discrete (Issue #2452).

This was made tricky by type checker not recognising that np.isscalar() effectively narrows type if it returns false; forcing to add assertions and minor refactoring.

One other thing that typing system has highlighted is that gym.Space probably shouldn't have ._shape / .shape: actually many subtypes don't use it. Arguably it belongs into `spaces.Box/MultiBinary/MultiDiscrete` only. Please let me know if such change is in line with the vision for gym API.

### DIFF
--- a/gym/spaces/box.py
+++ b/gym/spaces/box.py
@@ -1,3 +1,5 @@
+from typing import Tuple, SupportsFloat, Union, Type, Optional, Sequence, List
+
 import numpy as np
 
 from .space import Space
@@ -15,7 +17,7 @@ def _short_repr(arr):
     return str(arr)
 
 
-class Box(Space):
+class Box(Space[np.ndarray]):
     """
     A (possibly unbounded) box in R^n. Specifically, a Box represents the
     Cartesian product of n closed intervals. Each interval has the form of one
@@ -33,33 +35,29 @@ class Box(Space):
 
     """
 
-    def __init__(self, low, high, shape=None, dtype=np.float32, seed=None):
+    def __init__(
+        self,
+        low: Union[SupportsFloat, np.ndarray],
+        high: Union[SupportsFloat, np.ndarray],
+        shape: Optional[Sequence[int]] = None,
+        dtype: Type = np.float32,
+        seed: Optional[int] = None,
+    ):
         assert dtype is not None, "dtype must be explicitly provided. "
         self.dtype = np.dtype(dtype)
 
         # determine shape if it isn't provided directly
         if shape is not None:
             shape = tuple(shape)
-            assert (
-                np.isscalar(low) or low.shape == shape
-            ), "low.shape doesn't match provided shape"
-            assert (
-                np.isscalar(high) or high.shape == shape
-            ), "high.shape doesn't match provided shape"
         elif not np.isscalar(low):
-            shape = low.shape
-            assert (
-                np.isscalar(high) or high.shape == shape
-            ), "high.shape doesn't match low.shape"
+            shape = low.shape  # type: ignore
         elif not np.isscalar(high):
-            shape = high.shape
-            assert (
-                np.isscalar(low) or low.shape == shape
-            ), "low.shape doesn't match high.shape"
+            shape = high.shape  # type: ignore
         else:
             raise ValueError(
                 "shape must be provided or inferred from the shapes of low or high"
             )
+        assert isinstance(shape, tuple)
 
         # handle infinite bounds and broadcast at the same time if needed
         if np.isscalar(low):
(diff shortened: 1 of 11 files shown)


gym/spaces/box.py
```diff
@@ -22,67 +24,71 @@ class Box(Space):
...
         self.bounded_below = -np.inf < self.low
         self.bounded_above = np.inf > self.high
 
         super().__init__(self.shape, self.dtype, seed)
 
-    def is_bounded(self, manner="both"):
-        below = np.all(self.bounded_below)
-        above = np.all(self.bounded_above)
+    @property
+    def shape(self) -> Tuple[int, ...]:
```
RedTachyon, line 85: Is it possible to only add a type hint for that case, and not override the implementation? It's the same so ultimately it doesn't matter much, but it looks a bit odd.

My intuition is that only something like a type stub could do that which would be a bit of a hassle and not worth it in that case imo.

gym/spaces/box.py
```diff
@@ -22,67 +24,71 @@ class Box(Space):
...
+        """Has stricter type than gym.Space - never None."""
+        return self._shape
+
+    def is_bounded(self, manner: str = "both") -> bool:
```
RedTachyon, line 89: I'm not a fan of the string argument here. Either an enum, or maybe some booleans instead would be nice. The current approach slightly defeats the point of adding types, since most strings don't actually make any sense as an input.

gym/spaces/dict.py
```diff
@@ -1,10 +1,12 @@
 from collections import OrderedDict
 from collections.abc import Mapping, Sequence
+from typing import Optional, Union
+import typing as t
```
RedTachyon, line 4: Not a huge fan of this import, I'm assuming it's for `t.Dict`? I think it'd be better to do something like `from typing import Dict as DictType` or something like that; or just `typing.Dict` in the code.

Or commit to the newer style and use `dict` for type hints, see my later comments

gym/spaces/dict.py
```diff
@@ -57,10 +64,10 @@ def __init__(self, spaces=None, seed=None, **spaces_kwargs):
                 space, Space
             ), "Values of the dict should be instances of gym.Space"
         super().__init__(
-            None, None, seed
+            None, None, seed  # type: ignore
         )  # None for shape and dtype, since it'll require special handling
 
-    def seed(self, seed=None):
+    def seed(self, seed: Optional[Union[t.Dict, int]] = None) -> list:
```
RedTachyon, line 70: For the return type here, I'd say either use `typing.List`, or commit to using `list`, `dict` and `tuple` everywhere. Since 3.7 (earliest we support), those can be made generic through `from __future__ import annotations`

gym/spaces/discrete.py
```diff
@@ -1,8 +1,10 @@
+from typing import Optional
+
 import numpy as np
 from .space import Space
 
 
-class Discrete(Space):
+class Discrete(Space[int]):
```
RedTachyon, line 7: Is this actually correct? `space.sample()` will return a np.int64 by default, and from a quick check it seems that this might cause type checkers to throw some false warnings. Can you double check if this will actually work in terms of subtypes etc between python `int` and numpy `int`s?

gym/spaces/tuple.py
```diff
@@ -1,25 +1,29 @@
+from typing import Iterable, List, Optional, Union
+import typing as t
```
RedTachyon, line 2: Same comment as with Dict

RedTachyon (commented):
Overall good progress, I left a few comments, I think one for the actual functionality, and a few style things

gym/spaces/dict.py
ikamensh: I didn't realize `dict` works with __future__ annotations in 3.7. Will use that.

gym/spaces/dict.py
ikamensh: ok, will use native types whenever possible.

gym/spaces/box.py
ikamensh: Yeah I also believe type stubs are the only way to add types without touching the implementation.

gym/spaces/discrete.py
ikamensh: As far as I can tell, Discrete produces native int on sample method now:
```python
Python 3.8.8 (default, Apr 13 2021, 12:59:45) 
[Clang 10.0.0 ] :: Anaconda, Inc. on darwin
>>> from gym.spaces import Discrete
>>> d = Discrete(3)
>>> type(d.sample())
<class 'int'>
```

gym/spaces/box.py
ikamensh: The usability issue with an enum is that users need to import it:
if we have
```python
class Bound(Enum):
    BELOW = auto()
    ABOVE = auto()
    BOTH = auto()
```

Then either every user needs to import this Enum, or we should put it inside the Box class to be imported and available with it. We could also support both strings and Enum, but it's not a big win then. Overall the pattern of using strings in places that actually should have enums is widespread in ML, unfortunately - from losses and activations in Keras / Scikit-learn to environment creation in gym.

I would in general be in favor of revamping this part and using Enums, but it's better done in a coordinated effort and with preliminary design decision on which pattern exactly should be used.

So I suggest letting it be as it is right now, but maybe creating an issue to gather feedback on removing this pattern in gym in general. what do you think?

ikamensh:
Style comments are addressed

gym/spaces/discrete.py
RedTachyon: Correct, my bad

RedTachyon:
TIL the future import also allows using `a | b` for `Union[a, b]`

In any case, everything has been adressed so I'm happy with this being merged now

jkterry1:
@ikamensh could you please fix merge conflicts and then I'll merge?

ikamensh:
Rebase done. Note: I needed to add code types to the new functionality, which worked best with a minor refactoring.

@RedTachyon I've found the way to formulate restrictions on specific string constants that are supported: `Literal["both", "below", "above"]`. Unfortunately it's not available in python37.

ikamensh:
Hm, type checker fails in the cloud but not locally. Will investigate further.

ikamensh:
Type error reported as probable bug https://github.com/microsoft/pyright/issues/2852 and ignored for now. Build passes.

jkterry1:
@ikamensh according to github there's also still merge conflicts that can't be automatically handled

ikamensh:
They must have been new one. Rebased again.

jkterry1:
@ikamensh Ariel just sat down to review this again and there's still merge conflicts or new ones, could you please rebase?

ikamensh:
Rebased. Kind request to process this before merging more conflicting PRs :)

jkterry1:
Merged, my apologies

ikamensh:
Yay that was a big one!

PR #2544, opened 2021-12-26
Update box.py

modanesh:
At the end of the discussion in this PR  #2374 , @zlig pushed a commit to fix a docstring but it was after the PR was merged. So it basically didn't make it into the master. As I looked into the modified files, `test_flatten_observation.py` and `test_spaces.py` were already updated. Also, as per our discussion with @jkterry1 ,  the `kellycoinflip` env has been removed from the gym. So, there's only this file `gym/spaces/box.py` left to be updated, which is what this PR does.

### DIFF
--- a/gym/spaces/box.py
+++ b/gym/spaces/box.py
@@ -57,7 +57,7 @@ class Box(Space):
         if np.isscalar(high):
             high = np.full(shape, high, dtype=dtype)
 
-        self.shape = shape
+        self._shape = shape
         self.low = low
         self.high = high
 
@@ -81,7 +81,7 @@ class Box(Space):
         self.bounded_below = -np.inf < self.low
         self.bounded_above = np.inf > self.high
 
-        super(Box, self).__init__(self.shape, self.dtype)
+        super(Box, self).__init__(self._shape, self.dtype)
 
     def is_bounded(self, manner="both"):
         below = np.all(self.bounded_below)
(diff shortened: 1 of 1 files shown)


gym/spaces/box.py
```diff
@@ -70,15 +71,17 @@ def _get_precision(dtype):
         high_precision = _get_precision(self.high.dtype)
         dtype_precision = _get_precision(self.dtype)
         if min(low_precision, high_precision) > dtype_precision:
-            logger.warn(f"Box bound precision lowered by casting to {self.dtype}")
```
ikamensh, line 73: Not a useful change?

ikamensh:
Looking at the changelog it's not obvious what is the intention of this change. Is it to use `._space` instead of `.space`? Why is it a good idea?

modanesh:
Basically it's based on the pushed commit at the end of this discussion thread: #2374 . The main idea to fix a docstring.

ikamensh:
I'm looking at the diff, and can't see a docstring changed, but instead a bunch of other changes.

RedTachyon:
I'm confused, what's the point of this PR? It removes the seeding mechanism from init, which is in general a messy affair, but having the argument in the constructor is the least of its problems. It changes the logging to sometimes use `warnings` instead of always `gym.logger`, introduces an antiquated `str.format` instead of an existing f-string, and needlessly changes calls to `self.shape` to `self._shape`, even though the former is the getter for the latter.

I also don't see any docstring changes?

jkterry1:
Thanks ariel, I assume this was a miscommunication on my end then

Issue #2531, opened 2021-12-16
TypeError: reset() got an unexpected keyword argument 'seed'


deepboltzer:
There is an issue with the seeds after the Seeding update (#2422).

```
File "/home/joe/reinforce/rlenv/lib/python3.8/site-packages/gym/wrappers/time_limit.py", line 29, in reset
    return self.env.reset(seed=seed, **kwargs)
TypeError: reset() got an unexpected keyword argument 'seed'
```

*Code for Imitation:* 

```
import gym
import ale_py
import time

print('gym:', gym.__version__)
print('ale_py:', ale_py.__version__)

env = gym.make("BreakoutNoFrameskip-v4")

print("Observation Space: ", env.observation_space)
print("Action Space       ", env.action_space)

obs = env.reset()

for i in range(1000):
    action = env.action_space.sample()
    obs, reward, done, info = env.step(action)
    env.render()
    time.sleep(0.01)
env.close()
```
I am using gym: 0.21.0 and ale_py: 0.7.3.

vwxyzjn:
In pybullet I just hit the same issue. It's likely going to cause all third-party envs to fail.

vwxyzjn:
@RedTachyon would you mind looking into this?

RedTachyon:
I'm confused, are you sure it's gym release 0.21? This was released long before any seeding changes made their way into master.

RedTachyon:
Ah shit, I managed to replicate it with pybullet, I think I know what's up. Basically wrappers forward the arguments to the inside environment, and while "new style" environments can accept anything in `reset`, old environments can't. So even if you don't do anything, it's trying to pass the default `None` onward to the environment. 

Thanks for the catch, I think I have an idea on how to fix it, which will be possible thanks to the other change to the reset signature.

RedTachyon:
The latest commit in #2515 fixes it for me with pybullet, can someone check it with atari? (for some reason a simple install doesn't work when I try to use it, and I never used atari, so probably some of the ROM stuff is lost on me)

deepboltzer:
Thank you so much for fixing this issue @RedTachyon! The latest commit in #2515 also fixes the problem for atari.

vwxyzjn:
Hey I am still having this issue with the latest gym. @jkterry1 could u re-open the issue? See the following snippet:

```
import gym
import numpy as np
import pybullet_envs

env = gym.make("HopperBulletEnv-v0")
env.reset()

[...]

PR #2546, opened 2021-12-27
Adding return_info argument to reset to allow for optional info dict as a second return value

balisujohn:
This is an initial draft to make the change discussed in #1683. All changes made here are concerning the definition and implementations of the reset function in the env class in core.py.

I changed the abstract env class in core.py , so that it declares the optional argument return_info in the header

I Implemented the change for the cartpole environment.

The default behavior is to not return the info, but if the optional argument return_info is set to true, reset returns both an observation and the info dictionary.

The tests still seem to pass, and best I can tell it doesn't break subclasses of env which don't declare the optional return_info argument in the reset function header. This is an important consideration for not suddenly breaking existing third-party gym envs which subclass env.

I will try to implement this change for all the first party gym environments and extend the existing test to verify this functionality for all first party gym environments.

Any feedback is welcome.

TODO
- [x] Implement change for environments other than cartpole 
- [x] Extend test to environments other that cartpole
- [x] Make sure type checking passes
- [x] fix the merge conflict and figure out how to create a multi-case return type annotation
- [x]  Check if #2531 is going to be a problem
- [x] Ensure vectorized env compatibility.
- [x] Ensure compatibility of changes with stable-baselines3


jkterry1:
@balisujohn thanks a ton for this, I asked Ariel to review it when he's back. Can you please fix the merge conflict?

jkterry1:
@balisujohn can you please fix tests?

jkterry1:
@balisujohn its been a week so I'm just following up on this

jjshoots:
@jkterry1 Actually I think the only test that's failing is black, which isn't too hard to do if you'd like to merge it soon.

balisujohn:
This is not ready to merge. On my local branch I extended the change to a lot of other first party environments and it broke a lot of vectorized environment tests. Changes are necessary to the vectorized environment code to make the existing tests pass, and new tests will need to be added to make sure the vectorized environments also respond appropriately to the return_info argument in the reset function.

gym/core.py
```diff
@@ -30,7 +30,7 @@ class Env(Generic[ObsType, ActType]):
     And set the following attributes:
 
         action_space: The Space object corresponding to valid actions
-        observation_space: The Space object corresponding to valid observations
+        observation_spasce: The Space object corresponding to valid observations
```
RedTachyon, line 33: Typo

gym/core.py
```diff
@@ -70,7 +70,7 @@ def step(self, action: ActType) -> Tuple[ObsType, float, bool, dict]:
         raise NotImplementedError
 
     @abstractmethod
-    def reset(self, seed: Optional[int] = None) -> ObsType:
+    def reset(self, seed: Optional[int] = None , return_info: Optional[bool] = False) -> Union[ObsType,Tuple[ObsType, dict]]:
```
RedTachyon, line 73: It shouldn't be Optional[bool], just bool (but still with a default of False)

Despite the appearances, `Optional` doesn't mean that the argument is optional, but that it can also take the value `None`, which doesn't make sense here

gym/envs/classic_control/cartpole.py
```diff
@@ -155,12 +155,14 @@ def step(self, action):
...
         return np.array(self.state, dtype=np.float32), reward, done, {}
 
-    def reset(self, seed: Optional[int] = None):
+    def reset(self, seed: Optional[int] = None, return_info: Optional[bool] = False):
         super().reset(seed=seed)
         self.state = self.np_random.uniform(low=-0.05, high=0.05, size=(4,))
         self.steps_beyond_done = None
-        return np.array(self.state, dtype=np.float32)
-
+        if not return_info:
```
RedTachyon, line 162: So this will be the problematic part, since we'll need to patch this in all environments (not a big deal), but also make sure that third-party environments won't have problems with it (and I don't think there's anything we can do to automatically make them obey the new contract of `reset`)

RedTachyon (commented):
I'm not sure what the issues with vector API are, but I can very easily believe there are many. Overall I'm not fully convinced it's good to introduce this in a 0.x release, since it is a major change of the core behavior, even if it's optional

gym/core.py
balisujohn: Good catch, I will fix this.

gym/envs/classic_control/cartpole.py
balisujohn: I've started work on extending this patch to all first-party environments, so far, I haven't found any which return non-empty metadata, which is interesting.

gym/core.py
balisujohn: good catch, will fix

jkterry1:
@balisujohn Hey man, I'm just following up on the status of this. Relatedly, tests still aren't passing

balisujohn:
@jkterry1 I'll see if I can get what I have pushed by the end of the week. The sticking point is the AsyncVectorEnv, it's proving a bit hard to get the worker threads to also return a the metadata dict, but I expect I can fix it.

jkterry1:
@balisujohn could you please push whatever you have in the fairly near future? We need to include this in the next release of Gym, so I'm going to have Ariel finish whatever you haven't sometime next week

jkterry1:
(and could you please fix the new merge conflicts too?)

balisujohn:
So basically the return_info argument is implemented for all first party environments, and all implemented first party environments aside from the unit-test environments are tested in the test test_reset_info in test_envs.py. 

There is some work towards extending this functionality to both wrappers and vector environments, though tests still need to be written to check if the wrappers and vector environments correctly return metadata from the reset function. 

async_vector_env.py proved particularly difficult to instrument to optionally return metadata on reset, as the metadata must be passed back from worker threads conditionally, based on the the value of the return_info argument being sent to worker threads, so for now it is left unimplemented.

- [ ] return_info = True test for sync_vector_env.py
- [ ] return_info = True test for async_vector_env.py
- [ ] return_info = True implementation for async_vector_env.py
- [ ] return_info = True test for normalize.py 
- [ ] decide if record_episode_statistics should support return_info argument (should we return the initial info from calling reset)

balisujohn:
Also of note the only first party environments that return non-empty metadata dicts under any circumstance as far as I could tell were a subset of the toytext environments(frozen lake for example). It might be worth reviewing my decision to return a probability 1 as the metadata value returned from the reset function. It seems to be logically consistent with the meaning of the "prob" metadata parameter, but it's worth double checking.

balisujohn:
And a final note: it is of concern whether or not the wrappers/vector envs assume that any particular optional arguments are available the reset functions of third party environments they are wrapping. It seems like the sync_vector_env assumes that  the 'seed' keyword arguments will be available, which is a breaking change for third party environments which predate the creation of the 'seed' keyword argument in reset. So it's important to note I believe we have already made breaking changes to at least sync_vector_env.

RedTachyon:
@balisujohn I can't test it right now, but when I was editing the code for vector envs, the intention was to keep it backwards compatible, and when I'm reviewing it now, I'm like 80% that it is. If someone's using an old style environment that doesn't support these arguments, as long as they don't try to explicitly pass a seed, it should all be fine. 

To make old environments compatible with the changed API functionality, they need to be updated, perhaps we'll also make a wrapper to do it automatically in gym.make.

balisujohn:
@RedTachyon I'll see if I can create a minimal example that either confirms or disproves my claim. The main things left for this pull request other than a general review for quality are to decide if/how to implement the optional return_info behavior for atari_preprocessing.py and also to implement and test the return_info behavior in async_vec_env.py. I'm curious what your opinion is regarding the atari_preprocessing.py wrapper(like where are the atari envs implemented and do we have the ability to make them conform to the new standards?).

RedTachyon:
@balisujohn So I did a test with pybullet envs, and it works the way I expected. A code snippet like this (with the new version of gym) works:

```python
import pybullet_envs
import gym
env = gym.vector.make("HopperBulletEnv-v0")
env.reset()
```

but calling
```python
env.reset(seed=0)
```

causes a crash. So old-style environments still work with old-style resets. And if you want to do a new-style reset with an old-style behavior, that requires some more thought, but isn't strictly speaking necessary for backwards compatibility. 


As for atari - I don't use it and can't really test it right now, but the maintainer of ALE is in the loop and will update it when the time comes. For the moment it should work the same way the pybullet example does.

balisujohn:
I think this is ready for final review. I checked and at least for cartpole stable-baselines3 works fine. It seems like my fears about compatibility with out of data environments being broken by sync_vector_env were unfounded. Barring any mistakes, there are implementations and tests of the optional return_info behavior for all first-party environments, wrappers, and vector environments, except for atari_preprocessing.py. I think we should get feedback from someone who's more focused on the atari side of things, because it seems like the existing tests for atari_preprocessing.py have been disabled since the folder atari in the envs folder was deleted in a previous commit.

balisujohn:
It seems like the testing environment doesn't have mujoco-py installed, meaning the tests fail when testing the reset behavior on the mujoco environments. Shall I remove the tests for the first part mujoco environments, or would it be preferred to add mujoco-py as a testing environment dependency?

jkterry1:
@balisujohn all tests are currently disabled for mujoco environments AFAIK

balisujohn:
I added a test that tests the reset behavior for all first party environments including the mujoco environments, so I can remove those environments from the test.

jkterry1:
Right, please pull the mujoco tests. They'll be removed from Gym in some number of months anyways

jkterry1:
@balisujohn tests failed due to linting

balisujohn:
It seems like my version of black from python3.9 disagrees about how to format  gym/utils/atomic_write.py
 with the version being used in the pre-commit deployment step. (black version 21.12b0)

jkterry1:
@balisujohn I think that file is dead code, right? Why don't you just remove it.

balisujohn:
The linting should be fixed; I'm not sure what atomic_write.py does, this was the first time I noticed it. Maybe removing it would be a good first issue, to get new contributors familiar with the pull request process?

balisujohn:
@jkterry1  do you want this functionality added for atari_preprocessing.py? Otherwise, I think this pull request is ready for final review.

jkterry1:
@balisujohn let's have Jesse do that after this is merged

gym/core.py
```diff
@@ -71,7 +71,11 @@ def step(self, action: ActType) -> Tuple[ObsType, float, bool, dict]:
 
     @abstractmethod
     def reset(
-        self, *, seed: Optional[int] = None, options: Optional[dict] = None
+        self,
+        *,
+        seed: Optional[int] = None,
+        return_info: Optional[bool] = False,
```
XuehaiPan, line 77:
```suggestion
        return_info: bool = False,
```

`Optional[bool]` means `Union[bool, None]`. Same changes to the remaining type hints.

tests/envs/test_envs.py
```diff
@@ -55,6 +55,64 @@ def test_env(spec):
     env.close()
 
 
+def test_reset_info():
+    # blackjack gets its own assert statements because it returns a tuple as an obs, creating ambiguity for the normal assert statements
+    env_names = [
```
RedTachyon, line 60: I'm pretty sure there's a pattern somewhere in tests to automatically run tests on all valid envs, I think it's better to use that. This would need to be updated whenever one of these envs needs a version bump.

tests/wrappers/test_record_episode_statistics.py
```diff
@@ -24,6 +26,17 @@ def test_record_episode_statistics(env_id, deque_size):
     assert len(env.length_queue) == deque_size
 
 
+def test_record_episode_statistics_reset_info():
+    env = gym.make("CartPole-v1")
+    env = RecordEpisodeStatistics(env)
+    obs = env.reset()
+    assert isinstance(obs, np.ndarray) or isinstance(obs, tuple) or isinstance(obs, int)
```
RedTachyon, line 33: So two points, applying to this specific line and the ones similar in other parts of the tests.
1. As mentioned in another comment, you can parametrize the test so that it runs for all environments, so that they're all checked and not just CartPole
2. I think a better to check for validity of the observation would be `env.observation_space.contains(obs)`. It generalizes to all other weirder action spaces (for example this version would fail with a Dict action space)

tests/wrappers/test_record_video.py
```diff
@@ -29,6 +31,36 @@ def test_record_video_using_default_trigger():
     shutil.rmtree("videos")
 
 
+def test_record_video_reset_return_info():
+    env = gym.make("CartPole-v1")
+    env._max_episode_steps = 20
```
RedTachyon, line 36: I might be missing something, does this line actually do anything? Directly accessing a `_parameter` is a bit iffy

RedTachyon (changes requested):
Pretty good job overall, but there's a few changes that need to be done. Particularly important I think is making the tests run for all envs, since this can be an easy-to-break thing when we change/add envs.

balisujohn:
Thanks for the code review! I'll try to make the requested fixes soon.

jkterry1:
@balisujohn there's also merge conflicts

jkterry1:
@balisujohn Do you know how much longer making the fixes here will take? This is something that we need for the next upcoming release.

gym/core.py
balisujohn: Sounds good, it looks like I missed that one; will fix.

gym/core.py
```diff
@@ -71,7 +71,11 @@ def step(self, action: ActType) -> Tuple[ObsType, float, bool, dict]:
...
+        options: Optional[dict] = None,
     ) -> ObsType:
```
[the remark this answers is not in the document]
balisujohn: Fixing this.

balisujohn:
I updated the code based on the recommendations made in the code review. I merged the code with the current version in master. All tests pass when 

````
images = env.call("render", mode="rgb_array")
````

and the corresponding asserts in test_call_async_vector_env are disabled. When enabled, test_call_async_vector_env passes when only the tests in test_async_vector_env are called but fails with the following output:

````
tests/vector/test_async_vector_env.py ........XIO:  fatal IO error 11 (Resource temporarily unavailable) on X server ":1"
      after 766 requests (766 known processed) with 1 events remaining.
XIO:  fatal IO error 11 (Resource temporarily unavailable) on X server ":1"
      after 766 requests (766 known processed) with 4 events remaining.
XIO:  fatal IO error 11 (Resource temporarily unavailable) on X server ":1"
      after 767 requests (767 known processed) with 6 events remaining.
XIO:  fatal IO error 11 (Resource temporarily unavailable) on X server ":1"
      after 766 requests (766 known processed) with 7 events remaining.
FXIO:  fatal IO error 0 (Success) on X server ":1"
      after 766 requests (766 known processed) with 0 events remaining.
````

when being run with the entire test battery. This seems like an intermittent error related to rendering(potentially there is some complication making the render function fail in the vectorized asynchronous setting). Is it apparent to anyone why this is happening?

File: tests/wrappers/test_record_video.py (hunk 7 above)
Comment (balisujohn) [reply tests/wrappers/test_record_video.py]: This was just due to a naive copy of another test being used as a template.  It is now removed.

File: gym/vector/sync_vector_env.py [hunk 9]
```diff
@@ -99,19 +100,27 @@ def reset_wait(
...
         for env, single_seed in zip(self.envs, seed):
-            single_kwargs = {}
-            if single_seed is not None:
-                single_kwargs["seed"] = single_seed
-            if options is not None:
-                single_kwargs["options"] = options
-            observation = env.reset(**single_kwargs)
-            observations.append(observation)
+            if not return_info:
+                observation = env.reset(seed=single_seed, options=options)
```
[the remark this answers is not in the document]
balisujohn: You are totally right, It should now be fixed.

tests/envs/test_envs.py
```diff
@@ -55,6 +55,64 @@ def test_env(spec):
...
+        "FrozenLake-v1",
+        "FrozenLake8x8-v1",
+        "CliffWalking-v0",
+        "Taxi-v3",
+    ]
+    for env_name in env_names:
+        env = envs.make(env_name)
+        obs = env.reset()
+        assert isinstance(obs, np.ndarray) or isinstance(obs, int)
+        del obs
```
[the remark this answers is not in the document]
balisujohn: removed.

tests/envs/test_envs.py
balisujohn: Good catch, I made this change.

tests/wrappers/test_record_episode_statistics.py
balisujohn: 1. I switched the direct for all first party environments in test_envs to the parameterization decorator. For now I'm leaving the wrapper tests as single-environment tests, since many existing tests are like this, and it might make the tests take much longer. If people think it's worthwhile, I'm happy to add more coverage such that we try every first party env with every wrapper/vector env.

2. I made this switch for all the direct first party env tests  wrapper tests except normalize.py because it seems to not contain the observation in the observation space! This might be an error worth investigating. I also didn't implement this for the vector env tests.

RedTachyon:
I'm slightly confused, I don't see any rendering-related changes in the code here. Is the error only local on your machine, or did it also ever happen in the github pipeline?

If it's the former, I'd say if the error doesn't happen in the pipeline here, then it doesn't exist. I regularly get some errors on my local setup, but figured it's just a matter of my configuration not being good for some of the more esoteric features (shared memory, displays and all that nonsense)


Overall, if it's just a local error, my comments seem to be addressed so overall should be good to go.

balisujohn:
It's definitely local, because I was able to reproduce it with the version of gym currently in OpenAI gym master. Must be something weird with my Xorg installation.

Issue #2462, opened 2021-10-25
[Proposal] 3.6+ code style


ikamensh:
### Proposal 

Since gym no longer supports python2 or python below 3.7, there's an opportunity to use newer idioms: 
* f-strings
* implicit inheritance from object (`class Env(object)` -> `class Env`) 

and probably many others that are less wide spread.

[Optional] As this might touch many areas in the code, we could also create a list of commits to be ignored by git-blame (author annotations).

### Motivation

This will make source code slightly leaner and more readable.

### Pitch

I would suggest a series of PRs derived using linter tools like `pyupgrade` and `flynt` and any others. This way given that the tools are well tested, we can be more sure that the change doesn't introduce errors. I think this will be a good long-term investment into having modern source code that is nice to read and maintain.

### Alternatives

* Manual conversion of code as-you-go - has the drawback of manual effort and lack of consistency.
* Leave things as they are.

### Additional context

Naively running `find ./gym -name '*.py' -exec pyupgrade --py36-plus {} \;` and `flynt gym -ll 120` results in: 
`68 files changed, 216 insertions(+), 284 deletions(-)`. Here is a sample changelist https://github.com/ikamensh/gym/commit/b202b9b63d52e95d540b8d096127b1f67fd8ba7f - I could tune this more. 

Maybe we should omit modifying the environments or at least separate this to reduce the size of the change.

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/openai/gym/issues) in the repo (**required**)
Quick scan done, none found.

PR #2547, opened 2021-12-28
Py36+ code style in tests

ikamensh:
Arguably should have been done together with #2462. But better now than never :)

Derived using `pyupgrade --py36-plus` and `flynt -ll 120`, no manual modifications were done.


RedTachyon:
Looks good, pretty straight-forward

PR #2551, opened 2022-01-06
Move rendering.py to utils

carlosluis:
As discussed with @jkterry1, the file `rendering.py` is being used also outside `gym/envs/classic_control`, thus it is better suited to be under `utils`.


wookayin:
Note: #2599 completely removed pyglet rendering.

Issue #2497, opened 2021-11-23
[Proposal] Simplify the Box string representation


RedTachyon:
### Proposal 

Stop the `Box` string representation from printing out all the low/high values if they are all identical

### Motivation

In any reasonably large environment, `print(env.observation_space)` will spam the output with dozens of zero's or infs or whatever else. This information may be relevant if some of these values are different, but often is not necessary, and is just noise. And good luck printing a Dictionary observation space.

I think it'd be enough to change the __repr__ method on Box so that if `len(set(self.low)) == 1`, we just print the unique value, and the same thing for `self.high`.

### Pitch

Make Boxes printable again

### Alternatives

Grin and bear it

### Additional context

![image](https://user-images.githubusercontent.com/19414946/143067427-b2b999ba-e430-458a-81c3-f633db2a7b8f.png)
Help

This is partially going back on #2182 and #2183 , but I wasn't around to yell alarm at the full consequences of that PR. But since it will only simplify the representation when all limits are equal, it doesn't actually cause that same problem.

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/openai/gym/issues) in the repo (**required**)

jkterry1:
I mean the linked fix was less bad broken than before. I'm fine printing max/max only if both are uniform, that's definitely more readable, though a print statement clarifying that only single values are being displayed is needed to accompany a fix like this

Markus28:
I have implemented this feature and would be happy to raise a pull request. However, I'm not sure what you mean by a "clarifying print statement". In my implementation, only one scalar is printed if the limit is a multiple of the all-ones vector. This seemed quite natural to me since this representation is a valid way to instantiate an equivalent `Box` object (because the shape is also part of the `__repr__` string). Of course, one could also print something like `[inf, ..., inf]`, but this might still be hard to decipher if the shape is more complicated. Moreover, it wouldn't be obvious whether the omitted values are also `inf`s.

RedTachyon:
@Markus28 if you could make a PR with this it'd be great, I think a simple check whether all elements of the (lower/upper) limit are equal, and then choose the representation based on that. Probably computing this on initialization time, so then it's just a single boolean switch in the representation

jkterry1:
Closing in favor of PR so I can keep track easier

stefanbschneider:
Is there any way to enforce a more compact representation independent of the low/high values? Maybe even in versions 0.21 and before, where this MR was not yet merged?

PR #2554, opened 2022-01-11
Implement simplified Box representation

Markus28:
This pull request aims to simplify the string representation of Box spaces. 
If the lower (or analogously upper) limit is a multiple of the all-ones vector, we only print the multiplicand, instead of the entire vector. This was discussed in issue #2497.

The method `_short_repr` will return a possibly shortened string representation of an array, as outlined above. I will add a short comment to the method later.

Best,
Markus

### DIFF
--- a/gym/spaces/box.py
+++ b/gym/spaces/box.py
@@ -4,6 +4,12 @@ from .space import Space
 from gym import logger
 
 
+def _short_repr(arr):
+    if arr.size != 0 and np.min(arr) == np.max(arr):
+        return str(np.min(arr))
+    return str(arr)
+
+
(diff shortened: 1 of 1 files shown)


RedTachyon:
Since Box is de facto frozen, I'd suggest finding `low_repr` and `high_repr` to the initialization (so `self.low_repr = _short_repr(self.low)` etc.), and then using `self.low_repr` and `self.high_repr` to the actual `__repr__`

jkterry1:
@RedTachyon I believe the relevant changes you requested were made

RedTachyon:
@jkterry1 Yep, looks good now

Issue #2460, opened 2021-10-23
Remove reset enforcing aspect of time limit wrapper


jkterry1:
This is enforced for all environments in Gym now, so this is duplicate code

vwxyzjn:
I can take a closer look at this after 1/7 unless someone beats me to it, in which case I am happy to help review.

PR #2556, opened 2022-01-12
Remove reset enforceing in the `TimeLimit` wrapper

vwxyzjn:
Per #2460, the `OrderEnforcing` is already applied to all the envs, so we no longer need to enforce the order in the `TimeLimit` wrapper.


vwxyzjn:
Note that previously `OrderEnforcing` is not always applied. This PR always applies it so that you would get environments like 

```
<TimeLimit<OrderEnforcing<CartPoleEnv<CartPole-v1>>>>
```

RedTachyon:
Looks good, pretty straight-forward

PR #2565, opened 2022-01-19
Attempt to fix blackjack rendering tests

RedTachyon:
No idea if this will do it. We'll see.


RedTachyon:
image
But hey, tests are passing

It seems that setuptools didn't copy over the data files because they weren't explicitly listed. So I added them analogously to other files in `setup.py` and tests are passing (for now), so I'm happy

PR #2574, opened 2022-01-26
docs+credits

trigaten:
Update env docstrings


trigaten:
It should be noted that the docstrings added were written by the people accredited in the __credits__ strings.

PR #2575, opened 2022-01-26
replace mujoco-py with dm_control

jkterry1:


ikostrikov:
@jkterry1 I was considering adding a viewer. For example, there is one viewer for dm_control [here](https://github.com/google-research/robel/blob/5b0fd3704629931712c6e0f7268ace1c2154dc83/robel/simulation/dm_renderer.py#L37)

But I will have time to help with it only after this week.

PR #2583, opened 2022-01-29
Enable legacy warnings in case we screw up badly

jkterry1:
In case we do something very wrong in the future and need to issue warnings about old versions of Gym for reproducibility reasons or something, this will let us send tiny notices in import. Since its in a try except it can't cause errors for systems, and since it's pulling from a public git repo everything (including the immutable version history) is 100% public and auditable, so I can't do anything weird here.

Most of why I want this is that:

1) OpenAI doesn't give me very good PyPI release access
2) People aren't going to know to update to a x.x.1 version, or check their local machine version once installed.


trigaten:
Remove [gym/__init__.py](https://github.com/openai/gym/pull/2583/files#diff-4b2b225bea64f8d80bc46d9d586b22e692191bd0c669dc12cb073502f505ef4f) and merge.

PR #2585, opened 2022-01-29
Suppress PyGame import message

Markus28:
This pull request adds two lines to `__init__.py` to suppress a message that is printed whenever `pygame` is imported.
Exactly as in https://github.com/Farama-Foundation/PettingZoo/pull/577.

### DIFF
--- a/gym/__init__.py
+++ b/gym/__init__.py
@@ -13,5 +13,8 @@ from gym.envs import make, spec, register
 from gym import logger
 from gym import vector
 from gym import wrappers
+import os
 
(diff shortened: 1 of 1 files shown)


jkterry1:
@Markus28 did you manually check to make sure this actually works for things like the blackjack toytext environment?

Markus28:
Yes, I had checked it on Blackjack:

```
import gym
gym.make("Blackjack-v1")
```
produces no welcome message (opposed to the master branch).

PR #2587, opened 2022-01-29
Add version histories

trigaten:
Thanks to @kir0ul


gym/envs/box2d/lunar_lander.py
```diff
@@ -121,15 +121,18 @@ class LunarLander(gym.Env, EzPickle):
 
     ## Arguments
     To use to the _continuous_ environment, you need to specify the
-    `continuous"=True` argument like below:
+    `continuous"=True"` argument like below:
```
kir0ul, line 124: Looks like something is wrong here, I guess it should be:
```markdown
`continuous=True` argument like below:
```

trigaten: Thanks for the catch

Issue #2581, opened 2022-01-28
[Proposal] Make Env.np_random not optional


ikamensh:
### Proposal 

I'd like to initialise Env.np_random in it's `__init__` method. This way we don't need to deal with Optional (possibly None) type in the rest of the code. It's largely motivated by type checker compliance, but is also a good design.

### Motivation

Following code is present in many environments (specific one comes from CartPole, other classic_control envs have pretty much the same):
```python
def reset(self, *, seed: Optional[int] = None, options: Optional[dict] = None):
        super().reset(seed=seed)
        self.state = self.np_random.uniform(low=-0.05, high=0.05, size=(4,))  # "uniform" is not a known member of "None"
```

as an added comment on the last line highlights, it's not type compliant because `self.np_random` starts as None, and will be set later in `super().reset()`. Current code is not buggy, it's just that type checkers of today can't understand that type was changed deterministically in another method. 

I'd like instead to do complete initialisation in the `__init__` method. Furthermore, I'd suggest to rename the code living in the abstract `Env.reset` method to explicit `Env.set_seed` or `_set_seed`. 

While it so happens that you often want to set seed on the reset, the code can be more clearly labeled for what it does.

### Alternatives

There might be some reason why initialisation of Env.np_random must be delayed. Please let me know about it.

RedTachyon:
As mentioned in #2589, I'm fine with the solution there.

The reason not to initialize it in __init__ is because it's not always necessary, but the property approach should be fine, I think the cost of one extra `if` isn't that problematic.

PR #2589, opened 2022-02-01
Typecheck classic control environments

ikamensh:
Adds typechecks to classical control environments. Type checks are not comprehensive as typed signatures are not enforced, but should still be an improvement. This includes a possible solution to #2581. Part of #2452.

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -94,7 +94,7 @@ class Env(Generic[ObsType, ActType]):
         """
         # Initialize the RNG if the seed is manually passed
         if seed is not None:
-            self.np_random, seed = seeding.np_random(seed)
+            self._np_random, seed = seeding.np_random(seed)
 
     @abstractmethod
     def render(self, mode="human"):


gym/core.py
```diff
@@ -85,9 +92,9 @@ def reset(
         Returns:
             observation (object): the initial observation.
         """
-        # Initialize the RNG if it's the first reset, or if the seed is manually passed
-        if seed is not None or self.np_random is None:
-            self.np_random, seed = seeding.np_random(seed)
+        # Initialize the RNG if the seed is manually passed
+        if seed is not None:
+            self.seed(seed)
```
RedTachyon, line 97: What's the reason for this? `self.seed` is explicitly deprecated

gym/envs/classic_control/acrobot.py
```diff
@@ -171,6 +171,7 @@ def reset(self, *, seed: Optional[int] = None, options: Optional[dict] = None):
 
     def step(self, a):
         s = self.state
+        assert s is not None, "Call reset before using AcrobotEnv object."
```
RedTachyon, line 174: Isn't this enforced through a wrapper already? Or is this to make type checkers happy?

RedTachyon (commented):
One change seems to revert previous PRs (which came about after major and painful discussions), so should be undone afaik (the self.seed).

Other than that it mainly solves #2581 and adds only a few type annotations, but I think I'm fine with that overall.

gym/envs/classic_control/acrobot.py
ikamensh: yes, this is to make typechecker happy. It can't deduce the wrapper verification.

gym/core.py
ikamensh: with removal of uninitialized _np_random, this was pure code duplication, so I used self.seed(). Didn't know its deprecated.

PR #2590, opened 2022-02-01
Add typing to Wrapper signatures

ikamensh:
Work towards #2452 - adds typing to signatures of Wrapper methods, defines it's generic types.


RedTachyon:
Looks good

PR #2594, opened 2022-02-04
Add types to gym.make("CartPole-v1") and such

ikamensh:
This uses Literal to type all registered IDs of `gym.make` factory function.

Although `Literal` is not available in python3.7, it can bring much value to users of later python versions.
The overrides are done as close as possible to registration of aliases to ease future maintenance. This necessitated moving definion of `make` alias for `registry.make` to `__init__.py`. This is described in more detail in https://mypy.readthedocs.io/en/stable/literal_types.html ("overrides need to be followed by implementation when done outside of a stub file). 

## Rejected ideas
I've considered adding a stub file but this needs to duplicate all the typing we've already put in the source code, so this idea was rejected.

Instead of moving definition of make into dunder init, it might have worked to just import it after overrides, but this would break even more conventions.

## Roadmap

* There's already a PR for partially typing classical environments: #2589
* Add typing to other families of environments (Box2D?, Toy-Text, Mujoco?) - if some will be moved out of gym - let me know, I'll skip them
* We can make type checking more strict by e.g. requiring typing all signatures.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,7 +23,7 @@ repos:
     hooks:
       - id: flake8
         args:
-          - --ignore=E203,E402,E712,E722,E731,E741,F401,F403,F405,F524,F841,W503
+          - --ignore=E203,E402,E712,E722,E731,E741,F401,F403,F405,F524,F841,W503,E302,E704
           - --max-complexity=30
           - --max-line-length=456
           - --show-source
--- a/gym/envs/__init__.py
+++ b/gym/envs/__init__.py
@@ -1,26 +1,10 @@
-from __future__ import annotations
-
-from typing import Sequence, SupportsFloat, overload, Any
-import sys
-
-import numpy as np
-
-from gym.core import Env
 from gym.envs.registration import (
     registry,
     register,
+    make,
     spec,
     load_env_plugins as _load_env_plugins,
 )
-
-if sys.version_info >= (3, 8):
-    from typing import Literal
-else:
-
-    class Literal:
-        def __class_getitem__(cls, item):
-            return Any
-
 
 # Hook to load plugins from entry points
 _load_env_plugins()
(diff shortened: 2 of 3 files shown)


RedTachyon:
I don't like this tbh. I'm not actually sure if it is likely to do anything in the foreseeable future (type checkers aren't likely to check the exact value of the string in order to infer the type). And if it does, this would either break or mess up creating custom environments. This seems to effectively block registering custom environments (or rather, you'd have to overload the gym function in order to do that), and the benefit is pretty hypothetical and miniscule.

As much as I'm a fan of strict typing, I think this goes too far. The type of `gym.make` first argument is a string, which might be valid (and then the env is created) or invalid (=> exception), and I think that's the extent of type checking we want. 

tl;dr unless I get convinced otherwise, I'm moderately strongly against this.

ikamensh:
> I'm not actually sure if it is likely to do anything in the foreseeable future (type checkers aren't likely to check the exact value of the string in order to infer the type).

They do actually. I've verified this helps mypy and pyright infer exact type of the environment, just like if you would have constructed specific env via it's class, env = `CartPoleEnv(...)`. 

Drawbacks I agree exist:
1) It's unclear if envs outside of gym can expose their type the same way, now I can't think of how to do it.
2) It's verbose and would need a bit of love to stay in sync with any future changes to env versions. But given that most research uses the same old versions to be comparable to old results, maybe thats not that much effort after all.

> And if it does, this would either break or mess up creating custom environments. This seems to effectively block registering custom environments

I didn't understand this part, how does it mess up creating custom environments / block registering them? They will all default to the generic `make(x: str, **kwargs) -> Env` case, so they would have the same situation as right now, in no way worse. It's only a bonus for default gym environments being precisely known to type checkers. Or you've run some experiments / demo code that says otherwise?

RedTachyon:
> They will all default to the generic make(x: str, **kwargs) -> Env case

That's true, my bad. I'm a bit more on the fence now.

I think my main concern at this point is about just the design philosophy. The idea of gym is that we should have a common API, so `make` guarantees that we'll get an Env. I can see some very niche cases where this extra type information can be helpful, but it's also possible to just get those benefits by explicitly hinting the type in downstream code.

-------------------
Actually, scratch all of that, I wrote it and then realized something big -- is this actually *technically* correct in the first place? `gym.make` will, depending on some parameters, wrap the environment in a `TimeLimit` and/or `OrderEnforce` (or neither). 

Consider the following code snippet:
```python
import gym
from gym.envs.classic_control import CartPoleEnv

env = gym.make("CartPole-v1")  # According to the new type hint, it's a CartPoleEnv
print(isinstance(env, CartPoleEnv)) 
# False
```

ikamensh:
you are right, they return a wrapper, so it's currently not technically correct.

If we want precise typing: maybe Wrapper should be generic over the non-wrapper it is wrapping, i.e. `Wrapper[MaybeWrapped[E]]` where MaybeWrapped is either another Wrapper or the underlying non-wrapper type `E`. Or maybe we can hint at very specific wrapping sequence.

In fact I probably don't want to expose any of the wrapper details. What I do want to expose is the ObsType and ActType, as this is what user should pass into the environment.  So a simple solution would be to put more specific Env types in the return, like `Env[np.ndarray, np.ndarray | int]` for CartPole (or whatever the right ObsType and ActType is). How about that?

ikamensh:
I was thinking of being (this time) technically correct, meaning specifying that CartPole actually accepts int. Otherwise type checker will complain about working code, meaning people might prefer not to use type checker. Do you want to use type checker to "deprecate" using non-numpy actions?

After manually testing what inputs work, I got to:
```python
def make(id: Literal["CartPole-v0", "CartPole-v1"], **kwargs) -> Env[np.ndarray, int | np.ndarray]: ...
def make(id: Literal["MountainCar-v0"], **kwargs) -> Env[np.ndarray, int | np.ndarray]: ...
def make(id: Literal["MountainCarContinuous-v0"], **kwargs) -> Env[np.ndarray, list | tuple | np.ndarray]: ...
def make(id: Literal["Pendulum-v1"], **kwargs) -> Env[np.ndarray, list | tuple | np.ndarray]: ...
def make(id: Literal["Acrobot-v1"], **kwargs) -> Env[np.ndarray, int | np.ndarray]: ...
```

The observations are always produced by the environment so hopefully user has no way to mess up with them, meaning it's always `np.ndarray`

RedTachyon:
> Do you want to use type checker to "deprecate" using non-numpy actions?
Not what I meant, just that we get more specific type hints when using common envs. Although I could be convinced to deprecate using int's, at least via a warning or something. But might not be worth it, and definitely not a thing for this PR

In any case, I think this would be alright as long as it doesn't change the behavior, which it shouldn't.

jkterry1:
I haven't read this in detail, but a ton of new environments in gym won't be returning numpy arrays anymore with the hardware accelerated environments

RedTachyon:
So then these annotations will need to be updated as the envs get updated, not a big problem (and then this might actually become pretty useful)

Is the plan to use jax instead? I'd be very careful about moving away from numpy, even with accelerated envs it often makes sense to use it.

ikamensh:
I've updated the PR to use `Env` class but with specific obs/action types per our discussion.

RedTachyon:
So my one last doubt that I just noticed is whether anyone ever does `gym.registration.make(...)`. I'd assume no, but many such assumptions that I have made were already disproven. Which makes this technically a [breaking change](https://xkcd.com/1172/), which I think is better to avoid if there's no direct benefit. 

I think the easiest way around it copy-pasting all of the type hints into `envs/registration.py` and keeping the import structure as is.

PR #2596, opened 2022-02-06
Improvements to Reacher Documentation

Markus28:
This pull request aims to make some improvements to the documentation of Reacher. I have made the following changes:

### General
- Corrected some typos and some grammar. I'm not a native speaker and didn't want to introduce new mistakes, so I didn't make all the changes I believe would be necessary. Someone else might want to check this.

### Description
- Description was missing. I provided a brief description

### Action Space
- The first paragraph was quite confusing because it talked about an agent and denoted actions as `(action, action)`, implying that actions are only permitted to be on the main diagonal of R<sup>2</sup>.

### Observation Space
- The first paragraph wasn't accurate since observations also include the position of the target and the vector between the target and the reacher's fingertip 
- The second table of this section only listed cosines of the angles. This appears to be a copy-paste mistake, and the rows corresponding to Num 2,3 probably should be "sine".
- The third table of this section wasn't rendered correctly in HTML, seemingly because there is a missing line break
- Some closing parentheses were missing in the third table

### Rewards
- "Walker" was mentioned in the description of *reward_control*. Moreover, a weight for *reward_control* was mentioned. While such a weight exists for Walker, I don't believe it exists for Reacher. @RushivArora  correct me if I'm wrong.

### Starting State
- It was stated that the target is selected uniformly at random in [-0.2, 0.2]. However, it is chosen uniformly at random in a disk of radius 0.2 around the origin.
- "and the last index (“fingertip” - “target” distance)" -> "and the last element (“fingertip” - “target”)"
- The calculation *dt = 4 * 0.01 = 0.02* is obviously wrong. *dt=0.2* seems to be correct and I believe that the *4* should be a *2* (for frameskip) so I replaced it.

### Versions
- I don't believe there is a v3 of Reacher, so I removed *v3. The information for v1 is quite confusing since it is supposed to pertain to Reacher but explicitly excludes Reacher. I was not sure whether to remove the sentence "max_time_steps raised to 1000 for robot based tasks (not including reacher, which has a max_time_steps of 50)." and I left it in for now. I'm not sure how accurate the other version information is.

### DIFF
--- a/gym/envs/mujoco/reacher.py
+++ b/gym/envs/mujoco/reacher.py
@@ -10,7 +10,7 @@ class ReacherEnv(mujoco_env.MujocoEnv, utils.EzPickle):
     target that is spawned at a random position.
 
     ### Action Space
-    The action space is [-1, 1]<sup>2</sup>. An action `(a, b)` represents the torques applied at the hinge joints.
+    The action space is a `Box(-1, 1, (2,), float32)`. An action `(a, b)` represents the torques applied at the hinge joints.
 
     | Num | Action                                                                          | Control Min | Control Max | Name (in corresponding XML file) | Joint | Unit |
     |-----|---------------------------------------------------------------------------------|-------------|-------------|--------------------------|-------|------|
@@ -70,6 +70,11 @@ class ReacherEnv(mujoco_env.MujocoEnv, utils.EzPickle):
     Euclidean norm of the action, i.e. as *- sum(action<sup>2</sup>)*.
 
     The total reward returned is ***reward*** *=* *reward_distance + reward_control*
+
+    Unlike other environments, Reacher does not allow you to specify weights for the individual reward terms.
+    However, `info` does contain the keys *reward_dist* and *reward_ctrl*. Thus, if you'd like to weight the terms,
+    you should create a wrapper that computes the weighted reward from `info`.
+
 
     ### Starting State
     All observations start in state
--- a/gym/envs/mujoco/reacher.py
+++ b/gym/envs/mujoco/reacher.py
@@ -6,12 +6,11 @@ from gym.envs.mujoco import mujoco_env
 class ReacherEnv(mujoco_env.MujocoEnv, utils.EzPickle):
     """
     ### Description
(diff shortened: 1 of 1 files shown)


RushivArora:
Hey @Markus28 thanks for taking the initiative and working on typos :) ! Some general comments/discussions before we can merge it:

### Description

I intentionally did not write a description since I was delaying it. I couldn't find the original paper for reacher and it is important to give kudos to the original authors for it. Maybe you can find it? Other than that, I like what you wrote for the description

### Action Space

I kept the action space to be similar to the classical environments. Could you explain a little in detail to me how it was misleading, since I didn't fully get it? I'd say mentioning it as ``` [-1,1]<sup>2</sup> ``` could indicate a quadratic distribution over the action space when selecting random actions - which is not the case. We should consider reverting to the original for this one.

### Observation space

Ah these typos missed my eye in all the many gym docs - thanks for fixing it!

### Rewards

There is no such weight for reacher, but there is for other environments. I'd like to point out that that is an important difference (I do note right now that this is a fixed *static* value and is set to 1). This is necessary for users to realise that the coefficient in here is a higher value than other environments, and that it is fixed. Could you consider adding that line back?

### Starting State

The code for selecting the goal selects the ```(x,y)``` position of the target in uniformly random range ```[-0.2, 0.2]```. My understanding is that selecting two variables in a uniformly random range leads to a square. Could you elaborate on the disk? Maybe I missed something in the docs here?

### Version History

So the reason for phrasing v1 as such is: reacher *is* a robot environment and saying that raising the max_timesteps to 1000 for all robot environments to 1000 will also mean it was raised for reacher - which is not the case as reacher lasts 50 steps. Hence the "all environments have the 1000 time steps, except reacher which has 50". Looking at old PRs from years ago  one can see the max_timesteps being much lower (I believe it was 300).

The other version info was copied verbatim from the release notes, so I'd hope it is accurate!

Can you make the changes? I think then it is good to merge after that (particular emphasis on finding the paper, reverting action spaces, adding the coefficient lines back, and discussing the start state of the target)

jkterry1:
I'm going to merge this for expediency, please create another PR if it's determined that more changed are needed

PR #2598, opened 2022-02-07
Refactor Car Racing to use pygame

andrewtanJS:
This PR refactors the rendering function of the two environments to use pygame instead of pyglet. The refactored rendering is consistent with the previous pyglet rendering.

Here is a side by side comparison of the two environments. Left is the old pyglet version, and right is the new pygame version. 
image
image

### DIFF
--- a/gym/envs/box2d/car_dynamics.py
+++ b/gym/envs/box2d/car_dynamics.py
@@ -9,6 +9,7 @@ Created by Oleg Klimov
 
 import numpy as np
 import math
+import pygame.draw
 import Box2D
 from Box2D.b2 import (
     edgeShape,
@@ -42,9 +43,12 @@ HULL_POLY3 = [
     (-25, +20),
 ]
 HULL_POLY4 = [(-50, -120), (+50, -120), (+50, -90), (-50, -90)]
-WHEEL_COLOR = (0.0, 0.0, 0.0)
-WHEEL_WHITE = (0.3, 0.3, 0.3)
-MUD_COLOR = (0.4, 0.4, 0.0)
+WHEEL_COLOR = (0, 0, 0)
+WHEEL_WHITE = (77, 77, 77)
+MUD_COLOR = (102, 102, 0)
+
+SCALE = 6.0  # Track scale
+PLAYFIELD = 2000 / SCALE  # Game over boundary
 
 
 class Car:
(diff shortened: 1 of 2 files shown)


gym/envs/box2d/car_racing.py
```diff
@@ -438,235 +436,209 @@ def step(self, action):
...
-            -PLAYFIELD,
-            +PLAYFIELD,
-            0,
+        if mode == "rgb_array":
+            scaled_screen = pygame.transform.smoothscale(
+                self.screen, (VIDEO_W, VIDEO_H)
+            )
+            return np.transpose(
+                np.array(pygame.surfarray.pixels3d(scaled_screen)), axes=(1, 0, 2)
+            )
```
carlosluis, line 482: Could be worth it to offload this to a separate method:

```python
def _create_image_array(screen):
    return np.transpose(
                np.array(pygame.surfarray.pixels3d(screen)), axes=(1, 0, 2)
            )
```
and use it here in the two cases.

gym/envs/box2d/car_racing.py
```diff
@@ -438,235 +436,209 @@ def step(self, action):
...
                     ]
                 )
+        for poly in grass:
+            poly = [pygame.math.Vector2(c).rotate_rad(angle) for c in poly]
+            poly = [
+                (c[0] * zoom + translation[0], c[1] * zoom + translation[1])
+                for c in poly
+            ]
+            gfxdraw.aapolygon(self.surf, poly, (102, 230, 102))
+            gfxdraw.filled_polygon(self.surf, poly, (102, 230, 102))
```
carlosluis, line 528: These lines are pretty much duplicated above. Could be refactored as, e.g.:

```python
def draw_colored_polygon(poly, color, zoom, translation):
    poly = [pygame.math.Vector2(c).rotate_rad(angle) for c in poly]
    poly = [
                (c[0] * zoom + translation[0], c[1] * zoom + translation[1])
                for c in poly
            ]
    gfxdraw.aapolygon(self.surf, poly, color)
    gfxdraw.filled_polygon(self.surf, poly, color)

File: gym/envs/box2d/car_racing.py [hunk 3]
```diff
@@ -438,235 +436,209 @@ def step(self, action):
...
-        for x in range(-20, 20, 2):
-            for y in range(-20, 20, 2):
-                polygons_.extend(
+        trans_field = []
+        for coord in field:
+            coord = pygame.math.Vector2(coord).rotate_rad(angle)
+            coord = (coord[0] * zoom + translation[0], coord[1] * zoom + translation[1])
+            trans_field.append(coord)
+        gfxdraw.aapolygon(self.surf, trans_field, (102, 204, 102))
+        gfxdraw.filled_polygon(self.surf, trans_field, (102, 204, 102))
```
carlosluis, line 507: See comment below about refactoring. This might also benefit from it, since I see there's some duplication.

gym/envs/box2d/car_racing.py
```diff
@@ -438,235 +436,209 @@ def step(self, action):
...
+            # converting to pixel coordinates
+            poly = [(p[0] + PLAYFIELD, p[1] + PLAYFIELD) for p in poly]
+            poly = [pygame.math.Vector2(c).rotate_rad(angle) for c in poly]
+            poly = [
+                (c[0] * zoom + translation[0], c[1] * zoom + translation[1])
+                for c in poly
+            ]
+            color = [int(c * 255) for c in color]
+            gfxdraw.aapolygon(self.surf, poly, color)
+            gfxdraw.filled_polygon(self.surf, poly, color)
```
carlosluis, line 540: With the refactoring this would become:
```python
poly = [(p[0] + PLAYFIELD, p[1] + PLAYFIELD) for p in poly]
color = [int(c * 255) for c in color]
draw_colored_polygon(poly, color, zoom, translation)
```

gym/envs/box2d/car_racing.py
```diff
@@ -438,235 +436,209 @@ def step(self, action):
...
-        horiz_ind(30, -0.8 * self.car.hull.angularVelocity, (1, 0, 0))
-        vl = pyglet.graphics.vertex_list(
-            len(polygons) // 3, ("v3f", polygons), ("c4f", colors)
-        )  # gl.GL_QUADS,
-        vl.draw(gl.GL_QUADS)
-        vl.delete()
-        self.score_label.text = "%04i" % self.reward
-        self.score_label.draw()
+        # simple wrapper to render if the indicator value is above a threshold
+        def render_if_min(value, points, color):
```
carlosluis, line 571: Does this method change the behaviour of the indicators with respect to the pyglet version? I'm just wondering why this was not part of the rendering before (at least from what I can tell, this wrapping was not there before, was it?).

gym/envs/box2d/car_racing.py
```diff
@@ -438,235 +436,209 @@ def step(self, action):
...
+            self.isopen = False
 
+
+if __name__ == "__main__":
     a = np.array([0.0, 0.0, 0.0])
 
-    def key_press(k, mod):
-        global restart
-        if k == 0xFF0D:
-            restart = True
```
carlosluis, line 646: This case is not covered in your change. To preserve the functionality, a human needs to be able to restart the game by pressing ENTER.

carlosluis (changes requested):
Hey @andrewtanJS ! You did a pretty good job in replicating the rendering.

I left some small comments suggesting a few refactors and some other small functionalities that might have been removed during the port. Happy to iterate with you on this to get it merged.

gym/envs/box2d/car_dynamics.py
```diff
@@ -260,7 +264,82 @@ def step(self, dt):
...
+                white_poly = [
+                    (
+                        coords[0] * zoom + translation[0],
+                        coords[1] * zoom + translation[1],
+                    )
+                    for coords in white_poly
+                ]
+                pygame.draw.polygon(surface, color=WHEEL_WHITE, points=white_poly)
+
+    def draw1(self, viewer, draw_particles=True):
```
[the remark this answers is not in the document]
andrewtanJS: Ah yes, thanks for the catch!

gym/envs/box2d/car_racing.py
```diff
@@ -438,235 +436,209 @@ def step(self, action):
...
         angle = -self.car.hull.angle
-        self.transform.set_scale(zoom, zoom)
-        self.transform.set_translation(
-            WINDOW_W / 2
-            - (scroll_x * zoom * math.cos(angle) - scroll_y * zoom * math.sin(angle)),
-            WINDOW_H / 4
-            - (scroll_x * zoom * math.sin(angle) + scroll_y * zoom * math.cos(angle)),
-        )
-        self.transform.set_rotation(angle)
+        zoom = 0.1 * SCALE * max(1 - self.t, 0) + ZOOM * SCALE * min(self.t, 1)
```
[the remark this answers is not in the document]
andrewtanJS: Got it, it's been reinserted

gym/envs/box2d/car_racing.py
andrewtanJS: I added the an additional parameter which takes in the dimensions needed and scaled the screen accordingly. It's been refactored

gym/envs/box2d/car_racing.py
andrewtanJS: Yeah, this is not present in the previous rendering. One thing I noticed was that even with a totally stationary car, the indicators would still be rendered with an arbitrarily small width. This wrapper resolves it by only rendering if the value of the indicator is above a certain threshold.

carlosluis (approved):

carlosluis:
Thanks, LGTM

Issue #2600, opened 2022-02-07
[Bug Report] FrozenLake RGB rendering returns None


trigaten:
FrozenLake `render()` method in mode human always returns `None` and lacks an `rgb_array` render mode. For reference, Blackjack has recently been converted to pygame and implements these features correctly.

@younik I believe you converted FrozenLake to pygame?

younik:
@trigaten if you pass `mode='ansi'` it returns the old render mode; otherwise, it renders in GUI mode.

What should be the expected behaviour with `rgb_array`?

trigaten:
> @trigaten if you pass `mode='ansi'` it returns the old render mode; otherwise, it renders in GUI mode.
> 
> What should be the expected behaviour with `rgb_array`?

As in Blackjack, an rgb array should be returned.

younik:
> As in Blackjack, an rgb array should be returned.

Clear, thank you, I fix it now!

trigaten:
@younik There is still no RGB render mode.

PR #2604, opened 2022-02-08
Add rgb_array support for frozen lake env

younik:
From #2600, add rgb_array support for frozen lake env


trigaten:
Tested locally and works for me.

PR #2612, opened 2022-02-13
Fix `return_info` for wrappers

RedTachyon:
Included in the commit right now:
- ObservationWrapper
- AtariPreprocessing
- FrameStack (which pretends to be an ObservationWrapper, but that's a lie)

Most wrappers actually "just work" because `return self.env.reset(**kwargs)` will just handle everything. 

One weird case is the Monitor wrapper, because it passes the reset observation to a stats recorder, which then ignores that observation. Fun. But it should work as-is

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -300,8 +300,11 @@ class Wrapper(Env[ObsType, ActType]):
 
 class ObservationWrapper(Wrapper):
     def reset(self, **kwargs):
-        observation = self.env.reset(**kwargs)
-        return self.observation(observation)
+        if kwargs.get("return_info", False):
+            obs, info = self.env.reset(**kwargs)
+            return self.observation(obs), info
+        else:
+            return self.observation(self.env.reset(**kwargs))
 
     def step(self, action):
         observation, reward, done, info = self.env.step(action)
--- a/gym/wrappers/atari_preprocessing.py
+++ b/gym/wrappers/atari_preprocessing.py
@@ -133,6 +133,8 @@ class AtariPreprocessing(gym.Wrapper):
         # NoopReset
         if kwargs.get("return_info", False):
             _, info = self.env.reset(**kwargs)
+        else:
+            info = None
 
         noops = (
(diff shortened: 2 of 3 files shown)


Markus28:
- I think you are still missing `monitor.py`
- In the new API it should probably be possible to pass `info` to `StatsRecorder.after_reset`. But maybe this should be covered in a new PR 
- In `record_episode_statistics.py`: Maybe rename `observation` to something that implies that it might actually be `(observation, info)`.
- Same for `record_video.py`

Markus28:
Sorry, I missed that comment.
One other question: With your changes to `AtariPreprocessing` it seems that `reset` will return the `info` that was given by the initial reset of the Atari environment. 
Wouldn't it make more sense if it returned the `info` that was returned by the last NOOP step (or alternatively the last reset, if that is more recent than the last NOOP step)? Otherwise, `info` might be out of date.

I.e. replace `_, _, done, _` here:

https://github.com/openai/gym/blob/c6b722156dc659ffe7b509e9cd2465a7629946dd/gym/wrappers/atari_preprocessing.py#L145

by `_, _, done, info` and also get `observation, info` in this line:
https://github.com/openai/gym/blob/c6b722156dc659ffe7b509e9cd2465a7629946dd/gym/wrappers/atari_preprocessing.py#L147

jkterry1:
Can this please be merged into this PR if you haven't already? https://github.com/openai/gym/pull/2454 ?

RedTachyon:
@Markus28 That's a good point and franky, I have no clue what's the best way to handle it, my experience with Atari envs is zero. It'd also depend heavily on what the Atari people pass in their info dicts. Intuitively, I think passing the info from `env.reset()` makes more sense because it has a unique position in a -way (i.e. it doesn't necessarily have the same structure as the rest of infos coming from steps) 
@JesseFarebro any thoughts? (also maybe it'd be better to move the Atari Preprocessing wrapper to ALE?)

@jkterry1 Do you mean to just copy the changes from #2454 into here?

PR #2613, opened 2022-02-13
Add a compatibility wrapper for the new API

RedTachyon:
Not sure if we want this in right now since everything was designed to be backwards-compatible, but it might be good to have it and update it as time goes on. This should take an environment written for the old API (notably, no `reset` arguments), and convert it into a good, law-abiding citizen under the new order.


jkterry1:
Closing for now as there's properly breaking changes right now.

PR #2614, opened 2022-02-14
Add wrapper for making one hot observations

vwxyzjn:
For envs like `Taxi-v2`, their observation spaces are `Discrete` spaces that cannot be directly used in neural networks. As a result, RL libraries usually manually one-hot encode the environment such as in `openai/baselines` ([baselines/common/input.py#L43-L55](https://github.com/openai/baselines/blob/ea25b9e8b234e6ee1bca43083f8f3cf974143998/baselines/common/input.py#L43-L55))

This wrapper provides a convenient way to one hot encode discrete observation space:

```python
    env = gym.make('Taxi-v3')
    env = gym.wrappers.OnehotObservation(env)
    env.reset()
    assert isinstance(env.observation_space, gym.spaces.Box)
    assert env.observation_space.shape == (500,)
```


gym/wrappers/onehot_observation.py
```diff
@@ -0,0 +1,42 @@
...
+        Box(0, 1, (500,))
+
+    Args:
+        env (Env): environment
+    """
+
+    def __init__(self, env):
+        super().__init__(env)
+        self.n = None
+        if isinstance(self.env.observation_space, gym.spaces.Discrete):
```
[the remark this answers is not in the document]
vwxyzjn: This is a good point. Do you think it should be a `ValueError` or a warning?

gym/wrappers/onehot_observation.py
```diff
@@ -0,0 +1,42 @@
+import gym
+import numpy as np
+
+
+def one_hot(a, size):
+    b = np.zeros((size))
```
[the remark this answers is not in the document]
vwxyzjn: I have addressed this.

gym/wrappers/onehot_observation.py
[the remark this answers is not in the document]
vwxyzjn: I have made it a `ValueError`. This should be explicit to the users.

gym/wrappers/onehot_observation.py
```diff
@@ -0,0 +1,40 @@
...
+                "This wrapper can only apply to the Discrete observation space"
+            )
+        self.n = self.env.observation_space.n
+        self.observation_space = gym.spaces.Box(0, 1, (self.n,))
+        self.onehot_encoding = np.zeros(self.n, dtype=np.float32)
+
+    def observation(self, obs):
+        self.onehot_encoding[:] = 0
+        self.onehot_encoding[np.array(obs)] = 1
+        return obs
```
RedTachyon, line 40: Uh
This is one of the cases where I really don't want to sound mean because I make this mistake all the time. But I also feel a lot of... snarkiness coming on because I make this mistake *all the time*

`return self.onehot_encoding`

tests/wrappers/test_onehot_observation.py
```diff
@@ -0,0 +1,9 @@
+import gym
+
+
+def test_onehot_observation():
+    env = gym.make("Taxi-v3")
+    env = gym.wrappers.OnehotObservation(env)
+    env.reset()
```
RedTachyon, line 7: And relatedly to the previous comment, this could use some asserts on the actual `obs` from `obs = env.reset()`

RedTachyon (changes requested):
So there's one critical bug, simple to fix.

There might be another bug which I'm testing now, I'll report back in a few minutes

RedTachyon:
Ok, second bug is a bit more subtle and would cause problems sometimes, but these problems would be a major pain to find. Source: I once debugged a very similar bug

tl;dr don't reuse and return the same buffer array in the wrapper. Sure, it probably saves a bit of memory, but leads to stuff like this:

```python
all_obs = []
obs = env.reset()
all_obs.append(obs)

print(obs.argmax())
### Out: 332

obs, _, _, _ = env.step(env.action_space.sample())
all_obs.append(obs)

print(obs.argmax())
### Out: 432

print([o.argmax() for o in all_obs])
### Out: [432, 432]
```

Explanation: each time the environment actually returns the same exact array as the output, it just happens that its contents change over time. So if you accumulate the observations into a list (very common workflow), you'll end up with multiple references to the same array, with the value being the final value. 

Solution: either return a copy of the array, or just reinitialize the array each time. I don't think performance will hurt much either way.

RedTachyon:
Also a minor nitpick - intuitively I was writing `OneHotObservation` (capital H), which I think makes more sense with consistent camelcase, imo it should be renamed to that

gym/wrappers/onehot_observation.py
vwxyzjn: Ahh thanks for the check. I have also fixed the related test case

gym/wrappers/one_hot_observation.py
```diff
@@ -0,0 +1,43 @@
...
+        >>> env.observation_space
+        Box(0, 1, (500,))
+
+    Args:
+        env (Env): environment
+    """
+
+    def __init__(self, env: gym.Env):
+        super().__init__(env)
+        self.n: int = None
```
RedTachyon, line 34: I don't think this actually does anything? It will either be overwritten a few lines down, or there will be an exception

RedTachyon:
Overall looks good to me now, just one minor nitpick which doesn't really matter much either way

vwxyzjn:
That's a good pick! I have removed that unnecessary line. @jkterry1 I think this is ready to be merged.

tristandeleu:
It looks like you could leverage the [`flatten`](https://github.com/openai/gym/blob/master/gym/spaces/utils.py#L59) function (together with [`flatten_space`](https://github.com/openai/gym/blob/master/gym/spaces/utils.py#L161) for the `observation_space`) to handle more cases than just `Discrete` observation space (eg. nested spaces with `Dict` of `Discrete`, or `MultiDiscrete`).

PR #2615, opened 2022-02-14
Add a T-maze problem for testing out LSTM

vwxyzjn:
This PR implements the "Long-term dependency T-maze" from [Long-term dependency T-maze](https://proceedings.neurips.cc/paper/2001/file/a38b16173474ba8b1a95bcbc30d3b8a5-Paper.pdf). Such an environment can be used to test if the agent can memorize things, so it would be a useful addition to gym for downstream RL libraries to test out LSTM.

### DIFF
--- a/gym/envs/toy_text/memorytmaze.py
+++ b/gym/envs/toy_text/memorytmaze.py
@@ -1,0 +1,118 @@
+from typing import Optional
+import gym
+import numpy as np
+import pygame
+from pygame.constants import SRCALPHA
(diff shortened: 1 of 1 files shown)


jkterry1:
Hey, I'm closing this per a discord discussion I had with Costa. That being said, if you release this and can get all the major open source RL libraries to use it for LSTM testing, then we can merge this in. I just don't wan't more completely unused toytext environments in Gym

Issue #2627, opened 2022-02-18
[Proposal] fix linebreaks on gymlibrary.ml


Arcify:
After having changed the documentation in the frozen lake environment, I noticed that the linebreaks do not show up correctly at [https://www.gymlibrary.ml/pages/environments/toy_text/frozen_lake](url) because I did not know that the documentation was directly rendered onto the website

- at the bottom the version history is all on the same line.
- the explanation about the observation states can also need some linebreaks or periods.
- maybe also add an extra header for both the action space and the observation space

trigaten:
- [x] version history fixed
- [x] periods added
- [x] headers added 
Fixed here: https://github.com/openai/gym/pull/2628

PR #2628, opened 2022-02-18
Fix Frozen Lake docstring

trigaten:


trigaten:
Fixes https://github.com/openai/gym/issues/2627

PR #2630, opened 2022-02-19
Box Boundedness determined using get_inf instead of np.inf

### LINKED ISSUE #2629: [Bug Report] Box bounded below and above doesn't work when dtype is int
**Describe the bug**
A `Box` space that is created with dtype `int` and `low` (`high`) values being `-np.inf` (`np.inf`) will return `bounded_{above,below}` as True.

**Code example**
```python3
from gym.spaces import Box
import numpy as np

unbounded_below = Box(
    np.array([0, 13, -3, -np.inf]),
    np.array([0, 20, 0, 0]),
    dtype=int
)

unbounded_above = Box(
    np.array([0, 12, 20, 0]),
    np.array([np.inf, 20, 24, np.inf]),
    dtype=int
)

print(unbounded_above.bounded_above)
print(unbounded_below.bounded_below)

>>> [ True  True  True  True]
>>> [ True  True  True  True]
```

**My analysis**
At the [broadcast](https://github.com/openai/gym/blob/master/gym/spaces/box.py#L64), [get_inf](https://github.com/openai/gym/blob/master/gym/spaces/box.py#L226) turns `np.inf` into their integer equivalents. Determining [boundedness](https://github.com/openai/gym/blob/master/gym/spaces/box.py#L86) is comparing directly with `np.inf`, not with the integer equivalents given from `get_inf`.

**System Info**
 * Gym installed with pip
 * MacOS 10.15.7
 * Python version: 3.7.9
 * numpy 1.21.5

**Additional context**
I know it may seem strange to use `np.inf` with an integer-Box, but I use integer-Boxes all the time, and I need to check against `np.inf`.

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/openai/gym/issues) in the repo (**required**)

rusu24edward:
Resolves #2629

### DIFF
--- a/gym/spaces/box.py
+++ b/gym/spaces/box.py
@@ -61,8 +61,13 @@ class Box(Space[np.ndarray]):
             )
         assert isinstance(shape, tuple)
 
-        _low, _high = low, high # Store originals for determining boundedness
-        low = _broadcast(low, dtype, shape, inf_sign="-")  # type: ignore
+        # Capture the boundedness information before replacing np.inf with get_inf
+        _low = np.full(shape, low, dtype=dtype) if np.isscalar(low) else low
+        self.bounded_below = -np.inf < _low
+        _high = np.full(shape, high, dtype=dtype) if np.isscalar(high) else high
+        self.bounded_above = np.inf > _high
+
+        low = _broadcast(low, dtype, shape, inf_sign="-")
         high = _broadcast(high, dtype, shape, inf_sign="+")
 
         assert isinstance(low, np.ndarray)
@@ -84,8 +89,8 @@ class Box(Space[np.ndarray]):
         self.high_repr = _short_repr(self.high)
 
         # Boolean arrays which indicate the interval type for each coordinate
-        self.bounded_below = -np.inf < _low
-        self.bounded_above = np.inf > _high
+        # self.bounded_below = -np.inf < self.low
+        # self.bounded_above = np.inf > self.high
 
         super().__init__(self.shape, self.dtype, seed)
 
--- a/gym/spaces/box.py
+++ b/gym/spaces/box.py
@@ -62,9 +62,9 @@ class Box(Space[np.ndarray]):
         assert isinstance(shape, tuple)
 
         # Capture the boundedness information before replacing np.inf with get_inf
-        _low = np.full(shape, low, dtype=dtype) if np.isscalar(low) else low
+        _low = np.full(shape, low, dtype=float) if np.isscalar(low) else low
         self.bounded_below = -np.inf < _low
-        _high = np.full(shape, high, dtype=dtype) if np.isscalar(high) else high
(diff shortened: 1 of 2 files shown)


rusu24edward:
@RedTachyon I made some adjustments in order to pass the test. In this design, unbounded space is distinguished by the presence of `np.inf` rather than the `min/max` value in integer types. This means a bit of extra processing to determine the boundedness, but this only happens once during the initialization.

This changes the boundedness behavior of integer-type spaces to behave just like float spaces. This required changing a test.

RedTachyon:
Can you explain the change to the tests? Overall it should be good by now, but it seems that some of the test asserts were removed, which I'm not super happy about. The current tests should still be passing, and then you should add an extra test for this use case

jkterry1:
@rusu24edward lint tests are still failing

rusu24edward:
@RedTachyon The diff doesn't show it quite right. [Hiding the whitespace](https://github.com/openai/gym/pull/2630/files?diff=split&w=1) is a little nicer. In the test, there is an `if block` for capturing the float type, and an `elif` for capturing the integer type. These cases are handled differently. In this PR, I remove the `if block` but keep all the same tests as were in the float case, so now the "boundedness" of integer type is treated the same as the float type.

rusu24edward:
@jkterry1 I believe I have fixed the issue with linting, can you please re run the workflows for this PR?

jkterry1:
@rusu24edward just following up on this again, we'll need to include this fix in the 0.22.1 release thats going to go out soon

rusu24edward:
@jkterry1 I could use a bit of guidance on the failing test. It looks like the 'black' test fails in the pre-commit workflow. But I don't know what that means...
Screen Shot 2022-02-27 at 8 11 11 PM

kir0ul:
> @jkterry1 I could use a bit of guidance on the failing test. It looks like the 'black' test fails in the pre-commit workflow. But I don't know what that means...

Hey @rusu24edward, it means [the `Black` autoformatter tool](https://github.com/psf/black) would have reformatted your code, so your changes don't comply with the style of the existing codebase. You need to run Black locally to make it compliant.
But a better approach would be to install `pre-commit` so that it'll run each pipeline (not only Black) automatically before each commit. You can follow the following steps to install it:
1. Install `pre-commit` on your local machine: https://pre-commit.com/#install
2. Run `pre-commit install` to install the Git hooks in your local Gym repository
3. Run `pre-commit run --all-files` which will run Black and other hooks
4. Commit & push the changes

Now you should have exactly the same checks on your local machine as in CI.

jkterry1:
If you can't figure it out, once @RedTachyon finally approves this I'll fix it myself

rusu24edward:
@kir0ul Thank you for the detailed instructions, and @jkterry1 for offering to help. I believe I worked it out and just pushed up the changes. Please rerun the workflows :)

RedTachyon:
Looks good now after CI passes

PR #2637, opened 2022-02-21
gym.spaces.Tuple inherits from collections.abc.Sequence

duburcqa:
Following the PR I did a few months back (https://github.com/openai/gym/pull/2446), the tuple wrapper of gym should inherits from the abstract interface of Python. It is important for type check via `isinstance` and enable using such objects transparently with libraries such as [dmtree](https://github.com/deepmind/tree). 

It will bring a way helper methods along the way but it cannot be avoided to interoperability: : `__iter__`, `__reversed__`, `index`, and `count`
Personally I don't think it is an issue since it is new features and it is not conflicting.

As the previous PR, this patch is NOT removing any existing feature and should not break backward compatibility.

### DIFF
--- a/gym/spaces/tuple.py
+++ b/gym/spaces/tuple.py
@@ -1,9 +1,10 @@
+from collections.abc import Sequence
 from typing import Iterable, List, Optional, Union
 import numpy as np
 from .space import Space
 
 
-class Tuple(Space[tuple]):
+class Tuple(Space[tuple], Sequence):
     """
     A tuple (i.e., product) of simpler spaces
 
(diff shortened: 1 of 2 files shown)


RedTachyon:
Will the additional methods work properly? (`__reversed__` etc). I think if we're adding this abstract class, better make sure it will actually work.

duburcqa:
If the abstract methods are implemented are working, which is the case here, them yes it will without any doubt 👍 like `__iter__`, it does nothing but providing an iterator (in reverse order). It does not actually reverse the container itself.

jkterry1:
We still need to add tests though

duburcqa:
Ok now it should be fine. All tests pass. Just an issue with black but I don't know what it is. It seems unrelated.

RedTachyon:
This looks alright now, it's pretty minor and inoffensive unless I'm missing something very weird.

PR #2649, opened 2022-02-28
Locked framerate in human-mode rendering with PyGame

Markus28:
At the moment, the framerate in `mode="human"` rendering is not locked for many pygame environments. Thus, it will often run at high fps. Since timesteps are fixed, this leads to animations that are faster than real time. This pull request aims to fix the problem for Box2D and continuous control tasks.

Before fix:

After fix:

This fix is not applied to the toy text environments because they don't really have a concept of continuous time.


trigaten:
Why use `hasattr` in some envs, but not others?

Markus28:
I followed the pattern of each class. Usually `self.screen` is initialized to None in `__init__` and we check whether it is still None in `render`. However, in Blackjack, `self.screen` is only initialized in `render`, if the attribute doesn't exist yet. So I did the same thing with `self.clock` to make everything consistent within the class.

Issue #2564, opened 2022-01-19
[Proposal] Autoreset when stepping in done state


jkterry1:
@JesseFarebro proposed a future which I actually think we should do- have environments automatically reset when you take an action in a terminal state, and the observation returned is the first one for each state. This would mean you only need to call reset once (though you could call if more often if desired for seeding), and is analogous to what dm_env does. I think this would result in a big quality of life improvement for users and involve very little code changes, and is standard amongst other APIs. Unless there's a huge issue here someone points out, I think this should be done.

balisujohn:
I've been looking into this, and the big question that comes up for this feature is: "what do we do with the action being input in the .step() call which causes the reset(.step() called when done=1)"

In my current draft, the action is ignored, and the .step() fuction returns the result of a call to .reset() as well as a hard-coded zero reward and done=0. @jkterry1  Does this seems alright, or is there another behavior that would be ideal here? 

My worry is if we first reset, then apply the action, we now have two new observations from the environment and can only return one of them.

RedTachyon:
Can this actually even be done for environments in general? I think the only realistic implementation would be through a wrapper, and adding a wrapper like this is definitely not a big deal. Whether it should be default in `gym.make` is another question.

The way I'd see it implemented is having a guard in the `step` code, after performing the step logic:

```python
if done:
    obs, info = env.reset(...)
```

and probably adding the final observation to the info dict.

I don't see why an action should ever be ignored. After you get a `done==True`, the next action goes to the next episode.

Overall I think I'm fine adding this as a wrapper, and then whether or not it will be used, is down to the user.

balisujohn:
@RedTachyon 
A wrapper is a good idea, my original plan was to modify all the first party environments, but I think a wrapper would take less work and would be able to capture the behavior adequately.

I think the final observation in the info dict would be a bit non-straightfoward assuming people are trying to reconstruct trajectories, since adding the extra state from the last info is an extra thing to remember to do.  Also, there might already be arbitrary keys in the info dict from the environment, meaning there could be key overlap unless we use a nested info dict. 
 
So if you think it's fine to use an info dict to return the final observation, let me know if you'd prefer
 
 ```` 
info =  {env_info: {...}, final_obs: [...]}
````
or
 ```` 
 {env_info..., final_obs: [...]}
````

and I'll proceed with implementing the wrapper.

JesseFarebro:
Indeed, my thoughts for this was a wrapper. It would look something like this:

```py
[...]

PR #2650, opened 2022-02-28
autoreset wrapper

balisujohn:
This pull request implements the proposal in issue #2564.  Here is a draft of a wrapper which will automatically reset its environment when an additional call to step is made after an environment has returned done = True. There is a test provided for this wrapper using a dummy environment. 

For the call to step after done=True, self.env.reset() is called, then 

````obs, None, None, info````

is returned by the wrapper.
`
I think this is ready for review.

### DIFF
--- a/gym/wrappers/autoreset.py
+++ b/gym/wrappers/autoreset.py
@@ -39,9 +39,6 @@ import gym
 
 
 class AutoResetWrapper(gym.Wrapper):
-    def reset(self, **kwargs):
-        return self.env.reset(**kwargs)
-
     def step(self, action):
         obs, reward, done, info = self.env.step(action)
         if done:
--- a/gym/wrappers/autoreset.py
+++ b/gym/wrappers/autoreset.py
@@ -17,7 +17,8 @@ final_done is always True
 
 info is a dict of the form {info:{<self.env info>}, "final_obs":<the
 observation after calling self.env.step(), prior to calling
-self.env.reset()>}
+self.env.reset()>,"final_info":<the info after calling 
+self.env.step(), prior to calling self.env.reset()>}
 
 If done is not true when self.env.step() is called, self.step() returns
 obs, reward, and done as normal, and wraps the info returned
@@ -45,6 +46,11 @@ class AutoResetWrapper(gym.Wrapper):
         obs, reward, done, info = self.env.step(action)
         if done:
             new_obs, new_info = self.env.reset(return_info=True)
-            return new_obs, done, reward, {"info": new_info, "final_obs": obs}
+            return (
+                new_obs,
+                done,
+                reward,
+                {"info": new_info, "final_obs": obs, "final_info": info},
+            )
         else:
             return obs, reward, done, {"info": info}
--- a/gym/wrappers/autoreset.py
+++ b/gym/wrappers/autoreset.py
@@ -1,0 +1,26 @@
+from typing import Optional
+
+import numpy as np
+import gym
+
+
+class AutoResetWrapper(gym.Wrapper):
+    def __init__(self, env):
+        super().__init__(env)
+        self._env_done = False
+
+    def reset(self, **kwargs):
+        self._env_done = False
+        return self.env.reset(**kwargs)
+
+    def step(self, action):
+        if self._env_done:
+            obs, info = self.reset(
+                return_info=True
+            )  # we are assuming the return_info behavior is implemented in environments using this wrapper
+            return obs, None, None, info
+
+        obs, reward, done, info = self.env.step(action)
+        self._env_done = done
+
+        return obs, reward, done, info
--- a/gym/wrappers/autoreset.py
+++ b/gym/wrappers/autoreset.py
@@ -1,3 +1,22 @@
+"""
+A class for providing an automatic reset functionality for gym environments when calling step.
+
+If done was not true for the previous call to step, step returns
+
+obs, reward, done, info
+
+as normal.
+
+If done was true for the previous call to step, the action is ignored, and self.env.reset is called, and step returns
+
+obs, None, None, info
+
+Warning: When using this wrapper to collect rollouts, note that the action given to .step() which causes the reset (first call to 
+.step() after done = True) will have no effect on any observation, and should likely be excluded from logging.
+
+"""
+
+
 from typing import Optional
 
 import numpy as np
--- a/gym/wrappers/autoreset.py
+++ b/gym/wrappers/autoreset.py
@@ -1,54 +1,56 @@
-"""
-A class for providing an automatic reset functionality
-for gym environments when calling self.step().
-
-When calling step causes self.env.step() to return done,
-self.env.reset() is called,
-and the return format of self.step() is as follows:
-
-new_obs, final_reward, final_done, info
-
-new_obs is the first observation after calling self.env.reset(),
-
-final_reward is the reward after calling self.env.step(),
-prior to calling self.env.reset()
-
-final_done is always True
-
-info is a dict of the form {info:{<self.env info>}, "final_obs":<the
-observation after calling self.env.step(), prior to calling
-self.env.reset()>,"final_info":<the info after calling
-self.env.step(), prior to calling self.env.reset()>}
-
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by jkterry1** [comment]
Hey thanks a ton! Could you please add documentation for this and rebase/force push to get rid of the 30+ item long commit log from your previous PR?

**Comment by balisujohn** [comment]
I cleaned up the commit structure and added inline documentation, ready for review again :^)

File: gym/wrappers/autoreset.py [hunk 1]
```diff
@@ -0,0 +1,54 @@
...
+        if done:
+            new_obs, new_info = self.env.reset(return_info=True)
+            return (
+                new_obs,
+                reward,
+                done,
+                {"info": new_info, "final_obs": obs, "final_info": info},
+            )
+        else:
+            return obs, reward, done, {"info": info}
```
RedTachyon, line 54: What happened here? Why is `info` a single-element dictionary? In the base case, the output info dict should just be passed along without modification. If we're doing a reset, I'd just add `info["final_obs"] = obs` etc.

gym/wrappers/autoreset.py
```diff
@@ -0,0 +1,54 @@
...
+"""
+
+
+import gym
+
+
+class AutoResetWrapper(gym.Wrapper):
+    def step(self, action):
+        obs, reward, done, info = self.env.step(action)
+        if done:
```
RedTachyon, line 45: I think I'd prefer a single-return implementation, something like this:

```python
if done:
    info["final_obs"] = obs
    info["final_info"] = info
    obs, new_info = self.env.reset(...)
return obs, reward, done, info
```

(except properly patch through all the infos in the right way, which doesn't happen in this code snippet - but I hope the idea is clear)

tests/wrappers/test_autoreset.py
```diff
@@ -0,0 +1,93 @@
...
+    ):
+        self.count = 0
+        if not return_info:
+            return np.array([self.count])
+        else:
+            return np.array([self.count]), {"count": self.count}
+
+
+def test_autoreset_reset_info():
+    env = gym.make("CartPole-v1")
```
RedTachyon, line 44: You can parametrize the tests to run for all environments, the pattern appears in some other tests

tests/wrappers/test_autoreset.py
```diff
@@ -0,0 +1,93 @@
...
+    env = AutoResetWrapper(env)
+    ob_space = env.observation_space
+    obs = env.reset()
+    assert ob_space.contains(obs)
+    del obs
```
RedTachyon, line 49: Redundant

tests/wrappers/test_autoreset.py
```diff
@@ -0,0 +1,93 @@
...
+import numpy as np
+
+import gym
+from gym.wrappers import AutoResetWrapper
+
+
+class DummyResetEnv(gym.Env):
+    metadata = {}
+
+    def __init__(self):
```
RedTachyon, line 13: Add a docstring describing what this environment does please

RedTachyon (changes requested):
This isn't necessarily a comprehensive review, in the sense that I might have some more comments after everything here is addressed, but these things are the main things to fix.

tests/wrappers/test_autoreset.py
balisujohn: @RedTachyon  The problem with testing the wrapper for all environments is for the real environments often they have non-deterministic starting state, and non-deterministic transitions, meaning it is difficult to know what the correct return values should be for various cases, and when done will be true, without setting a seed and then recording the behavior of each environment and hard coding a test for the recorded behavior. That was my motivation for creating the deterministic dummy environment.

gym/wrappers/autoreset.py
balisujohn: @RedTachyon  So the big thing here is I don't want to overwrite the keys if there are already "final_obs" or "final_info" keys in info from the environment itself. I'm happy with adopting this style if we have an assert that the info does not already contain these keys.

tests/wrappers/test_autoreset.py
balisujohn: I didn't do that for the autoreset wrapper since a lot of the other wrappers are not tested with all environments. My thinking is that if we test every wrapper with every first party environment, it could make the tests take a lot longer.

tests/wrappers/test_autoreset.py
balisujohn: done

tests/wrappers/test_autoreset.py
balisujohn: removed

gym/wrappers/autoreset.py
balisujohn: My two design goals were consistent return format across done = true and done = false cases, and preventing keys from getting overwritten if there were already keys called "final_obs" or "final_info" in the environment's info. I think my new version addresses your design goals while protecting against accidental key overriding.

balisujohn:
Hi @RedTachyon Thanks for your code review! I think I have either responded to or fixed each point made in the review. Let me know if you would like additional changes.

jkterry1:
So I took a moment to sit down and thing about how this should work wrt. other environments. I think that the changes this wrapper makes, in its current state, make sense. I also think that this should be applied to the default API, with the exception that the option to turn it off remain after the 1.0 release. The remaining question then is if this should be applied to all environments as a hard coded change or via a universally applied wrapper in the `make` method. My preference is for this to be a universally applied wrapper and document it in the env creation docs, however if there's a reason not to do it this way, I'd like to know.

RedTachyon:
(I'll go through the updated code soon)

One thing to keep in mind is that doing this without a wrapper might be borderline impossible without (a) some extremely cursed features (I can't think of a specific one right now, but I can imagine one exists), (b) requiring users to include additional code just for this feature to work, or (c) reinventing wrappers.

This is because we'd need to call some custom code at the end of `env.step`, which is defined entirely by the user, and by default is a completely abstract method. A cursed method might involve something with metaclasses, maybe making `env.step` call another method under the hood. A simpler (but still cursed) way would be adding this to `Env.step` and requiring users to add a `super` call at the *end* of their `env.step` implementation. 
And if you put me in a room with a computer, without the existence of wrappers, and asked me to implement this feature, I'd probably write a wrapper.

Unless I'm missing some really simple way to implement this, I'd be very interested to see that. But we're probably stuck with keeping this as a wrapper because there's just no other sane way.

vwxyzjn (changes requested):
Everything else looks good but let's align the info keys :D

vwxyzjn:
> So I took a moment to sit down and thing about how this should work wrt. other environments. I think that the changes this wrapper makes, in its current state, make sense. I also think that this should be applied to the default API, with the exception that the option to turn it off remain after the 1.0 release. 

A quick thought on this. If you want to do this, the API would be really similar to the vector env API. It raises the question for me: why not just make every environment a vectorized environment in this case?

gym/wrappers/autoreset.py
```diff
@@ -0,0 +1,56 @@
...
+    def step(self, action):
+        obs, reward, done, info = self.env.step(action)
+
+        if done:
+
+            new_obs, new_info = self.env.reset(return_info=True)
+            assert "final_obs" not in new_info
+            assert "final_info" not in new_info
+
+            new_info["final_obs"] = obs
```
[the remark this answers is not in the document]
balisujohn: done

jkterry1:
The goal is to keep the basic API for educational and related uses and have a new make_vec method for returning efficient vectorized environments natively in the future.

jkterry1:
Per Ariel's comments, my corresponding vote would be to shim this as a wrapper in the make method then unless I'm missing something.

balisujohn:
Do you want this added to gym.make in the scope of this pull request, or will that be later on?

jkterry1:
I went ahead and merged this, please create a second PR adding the gym.make functionality

jkterry1:
@balisujohn I'm just following up on the make argument PR?

balisujohn:
@jkterry1 I should be able to get to it this weekend.

PR #2651, opened 2022-02-28
Removed Striker and Thrower

RushivArora:
This PR removes the striker and thrower Mujoco environments from Gym.

While creating the PR I realised, would it make sense to add them to an archive directory in the Mujoco Envs? I do not currently do that, but can add it here @jkterry1


RushivArora:
Backed up the thrower and striker environments at: https://github.com/RushivArora/Gym-Mujoco-Archive

This PR now completes removes for these two environments: 2 .xml asset files, 2 .py files for the environment, init calls in `gym.envs.mujoco.__init__`, and register commands in `gym.envs.__init__`

PR #2654, opened 2022-02-28
Rename render modes/fps

trigaten:


ZhiqingXiao:
@trigaten @jkterry1   Renaming `"render.modes"` to `"render_modes"` is unnecessary, and it will introduce incompatibility to many  extension libraries.

pzhokhov:
@trigaten what is the reason for this change?

jkterry1:
@pzhokhov the variable names here are going to become main features of the gym API in the next release instead of being minor features as part of the roadmap to the 1.0 release, so we changed them to a more standardized naming in advance of making them a core feature.

pseudo-rnd-thoughts:
Running the pytests I get the following warnings, could you fix these

```
tests/wrappers/test_record_video.py: 10 warnings
tests/wrappers/test_video_recorder.py: 3 warnings
  /gym/wrappers/monitoring/video_recorder.py:115: DeprecationWarning: WARN: `env.metadata["video.frames_per_second"] is marked as deprecated and will be replaced with `env.metadata["render_fps"]` see https://github.com/openai/gym/pull/2654 for more details

```

Issue #2647, opened 2022-02-27
[Bug Report] ~1500 missing warnings with pytest --forked option


kir0ul:
If you are submitting a bug report, please fill in the following details and use the tag [bug].

**Describe the bug**
I don't know if I'm missing something but it seems to me the pytest `--forked` option is missing to report **a lot** of warnings.

**Code example**
With the `--forked` option I get only **2 warnings** (as it is currently the case in CI):
```
❯ docker run gym-docker pytest --forked --import-mode=append
+ export DISPLAY=:0
+ DISPLAY=:0
+ display=0
+ file=/tmp/.X11-unix/X0
+ Xvfb -screen 0 1024x768x24
++ seq 1 10
+ for i in $(seq 1 10)
+ '[' -e /tmp/.X11-unix/X0 ']'
+ echo 'Waiting for /tmp/.X11-unix/X0 to be created (try 1/10)'
+ sleep 1
Waiting for /tmp/.X11-unix/X0 to be created (try 1/10)
+ for i in $(seq 1 10)
+ '[' -e /tmp/.X11-unix/X0 ']'
+ break
+ '[' -e /tmp/.X11-unix/X0 ']'
+ exec pytest --forked --import-mode=append
============================= test session starts ==============================
platform linux -- Python 3.10.2, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /usr/local/gym
plugins: forked-1.4.0
collected 623 items / 5 skipped / 618 selected

tests/test_core.py ............                                          [  1%]
tests/envs/test_action_dim_check.py s                                    [  2%]
tests/envs/test_determinism.py ................                          [  4%]
tests/envs/test_envs.py ..................................               [ 10%]
tests/envs/test_frozenlake_dfs.py .                                      [ 10%]
tests/envs/test_lunar_lander.py ..                                       [ 10%]
tests/envs/test_mujoco_v2_to_v3_conversion.py s                          [ 10%]
tests/envs/test_registration.py ......................................   [ 16%]
tests/spaces/test_spaces.py ............................................ [ 23%]
............................................................             [ 33%]
tests/spaces/test_utils.py ............................................. [ 40%]
....................................                                     [ 46%]
tests/utils/test_atexit.py .                                             [ 46%]
tests/utils/test_env_checker.py .                                        [ 46%]
tests/utils/test_seeding.py ..                                           [ 47%]
tests/vector/test_async_vector_env.py ..............................     [ 52%]
[...]

PR #2660, opened 2022-03-02
Reduces warnings produced by pytest from ~1500 to 13 

pseudo-rnd-thoughts:
Addresses a majority of warnings noted in issue #2647
These changes reduced the number of warnings pytest (with --forked) from ~1500 to 13

Remaining warnings
* Bipedal walker and lunar lander observation spaces' low and high values are np.inf, env_checker.py raises warnings that these values should be lower. I see no reason why we can't set these values lower however to what number? (8 warnings)
* In testing spaces, test_iterable and test_iterable_custom_space use spaces that return a single value causing spaces.Box to raise a warning when 

Errors
* tests/vector/test_async_vector_env.py::test_custom_space_async_vector_env_shared_memory raises an exception that is uncaught currently. 
* tests/vector/test_async_vector_env.py::test_call_async_vector_env causes a failure of the tests

Both are currently being solved

### DIFF
--- a/gym/envs/box2d/car_racing.py
+++ b/gym/envs/box2d/car_racing.py
@@ -170,6 +170,8 @@ class CarRacing(gym.Env, EzPickle):
             shape=polygonShape(vertices=[(0, 0), (1, 0), (1, -1), (0, -1)])
         )
 
+        # This will throw a warning in tests/envs/test_envs in utils/env_checker.py as the space is not symmetric
+        #   or normalised however this is not possible here so ignore
         self.action_space = spaces.Box(
             np.array([-1, 0, 0]).astype(np.float32),
             np.array([+1, +1, +1]).astype(np.float32),
--- a/gym/envs/classic_control/pendulum.py
+++ b/gym/envs/classic_control/pendulum.py
@@ -95,6 +95,9 @@ class PendulumEnv(gym.Env):
         self.screen_dim = 500
 
         high = np.array([1.0, 1.0, self.max_speed], dtype=np.float32)
+        # This will throw a warning in tests/envs/test_envs in utils/env_checker.py as the space is not symmetric
+        #   or normalised as max_torque == 2 by default. Ignoring the issue here as the default settings are too old
+        #   to update to follow the openai gym api
         self.action_space = spaces.Box(
             low=-self.max_torque, high=self.max_torque, shape=(1,), dtype=np.float32
         )
@@ -186,7 +189,7 @@ class PendulumEnv(gym.Env):
             scale_img = pygame.transform.smoothscale(
                 img, (scale * np.abs(self.last_u) / 2, scale * np.abs(self.last_u) / 2)
             )
-            is_flip = self.last_u > 0
+            is_flip = bool(self.last_u > 0)
             scale_img = pygame.transform.flip(scale_img, is_flip, True)
             self.surf.blit(
                 scale_img,
--- a/gym/envs/toy_text/cliffwalking.py
+++ b/gym/envs/toy_text/cliffwalking.py
@@ -44,7 +44,7 @@ class CliffWalkingEnv(Env):
         self.nA = 4
 
         # Cliff Location
-        self._cliff = np.zeros(self.shape, dtype=np.bool)
+        self._cliff = np.zeros(self.shape, dtype=bool)
         self._cliff[3, 1:-1] = True
 
         # Calculate transition probabilities and rewards
--- a/gym/spaces/discrete.py
+++ b/gym/spaces/discrete.py
@@ -25,7 +25,7 @@ class Discrete(Space[int]):
         super().__init__((), np.int64, seed)
 
     def sample(self) -> int:
-        return self.start + self.np_random.randint(self.n)
+        return int(self.start + self.np_random.integers(self.n))
 
     def contains(self, x) -> bool:
         if isinstance(x, int):
--- a/gym/wrappers/record_video.py
+++ b/gym/wrappers/record_video.py
@@ -116,3 +116,9 @@ class RecordVideo(gym.Wrapper):
             self.video_recorder.close()
         self.recording = False
         self.recorded_frames = 1
+
+    def close(self):
+        self.close_video_recorder()
+
+    def __del__(self):
+        self.close_video_recorder()
--- a/tests/envs/test_envs.py
+++ b/tests/envs/test_envs.py
@@ -10,6 +10,7 @@ from gym.utils.env_checker import check_env
 # This runs a smoketest on each official registered env. We may want
 # to try also running environments which are not officially registered
 # envs.
+@pytest.mark.filterwarnings('ignore:.*We recommend you to use a symmetric and normalized Box action space.*')
 @pytest.mark.parametrize("spec", spec_list)
 def test_env(spec):
     # Capture warnings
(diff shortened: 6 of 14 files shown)


jkterry1:
Thanks so much, this is really great.

-Can you please remove the forked property from CI that caused this disaster in the first place and get fix linting?
-For the lunar lander and bipedal walker environments, could you please do the physics math and figure out the maximum observation ranges? (If you don't have the background it's okay and I can have someone else deal with it, it's just everyone with the relevant background here is really over capacity right now)
-You're currently working on fixing the error in the sync test, right?

pseudo-rnd-thoughts:
* I will look at computing the maths for the lunar lander and bipedal walker. If I can't do it, then I will comment back
* For the async test, I had a look at it yesterday and couldn't solve it. Will look at it again now

jkterry1:
Please do the unused imports in a separate PR, there's a bunch of imports that static analysis often thinks are unused by aren't. If that's real that's obviously great to remove though (and flake8 should've caught them....)

pseudo-rnd-thoughts:
@jkterry1 Both the build and pre-commit are failing on test_async_vector_env::test_call_async_vector_env
Strangely, when I just run pytest tests/vetor/test_async_vector_env.py then I never get an error but when I run all of the tests (pytest tests) then an error is thrown

This is an issue because, I can't replicate the issue easily outside of pytest. Any ideas of why this could be happening?

kir0ul:
It seems there's some strange interaction going on in both (with and without shared memory) `test_call_async_vector_env` tests, which results in a segfault at line 113: `images = env.call("render", mode="rgb_array")`, but I don't know why (if anyone else have any idea, feel free to chime in).

As a workaround, I created a patch (based on https://github.com/pseudo-rnd-thoughts/gym/commit/addbaf0f28c86e6839b2b113ce75358099e0eb25) that adds a mark on the failing test, and modified the CI build so that it runs the tests in two steps:

1. the "main' tests without the `--forked` option, so that it keeps the useful warnings,
2. the test that require the `--forked` option.

It's not super pretty but at least all the CI pipelines passed.
@pseudo-rnd-thoughts Feel free to cherry pick my patch if you think it's useful: https://github.com/kir0ul/gym/commit/ca4e9f78993db10a3dd3714d60bdba43f7780134

benblack769:
So to me this seems like typical GUI nonsense. The way that GUI code opens and closes system resources controlling window management and how that interacts with multiprocessing is quite fragile.

Here is my understanding of the goings on here. Simply put, the env tests are creating system resources that are breaking the asyc vec tests.

Non-forked verion:

1. Env tests open and close connections to system resources controlling window management (x11 stuff) in main thread. Some of these resources are maintained, despite calling pygame.quit, etc.
2. Process fork in async vec env also tries to open render window. System is very confused because the process that originally allocated those resources (the main process) is not the same as the new process (the fork). Segfaults ensue.

While in the forked version, each test gets its own fork of the main process, and so resouces are cleaned up properly and you never fork a process which has already opened up a x11 connection.

One way of fixing this is to simply not open a window during rendering. This is possible with pygame via something like this:

```
def render(...):
      ....
            if mode == "human":
                pygame.init()
                self.screen = pygame.display.set_mode((screen_width, screen_height))
				self.window_is_opened = True
            else:
                self.screen = pygame.Surface((screen_width, screen_height))
				self.window_is_opened = False

def close(self):
	if self.screen is not None:
		if self.window_is_opened:
			pygame.quit()
		self.isopen = False
```
This is the solution we have adopted in pettingzoo, but it is a bit problematic because it is impossible to switch between "rgb_array" and "human" render modes. I have confirmed that this fixes the problem, but I would not recomend changing this without more discussion.

Another solution is to simply have the vector env tests run in a different pytest run than the env tests.

Another solution is to figure out how to properly clean up the process so that the OS has no idea that this process ever openend up a window (no idea how where to start here)

jkterry1:
Another option, as proposed by @RedTachyon is to remove this functionality from async entirely as it's so incredibly rarely used (if used at all)

pseudo-rnd-thoughts:
Thank you all for the suggestions and for the code idea @kir0ul but I actually solved the issue  
@benblack769 Your suggestion sparked a brain wave for me that maybe some resources were not being cleaned up properly causing the issue
Linking this with the idea that we were only getting an error when running pytest tests but no error for pytest /tests/vector

In short, pygame and CarRacing were causing the issue so this only happened when tests/envs/test_determinism.py tests/envs/test_envs.py are run with test_call_async_vector_env

The fix that I implemented was on CarRacing close, then pygame are quit (releasing all of its resources I think)
```
from tests.envs.spec_list import spec_list
from tests.vector.test_async_vector_env import test_call_async_vector_env

print('Pre env call')
test_call_async_vector_env(True)

print(spec_list[4])
env = spec_list[4].make()
env.close()

print('Post env call')
test_call_async_vector_env(True)
``` 
However, when I run this example code I get the following warning `ALSA lib pcm_dmix.c:1089:(snd_pcm_dmix_open) unable to open slave` (I'm on Ubuntu 20.04)
Is anyone able to check if a similar error is thrown on other systems 

I think that this requires more investigation of why this happened if we care about Async vector environments to generate a simplified version of the problem to test with for pytest if it is an issue on their own

jkterry1:
Alsa is a linux audio server that pygame uses by default and shouldn't, that occasionally causes errors. PettingZoo changes the default flag to a sensible one on load to prevent these things (grep alsa in the library should do I believe). This should've been applied to gym before but I forgot about it until just now.

jkterry1:
@pseudo-rnd-thoughts after some discussion, none of the maintainers have any earthly idea what's going on with the iterator thing you mentioned at the beginning, and that's in code that's set to be replaced later this year. Unless you can figure it out in a reasonable length of time, I think we should just live with it.

pseudo-rnd-thoughts:
@jkterry1 Thanks for exploring that, the problem seems to be that the iterators are returning a value instead of an array which is why Box is raising a warning. However, if that code is being replaced soon, I won't worry about it

For the ASLA, you mentioned that pettingzoo had a solution for pygame to stop the warning. Could you link the commit or the solution for me to implement? 

If we ignore the iterator issues then the final issue is with the observation space for the lunar lander and bipedal walker environments. 
Looking at the maths, it is too complex for me to find the max and min values. 
So I have run the environment with random actions, 10000  times to find the maximum and value observations the agent encountered and got around 7 (I dont have the exact values to hand) for both. 
In addition, I'm going to download some pre-trained models for both environments and repeat the same for that. 

Do you think that would be acceptable to use as the upper and lower bounds of the environment? We could add some additional tolerance just in case there are some more extreme values. Or do you think we can compute the true max and min values of the environment? 

I'm hoping to have this finish this weekend so we can close the PR

jkterry1:
The ALSA hack is in this file:  https://github.com/Farama-Foundation/PettingZoo/blob/master/pettingzoo/__init__.py

The CI error is strange, does black (using the same version of it as CI) pass for you locally?

For the math, I'm going to ask Ariel to take a crack at them. He has the background to do so.

jkterry1:
We'll have to merge this PR without the lunar lander and multiwalker bounds and those will be handled in a seperate PR

pseudo-rnd-thoughts:
@jkterry1 I have used your pygame fix and run black locally and have committed the reformatted files which stopped black throwing errors locally at least  
Also, I seemed to have messed up trying to fix a conflict with the updated master. Do I need to change something?

RedTachyon:
Quick comment about observation space bounds for LunarLander and BipedalWalker - I'll try to get at least some bounds, but I suspect some components might have to remain infinite. Specifically the velocity-related terms, and especially angular velocity *I think* would be unbounded. So the warnings would still be there.

PR #2662, opened 2022-03-03
Add SpaceWrapper, FilteredSpace and Space.where

lebrice:
This PR adds a new feature that makes the gym spaces a lot more expressive: **Filtered** spaces!

There are 3 components in this PR:
- `SpaceWrapper` (in `gym/spaces/other/wrapper.py`):
    - Same idea as `gym.Wrapper`, but for spaces! This can serve as the basis for a bunch of cool adapters for gym spaces in the future! For example, something like a `TensorSpaceWrapper`, that converts the samples to/from Tensors in the `sample` and `contains`, respectively.
- `FilteredSpace`  (in `gym/spaces/other/filter.py`):
    - Adds assumptions (predicates) to a space, limiting the samples that are considered valid.
    - Uses rejection sampling with a maximum of `max_sampling_attempts` attempts.
- `Space.where(predicate)`:
    - Adds filters to the space, returning a `SpaceWrapper`.

```python
>>> from gym.spaces import Discrete, Box, Space
>>> import numpy as np

>>> odd_digits = Discrete(10, seed=123).where(lambda n: n % 2 == 1)
>>> 1 in odd_digits
True
>>> 2 in odd_digits
False
>>> odd_digits.sample()
5
```
Other cool examples:
```python
>>> unit_circle = Box(-1, 1, shape=(2,), dtype=np.float64, seed=123).where(lambda xy: (xy ** 2).sum() <= 1)
>>> np.array((0.1, 0.2)) in unit_circle
True
>>> np.array((1, 1)) in unit_circle
False
>>> unit_circle.sample()
array([ 0.36470373, -0.89235796])

>>> def hypersphere(radius: float, dimensions: int) -> FilteredSpace[np.ndarray]:
...     return Box(-radius, +radius, shape=(dimensions,)).where(
...         lambda v: v.pow(2).sum() <= radius**dimensions
...     )
...
>>> unit_sphere = hypersphere(radius=1, dimensions=3)
```

### DIFF
--- a/gym/spaces/other/filter.py
+++ b/gym/spaces/other/filter.py
@@ -38,27 +38,27 @@ class FilteredSpace(SpaceWrapper[T_cov]):
 
     Other cool examples:
 
-    >>> unit_circle = Box(-1, 1, shape=(2,), dtype=np.float64, seed=123).where(lambda xy: (xy ** 2).sum() <= 1)
-    >>> np.array((0.1, 0.2)) in unit_circle
+    >>> unit_disk = Box(-1, 1, shape=(2,), dtype=np.float64, seed=123).where(lambda xy: (xy ** 2).sum() <= 1)
+    >>> np.array((0.1, 0.2)) in unit_disk
     True
-    >>> np.array((1, 1)) in unit_circle
+    >>> np.array((1, 1)) in unit_disk
     False
-    >>> unit_circle.sample()
+    >>> unit_disk.sample()
     array([ 0.36470373, -0.89235796])
 
-    >>> def hypersphere(radius: float, dimensions: int) -> FilteredSpace[np.ndarray]:
+    >>> def hyperball(radius: float, dimensions: int) -> FilteredSpace[np.ndarray]:
     ...     return Box(-radius, +radius, shape=(dimensions,)).where(
-    ...         lambda v: v.pow(2).sum() <= radius**dimensions
+    ...         lambda v: v.pow(2).sum() <= radius ** 2
     ...     )
     ...
-    >>> unit_sphere = hypersphere(radius=1, dimensions=3)
+    >>> unit_ball = hyperball(radius=1, dimensions=3)
     """
 
     def __init__(
         self,
         space: Space[T_cov],
         predicates: Sequence[Predicate[T_cov]],
-        max_sampling_attempts: int | None = None,
+        max_sampling_attempts: int | None = 1_000,
     ):
         super().__init__(space=space)
         self.space = space
--- a/gym/spaces/other/filter.py
+++ b/gym/spaces/other/filter.py
@@ -1,0 +1,171 @@
+from __future__ import annotations
+
+import functools
+import inspect
+from typing import Any, Callable, List, Sequence, TypeVar
+
+import numpy as np
+from gym.spaces.box import Box
+from gym.spaces.dict import Dict as DictSpace
+from gym.spaces.discrete import Discrete
+from gym.spaces.multi_binary import MultiBinary
+from gym.spaces.multi_discrete import MultiDiscrete
+from gym.spaces.other.wrapper import SpaceWrapper
+from gym.spaces.space import Space
+from gym.spaces.tuple import Tuple as TupleSpace
+
+T = TypeVar("T")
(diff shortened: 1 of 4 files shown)


Markus28:
Some very superficial comments
- `v.pow(2).sum() <= radius**dimensions` needs to be changed to `v.pow(2).sum() <= radius**2` in the documentation (or alternatively use `np.linalg.norm`).
- This is rather pedantic remark, but consider changing "unit circle" and "hypersphere" to "unit disk" and "hyperball", respectively. The former is usually understood to be the boundary of the latter
- At the moment, it seems that rejection sampling will not terminate after a fixed number of steps by default. It might be a good idea to issue a warning after a certain number of samples anyways. The curse of dimensionality is quite unintuitive, so I'd be concerned about people falling into that trap (in your hyperball example, for instance, you can expect >100 samples per `.sample()` call for `dimensions=9` and it gets worse *very* quickly)

RedTachyon:
I'm pretty conflicted about this, it kinda makes sense, especially from a mathematical point of view, but I don't see how this helps anything on the actual programming side.

The only things it really seems to affect are `.contains` and `.sample`, both of which are a little bit useful but not that much. `contains` is just validation, so ideally it should never actually "fire up" if everything's designed well; `sample` is more of a convenience function for quick testing and probably shouldn't be used in any serious computation.

What exactly is the benefit and usecases of this? Current spaces have the philosophy that they contain all valid observations, but it's not the case that every observation within the space will necessarily occur. (injective vs surjective) This adds a very limited surjective property, but is this ever actually useful?

lebrice:
Hey @RedTachyon, thanks for taking a look.
I'm assuming you are hesitant about the `FilteredSpace`, but what do you think of `SpaceWrapper`? I can split this into two PRs if need be, LMK.

You are right that these toy examples aren't terribly useful. The real value of `FilteredSpace` and `Space.where` to me is the expressive power that this gives to all the gym spaces, as well as any spaces that are added in the future, since you can define arbitrarily specific subspaces that describe exactly what your environment / algorithm expects to receive / produce.


One potential use-case that comes to mind would be environments where you have some type of constraint on the actions you can take, for example, an environment where you have an allocation problem, and the sum of all the action terms should be 1:
```python
action_space = Box(0, 1, shape=(4,)).where(lambda v: np.isclose(sum(v), 1))
```

Other use-cases that I have in mind involve other subclasses of `gym.Space` that I've been creating outside of gym. In these kinds of spaces, the `Space.contains` and `Space.sample` methods are quite a bit more important than in the current gym spaces, since they are used to specify the interfaces between algorithms and the environments they can be applied to. It therefore becomes very useful to check if a given environment, for instance, can fit within the spaces of environments that the algorithm supports, something like this:
```python
>>> Envs: Space[gym.Env]
>>> discrete_action_envs = Envs().where(lambda env: isinstance(env.action_space, Discrete))
>>> gym.make("CartPole-v1") in discrete_action_envs
True
>>> gym.make("Acrobot-v1") in discrete_action_envs
False
>>> discrete_action_envs.sample()
<TimeLimit<OrderEnforcing<MountainCarEnv<MountainCar-v0>>>>
```

I'll think of more practical use-cases that only involve the current spaces / environments in gym. I believe that this feature would be a good addition to gym, even if it turns out that the current codebase would have little need for it.

RedTachyon:
Overall I still don't really see an actual use case for current or foreseeable-future gym code, except some very niche stuff (and there's so much niche stuff in RL that any feature would find its use case). Env-typed spaces are an admittedly pretty cute abuse of existing structures, but it's so niche that I don't see why it should be combined with `gym.Space` which has a very clear-cut purpose.

And SpaceWrapper in its current form has exactly the same issues. I can kinda see the concept (I'm a sucker for anything Haskell-y, and I wouldn't be surprising if you had some functional programming inspiration), and could get behind it if I see it being useful, but right now the only actual wrapper is FilteredSpace.

tl;dr I don't really see this PR being approved anytime soon, it might be closed or kept in a thinking box for a while. If you're still convinced that this is a good idea (it might be!), I'd recommend opening an issue with a proposal, and focusing more on the justification of the actual feature. And if then the feature itself is approved, the implementation will be the trivial part.

PR #2663, opened 2022-03-03
Add ImageSpace (subclass of Box with convenience properties)

lebrice:
This PR does two things:
- Adds `ImageSpace` which is is just a `Box` space with some added convenience methods and properties:
    ```python
    >>> mnist_images = ImageSpace(0, 255, (1, 28, 28), dtype=np.uint8)
    >>> mnist_images
    ImageSpace(0, 255, (1, 28, 28), uint8)
    >>> mnist_images.is_channels_first
    True
    >>> mnist_images.channels, mnist_images.height, mnist_images.width
    (1, 28, 28)
    
    >>> image_batch = ImageSpace(0, 1, (32, 3, 256, 256), dtype=np.float64)
    >>> image_batch
    ImageSpace(0.0, 1.0, (32, 3, 256, 256), float64)
    >>> image_batch.is_channels_first
    True
    >>> image_batch.batch_size
    32
    >>> image_batch.channels
    3
    >>> image_batch.height, image_batch.width
    (256, 256)
    ```

- minor clean-up of the `batch_space` function from `gym.vector.utils.spaces`

### DIFF
--- a/gym/spaces/box.py
+++ b/gym/spaces/box.py
@@ -167,7 +167,8 @@ class Box(Space[np.ndarray]):
         return [np.asarray(sample) for sample in sample_n]
 
     def __repr__(self) -> str:
-        return f"Box({self.low_repr}, {self.high_repr}, {self.shape}, {self.dtype})"
+        cls = type(self)
+        return f"{cls.__qualname__}({self.low_repr}, {self.high_repr}, {self.shape}, {self.dtype})"
 
     def __eq__(self, other) -> bool:
         return (
--- a/gym/spaces/other/image.py
+++ b/gym/spaces/other/image.py
@@ -1,0 +1,111 @@
+from __future__ import annotations
+from typing import Optional, SupportsFloat
+import numpy as np
+from gym.spaces.box import Box
+
+
+class ImageSpace(Box):
+    """ Space of image arrays.
+
+    This is just a `Box` space with some added convenience methods and properties.
+    
+    >>> mnist_images = ImageSpace(0, 255, (1, 28, 28), dtype=np.uint8)
+    >>> mnist_images
+    ImageSpace(0, 255, (1, 28, 28), uint8)
+    >>> mnist_images.is_channels_first
+    True
+    >>> mnist_images.channels, mnist_images.height, mnist_images.width
+    (1, 28, 28)
+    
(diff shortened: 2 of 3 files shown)


jkterry1:
Hey, I discussed this with @RedTachyon . I don't think that the couple of extra features added are worth the added complexity to users of having an additional space for people to deal with? I'm going to close this for the moment, but feel free to push back.

lebrice:
Hey @jkterry1, thanks for replying. 
I placed this feature in a new `gym.spaces.other` package, with this kind of consideration in mind. Don't you think that kind of separation is sufficient for people to both have the spaces they are used to in `gym.spaces` and other newer/different spaces in `gym.spaces.other`? Where/how would you like new contributions to gym.spaces supposed to be placed/organized?

RedTachyon:
@lebrice I think the key question when adding any feature is "Is this feature actually necessary?" Adding it into an isolated location (which itself is a new feature in this case) doesn't bypass that. This seems to have been added out of nowhere, without any previous issue raised with a discussion on the topic?

lebrice:
Hey @RedTachyon, you are correct, this does come "out of nowhere", and with the history of Gym, taking a conservative  stance towards new features is totally understandable.

However, I'd argue that there still needs to be a place for this kind of spontaneous, low-risk/potentially-high-reward features to be proposed, wouldn't you agree? Limiting ourselves to strictly considering  "actually necessary" changes prevents a bunch of creative and potentially useful ideas from being proposed by members of the community!

> without any previous issue raised with a discussion on the topic?

In this particular case, I feel like a PR is as good a place as any to discuss the feature, since the contribution is tiny, and all the code is here.

Practical use-cases for this feature are easy to imagine:  `ImageSpace`s could be used as the observation_space of the `PixelObservationWrapper` wrapper, for instance, and any env which outputs images. This would probably remove a bunch of boilerplate logic for checking the image format, or image width, height, etc from a bunch of algorithms in a bunch of repos.

lebrice:
> having an additional space for people to deal with

I'd argue that `ImageSpace` being a subclass of Box, this is more like 1.25x the number of spaces to deal with, since everything about Box is exactly the same with `ImageSpace`. All the generic functions also work exactly the same way.

tristandeleu:
Would `ImageSpace` be any less useful than `MultiDiscrete` or `MultiBinary`?

Issue #2540, opened 2021-12-22
[Proposal] New rendering API


jkterry1:
As mentioned in #2524, the rendering API is the last breaking change I think should be made.

Right now, the rendering API has a few problems problems:

- When using frame skipping or similar wrappers, calling .`render()` after each step does allow you to extract a smooth video
- Many environments can't render after each step (or if they can, doing so is super hacky), but are able to easily return an output at the end. This the case for things like physics simulation environments, where it makes more sense to save state logs that can then be handed off to separate rendering logic at the end.
- Relatedly, certain environments can only render after being initialized in a certain way, resulting in weird environment logic to make sure they're always renderable.
- While this isn't an issue for Gym, in PettingZoo (which is based on Gym's API), its very unclear to users which agents step to call .render().

What I think would be a solution to all these problems that would also be more user friendly is the following:

1. Make `render` an argument to `init`, defaulting to `render=None`, where people can specify if they want to render, and if so how, at initialization. This could also be put in reset instead of init, I'm not certain which makes more sense.
2. Add a collection method to call after you're done acting in the environment, where if render should return something (e.g. a list of numpy frames or a video file as specified by the initialization argument), this is returned to you. We could call this `.collect()`, though I'm open to better names for the function.

We could also create a Gym wrapper, say `gym.wrappers.legacy_render ` that makes calls to existing `.render()` methods and supports common render arguments (presumably `live`, `video` and `frames`), such that there old environments wouldn't even have to change their code to still be useable.

vwxyzjn:
I agree for fast throughput envs we need a different rendering mechanism and thus some changes are warranted. Here are some thoughts.

> When using frame skipping or similar wrappers, calling .render() after each step does allow you to extract a smooth video

Do you mean "does not"? This is not necessarily true. For example, the following can extract a smooth video.

```python
env = gym.make(gym_id)
env = gym.wrappers.RecordEpisodeStatistics(env)
env = gym.wrappers.RecordVideo(env, f"videos/{run_name}")
env = NoopResetEnv(env, noop_max=30)
env = MaxAndSkipEnv(env, skip=4)
env = EpisodicLifeEnv(env)
if "FIRE" in env.unwrapped.get_action_meanings():
    env = FireResetEnv(env)
env = ClipRewardEnv(env)
env = gym.wrappers.ResizeObservation(env, (84, 84))
env = gym.wrappers.GrayScaleObservation(env)
env = gym.wrappers.FrameStack(env, 4)
```
[...]

PR #2671, opened 2022-03-06
Render API

younik:
## New render API
Following this discussion: #2540

This PR extends the `render()` method, allowing the user to specify `render_mode` during the instantiation of the environment. The default value of `render_mode` is None; in that case, the user can call render with the preferred mode as before. In this way, these changes are backwards compatible. 
In case `render_mode != None`, the argument _mode_ of `.render()` is ignored.
In case `render_mode = "human"`, the rendering is handled by the environment without needing to call `.render()`.
With other render modes, `.render()` returns a proper `List` with all the renders since the last `.reset()`or `.render()`. For example, with `render_mode = "rgb_array"`, `.render()` returns a `List` of `np.ndarray`, while with `render_mode = "ansi"` a `List[str]`.

### TODO
- [x] Add deprecation warnings to mode arg in .render() and VideoRecorder

## Examples

```
import gym

env = gym.make('FrozenLake-v1', render_mode="human")
env.reset()
for _ in range(100):
    # env renders automatically, no needs to call .render()
    
env.render()
> None
```

```
import gym

env = gym.make('CartPole-v1', render_mode="rgb_array")
env.reset()

for _ in range(10):

frames = env.render()
type(frames)
> <class 'list'>
len(frames)
> 11
len(env.render()) # expect 0 because frames are popped by previous .render() call
> 0

env.reset()
len(env.render())
> 1
```

### Example of backward compatibility:

```
import gym

env = gym.make('FrozenLake-v1')  # default render_mode=None
env.reset()
for _ in range(100):
    # no rendering handled by the environment since render_mode = None
    env.render()  # render with human mode (default)
   
```


andrewtanJS:
Hey @younik, overall it looks great! I only really have one major comment/suggestion regarding the render API changes, but besides that everything looks solid to me.

younik:
Hello @andrewtanJS, thanks for your review!

I want also to discuss this: currently, VideoRecorder is a wrapper, you wrap an environment and after each step (or more) you call `capture_frame()`. This makes sense with the previous API, but with `collect_render()`, maybe it makes more sense to just have a function that receives a list of frames (the output of `collect_render()`) and save a video/JSON.
I kept the previous implementation of VideoRecorder as a wrapper to change as less as possible in this big PR (and inside it, it calls `collect_render()[-1]` instead of `render()`)

andrewtanJS:
Yeah, keeping the previous implementation for now is sound, and I think it could remain as such for now. I think the function you're suggesting can just be added as a util function, but that can come later; as it is, collect_render should be sufficient for users to perform whatever they need to do with the output.

jkterry1:
A few comments:

-There needs to be an accompanying documentation PR for this (in the gym-docs repo)
-My intention was for this remove the video render wrapper and have it be a standard util that environments could call, and to have the arguments for render be ones that enable human mode, return the rgb array, or return a video and have the video argument use that tooling

norabelrose:
Hi, sorry to barge in here. With this new render API, it doesn't appear that there's any documented/standardized way to 1) turn rendering off or on without creating a new environment object from scratch, or 2) "pop" rendered frames so that they're no longer included in the return value of `collect_render()`.

Is my understanding of this PR correct? If so, would it be relatively easy to fix either or both of these concerns? My specific use case is a Gym-based library for preference-based RL (under construction [here](https://github.com/norabelrose/classroom)). For PbRL it's quite helpful to be able to render snippets of episodes for later evaluation by humans without rendering the entire trajectory. This type of fine-grained rendering is possible in stable gym, and it seems like it would be a considerable step backward in functionality if this became impossible.

I'd be willing to contribute code if necessary to make sure fine-grained rendering is still possible in Gym 1.0. Thank you for your consideration.

younik:
> I'd be willing to contribute code if necessary to make sure fine-grained rendering is still possible in Gym 1.0. Thank you for your consideration.

Hello @norablerose, thanks for your important comment

> Is my understanding of this PR correct? 
Yes, it is correct

> If so, would it be relatively easy to fix either or both of these concerns?
Yes, both changes should not cause any problem

I agree with your considerations, it should just be an enhancement and I don't see major drawbacks.

cc: @jkterry1

vwxyzjn (commented):
Hey @younik thanks for this PR! Would you mind listing out the 1) PR description and 2) example code to test out the new API? These will really help future maintainers and users when they look back to this PR: they can just read what does the PR do instead of looking through the source code.

younik:
> Hey @younik thanks for this PR! Would you mind listing out the 1) PR description and 2) example code to test out the new API? These will really help future maintainers and users when they look back to this PR: they can just read what does the PR do instead of looking through the source code.

Hello @vwxyzjn, sure! I changed the first message

younik:
@norabelrose If we want to discuss more on int, maybe it is better to open an issue

What is the advantage of having both` env.requires_render` and `render` in init?
If someone needs to turn off render for a little piece of code, they can use `with env.no_render()`, in other cases maybe it suffices to use `render=False` and eventually `with env.enable_render()`?

Also, another possibility is to avoid `render` in init and use `render_mode=None`for no rendering.

vwxyzjn:
Hi, @younik thanks so much for updating the first message. I have a follow-up question. How does the new API play with `RecordVideo` wrapper?

younik:
> Hi, @younik thanks so much for updating the first message. I have a follow-up question. How does the new API play with `RecordVideo` wrapper?

Currently I made as little changes as possible and just changed `render()` with `collect_render()[-1]` inside `RecordVideo` wrapper. This means that the user still needs to call `capture_frame()` after each step (or whenever they want).
Nevertheless, we were discussing with @jkterry1 on making `RecordVideo` a util function that accepts a list of frames (like the output of `collect_render()`) and produces save the video/JSON that make more sense with this API.

jkterry1:
@younik can you please fix merge conflicts?

norabelrose:
@younik I'd agree that it would be cleaner to not have an init flag at all and simply use the context managers, but according to Jordan there are some environments where you need to know at initialization time whether you're going to be doing any rendering. Maybe that could resolved with lazy loading or some other approach instead.

If it's ok with everyone I'll make a comment on #2540 instead of creating a new issue for this— it seems like it would be cleaner to keep the discussion in one place.

jkterry1:
@norabelrose 

- The collect render method is intended be a monster function that has behavior wholly defined by the environment. If the environment wants to pop the stack, that's perfectly allowed by the API, that's just not what you'd want to happen for any of the environments built into Gym itself (IMO). The docs will reflect this. This also resolves the concern about passing a no render argument.
- We can't put render in reset and not init or allow for changing render modes on the fly and have the API be general to all environments, though I would like to if it were possible. Some complex simulators used in the robotics etc. community have to be initialized in certain ways on boot to have rendering function in the expected manner. This is why render is in init, and all the other functions we've added as part of API changes have been putting things into reset. Render is special here.

norabelrose:
@jkterry1 Thank you for the feedback.

I understand that changing the render _mode_ on the fly could be problematic for many environments (i.e. from "human" to "rgb_array" or vice versa), but it's not clear to me why it would be a problem to allow simply turning rendering _on_ or _off_ without re-initializing the env. Are we supposed to keep rendering on constantly, incurring the rendering overhead for all time steps, and discard the rendered time steps that we don't want to use? That seems really sub-optimal and could greatly slow down training for a use case like preference-based RL, where you only need to render on the order of ~1% of the time steps. It seems like we could keep render or render_mode as an init flag and still allow an escape hatch to toggle rendering at the time step level, or at least at the trajectory level (i.e. between calls to `reset()`). Perhaps not all environments would even support that, but we could include a metadata field that specifies whether fine grained toggling of rendering is possible or not. What do you think?

jkterry1:
You could just re-initialize the environment (which is normally very cheap), and define collect render to only return the render modes you need right? Or am I missing something?

norabelrose:
@jkterry1 The problem is that re-initializing the environment wouldn't be viable for cases where you want to render one or more selected segments of a potentially long episode, since re-initializing the environment would destroy the environment state. In those cases you would have to render all the frames for an episode and discard all but the ones you actually want— which is what the implementation of `RenderVideo` and `VideoRecorder` is forced to do in this PR.

Also, having to re-initialize the environment for this just seems like a "hack" in general, one which _could_ lead to noticeable performance degradations in cases where initializing the environment is expensive. 

This problem could be fixed, I think, by simply adding a Boolean `rendering` or `requires_render` property to `Env` which can be toggled on or off.

younik:
> Are we supposed to keep rendering on constantly, incurring the rendering overhead for all time steps, and discard the rendered time steps that we don't want to use? That seems really sub-optimal and could greatly slow down training for a use case like preference-based RL, where you only need to render on the order of ~1% of the time steps.

Right now, rendering computation is performed even if you don't call `collect_render()`, so I agree at least to allow disabling of rendering through `requires_render` or `render_mode = None`

younik:
@norabelrose I will add `None` as `render_mode` to all environments; this will be the default option.

norabelrose:
@younik Just to clarify, does that mean `render_mode` would be a settable property on `Env` that could change the rendering mode without creating a new env object?

That would _almost_ be enough to address my concerns (although I still think that at least the built-in envs need to "pop" timesteps off the stack when `collect_render` is called, otherwise we'll waste RAM for no reason). But I'm actually worried that making `render_mode` a settable property goes _too_ far— I think @jkterry1 is probably right that allowing users to toggle between `'human'` and `'rgb_array'` would add unnecessary complexity for many environments. It seems like a simple binary `requires_render` property, by analogy to PyTorch's `requires_grad`, would be best.

younik:
> Just to clarify, does that mean render_mode would be a settable property on Env that could change the rendering mode without creating a new env object?

No, the `render_mode` will still be fixed at initialization; this should solve the problem of performances in training that you pointed out.
Personally, I cannot think of a common use case in which you need to toggle on / off the rendering, is this important?

norabelrose:
@younik Okay, thank you. I still don't really understand what exactly you're planning to do to the API but I guess I can read the updated PR once you commit the changes. As long as it's easy to toggle rendering on and off at the timestep level I'll be happy with that, basically.

Well, by far the most common use case for toggling the rendering on and off is simply the `RecordVideo` wrapper and associated `VideoRecorder` in stable gym, which currently only renders frames when `episode_trigger` or `step_trigger` indicate that this is necessary. In general, being able to toggle rendering on or off is a pretty useful feature for anyone who wants to look at video clips of their agent sampled at a certain interval, or at random intervals, during training without the overhead of rendering every single timestep.

In my case, I kind of need this feature because I'm building a library for preference-based RL which will only render clips of the agent's behavior for later human evaluation, and it's only necessary to render a small fraction (~1%) of all trajectories/timesteps for that purpose. Even episode-level toggling isn't really fine grained enough, since episodes might be long and you want to be able to get roughly i.i.d. clips from all points in the trajectory. I'll admit that preference-based RL isn't very popular right now, but I think it's reasonable to expect that it will get more popular as people increasingly want to apply their RL agents to complex goals that aren't specifiable in terms of a hard-coded reward function.

vwxyzjn:
Thanks for all the work @younik. I agree with @norabelrose that having more fine-grained control would be helpful.

The PR has a lot of moving parts, and I suggest explaining things more explicitly in the PR description. I have a number of questions:

1. How is the new API made backward-compatible?
> by far the most common use case for toggling the rendering on and off is simply the RecordVideo wrapper
2. How does the new API work with the most common use case of using the `RecordVideo` to record agents' gameplay?

Lastly, if we are collecting frames, we can even outsource the video rendering to external libraries such as tensorboard. See [SummaryWriter.add_video](https://pytorch.org/docs/stable/tensorboard.html#torch.utils.tensorboard.writer.SummaryWriter.add_video) and [moviepy](https://zulko.github.io/moviepy/getting_started/videoclips.html)

younik:
> I agree with @norabelrose that having more fine-grained control would be helpful.

Do you suggest adding the attribute `requires_grad`, the with statement or both?

> 1. How is the new API made backward-compatible?

The `render` method is kept for now, with a deprecation warning. This just calls a private `_render` method that is almost identical to the previous `render` one. In the constructor, I set `None` as the default `render_mode`. In this way, legacy codes still work as expected (with no rendering without `render` calls).


> 2. How does the new API work with the most common use case of using the `RecordVideo` to record agents' gameplay?

The RecordVideo should still be done. For now, it just calls `collect_render()[-1]` (with a check on the `render_mode`) instead of `render()`, but as we were discussing, this will be transformed into a util method for the output of `collect_render()`. As you pointed out, this could be easily done with external libraries.

Actually, I just noticed that RecordVideo is not backwards-compatible since you have to specify the `render_mode` to obtain the video. As we plan to deprecate the entire class, I can just unroll the changes, add a deprecation warning and use `_render` instead of `render`.

jkterry1:
@younik can you please also fix conflicts?

gym/envs/box2d/bipedal_walker.py
```diff
@@ -154,9 +154,9 @@ class BipedalWalker(gym.Env, EzPickle):
 
     """
 
-    metadata = {"render_modes": ["human", "rgb_array"], "render_fps": FPS}
+    metadata = {"render_modes": [None, "human", "rgb_array"], "render_fps": FPS}
 
-    def __init__(self, hardcore: bool = False):
+    def __init__(self, render_mode=None, hardcore: bool = False):
```
pseudo-rnd-thoughts, line 159: Could you add a type hint? Possible a literal if we know the exact values?

pseudo-rnd-thoughts:
@younik This looks good, my only note that I dont think anyone else has mentioned yet would be to add type hinting and pydocs for the all of the function. However, I think it could be reasonable to do this separately in another PR.

younik:
@norabelrose We are stepping back on the breaking change and we are not planning anymore to deprecate `.render()`.
I just extended it to allow render mode specification during initialization (see the PR message for more details).
This should solve your concerns since you can use the old workflow with `render_mode = None`.

If you can have a look and comment on this, it is highly appreciated!

norabelrose:
@younik Thanks for your work on this. I looked through the changes and it looks like all my original concerns are resolved.

The one thing I'll say is that it's honestly not clear to me why any of the remaining changes are strictly necessary– it seems like added complexity with little benefit– but it's not a big deal and I wouldn't be upset if this gets merged.

Thanks for listening to my feedback.

gym/envs/box2d/bipedal_walker.py
```diff
@@ -154,12 +155,10 @@ class BipedalWalker(gym.Env, EzPickle):
 
     """
 
-    metadata = {"render_modes": ["human", "rgb_array"], "render_fps": FPS}
+    metadata = {"render_modes": [None, "human", "rgb_array"], "render_fps": FPS}
```
RedTachyon, line 158: This will be just a matter of convention, but do we need to include `None` as a render mode? I think any environment naturally supports a `None` mode, since it basically just says "do whatever, I don't care about rendering".

gym/envs/classic_control/cartpole.py
```diff
@@ -179,94 +186,105 @@ def reset(
...
-        gfxdraw.aapolygon(self.surf, cart_coords, (0, 0, 0))
-        gfxdraw.filled_polygon(self.surf, cart_coords, (0, 0, 0))
-
-        l, r, t, b = (
-            -polewidth / 2,
-            polewidth / 2,
-            polelen - polewidth / 2,
-            -polewidth / 2,
-        )
+        if self.renderer.mode is not None:
```
RedTachyon, line 197: Just wanted to say I actually like this pattern quite a bit, relegating the nasty bits of rendering to the `Renderer`. Maybe we should lean on it a bit more?

gym/utils/renderer.py
```diff
@@ -0,0 +1,25 @@
+from typing import Any, Callable, List, Optional
+
+NO_RETURNS_RENDER = ["human", None]
+
+
+class Renderer:
```
RedTachyon, line 6: This requires a bunch of comments describing what exactly it does, since it seems to have a pretty important role in the whole rendering setup

tests/envs/test_envs.py
```diff
@@ -92,12 +94,13 @@ def test_random_rollout():
 
 def test_env_render_result_is_immutable():
```
RedTachyon, line 95: I know this is an existing text, but can we parametrize it over all envs? I'm not sure which ones support `ansi`, but doing the test for all of them would be nice. (you can probably just explicitly skip envs which don't have `ansi in self.metadata['render_modes']`

tests/utils/test_env_checker.py
```diff
@@ -12,6 +12,9 @@ class ActionDictTestEnv(gym.Env):
     action_space = Dict({"position": Discrete(1), "velocity": Discrete(1)})
     observation_space = Box(low=-1.0, high=2.0, shape=(3,), dtype=np.float32)
 
+    def __init__(self, render_mode=None):
```
RedTachyon, line 15: Type hint

RedTachyon: (I know this might sound a bit nitpicky, especially since this is an internal piece of testing code, but it's helpful when you go back to this code after a 6 months long break and try to figure out what this was supposed to be)

RedTachyon (commented):
So there seems to have been a change to the API which was described on Discord, but not here, but I'll just put it here.

Last we were discussing, there was an explicit argument `single_rgb_array` that can be used in `__init__`, this has been now removed in favor of `render_mode=None` + `env.render(mode='rgb_array)`. I don't think this is a good idea, since in some cases, you might need to set up some explicit machinery for rendering during initialization, even if you're only planning to use `rgb_array`. 

Consider for example an environment which only displays stuff on screen, which you then capture with a virtual display - you need to set this up ahead of time, and then you can call `img = env.render(...)` to get the actual frame. But you might want to run some episodes without setting it up, because you don't actually need the rendering right now - in that case it'd be useful to keep the option to have both `render_mode=None` (which says "don't do anything to support rendering") and `render_mode=single_rgb_array` (which says "get ready to render individual frames)

That being said, maybe shuffling the names around might be in order. Is there actually a strong reason not to use `rgb_array` and `rgb_array_list` instead, for example? This way all the intuitions of the current `rgb_array` can stay the same, and the names also make more sense.

I'd also consider adding a default `Env.__init__` to do some of the rendering boilerplate there, if we can do it in a way that generalizes to all/most environments. That would require people to start doing a `super().__init__(...)`, but we already added that requirement in `reset`, so it might be good to just make it into a habit. That being said, this might be better left for 1.0?

I'm not currently able to go through the changes to all environments in detail, so I'll just assume that the changes are consistent, and that the tests would catch anything out of the ordinary.


Also, one thing I'd like to see either here or in the docs is an example of how to properly handle all of this stuff in a custom environment. We already have an example custom environment creation tutorial, so it could be integrated into that (either as an extension, i.e. another section there, or just included in the middle of the code; I'm leaning towards the former option). Having the `Renderer` class seems pretty handy for that and should be explained well

younik:
> @younik Thanks for your work on this. I looked through the changes and it looks like all my original concerns are resolved.
> 
> The one thing I'll say is that it's honestly not clear to me why any of the remaining changes are strictly necessary– it seems like added complexity with little benefit– but it's not a big deal and I wouldn't be upset if this gets merged.
> 
> Thanks for listening to my feedback.

@norabelrose The idea is to deprecate the _mode_ argument of `.render()` then. With these changes, we make the rendering mode known at initialization and unchangeable on the fly. This is needed for some environments and, after removing completely the _mode_ argument, the code will benefit. 
The problems that you pointed out should not arise since we will maintain single fine-grained renderings with special render_mode (e.g. _single_rgb_array_)

gym/envs/box2d/bipedal_walker.py
younik: I added it to match `assert render_mode in self.metadata["render_modes"]`, but we can change those lines instead.
Also in tests, if you want to try all the combinations, you can do it with `for mode in env.metadata["render_modes"]`.
Are we sure that all environments want to support None as render mode?

Pointed these out, I don't have a strong preference. I agree that repeating it in all environments is not nice.

tests/envs/test_envs.py
younik: Yes, should it be done in another PR?

pseudo-rnd-thoughts: I’m looking at upgrading the environment tests, so I can include it in my PR

younik: @pseudo-rnd-thoughts it makes more sense, thanks

gym/envs/box2d/bipedal_walker.py
younik: I removed None in render_modes and changed asserts

gym/envs/classic_control/cartpole.py
younik: Thanks!

Do you mean moving this code (that is repeated in every environment) inside Renderer?
```
if self.renderer.mode is not None:
    return self.renderer.get_renders()
else:
    return self._render(mode)
```

I avoided it because this would mean to also pass `mode` arg, thus changing the signature after the deprecation. I prefer to keep the `Renderer` interface fixed. What do you think?

Markus28:
@younik Is my understanding correct that this API does not allow users to find out what the render_mode of a given environment instance is? I think that's somewhat problematic because the behavior of the render method becomes completely unpredictable (unless you know how the env was instantiated). For example, in the upcoming human-rendering wrapper we need to check that the render_mode is compatible with the wrapper. But the only way to do it atm is to actually call the render method, which is kinda smelly.

younik:
> @younik Is my understanding correct that this API does not allow users to find out what the render_mode of a given environment instance is? I think that's somewhat problematic because the behavior of the render method becomes completely unpredictable (unless you know how the env was instantiated). For example, in the upcoming human-rendering wrapper we need to check that the render_mode is compatible with the wrapper. But the only way to do it atm is to actually call the render method, which is kinda smelly.

This is a good point; we can force the environment to define the attribute `render_mode`? This could be done with env_checker and explaining it clearly in the documentation

pseudo-rnd-thoughts (changes requested):
This looks generally good, the only thing this is missing is a test that runs all of the environment with all of these rendering modes. Could you add this?

gym/core.py
```diff
@@ -130,42 +179,32 @@ def reset(
         if seed is not None:
             self._np_random, seed = seeding.np_random(seed)
 
-    def render(self, mode="human"):
-        """Renders the environment.
+    # TODO: remove kwarg mode with gym 1.0
+    def render(self, mode="human") -> Optional[Union[RenderFrame, List[RenderFrame]]]:
```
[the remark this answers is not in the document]
younik: @pseudo-rnd-thoughts to maintain backward compatibility, the mode argument is *not* ignored if render_mode is None.
In this way, legacy codes still work, i.e. if you don't specify the render_mode, but you call render you will get human rendering.

younik:
@pseudo-rnd-thoughts thanks for the comments; I added the tests for rendering

DDDOH:
I can't run the example given in [comment](https://github.com/openai/gym/pull/2671#issue-1160674687), and I'm sure my gym version is the latest (0.24.1), the error message is ```FrozenLakeEnv.__init__() got an unexpected keyword argument 'render_mode'```. Will the document get updated so that users can know clearly what's happening with the render API? Thanks

younik:
> I can't run the example given in [comment](https://github.com/openai/gym/pull/2671#issue-1160674687), and I'm sure my gym version is the latest (0.24.1), the error message is `FrozenLakeEnv.__init__() got an unexpected keyword argument 'render_mode'`. Will the document get updated so that users can know clearly what's happening with the render API? Thanks

The new render API is not in the release version yet, if you want to use it, you should install gym from source

DDDOH:
That makes sense. Looking forward to the formal released version.

wookayin:
Why did the authors decide to change the render API -- the removal of argument `render_mode` -- this will break **all** existing codes that would use offline rendering (`render("rgb_array")`). Could you please reconsider adding it back unless the breaking change and removal of the argument is really necessary?

younik:
> Why did the authors decide to change the render API -- the removal of argument `render_mode` -- this will break **all** existing codes that would use offline rendering (`render("rgb_array")`). The core `gym.Env` interface has `mode` (or `render_mode`) as in https://github.com/openai/gym/blob/master/gym/core.py#L185, which shall not be removed until v1.0 after enough deprecation period.

Hello, for a detailed explanation of why this render API was introduced see https://github.com/openai/gym/issues/2540
We are aware this is a breaking change, thus as you pointed out we introduced a deprecation warning that will be there for enough period. For now, the code is backwards compatible. 

> In addition, there seems many bugs around: As discussed in the previous comments and #2825, `render_mode` constructor is not even properly implemented (which is already fixed in the HEAD). Also it returns a **different type/shape** (which I think is wrong) than what would be in the previous versions. (Update: I've filed an issue for this: #2889)

The constructor is properly implemented in source; as I answered in the issue, we mistakenly merge the update documentation before the release, sorry about that. Also, #2889 is the expected behaviour, I will answer your issue soon.

wookayin:
**If the maintainers intended such a breaking changes, this should be done in v1.0 release or after some enough deprecation/announcement period.** I feel release of this new API went a bit hasty without enough discussion. I would argue that we should **keep the backward compatibility** as much as possible. As a minority opinion, with all due respect I don't think the new API is necessarily better --- I understand the purpose and intention discussed since #2540 but they could've been implemented in a different way without hurting the sensible interface design and backward compatibility.

All that said, I do understand that in some cases breaking changes and radical changes are necessary. However, this should be done rather carefully with all other relevant changes together. What other people think For example, it is said (#2891) `render(**kwargs)` are deprecated and will be removed at gym v1.0 --- but what's the point of doing this if you have already broke the API and suddenly changed the behavior in v0.24.0? They could have done together so it can be fool-proof and could prevent all the unwanted use cases by normal users who could never be aware of the change.

younik:
> they could've been implemented in a different way

Even if it had already been merged, I would be sincerely happy to implement a better proposal that will address the problems explained in https://github.com/openai/gym/issues/2540

> we should keep the backward compatibility
> if you have already broke the API

What do you mean? Legacy codes still work (with just a warning), the current changes are backward compatible

wookayin:
The current behavior is *different* from older versions in the following sense. I would appreciate why you would think this is backward compatible. In my view, the changes are "on purpose", so it obviously has broken backward compatibility:

```
>>> env.render('rgb_array')
Wrapper.render() takes 1 positional argument but 2 were given
```

```
>>> env.render(mode='rgb_array')
```
- Now returns a list, as explained/proposed in this PR #2671
- Previously it returns the rendered frame of w.r.t the current state (see #2891)

```
>>> env.render(mode='rgb_array', width=300, height=300)
```
The additional parameters to renderer is at this point buggy (#2891), so I can't tell right now whether it will behave the same, but I'm not sure whether those keyword arguments will be still supported (especially, when no errors are thrown and they are silently ignored).

wookayin:
I looked at the code and figured out what @younik meant by backward compatibility.

```python	
>>> env = gym.make("HalfCheetah-v4")
>>> env.reset();
>>> env.render(mode='rgb_array', width=300, height=300).shape  
(480, 480, 3)
```

The only change is that the env wrapper requires keyword argument (no positional arguments, i.e., `env.render("rgb_array")`). The behavior is backward compatible with the default behavior not having changed. So the conclusion is the new versions of gym would have no a disaster except for a very straightforward signature change.

However, I think `render_mode="rgb_array"` needs to behave exactly same as if the same `mode=...` parameter were given to each `env.render()` call. This would be more consistent and the only reasonable behavior.

younik:
@wookayin sorry, I was answering you, but your message make me realize that the mode as argument was broken and I worked for the fix (https://github.com/openai/gym/pull/2893)

So, all old-style code work as before.

>However, I think render_mode="rgb_array" still needs to behave exactly same as if the same mode=... parameter were given to each env.render() call. This would be more consistent and reasonable behavior; can we consider having a different name for render_mode?

In general, I think we want to encourage the use of `rgb_array` over `single_rgb_array` since not every environment would support it. 
Since one should be aware of the new render API to use it; this should not introduce confusion. We can obviously discuss it; `rgb_array` and `single_rgb_array` vs `multiple_rgb_array` and `rgb_array` doesn't make too much different. 
But as I said, the standard behaviour should be returning a collection of frames.

wookayin:
@younik @pseudo-rnd-thoughts  I still think the new API doesn't make much sense. For the sake of **consistency**, this should have been renamed as something like `rgb_arrays` or `rgb_array_list` when used in the constructor. It is extremely confusing that `env.render(mode="rgb_array")` returns a single RGB frame but `Env(mode="rgb_array").render()` does a list -- **which is even stateful** 🤯.

Any chance to revisit this? This is perhaps the only, last time to have it fixed before the breaking change. Hope this can be addressed before v0.26.0 release (#3056).

PR #2673, opened 2022-03-08
Added event.pump to render

MichaelJ-T:
Added `pygame.event.pump()` to `render()` in lunar_lander.py - to allow pygame to handle internal actions

This action solves window freeze on Windows:
- when trying to move window 
- when window is clicked


jkterry1:
Closing in favor of #2684

Issue #2675, opened 2022-03-09
[Bug Report] `gym.wrappers.RecordVideo` no longer render videos for Atari environments.


vwxyzjn:
`gym.wrappers.RecordVideo` no longer render videos for Atari environments. Try running the following script with `gym==0.23.0` and  `gym==0.22.0`

```python
import gym

env = gym.make("BreakoutNoFrameskip-v4")
env = gym.wrappers.RecordEpisodeStatistics(env)
env = gym.wrappers.RecordVideo(env, f"videos/{gym.__version__}")

env.reset()
for _ in range(2000):
    _, _, done, _ = env.step(env.action_space.sample())
    if done:
        env.reset()
```

the `videos/0.22.0` contains videos but not the `videos/0.23.0`:
![image](https://user-images.githubusercontent.com/5555347/157367532-5bf29b22-8a2c-4b14-9230-35bacfb9db18.png)

vwxyzjn:
I think it's related to https://github.com/openai/gym/commit/35b78ac0ab13a6f91f7ebadad510651f346a6922#diff-90696e71aa0337431a4337e5383a2e082bc301b183de550ae1c088d3aa22a8da: the `render.modes` in the atari envs also needed to be changed to `render_modes`.

vwxyzjn:
I can confirm this works after https://github.com/mgbellemare/Arcade-Learning-Environment/pull/448

![image](https://user-images.githubusercontent.com/5555347/157369057-a0e35e14-c217-483b-a921-1f6662ded7ba.png)

This is also likely a breaking change to existing environments such as pybullet envs. @jkterry1 Maybe it's worth submitting a PR for `RecordVideo` wrapper to be backward-compatible (supporting both `render.modes` and `render_modes`) for the next 6 months?

jkterry1:
Closing due to merged PR

PR #2678, opened 2022-03-09
Make `VideoRecorder` backward-compatible to `gym<0.23`

vwxyzjn:
This PR makes `VideoRecorder` backward-compatible to `gym<0.23`. See #2675.

### DIFF
--- a/gym/wrappers/monitoring/video_recorder.py
+++ b/gym/wrappers/monitoring/video_recorder.py
@@ -36,7 +36,7 @@ class VideoRecorder:
 
     def __init__(self, env, path=None, metadata=None, enabled=True, base_path=None):
         modes = env.metadata.get("render_modes", [])
-
+        
         # backward-compatibility mode:
         backward_compatible_mode = env.metadata.get("render.modes", [])
         if len(modes) == 0 and len(backward_compatible_mode) > 0:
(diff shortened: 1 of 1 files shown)


vwxyzjn:
Hey @RedTachyon could you help review this please?

gym/wrappers/monitoring/video_recorder.py
```diff
@@ -93,6 +103,27 @@ def __init__(self, env, path=None, metadata=None, enabled=True, base_path=None):
...
+        )
+        self.backward_compatible_output_frames_per_sec = env.metadata.get(
+            "video.output_frames_per_second", self.frames_per_sec
+        )
+        if self.frames_per_sec != self.backward_compatible_frames_per_sec:
+            logger.deprecation(
+                '`env.metadata["video.frames_per_second"] is marked as deprecated and will be replaced with `env.metadata["render_fps"]` '
+                "see https://github.com/openai/gym/pull/2654 for more details"
+            )
+            self.frames_per_sec = self.backward_compatible_frames_per_sec
```
witoong623, line 119: These assignment will create a new bug if an user implements only new `render_fps`.
What if we use `get('render_fps')` so that when it is not available, it is `None` and then if the user doesn't have that value, then we go back to use `video.frames_per_second` (which default to 30) and warn the user.

pseudo-rnd-thoughts: This was fixed in https://github.com/openai/gym/pull/2703

Issue #2674, opened 2022-03-08
[Question] Where is `gym.wrappers.Monitor` in `gym==0.23`?


unsignedrant:
### Question

So, it works perfectly fine in `0.22` but not in `0.23` and I don't see any deprecation warning in earlier versions, any suggestions/alternatives?

```
AttributeError: module 'gym.wrappers' has no attribute 'Monitor'
```

balisujohn:
It was removed. I think the intended replacements are the RecordEpisodeStatistics wrapper and the RecordVideo wrapper.

jkterry1:
Before being removed, importing it also raised a depreciation warning.

unsignedrant:
@jkterry1

```
(gymenv) unsignedrant@MacBook-Pro Desktop % pip install gym==0.22
Collecting gym==0.22
  Using cached gym-0.22.0-py3-none-any.whl
Collecting numpy>=1.18.0
  Using cached numpy-1.22.3-cp310-cp310-macosx_10_14_x86_64.whl (17.6 MB)
Collecting gym-notices>=0.0.4
  Using cached gym_notices-0.0.5-py3-none-any.whl (2.7 kB)
Collecting cloudpickle>=1.2.0
  Using cached cloudpickle-2.0.0-py3-none-any.whl (25 kB)
Installing collected packages: gym-notices, numpy, cloudpickle, gym
Successfully installed cloudpickle-2.0.0 gym-0.22.0 gym-notices-0.0.5 numpy-1.22.3
(gymenv) unsignedrant@MacBook-Pro Desktop % python
Python 3.10.2 (main, Feb  2 2022, 06:19:27) [Clang 13.0.0 (clang-1300.0.29.3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from gym.wrappers import Monitor

>>> 
```
No deprecation warning is given or alternatives suggested. So, what should I use instead?

vwxyzjn:
Hey, @unsignedrant we are deprecating `Monitor` in favor of `RecordEpisodeStatistics and `RecordVideo` wrappers. Could you give them a try? Note the `RecordVideo` wrapper is temporarily broken for third party envs (see #2678)

vwxyzjn:
@unsignedrant thanks for giving it a try. Could you try `RecordVideo` with `CartPole-v1`? `RecordVideo` does not work with Pong at the moment because of #2678.

jkterry1:
[...]

PR #2685, opened 2022-03-11
Actually display deprecation warnings

vwxyzjn:
Hey folks, it turns out our deprecation warnings are not shown to the users most of the time. See https://github.com/openai/gym/issues/2674. You can try this out with the following command, which clearly shows the `env.seed()` warning is not being displayed at all. Pretty wild stuff, right? XD It turns out the deprecations are only shown in you run `python -Wall`, which 99% of the users would not do.

```python
(gym-vSAllIMi-py3.9) ➜  gym git:(deprecationwarning) ✗ python
Python 3.9.5 (default, Jul 19 2021, 13:27:26) 
[GCC 10.3.0] on linux
>>> import gym
>>> e = gym.make("CartPole-v1")
>>> e.seed(1)
[1]
>>> exit()
```

This PR actually displays the deprecation warnings (only once so the users are not spammed):

```python
(gym-vSAllIMi-py3.9) ➜  gym git:(deprecationwarning) ✗ python
Python 3.9.5 (default, Jul 19 2021, 13:27:26) 
[GCC 10.3.0] on linux
>>> import gym
>>> e = gym.make("CartPole-v1")
>>> e.seed(1)
/home/costa/Documents/go/src/github.com/vwxyzjn/gym/gym/core.py:172: DeprecationWarning: WARN: Function `env.seed(seed)` is marked as deprecated and will be removed in the future. Please use `env.reset(seed=seed) instead.
  deprecation(
[1]
>>> e.seed(1)
[1]
>>> e.seed(1)
[1]
```

### DIFF
--- a/gym/logger.py
+++ b/gym/logger.py
@@ -13,6 +13,9 @@ DISABLED = 50
 min_level = 30
 
 
+warnings.simplefilter("once", DeprecationWarning)
+
+
 def set_level(level: int) -> None:
     """
     Set logging threshold on current logger.


wookayin:
This has caused a bug in controlling DeprecationWarning (see #3059).

FYI, The deprecation warning mentioned in the above example is gone, the deprecated API has been removed in the latest master branch: 3a8daafc.

PR #2703, opened 2022-03-17
Fixed warnings produced by test_record_video.py and test_video_recorder.py

pseudo-rnd-thoughts:
With https://github.com/openai/gym/pull/2654, test_record_video.py and test_video_recorder.py were throwing new warnings
This was caused by the new render_fps value being different from the default backward compatible value. 

Updated the default backward compatible values to be the up to date parameters


vwxyzjn:
Thank you for this PR. @jkterry1 this looks good to me.

PR #2707, opened 2022-03-19
Reduce pytest warnings to zero

pseudo-rnd-thoughts:
As I noted in #2602, it would be nice if the pytest warnings were thrown as errors so if new changes caused issues
Building upon #2602 that updated the CI and #2703 that fixed warnings (but hasn't been PR yet) 

This fixes all warnings from #2660, mostly through added filterwarnings to prevent the warnings from being thrown


kir0ul:
I'm not sure it's a good idea to transform warnings into errors, I think warnings and errors are two different things and having the flexibility of allowing warnings to pass the CI can sometimes be a good thing.
One concrete example could be that someone is working on a feature and gets a deprecation warning. Since this warning has become an error, CI will not pass, and they'll have to fix the warning to make it pass. But maybe this deprecation warning should be fixed in its own dedicated issue as it could impact other things which are independent of the feature they're working on.
I think we should tend to zero warning, but we should allow warnings to pass the CI in case there's the need for some flexibility.

pseudo-rnd-thoughts:
Thanks for the comment. I agree with you but in your example, couldn't they just create a separate PR to address the DeprecatedWarnings? Do you think we should only be throwing errors for user warnings? 
I included DeprecationWarnings because when I was doing #2660 a majority of the warnings I was getting were DeprecationWarnings and can be the most critical if left unchanged for a long period of time.

pseudo-rnd-thoughts:
Im setting this as a draft because it requires #2602 and #2703 to be merged first

pseudo-rnd-thoughts:
With #2602 and #2703 merged, the PR is ready for review @jkterry1 
Would this PR be of interest? It should help prevent warnings being thrown and not being fixed

pseudo-rnd-thoughts:
Thanks for the feedback @kir0ul, I agree that this is an important PR that we should consider carefully before merging and I think it would be reasonable not to do so. However, I think there are several important advantages to it

> maybe they're working on something unrelated to this warning, which could be fixed in its own specific PR.

I agree with this however, if it is a separate issue, then it would naturally make sense for another PR to make such changes. Or if necessary I believe a force merge could ignore the failing tests

As well, as this commit, sets the number of warnings to zero, then any new warnings that arise should be caused by the changes made by the PR. Meaning that the PR would address their own issues.

This is one of the advantages of stopping warnings early is that the person causing the issue, also has to solve them. Rather than waiting for a while before fixing, by possibly someone new who may misunderstand the original code or the original author who may have forgotten their implementation. 

> let's say we find another bug and get another time 1500 warnings all at once, if they are now converted to errors you wouldn't have been able to fix them in several chunks as you did.

This is fair, but as I explained above, I think that is very unlikely get ~1500 warnings again, due to eliminating them as they occur and that they would only have to address the warnings caused by themselves most likely. 

> warning checkbox in the PR template 

I hadn't noticed that changes,  thanks for doing that, the problem is that the github build actions don't show the number of warnings so it is easy to forget to check if the person has removed warnings. Allowing the number of warnings to increase again, causing the issue above.

Is there something I missing? If this is too racial a change to make then I can understand if we don't want to move forward with this PR

jkterry1:
@pseudo-rnd-thoughts could you please fix merge conflicts?

pseudo-rnd-thoughts:
@kir0ul Talking to @jkterry1, while I still think that the idea of reducing warnings is good, this method of doing it could cause too many issues to be helpful. 
I have removed the warnings to errors part and removed some additional warnings that I found such that the number of warnings should be now 0!

jkterry1:
Hey, I'm closing this per our discussion, please create a new PR with the changes that are still needed

PR #2710, opened 2022-03-21
Hide PyGame window when rendering with mode="rgb_array"

Markus28:
PyGame will currently open a blank window if we are using mode="rgb_array". ~I propose a fix where we keep track of the last mode that was used in the call to `render` and initialize a new display if the mode changes. We can use `flags=pygame.HIDDEN` to hide the display if `mode=="rgb_array"`.~

I have implemented this for LunarLander. I'm not really familiar with PyGame so someone should confirm that this approach actually makes sense. Then I will implement it for the other environments.

### DIFF
--- a/gym/envs/box2d/bipedal_walker.py
+++ b/gym/envs/box2d/bipedal_walker.py
@@ -536,7 +536,7 @@ class BipedalWalker(gym.Env, EzPickle):
         return np.array(state, dtype=np.float32), reward, done, {}
 
     def render(self, mode="human"):
-        if self.screen is None:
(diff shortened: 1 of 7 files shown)


Markus28:
On second thought, why do we need `screen` for `mode="rgb_array"` anyways? I have changed the fix to only use `screen` if `mode` is "human". This has the big advantage that alternating calls with `mode="rgb_array"` and `mode="human"` have less overhead and the window remains responsive.

jkterry1:
Closed per discord chat

PR #2711, opened 2022-03-21
call close method of parent when overriding it in RecordVideo

witoong623:
`gym.Wrapper`'s `close` method calls the environment that it wraps.  However `RecordVideo`, which subclass `Wrapper`, overrode `close` method but didn't call parent `close` method, this can cause resource leak in a environment that relies on `close` method to clean up.

Previously (before https://github.com/openai/gym/commit/850247f888cc4642561488b7c17f8b29873b31bd), `close` method wasn't overridden in `RecordVideo`, so the environment can close properly. This PR add a function call to `close` method of parent class (Wrapper).

### DIFF
--- a/gym/wrappers/record_video.py
+++ b/gym/wrappers/record_video.py
@@ -118,6 +118,7 @@ class RecordVideo(gym.Wrapper):
         self.recorded_frames = 1
 
     def close(self):
+        super().close()
         self.close_video_recorder()
 
     def __del__(self):


witoong623:
@pseudo-rnd-thoughts Since `Wrapper` and `Env` doesn't override `__del__` and call `self.close()` inside, is it appropriate to call `self.close()` in this case?

However, I think I missed to add call to `super().__del__()` because some wrapper classes override it.

pseudo-rnd-thoughts:
Your right to call the super delete function but my suggestion was to just use the close function to ensure that the wrapper closes properly and to limit the repeated code 
```
def __del__(self):
    super().__del__()
    self.close()
```

Wrapper doesn't seem to have a `__del__` function so current code doesn't close the environment and to prevent repeating code then just calling `close` is easier I think

witoong623:
I understand your suggestion but in my opinion a class should only take care of clean up its resource, If the class is a sub class, it shouldn't clean up resource of its super class (unless it is specifically designed to do so), its only responsibility is to maintain behavior of its super class by calling super method, it is the super class's responsibility to decide whether it needs to use `__del__` to clean up resource.

Calling `close` inside `__del__` doesn't conform to behavior of `Env` and `Wrapper` that doesn't use `__del__` to call `close` inside, if `RecordVideo` calls `close` inside `__del__`, it will make `RecordVideo` the only class that exhibit this behavior (apart for environment that sub class `VectorEnv`). So I think it is more important to maintain core API behavior than avoiding repeat code.

I understand your concern about the users don't close the environment properly (whether they forget to call close, or an exception is thrown and there is no try-except to catch it), my custom environment can have devastating effect too if I fail clean up its resource (and I will mitigate it by overriding `__del__` in the environment class itself), but I think it's not my place to decide to add this behavior, I just want to fix the bug and maintain the core API behavior.
PS, I side with avoiding using `__del__` unless I need it)

vwxyzjn:
LGTM @jkterry1.

jkterry1:
@witoong623 thanks a ton!

PR #2712, opened 2022-03-22
Make pygame optional in classic control and toy text envs

### LINKED ISSUE #2691: [Proposal] Make pygame dependency optional for classic control
### Proposal 

If I understand well, pygame is only used for rendering, so it should be possible to install gym and use classic control environments without installing pygame.

### Motivation

Fewer dependencies are always nice.
Classic control envs are usually used to debug things/try things quickly without necessary the need to render them.
This would also prevent unexpected/inconsistent behavior, as in https://github.com/openai/gym/issues/2634 https://github.com/openai/gym/issues/2641 or https://github.com/openai/gym/issues/2642 (where previously a `pip install gym` would have sufficed).

### Pitch

Make pygame dependency optional, throw an error if the user wants to render and pygame is not installed.

### Alternatives

Do nothing.

### Checklist

RedTachyon:
Closes #2691

Black seems to have made some formatting changes, so we'll see if CI flags that. Worst case I'll wait for the pre-commit PR to go through


pseudo-rnd-thoughts:
Does this fix this issue? https://github.com/mgbellemare/Arcade-Learning-Environment/issues/450

JesseFarebro:
@pseudo-rnd-thoughts no this won't fix the issue in the ALE. You'll have to wait for an updated release of ale-py. For the time being you can use the `render_mode` keyword argument. See https://github.com/mgbellemare/Arcade-Learning-Environment/issues/450 for more info.

gym/envs/classic_control/acrobot.py
```diff
@@ -267,6 +265,9 @@ def _dsdt(self, s_augmented):
         return (dtheta1, dtheta2, ddtheta1, ddtheta2, 0.0)
 
     def render(self, mode="human"):
+        import pygame
```
andrewtanJS, line 268: I think we could instead shift the imports into the if clause: `if self.screen is None:` so we don't need to call the import statement every time we call `render()`

gym/envs/classic_control/acrobot.py
```diff
@@ -340,6 +341,8 @@ def render(self, mode="human"):
 
     def close(self):
         if self.screen is not None:
+            import pygame
```
andrewtanJS, line 344: I think this shouldn't be necessary? if `self.screen is not None`, then pygame would already be imported through `render()`

andrewtanJS (commented):
Overall this is a good change, but I think there are some minor adjustments that will improve it.

gym/envs/classic_control/acrobot.py
RedTachyon: In terms of actually running the code, it's not necessary, but any static code checker will freak out if this is not present

gym/envs/classic_control/acrobot.py
RedTachyon: The same comment about static code checkers applies here. The repeated import shouldn't be a performance problem fortunately, python is smart enough to avoid actually doing the whole import again.

PR #2713, opened 2022-03-23
add pygame GUI render for taxi env

younik:
I added pygame GUI render for Taxi env with `mode="human"`

Result:
Screenshot from 2022-03-23 10-22-49


gym/envs/toy_text/taxi.py
```diff
@@ -1,11 +1,13 @@
 import sys
 from contextlib import closing
 from io import StringIO
+from os import path
 from typing import Optional
 
 import numpy as np
 from gym import Env, spaces, utils
 from gym.envs.toy_text.utils import categorical_sample
+import pygame
```
andrewtanJS, line 10: I think we could defer the pygame import to the `_render_gui()` method, so that #2691 is preemptively solved (similar to #2712)

gym/envs/toy_text/taxi.py
```diff
@@ -215,10 +227,139 @@ def reset(
...
+        if self.background_img is None:
+            file_name = path.join(path.dirname(__file__), "img/taxi_background.png")
+            self.background_img = pygame.transform.scale(
+                pygame.image.load(file_name), self.cell_size
+            )
+
+        desc = self.desc
+        background = pygame.Surface((WINDOW_SIZE[0], WINDOW_SIZE[1]))
+        background.fill((20, 24, 35))
+        self.window.blit(background, background.get_rect())
```
andrewtanJS, line 293: Do we need to initialize another surface for the background, or can we just do `self.window.fill((20, 24, 35))` to begin with?

andrewtanJS (commented):
Overall looks good, and I only have a few minor comments

Markus28 (changes requested):
- It seems that the framerate is not locked (see other PyGame envs how to do that)
- We should probably have a `close` method as in the other PyGame environments

Some remarks that are not really that are not that important/maybe not worth fixing:
- The destination can occasionally block out the taxi
- The way the destination is displayed suggests that it also presents an obstacle if it is at one of the bottom locations. But actually, the taxi can "move right through it". Idk if it is actually possible or worth fixing, but I think one might avoid misunderstandings if one could manage to make the destination always line up with a median.
- Might make sense to reset `self.taxi_orientation` in `reset` but that doesn't really matter I guess

younik:
Hello @Markus28, thanks for your review!

I integrated points 1, 2 and 5.

> The destination can occasionally block out the taxi

This problem is only graphically, as pointed out in point 4th, the taxi can pass through it.

> The way the destination is displayed suggests that it also presents an obstacle if it is at one of the bottom locations. But actually, the taxi can "move right through it". Idk if it is actually possible or worth fixing, but I think one might avoid misunderstandings if one could manage to make the destination always line up with a median.

I would like to make it nicer, but I have no idea how to do it (without having complicated code/particular cases) with the current hotel image.

trigaten:
From what I understand, this looks good. I second the confusion with the Taxi being hidden behind the building/it being an obstacle. Maybe reduce the building alpha?

younik:
@trigaten 

Reducing the building alpha, the result is like this:
Screenshot from 2022-03-27 21-03-23

We need something to specify which one is the destination, this is the aim of the building. Moreover, the taxi can stay in the destination tile.

trigaten:
@younik I like the reduced alpha. It also sort of shows that the taxi can pass through. Maybe try reducing a bit more?

younik:
> @younik I like the reduced alpha. It also sort of shows that the taxi can pass through. Maybe try reducing a bit more?

I pushed the version with alpha
taxi

younik:
Is the pre-commit failure my fault? I run it locally and it is ok

Markus28:
Probably this: 
https://github.com/psf/black/issues/2964#issue-1183788875

pseudo-rnd-thoughts:
This PR fixed this issue now I believe, https://github.com/openai/gym/pull/2724, so we just need to rerun the pre-commit

PR #2714, opened 2022-03-23
make error message on accessing private attributes more representative

H4sh3:
This error message was kinda misleading.
The code is checking if a private attributes is accessed and then raises an error.


RedTachyon:
Makes sense, can be merged imo

There's a different thing to consider whether we actually want to strictly forbid accessing private attributes, but that's a discussion for another issue.

PR #2717, opened 2022-03-24
Update check_env

Markus28:
I have updated two aspects of `check_env`
- I believe that it should only complain about missing `"render_fps"` data if human-mode is actually available. Moreover, I believe that "may not occur at inconsistent fps" should be "may ~not~ occur at inconsistent fps".
- Currently we do not check for `return_info`. I have added a function that checks for the existence of `return_info` parameter in `reset` (or something like `**kwargs`), runs `env.reset(return_info=True)` and makes sure that the result makes sense.

### DIFF
--- a/gym/utils/env_checker.py
+++ b/gym/utils/env_checker.py
@@ -255,15 +255,6 @@ def _check_render(
     :param headless: Whether to disable render modes
         that require a graphical interface. False by default.
     """
-    render_fps = env.metadata.get("render_fps")
-    if render_fps is None:
-        if warn:
-            logger.warn(
-                "No render fps was declared in the environment "
-                " (env.metadata['render_fps'] is None or not defined), "
-                "rendering may not occur at inconsistent fps"
-            )
-
     render_modes = env.metadata.get("render_modes")
     if render_modes is None:
         if warn:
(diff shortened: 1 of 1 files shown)


RedTachyon:
I'm not sure if I like the `render_fps` change. Even without human mode, it's useful to have the FPS for `rgb_array` rendering, so that it can be later converted into a video. So I'd revert that change. 

Warning message change is fine.

`return_info` update is also fine

Markus28:
That's an excellent point I haven't thought of, especially considering `RecordVideo`.
But I would still argue that a missing `render_fps` entry doesn't need to be flagged as a problem if rendering isn't implemented at all. Especially because this will be the case for many/most custom environments. 

What do you think about this alternative:

```
if render_fps is None and render_modes is not None and len(render_modes)>0:
```

The only problem with this approach is that it assumes that the user will actually list the render modes in metadata (but if they don't, `check_env` will also issue a warning about that). 

But I can also see the appeal of enforcing the use of `render_fps` in all cases if you'd prefer that :)

RedTachyon:
Yea, I think it's fine not to include this warning if `render` itself is not implemented (as indicated by the metadata)

trigaten:
@jkterry1 render_fps check looks fine to me

PR #2719, opened 2022-03-25
Pitch: Wrapper that adds human-mode rendering to any environment with rgb_array-rendering

Markus28:
This PR is supposed to provide a wrapper that adds human-rendering functionality to environments that only support rgb_array rendering out of the box

---
### Pitch
Oftentimes I have found myself in a situation where I had implemented an environment that could produce RGB images, but that could not actually render the environment in real-time. It would be convenient to have a wrapper that adds this functionality. It's not that hard to add this feature yourself, but if you don't use PyGame for rendering in the first place, it may be *a lot* of boilerplate code.

---

If someone thinks that this might be a good idea, I will start adding documentation and tests if necessary. Not sure whether I'm missing something, but this seems like a wrapper that could be quite useful (at least I will be using it personally).


jkterry1:
On hold until new rendering API is merged, also has merge conflicts

pseudo-rnd-thoughts:
Could you add some tests for this?

pseudo-rnd-thoughts (changes requested):
It generally looks good, could add a some tests? More to check that the code runs than to check the render output

gym/__init__.py
```diff
@@ -1,23 +1,25 @@
 """Root __init__ of the gym module setting the __all__ of gym modules."""
 # isort: skip_file
+import os
+
+os.environ["PYGAME_HIDE_SUPPORT_PROMPT"] = "hide"
```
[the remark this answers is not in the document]
Markus28: It was already there, this will hide a welcome message from PyGame. I moved it up because `from gym.envs import make, spec, register` will import gym.wrappers, which will import PyGame in human_rendering.py. So we need to set the environment variable before importing `gym.envs`. The `noqa`s are then required because flake8 will complain otherwise. But I guess this will all be solved by moving PyGame into the Wrapper.

Markus28:
I'm no longer sure whether using a wrapper is actually the best approach. It would have worked well with the old API but the new API leads to two new restrictions:

- We cannot change the render mode to something else when initializing the wrapper (i.e. the wrapper can only use human-rendering)
- The wrapped environment must use rgb_array rendering

Maybe it might be more convenient if we somehow include the functionality in `gym.utils.renderer.Renderer`. People who use `Renderer` wouldn't have to worry about wrapping the environment and the two issues above would be solved.

gym/__init__.py
```diff
@@ -1,23 +1,25 @@
...
 
-from gym import error
-from gym.version import VERSION as __version__
+from gym import error  # noqa: E402
```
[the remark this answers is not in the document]
Markus28: See the other comment

pseudo-rnd-thoughts:
Would you be able to add some tests for this as it affects a fairly critical part of the code
This may be easier once https://github.com/openai/gym/pull/2900 is merged

tests/wrappers/test_human_rendering.py
```diff
@@ -0,0 +1,26 @@
...
+        for _ in range(75):
+            _, _, done, _ = env.step(env.action_space.sample())
+            if done:
+                env.reset()
+
+        env.close()
+
+    with pytest.raises(
+        AssertionError,
+        match="Expected env.render_mode to be one of 'rgb_array' or 'single_rgb_array' but got 'human'",
```
pseudo-rnd-thoughts, line 24: Wrap with `re.escape("...")` to avoid regex issues

gym/envs/registration.py
```diff
@@ -589,15 +589,48 @@ def make(
...
+            )
+
+    try:
+        env = env_creator(**_kwargs)
+    except TypeError as e:
+        if str(e).find("got an unexpected keyword argument 'render_mode'"):
+            raise AttributeError(
+                f"You passed render_mode='human' although {id} doesn't implement human-rendering natively. "
+                "Gym tried to apply the HumanRendering wrapper but it looks like your environment is using the old "
+                "rendering API, which is not supported by the HumanRendering wrapper."
```
pseudo-rnd-thoughts, line 627: Would we want to include the actual error, just in case it is not this error?

gym/wrappers/human_rendering.py
```diff
@@ -0,0 +1,117 @@
...
+
+        Args:
+            env: The environment that is being wrapped
+        """
+        super().__init__(env)
+        assert env.render_mode in [
+            "single_rgb_array",
+            "rgb_array",
+        ], f"Expected env.render_mode to be one of 'rgb_array' or 'single_rgb_array' but got '{env.render_mode}'"
+        assert (
```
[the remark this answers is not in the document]
Markus28: We use it in the _render_frame method in self.clock.tick

gym/envs/registration.py
Markus28: I think it should appear in the stacktrace

pseudo-rnd-thoughts: Fair then ignore my comment

Markus28:
Thanks, I changed both things :)

Issue #2680, opened 2022-03-09
[Bug Report] VectorEnv action space seed problem


vwxyzjn:
**Describe the bug**
`envs.action_space.sample()` is not reproducible after seeds.

**Code example**
```python
import gym
import numpy as np
def make_env(env_id, seed):
    def thunk():
        env = gym.make(env_id)
        env.seed(seed)
        env.action_space.seed(seed)
        env.observation_space.seed(seed)
        return env

    return thunk
seed = 1
envs = gym.vector.SyncVectorEnv([make_env("CartPole-v1", seed)])
sampled_actions = []
for _ in range(20):
    sampled_actions += [envs.action_space.sample()]
print(np.array(sampled_actions).flatten())
```
```
(gym-vSAllIMi-py3.9) ➜  gym git:(master) ✗ python w.py
[1 0 0 1 0 1 1 0 1 1 1 1 1 0 0 1 0 1 0 0]
(gym-vSAllIMi-py3.9) ➜  gym git:(master) ✗ python w.py
[0 1 1 1 0 0 1 0 1 0 0 0 1 1 0 0 1 1 1 1]
(gym-vSAllIMi-py3.9) ➜  gym git:(master) ✗ python w.py
[0 1 1 0 1 1 1 1 1 0 1 0 0 1 1 1 1 0 1 1]
(gym-vSAllIMi-py3.9) ➜  gym git:(master) ✗ python w.py
[0 0 0 1 1 1 1 1 0 1 1 0 1 1 0 1 0 0 0 0]
(gym-vSAllIMi-py3.9) ➜  gym git:(master) ✗ python w.py
[0 1 0 1 0 0 0 0 1 0 1 1 0 1 0 1 1 1 1 0]
```

I think the problem is https://github.com/openai/gym/pull/2280 batches a new unseeded action space... So if I seed this action space again it will work:

```python
import gym
import numpy as np
def make_env(env_id, seed):
    def thunk():
        env = gym.make(env_id)
        env.seed(seed)
        env.action_space.seed(seed)
        env.observation_space.seed(seed)
        return env

    return thunk
seed = 1
envs = gym.vector.SyncVectorEnv([make_env("CartPole-v1", seed)])
envs.action_space.seed(seed)
envs.observation_space.seed(seed)
sampled_actions = []
for _ in range(20):
    sampled_actions += [envs.action_space.sample()]
print(np.array(sampled_actions).flatten())
```
```bash
[...]

PR #2727, opened 2022-04-02
Fixed batch spaces where the original space's seed was ignored. Issue 2680

pseudo-rnd-thoughts:
Provides a fix for issue 2680 where the SyncVectorEnv ignores the action space seed

This required changes to the batch_space function for each of the spaces to pass the seed for the batched space
The space constructor seed parameter allows a seeding.RandomNumberGenerator that ignores the seed function if passed


tests/vector/test_spaces.py
```diff
@@ -95,6 +95,32 @@ def test_batch_space(space, expected_batch_space_4):
     assert batch_space_4 == expected_batch_space_4
 
 
+def test_batch_space_seed():
+    for space in [
```
lebrice, line 99: This seems unrelated to the rest of the test, should probably be a different test, and use a parametrized `space` argument.

tests/vector/test_spaces.py
```diff
@@ -95,6 +95,32 @@ def test_batch_space(space, expected_batch_space_4):
...
+        Box(0, 10, seed=123),
+        Tuple([Box(0, 5, seed=123), Box(0, 3, seed=123)], seed=123),
+        Dict(
+            {"space-1": Box(0, 5, seed=123), "space-2": Box(0, 10, seed=123)}, seed=123
+        ),
+    ]:
+        batched_space = batch_space(space)
+        assert space.np_random == batched_space.np_random
+        for _ in range(100):
+            assert np.all(space.sample() == batched_space.sample())
```
lebrice, line 109: Wait a minute, the compared spaces won't have the same number of dimensions, so this equality check uses broadcasting, right? Does this still work when comparing dicts? If so, that's something I wasn't aware of!

tests/vector/test_spaces.py
```diff
@@ -95,6 +95,32 @@ def test_batch_space(space, expected_batch_space_4):
...
+
+    space = MultiDiscrete([5, 3], seed=123)
+    batched_space = batch_space(space)
+    assert space.np_random == batched_space.np_random
```
lebrice, line 113: Would perhaps be a good idea to check that the rng state isn't being replicated at other indices in the batched space, when `n > 1`. You wouldn't want the batched space to just return `n` identical samples, right?

gym/spaces/dict.py
```diff
@@ -134,9 +136,9 @@ def __len__(self) -> int:
 
     def __repr__(self) -> str:
         return (
-            "Dict("
-            + ", ".join([str(k) + ":" + str(s) for k, s in self.spaces.items()])
-            + ")"
+            "Dict({"
+            + ", ".join([str(k) + ": " + str(s) for k, s in self.spaces.items()])
```
lebrice, line 140: I'd suggest removing such changes, since they don't really have anything to do with the issue you're solving

gym/spaces/box.py
```diff
@@ -56,6 +57,8 @@ def __init__(
             shape = low.shape  # type: ignore
         elif not np.isscalar(high):
             shape = high.shape  # type: ignore
+        elif np.isscalar(low) and np.isscalar(high):
```
lebrice, line 60: This makes sense, but doesn't belong in this PR

gym/spaces/box.py
```diff
@@ -44,7 +45,7 @@ def __init__(
         high: Union[SupportsFloat, np.ndarray],
         shape: Optional[Sequence[int]] = None,
         dtype: Type = np.float32,
-        seed: Optional[int] = None,
+        seed: Optional[int | seeding.RandomNumberGenerator] = None,
```
lebrice, line 48: Isn't this a bit too strict of a requirement? Couldn't one pass a `np.random.Generator` here?

gym/spaces/multi_discrete.py
```diff
@@ -31,7 +33,12 @@ class MultiDiscrete(Space[np.ndarray]):
 
     """
 
-    def __init__(self, nvec: list[int], dtype=np.int64, seed=None):
+    def __init__(
+        self,
+        nvec: list[int] | np.ndarray,
```
lebrice, line 38: Unrelated change:
```suggestion
        nvec: list[int],
```

gym/vector/utils/spaces.py
```diff
@@ -88,13 +108,14 @@ def _batch_space_dict(space, n=1):
...
-        )
+        ),
+        seed=copy(space.np_random),
     )
 
 
 @batch_space.register(Space)
 def _batch_space_custom(space, n=1):
-    return Tuple(tuple(space for _ in range(n)))
+    return Tuple(tuple(space for _ in range(n)), seed=copy(space.np_random))
```
lebrice, line 118: I believe this change here introduces a subtle bug:
Each subspace is using the a Generator with the same state for each subspace, which I think will lead to each space giving back identical samples.
Correct me if I'm wrong about this, (preferably with a unit test :wink: )!

tests/vector/test_sync_vector_env.py
```diff
@@ -167,3 +167,17 @@ def test_custom_space_sync_vector_env():
         "step(action-5)",
         "step(action-7)",
     )
+
+
+def test_sync_vector_env_seed():
+    env = make_env("BipedalWalker-v3", seed=123)()
```
lebrice, line 173: I suggest parametrizing the env name to get some different types of spaces in.
There's also some space fixture you can reuse from the vectorenv tests, so you also cover batching of custom spaces as well.

tests/vector/utils.py
```diff
@@ -109,6 +109,7 @@ def step(self, action):
 def make_env(env_name, seed):
     def _make():
         env = gym.make(env_name)
+        env.action_space.seed(seed)
```
lebrice, line 112: Why seed the action space, but not the observation space?

lebrice (commented):
Interesting stuff @pseudo-rnd-thoughts !

I left some comments here, hopefully they are useful to you.
I'm not a "real" contributor to Gym, so I'm not approving this PR or anything, just commenting.

Have a good one!

tests/vector/test_spaces.py
pseudo-rnd-thoughts: Thanks for the comments

I didn't use a parameterized pytest because this test is not possible for all of the spaces (Discrete and MultiDiscrete) as there randomness is different then we can't check the values produced are the same. 
So to test all of the Spaces in one functions, this was the only way.

I agree, if we could find a way that the batched spaces produced the same random values as the original space then we should change to parameterized. Does that make sense or do you think we should still change it?

tests/vector/test_spaces.py
pseudo-rnd-thoughts: I only did this testing for n=1 as then the sample values should have the same size, though sometimes with an additional dimension (numpy can deal with that fine). 
I haven't done any testing to check if the equivalence check if ok when you have n>1

tests/vector/test_spaces.py
pseudo-rnd-thoughts:
```
space = MultiDiscrete([5, 3], seed=123)
batched_space = batch_space(space, n=3)
print(f'{space.sample()=}')
print(f'{batched_space.sample()=}')
print(f'{space.sample()=}')
print(f'{batched_space.sample()=}')

space.sample()=array([3, 0])
batched_space.sample()=array([[3 0]
 [1 0]
 [0 2]])
space.sample()=array([1, 0])
batched_space.sample()=array([[4, 0],
       [4, 2],
       [2, 0]])
```
I dont totally understand your point, is this experiment what you mean?

gym/spaces/dict.py
pseudo-rnd-thoughts: Very true, I will remove it, I planning on creating a PR with typing for batched_space where it means more sense there

gym/spaces/box.py
pseudo-rnd-thoughts: True, I will add it to another PR

gym/spaces/box.py
pseudo-rnd-thoughts: As seeding.RandomNumberGenerator is a subclass of np.random.Generator and in the future, seeding.RandomNumberGenerator is planned to be removed (there isn't a planned date but I think Gym 1.0 would be the plan).
I can reduce the strictness but this is the current gym randomness method

gym/spaces/multi_discrete.py
pseudo-rnd-thoughts: Thanks, removed

tests/vector/test_sync_vector_env.py
pseudo-rnd-thoughts: Part of the problem, related to the testing batch space seeds is that the batched space of Discrete and MultiDiscrete do not produce the same randomness as the original space. For Discrete, MultiDiscrete can be modified to produce the same randomness but that didn't feel part of the original PR. 
This means that we can only test this PR with certain environments (i.e. ones without Discrete or MultiDiscrete action spaces). 
The previous test (test_batched_space_seed) should test that seeding works for all of the spaces and that sample works. So I dont see much point testing more here. Am I missing something?

tests/vector/utils.py
pseudo-rnd-thoughts: As we are testing that the SyncVectorEnv uses a seeded action space as the one passed by the environment functions then we only need to ensure that the action space is seeded correctly. 
I dont know any code that uses the randomness in the observation space so I haven't set it

gym/vector/utils/spaces.py
pseudo-rnd-thoughts: This is a tricky one, but I dont think so. Looking at the Tuple code, then the seed really doesn't do anything except if you are trying to set a new seed
As the Tuple uses the same space and the Tuple sample uses the subspace sample iteratively which updates each time, Tuple shouldn't produce the same random samples.
Im adding Space to the batch space tests

tests/vector/test_spaces.py
lebrice: I'd parametrize with only a few spaces first then. For loops like these don't belong in tests imo.

gym/vector/utils/spaces.py
lebrice: If you're only doing a shallow copy, and all the individual subspaces will actually share the same Generator, then yeah, as you describe (I think),  they will all produce different samples. However, that's quite weird behaviour IMO. I believe all the subspaces should be independently sampled, and each of them should produce different results. It's unclear to me how that could be achieved with a Generator, but something I'd do would probably look like this with `np.random.RandomState`:
```python
def batch_space(space: Space, n: int) -> Space:
    subspaces = [copy.deepcopy(space) for _ in range(n)]
    if space.rng:
        rng: np.random.RandomState = space.rng
        # Create n random states from the original random state.
        seeds = space.rng.integers(0, 1e8, n)
        for subspace, seed in zip(subspaces, seeds):
            space.rng = np.random.RandomState(seed)
    return spaces.Tuple(subspaces)
```

tests/vector/test_sync_vector_env.py
lebrice: > Part of the problem, related to the testing batch space seeds is that the batched space of Discrete and MultiDiscrete do not produce the same randomness as the original space

Yes, but that's perfectly ok. What you need to test is that the randomness of the batched space is deterministic, given the initial seed.

Something like:

```python
@pytest.mark.parametrize("space", <lots_of_diverse_spaces>)
@pytest.mark.parametrize("n", [1, 2, 5])
@pytest.mark.parametrize("base_seed", [123, 456])
def test_deterministic(space: Space, n: int, base_seed: int):
    space_a = space
    space_a.seed(base_seed)
    space_b =copy.deepcopy(space_a)
    
    # Use the rng of space_a, just to be able to check below that space_a_batched
    # doesn't depend on the original rng of space_a:
    space_a.sample()
    
    space_a_batched = batch_space(space_a, n)
    space_b_batched = batch_space(space_b, n)
    
    assert space_a_batched == space_b_batched
    # Note: Use something from https://numpy.org/doc/stable/reference/routines.testing.html
    # to compare deeply nested objects/dicts.
    assert space_a_batched.sample() == space_b_sample.sample()
```

>This means that we can only test this PR with certain environments (i.e. ones without Discrete or MultiDiscrete action spaces).

I think that's Incorrect. See above, plus, nested / deeply nested dicts, tuples, and custom spaces also need to be tested.

> The previous test (test_batched_space_seed) should test that seeding works for all of the spaces and that sample works. So I dont see much point testing more here. Am I missing something?

Yes:
- I believe there should also be a test to assert that the RNG from the original (unbatched) space isn't shared with the batched space. (see the comment in the snippet above)

tests/vector/utils.py
lebrice: Fair enough. I guess it could be considered slightly out-of-scope for this PR, even though it makes a lot of sense, in my opinion.

tests/vector/test_spaces.py
lebrice: Just to make sure I understand: Is your goal with this PR to make `stack([space.sample() for _ in range(5)]) == `batch_space(space, 5).sample()` ?
If so, that's neat! I suggest you test it like this, and with dicts and weird spaces as well.
The functions in here might be helpful for the nested spaces: https://numpy.org/doc/stable/reference/routines.testing.html

tests/vector/test_spaces.py
lebrice: > we can't check the values produced are the same.

See other thread for a more detailed response

tests/vector/test_spaces.py
lebrice: I suggest you use the `iterate` utility function (from [here](https://github.com/openai/gym/blob/e11231d84fd8b1386455709d77051addf8ded095/gym/vector/utils/spaces.py#L101)) to iterate over the batched sample, and check that all items are not the same. Something like:

```python
@pytest.mark.parametrize("space", <lots_of_diverse_spaces>)
@pytest.mark.parametrize("n", [1, 2, 5])
@pytest.mark.parametrize("base_seed", [123, 456])
def test_rng_different_at_each_index(space: Space, n: int, base_seed: int):
    space.seed(base_seed)
    batched_space = batch_space(space, n)
    batched_sample = batched_space.sample()
    slices = list(iterate(batched_space, batched_sample)
    assert not all(slice == slices[0] for slice in slices)
```

tests/vector/test_sync_vector_env.py
lebrice: Also, forgot to include it in my snippet above, but it would also be important to check explicitly that different seeds give different results in the batched spaces too.

tests/vector/test_spaces.py
pseudo-rnd-thoughts: Sadly not really, that I don't believe is possible due to the random number generator being updated every time that sample is called meaning that for all batched space if n > 1 then the samples are not equivalent. The PR is sadly quite a bit simpler, it is that the batched space currently ignores the original space's random number seed and this PR fixes that

tests/vector/test_spaces.py
pseudo-rnd-thoughts: Your idea of checking that the random number generator doesn't generate the same values for n > 1. I will add that

gym/vector/utils/spaces.py
pseudo-rnd-thoughts: There are two possible issues
1. the original space and the batched space random number generator are the same object meaning that sampling one space effects the other
2. The subspaces of the batched space produce the same random values on each sample. i.e. space.sample = [1], batched_space.sample = [1, 1, 1]

So I believe that I have solved both issues
For issue 1, I have made a new update where we just do a deepcopy of the original space to prevent the two spaces sharing the same random number generator
 ```
def batch_space(space: Space, n: int = 1) -> Space:
    # Without deepcopy, then the space.np_random is batched_space.spaces[0].np_random
    # Which is an issue if you are sampling actions of both the original space and the batched space
    space_copy = deepcopy(space)
    return Tuple(tuple(space_copy for _ in range(n)), seed=copy(space.np_random))
```

For issue 2, this issue is more complex but this issue has never been an issue due to the way that the tuple sample function works. As the subspaces all share the same random number generator, as each subspace is the same object, and as the tuple sample function is `return tuple(space.sample() for space in self.spaces)`. Then there isn't an issue as the space random number generator is updated each time sample is called. Meaning that the sample always produce different values

Does that address your issue or is it different?

**Comment by pseudo-rnd-thoughts** [comment]
@lebrice I think I have made all of the changes that you requested, could you review again please?

**Comment by jkterry1** [comment]
@pseudo-rnd-thoughts can you please fix tests?

**Comment by jkterry1** [comment]
merge conflicts*

**Comment by pseudo-rnd-thoughts** [comment]
@jkterry1 I have fixed the merge conflicts and passed all of the tests. 

However, running the tests locally (on mac) then I'm getting an error with `tests/vector/test_shared_memory::test_read_from_shared_memory` for all the spaces. 
But the problem is that Im getting the problem on the master branch so I dont think that I have introduced an issue. Do you want me to solve this issue in this PR or a separate one?

File: gym/spaces/box.py (hunk 6 above)
Comment (lebrice) [reply gym/spaces/box.py:50]: Why not allow a `np.random.RandomState`?

File: gym/spaces/dict.py [hunk 11]
```diff
@@ -44,7 +46,7 @@ class Dict(Space[TypingDict[str, Space]], Mapping):
     def __init__(
         self,
         spaces: dict[str, Space] | None = None,
-        seed: dict | int | None = None,
+        seed: Optional[int | seeding.RandomNumberGenerator] = None,
```
lebrice, line 49: Why remove the option of passing a dict of seeds here?

gym/spaces/tuple.py
```diff
@@ -16,7 +17,9 @@ class Tuple(Space[tuple], Sequence):
     """
 
     def __init__(
-        self, spaces: Iterable[Space], seed: Optional[Union[int, List[int]]] = None
+        self,
+        spaces: Iterable[Space],
+        seed: Optional[Union[int, seeding.RandomNumberGenerator]] = None,
```
lebrice, line 22: Why remove the option of passing a sequence of ints?
```suggestion
        seed: Optional[Union[int, Sequence[int], seeding.RandomNumberGenerator]] = None,
```

gym/utils/seeding.py
```diff
@@ -23,6 +23,12 @@ def np_random(seed: Optional[int] = None) -> Tuple["RandomNumberGenerator", Any]
 # TODO: Remove this class and make it alias to `Generator` in a future Gym release
 # RandomNumberGenerator = np.random.Generator
 class RandomNumberGenerator(np.random.Generator):
+    def __eq__(self, other):
```
lebrice, line 26: Q: Does `np.random.Generator` implement `__eq__`? If so, then shouldn't it also be called here?

gym/vector/utils/spaces.py
```diff
@@ -47,37 +48,56 @@ def batch_space(space, n=1):
 def _batch_space_box(space, n=1):
     repeats = tuple([n] + [1] * space.low.ndim)
     low, high = np.tile(space.low, repeats), np.tile(space.high, repeats)
-    return Box(low=low, high=high, dtype=space.dtype)
+    return Box(low=low, high=high, dtype=space.dtype, seed=copy(space.np_random))
```
lebrice, line 51: IDK exactly what the difference in behaviour is between `copy` and `deepcopy` in the case of a `np.random.Generator`, but I'd use `deepcopy` if I were you, just to be "safe".
```suggestion
    return Box(low=low, high=high, dtype=space.dtype, seed=deepcopy(space.np_random))
```

gym/vector/utils/spaces.py
```diff
@@ -47,37 +48,56 @@ def batch_space(space, n=1):
...
 
 
 @batch_space.register(Discrete)
 def _batch_space_discrete(space, n=1):
     if space.start == 0:
-        return MultiDiscrete(np.full((n,), space.n, dtype=space.dtype))
+        return MultiDiscrete(
+            np.full((n,), space.n, dtype=space.dtype),
+            dtype=space.dtype,
+            seed=copy(space.np_random),
```
lebrice, line 60: Here too
```suggestion
            seed=deepcopy(space.np_random),
```

gym/vector/utils/spaces.py
```diff
@@ -47,37 +48,56 @@ def batch_space(space, n=1):
...
+        )
     else:
         return Box(
             low=space.start,
             high=space.start + space.n - 1,
             shape=(n,),
             dtype=space.dtype,
+            seed=copy(space.np_random),
```
lebrice, line 68:
```suggestion
            seed=deepcopy(space.np_random),
```

gym/vector/utils/spaces.py
```diff
@@ -47,37 +48,56 @@ def batch_space(space, n=1):
...
 @batch_space.register(MultiDiscrete)
 def _batch_space_multidiscrete(space, n=1):
     repeats = tuple([n] + [1] * space.nvec.ndim)
     high = np.tile(space.nvec, repeats) - 1
-    return Box(low=np.zeros_like(high), high=high, dtype=space.dtype)
+    return Box(
+        low=np.zeros_like(high),
+        high=high,
+        dtype=space.dtype,
+        seed=copy(space.np_random),
```
lebrice, line 80:
```suggestion
        seed=deepcopy(space.np_random),
```

gym/vector/utils/spaces.py
```diff
@@ -47,37 +48,56 @@ def batch_space(space, n=1):
...
 
 @batch_space.register(MultiBinary)
 def _batch_space_multibinary(space, n=1):
-    return Box(low=0, high=1, shape=(n,) + space.shape, dtype=space.dtype)
+    return Box(
+        low=0,
+        high=1,
+        shape=(n,) + space.shape,
+        dtype=space.dtype,
+        seed=copy(space.np_random),
```
lebrice, line 91:
```suggestion
        seed=deepcopy(space.np_random),
```

gym/vector/utils/spaces.py
```diff
@@ -47,37 +48,56 @@ def batch_space(space, n=1):
...
+    )
 
 
 @batch_space.register(Tuple)
 def _batch_space_tuple(space, n=1):
-    return Tuple(tuple(batch_space(subspace, n=n) for subspace in space.spaces))
+    return Tuple(
+        tuple(batch_space(subspace, n=n) for subspace in space.spaces),
+        seed=copy(space.np_random),
```
lebrice, line 99:
```suggestion
        seed=deepcopy(space.np_random),
```

gym/vector/utils/spaces.py
```diff
@@ -88,13 +108,17 @@ def _batch_space_dict(space, n=1):
                 (key, batch_space(subspace, n=n))
                 for (key, subspace) in space.spaces.items()
             ]
-        )
+        ),
+        seed=copy(space.np_random),
```
lebrice, line 112:
```suggestion
        seed=deepcopy(space.np_random),
```

gym/vector/utils/spaces.py
```diff
@@ -88,13 +108,17 @@ def _batch_space_dict(space, n=1):
...
     )
 
 
 @batch_space.register(Space)
 def _batch_space_custom(space, n=1):
-    return Tuple(tuple(space for _ in range(n)))
+    # Without deepcopy, then the space.np_random is batched_space.spaces[0].np_random
+    # Which is an issue if you are sampling actions of both the original space and the batched space
+    space_copy = deepcopy(space)
+    return Tuple(tuple(space_copy for _ in range(n)), seed=copy(space.np_random))
```
lebrice, line 121:
```suggestion
    return Tuple(tuple(space_copy for _ in range(n)), seed=deepcopy(space.np_random))
```

tests/vector/test_spaces.py
```diff
@@ -129,3 +133,78 @@ def test_iterate_custom_space(space, batch_space):
...
+        {"space-1": Box(0, 5, (), seed=123), "space-2": Box(0, 10, (), seed=123)},
+        seed=123,
+    ),
+    Discrete(5, seed=123),
+    MultiDiscrete([5, 3], seed=123),
+]
+
+
+@pytest.mark.parametrize(
+    "space", seeded_spaces, ids=[space.__class__.__name__ for space in seeded_spaces]
```
lebrice, line 152: Watch out, the seeded_spaces are shared between tests, and so their rng state is affected!
(BTW this hasn't caused an issue here, because you're only using `seeded_spaces` in one place, and your tests aren't dependant on the precise rng spate of the `space` they get from `seeded_spaces`.
Just thought I'd point this out, in case it's useful to you :)

tests/vector/test_spaces.py
```diff
@@ -129,3 +133,78 @@ def test_iterate_custom_space(space, batch_space):
...
+)
+def test_batch_space_seed(space):
+    batched_space = batch_space(space)  # n=1
```
lebrice, line 155: Nit:
```suggestion
    batched_space = batch_space(space, n=1)
```

tests/vector/test_spaces.py
```diff
@@ -129,3 +133,78 @@ def test_iterate_custom_space(space, batch_space):
...
+    # Check that batched space a and b random number generator are not effected by the original space
+    space_a.sample()
+    space_a_batched_sample = space_a_batched.sample()
+    space_b_batched_sample = space_b_batched.sample()
+    for a_sample, b_sample in zip(
+        iterate(space_a_batched, space_a_batched_sample),
+        iterate(space_b_batched, space_b_batched_sample),
+    ):
+        if isinstance(a_sample, tuple):
+            for a_subsample, b_subsample in zip(a_sample, b_sample):
```
lebrice, line 207: Really subtle, but here you should probably assert that they have the same length, or use `itertools.zip_longest` instea of `zip`, otherwise the test could hypothetically silently pass when the samples are different.
```suggestion
            for a_subsample, b_subsample in itertools.zip_longest(a_sample, b_sample):
```
(edit: oh and you'd need to import itertools at the top of the file too)

tests/vector/test_sync_vector_env.py
lebrice: Hey @pseudo-rnd-thoughts, would you mind un-resolving the conversation that was here? I don't seem to be able to do so myself.

lebrice (commented):
Thanks for the changes @pseudo-rnd-thoughts !

gym/utils/seeding.py
lebrice: Why is this marked as resolved?

gym/spaces/dict.py
RedTachyon: +1 on the question (also for the sequence of ints later on in the code)

gym/spaces/dict.py
```diff
@@ -3,9 +3,11 @@
 from collections import OrderedDict
 from collections.abc import Mapping, Sequence
 from typing import Dict as TypingDict
+from typing import Optional
 
 import numpy as np
 
+from ..utils import seeding
```
RedTachyon, line 10: `from gym.utils`, similarly elsewhere in the code (relative imports are evil)

gym/utils/seeding.py
RedTachyon, line 26: Why do we need this? Both in the sense of where is it used, and why can't we just use the inherited `__eq__`?

gym/vector/utils/spaces.py
RedTachyon: +1 on using deepcopy instead of copy, unless you were very deliberate on using `copy` (in which case - explain, because I can kinda see it, but not necessarily)

tests/vector/test_spaces.py
```diff
@@ -1,7 +1,11 @@
+import copy
+
 import numpy as np
 import pytest
+from numpy.testing._private.utils import assert_array_equal
```
RedTachyon, line 5: `from numpy.testing import assert_array_equal`

tests/vector/test_spaces.py
RedTachyon: Either this, or remove the dead-code-comment (whichever is applicable)

tests/vector/test_spaces.py
RedTachyon: We expect them to be the same size, yes? I'd just put an assert on the length in that case

tests/vector/test_sync_vector_env.py
RedTachyon, line 173: Don't hard-code an env, use the parametrization over all envs.

tests/vector/utils.py
```diff
@@ -75,10 +75,10 @@ class CustomSpace(gym.Space):
     """Minimal custom observation space."""
 
     def sample(self):
-        return "sample"
```
RedTachyon, line 78: Was the old version of this custom space ever used? Does changing this change any other tests?

RedTachyon (commented):
Added a bunch of comments. Overall this breaks my brain a little bit, but I think the angle of the solution is fine, so I left some nitpicks for the implementation.

And for the ease of clicking for future me: #2680

RedTachyon:
Ah, and please name the PR in some more descriptive way.

lebrice:
Hey @pseudo-rnd-thoughts just a small remark:
I'd rather mark the conversations as resolved myself when reviewing. This makes it easier to see if my comments are addressed. 

Also, and this is perhaps a bit "extra", but I personally like to leave a comment with the hash of the commit that addresses the feedback, sometibg like "fixed in abcdef123", and GitHub will automatically create a link to view the diff for that commit.

This makes the reviewing experience much better. Also, could you please un-resolve the discussions where I left a comment? I don't believe I can do it myself atm.

pseudo-rnd-thoughts:
Thanks, this is my first time working on a project of this size so thanks for the recommendations, I will try to do all of them.

tests/vector/test_spaces.py
lebrice: I'd expect that `iterate` might return an iterator rather than a list or tuple, so calling `len` would exhaust it. That's why I'm thinking zip_longest might be safer, since IIRC the for loop over an exhausted iterator will just do nothing? or raise an exception? not sure.

pseudo-rnd-thoughts:
@lebrice I think that you have made all of the changes necessary. 
Would you be able to resolve all of the questions that you think have been solved?

gym/vector/utils/spaces.py
lebrice: Thanks for the clarification, however this doesn't really address the second issue. The subspaces shouldn't have implicitly-shared RNG. They should each have independent (and different!) RNG that are deterministic with respect to the base space.

Again, here's the solution:
Sample N seeds with the rng of the base space, and then pass these N seeds to each subspace (or a Generator for each subspace).

tests/vector/test_spaces.py
```diff
@@ -129,3 +133,79 @@ def test_iterate_custom_space(space, batch_space):
     for i, item in enumerate(iterator):
         assert item in space
     assert i == 3
+
+
+seeded_spaces = [
```
lebrice, line 138: I suggest you reuse the `spaces` and `expected_batched_spaces_4` and so-on fixtures from above, and just seed them.

tests/vector/test_sync_vector_env.py
lebrice: Still need to parametrize the env name here IMO.

tests/vector/utils.py
lebrice: This can be marked as resolved.

tests/vector/test_spaces.py
lebrice: This can be marked as resolved.

tests/vector/test_spaces.py
lebrice: This can be marked as resolved.

gym/spaces/box.py
lebrice: I guess this is probably out of scope for this PR, so no need to change this, but I'm really skeptical about the usefulness of having this `seeding.RandomNumberGenerator` thingy. If someone can point me to the rationale behind using it, I'd be curious to take a look.

RedTachyon: >If someone can point me to the rationale behind using it, I'd be curious to take a look.

It's a [compatibility layer](https://github.com/openai/gym/blob/master/gym/utils/seeding.py#L23) that supports both the new `np.random.Generator` API as well as the older `np.random.RandomState` which is now considered legacy

pseudo-rnd-thoughts:
@lebrice Sorry I messed up, I have responded to all of your messages but had them as a review and pending and thought they had gone through but they hadn't. Just did that now, I hope my "old" answers are helpful

tests/vector/test_spaces.py
pseudo-rnd-thoughts: I removed `seeded_spaces` in 296cf50def0ef62b3d47f56f2cd98628806e36e9

tests/vector/test_sync_vector_env.py
pseudo-rnd-thoughts: Done in b7fc60fdb8a0dc77e8615ffc18dbba668e7b7a14

pseudo-rnd-thoughts: Done in b7fc60fdb8a0dc77e8615ffc18dbba668e7b7a14

gym/utils/seeding.py
pseudo-rnd-thoughts: Removed in 41a568300a7d91a03e529aaf03ea24e5699bdd4f

tests/vector/test_spaces.py
pseudo-rnd-thoughts: New test in `tests/vector/test_spaces::test_rng_different_at_each_index`

gym/vector/utils/spaces.py
pseudo-rnd-thoughts: Changed in 4fdd037c6531e01fe4e003eae09c357f44778187

RedTachyon:
Looks good now

PR #2737, opened 2022-04-06
fix #2736

### LINKED ISSUE #2736: [Bug Report] Taxi env.render(mode='rgb_array') broken
```python
import gym
env = gym.make("Taxi-v3")
env.reset()
env.render(mode='rgb_array')
```
yields

```
  File "<stdin>", line 1, in <module>
  File "/opt/anaconda3/envs/vid/lib/python3.9/site-packages/gym/core.py", line 286, in render
    return self.env.render(**kwargs)
  File "/opt/anaconda3/envs/vid/lib/python3.9/site-packages/gym/core.py", line 286, in render
    return self.env.render(**kwargs)
  File "/opt/anaconda3/envs/vid/lib/python3.9/site-packages/gym/envs/toy_text/taxi.py", line 234, in render
    return self._render_gui(mode)
  File "/opt/anaconda3/envs/vid/lib/python3.9/site-packages/gym/envs/toy_text/taxi.py", line 269, in _render_gui
    self.destination_img = self.destination_img.convert_alpha()
pygame.error: No video mode has been set
```

@younik

younik:
Fix #2736, `pygame.display.set_mode` must be always called before  using `set_alpha`


Markus28:
I don't think this will allow rendering on headless systems though (and wouldn't it pop up a window in case of "rgb_array" again?)

younik:
> I don't think this will allow rendering on headless systems though (and wouldn't it pop up a window in case of "rgb_array" again?). Also, the problem is being caused by convert_alpha, not set_alpha. I'm not that familiar with PyGame but couldn't you just use set_alpha instead of convert_alpha?

I haven't thought about headless system case, thank you!
Actually you are right, it works even without `convert_alpha`.

PR #2743, opened 2022-04-10
Add test gym utils play. Fix #2729

### LINKED ISSUE #2729: [Proposal] gym.utils.play has no tests
I ran coverage over our pytest to see what percentage of the code base is currently being tested
It seems that gym.utils.play is not being tested at all currently (every other files is from my check)
Is it possible to run tests on this as the tests are normally headless I thought?

gianlucadecola:
Referring to #2729; to add test to gym.utils.play I have refactored the play() script and extracted a PlayableGame class to have more separation and easier testing. 
Running headless the whole game is possible but it seems that pygame events are not available (https://www.pygame.org/wiki/HeadlessNoWindowsNeeded) so I prefer to mock them in tests and test the PlayableGame class

### DIFF
--- a/gym/utils/play.py
+++ b/gym/utils/play.py
@@ -19,10 +19,13 @@ from pygame.locals import VIDEORESIZE
 
 
 class PlayableGame:
-    def __init__(self, env):
+    def __init__(self, env, keys_to_action=None):
         self.env = env
+        self.relevant_keys = self.get_relevant_keys(keys_to_action)
+        self.pressed_keys = []
+        self.running = True
 
-    def get_relevant_keys(self, keys_to_action=None):
+    def get_relevant_keys(self, keys_to_action):
         if keys_to_action is None:
             if hasattr(self.env, "get_keys_to_action"):
                 keys_to_action = self.env.get_keys_to_action()
@@ -37,6 +40,30 @@ class PlayableGame:
         relevant_keys = set(sum(map(list, keys_to_action.keys()), []))
         return relevant_keys
 
+    def get_video_size(self, zoom=None):
+        rendered = self.env.render(mode="rgb_array")
+        video_size = [rendered.shape[1], rendered.shape[0]]
+
+        if zoom is not None:
+            video_size = int(video_size[0] * zoom), int(video_size[1] * zoom)
+
+        return video_size
+
+    def process_event(self, event):
+        if event.type == pygame.KEYDOWN:
+            if event.key in self.relevant_keys:
+                self.pressed_keys.append(event.key)
+            elif event.key == 27:
+                self.running = False
+        elif event.type == pygame.KEYUP:
+            if event.key in self.relevant_keys:
+                self.pressed_keys.remove(event.key)
+        elif event.type == pygame.QUIT:
+            self.running = False
+        elif event.type == VIDEORESIZE:
+            video_size = event.size
+            screen = pygame.display.set_mode(video_size)
+            print(video_size)
 
 
 def display_arr(screen, arr, video_size, transpose):
@@ -104,15 +131,10 @@ def play(env, transpose=True, fps=30, zoom=None, callback=None, keys_to_action=N
         If None, default key_to_action mapping for that env is used, if provided.
     """
     env.reset()
-    game = PlayableGame(env)
-    
-    rendered = env.render(mode="rgb_array")
+    game = PlayableGame(env, keys_to_action)
 
     relevant_keys = game.get_relevant_keys(keys_to_action)
-
-    video_size = [rendered.shape[1], rendered.shape[0]]
-    if zoom is not None:
-        video_size = int(video_size[0] * zoom), int(video_size[1] * zoom)
+    video_size = game.get_video_size(zoom)
 
     pressed_keys = []
     running = True
@@ -137,6 +159,7 @@ def play(env, transpose=True, fps=30, zoom=None, callback=None, keys_to_action=N
 
         # process pygame events
         for event in pygame.event.get():
+            # game.process_event(event)
             # test events, set key states
             if event.type == pygame.KEYDOWN:
                 if event.key in relevant_keys:
--- a/gym/utils/play.py
+++ b/gym/utils/play.py
@@ -1,12 +1,13 @@
 import argparse
 
-import matplotlib
 import pygame
 
 import gym
 from gym import logger
 
 try:
+    import matplotlib
+
     matplotlib.use("TkAgg")
     import matplotlib.pyplot as plt
 except ImportError as e:
@@ -133,22 +134,19 @@ def play(env, transpose=True, fps=30, zoom=None, callback=None, keys_to_action=N
     env.reset()
     game = PlayableGame(env, keys_to_action)
 
-    relevant_keys = game.get_relevant_keys(keys_to_action)
     video_size = game.get_video_size(zoom)
 
-    pressed_keys = []
-    running = True
     env_done = True
 
     screen = pygame.display.set_mode(video_size)
     clock = pygame.time.Clock()
 
-    while running:
+    while game.running:
         if env_done:
             env_done = False
             obs = env.reset()
         else:
-            action = keys_to_action.get(tuple(sorted(pressed_keys)), 0)
+            action = keys_to_action.get(tuple(sorted(game.pressed_keys)), 0)
             prev_obs = obs
             obs, rew, env_done, info = env.step(action)
             if callback is not None:
@@ -159,19 +157,9 @@ def play(env, transpose=True, fps=30, zoom=None, callback=None, keys_to_action=N
 
         # process pygame events
         for event in pygame.event.get():
-            # game.process_event(event)
+            game.process_event(event)
             # test events, set key states
-            if event.type == pygame.KEYDOWN:
-                if event.key in relevant_keys:
-                    pressed_keys.append(event.key)
-                elif event.key == 27:
-                    running = False
-            elif event.type == pygame.KEYUP:
-                if event.key in relevant_keys:
-                    pressed_keys.remove(event.key)
-            elif event.type == pygame.QUIT:
-                running = False
-            elif event.type == VIDEORESIZE:
+            if event.type == VIDEORESIZE:
                 video_size = event.size
                 screen = pygame.display.set_mode(video_size)
                 print(video_size)
--- a/gym/utils/play.py
+++ b/gym/utils/play.py
@@ -20,13 +20,15 @@ from pygame.locals import VIDEORESIZE
 
 
 class PlayableGame:
-    def __init__(self, env, keys_to_action=None):
+    def __init__(self, env, keys_to_action=None, zoom=None):
         self.env = env
-        self.relevant_keys = self.get_relevant_keys(keys_to_action)
+        self.relevant_keys = self._get_relevant_keys(keys_to_action)
+        self.video_size = self._get_video_size(zoom)
+        self.screen = pygame.display.set_mode(self.video_size)
         self.pressed_keys = []
         self.running = True
 
-    def get_relevant_keys(self, keys_to_action):
+    def _get_relevant_keys(self, keys_to_action):
         if keys_to_action is None:
             if hasattr(self.env, "get_keys_to_action"):
                 keys_to_action = self.env.get_keys_to_action()
@@ -41,7 +43,7 @@ class PlayableGame:
         relevant_keys = set(sum(map(list, keys_to_action.keys()), []))
         return relevant_keys
 
-    def get_video_size(self, zoom=None):
+    def _get_video_size(self, zoom=None):
         rendered = self.env.render(mode="rgb_array")
         video_size = [rendered.shape[1], rendered.shape[0]]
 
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts:
Would you be able to add typing to the PlayableGame or pydocs for people using the class to be able to know what expected information is?

gianlucadecola:
Added type hint :smiley: 
As for testing the ```play()``` function, I'm quite unsure how to break the while loop since it does not seem possible to inject events in headless mode. One solution might be adding a ```testing``` parameter inside ```play()``` and tweaking a little bit the function but it seems like a sort of hacky solution to me :thinking:

pseudo-rnd-thoughts:
I'm not experienced with pygame so this may not work
My idea would be to use a custom testing callback function with the `play` function that would add a new event action to the event queue that would be processed next by the game. 
This could allow you to run the `play` function forward `n` time steps with a known input then compare the resulting game observation at `n` to a standard environment at `n`

gianlucadecola:
> I'm not experienced with pygame so this may not work My idea would be to use a custom testing callback function with the `play` function that would add a new event action to the event queue that would be processed next by the game. This could allow you to run the `play` function forward `n` time steps with a known input then compare the resulting game observation at `n` to a standard environment at `n`

Well, it actually worked pretty well :tada: Really good hint

pseudo-rnd-thoughts:
Wow, I was half expecting that to not work, thanks for doing that 

You probably have a reason but 
1. Why are you using dummy env rather than an actual gym environment, i.e. cartpole?
2. Why are you using a dummy key event rather than an actual pygame event?

Extra test ideas
3. Could you extend the current test_play to be longer and assert the final observation are equal as checking that the reward hasn't changed at all (as the initial value is 0) doesn't seem to test the function properly to me
4. This may be too much, but we could run the play for all of the `spec_list` environments with random actions being generated in the callback (key down for the new action, key up for the old action) 

Finally, if you think it is possible to add a test for `PlayPlot`? There doesn't seem much to test, maybe the data saved is correct, particularly for `len(plot_names)` > 1. 
Plus we should add a doc string and type hinting for the class if possible

gianlucadecola:
> Wow, I was half expecting that to not work, thanks for doing that
> 
> You probably have a reason but
> 
>     1. Why are you using dummy env rather than an actual gym environment, i.e. cartpole?
> 

I didn't like the idea of test being tied to a particular environment. It seems more generic to me to have the dummy just for testing purpose of the play function

>     2. Why are you using a dummy key event rather than an actual pygame event?

This can actually be removed and replaced with an actual pygame event, I will commit the fix

> 
> Extra test ideas 3. Could you extend the current test_play to be longer and assert the final observation are equal as checking that the reward hasn't changed at all (as the initial value is 0) doesn't seem to test the function properly to me 4. This may be too much, but we could run the play for all of the `spec_list` environments with random actions being generated in the callback (key down for the new action, key up for the old action)
> 
> Finally, if you think it is possible to add a test for `PlayPlot`? There doesn't seem much to test, maybe the data saved is correct, particularly for `len(plot_names)` > 1. Plus we should add a doc string and type hinting for the class if possible

Good point on .3; 
.4 seems an overkill to me, maybe it could be done in another PR :thinking: 

I will try to look into PlayPlot, also :+1:

jkterry1:
to answer an above question, removing old code that doesn't do anything is fine

gym/utils/play.py
```diff
@@ -18,6 +21,55 @@
 from pygame.locals import VIDEORESIZE
 
 
+class PlayableGame:
+    def __init__(self, env: Env, keys_to_action: dict = None, zoom: float = None):
```
RedTachyon, line 25: `keys_to_action: Optional[dict]`

gym/utils/play.py
```diff
@@ -18,6 +21,55 @@
...
+        self.env = env
+        self.relevant_keys = self._get_relevant_keys(keys_to_action)
+        self.video_size = self._get_video_size(zoom)
+        self.screen = pygame.display.set_mode(self.video_size)
+        self.pressed_keys = []
+        self.running = True
+
+    def _get_relevant_keys(self, keys_to_action: dict) -> set:
```
RedTachyon, line 33: `Optional[dict]`

gym/utils/play.py
RedTachyon: Even better if we can specify the key/value types on the dict (I guess values are arbitrary actions, but keys should be keypresses? how are they represented?)

gym/utils/play.py
```diff
@@ -18,6 +21,55 @@
...
+        if keys_to_action is None:
+            if hasattr(self.env, "get_keys_to_action"):
+                keys_to_action = self.env.get_keys_to_action()
+            elif hasattr(self.env.unwrapped, "get_keys_to_action"):
+                keys_to_action = self.env.unwrapped.get_keys_to_action()
+            else:
+                assert False, (
```
RedTachyon, line 40: Explicitly `raise` some `Exception` instead of a failing `assert`, you can either use one of the existing exceptions like `AttributeError` or something; or make a new one in this file to make it more descriptive.

gym/utils/play.py
```diff
@@ -18,6 +21,55 @@
...
+                    self.env.spec.id
+                    + " does not have explicit key to action mapping, "
+                    + "please specify one manually"
+                )
+        relevant_keys = set(sum(map(list, keys_to_action.keys()), []))
```
RedTachyon, line 45: `map` is nice (functional programming is great), but in Python it's typically more readable to use a comprehension, so it'd be something like
`set(sum(list(key) for key in keys_to_action), []))`

But is this actually what we want this to do? I'm not sure what each key is meant to be (see previous comment about types), so take another look at the logic here. Intuition tells me that you might have wanted to just do `set(keys_to_actions.keys())` or something like that (which might be equivalent to just `set(keys_to_actions)`)

gym/utils/play.py
```diff
@@ -18,6 +21,55 @@
...
+        return relevant_keys
+
+    def _get_video_size(self, zoom: float = None) -> Tuple[int, int]:
+        rendered = self.env.render(mode="rgb_array")
```
RedTachyon, line 49: Can you put like a TODO here so that we remember to update this when the render API change goes through?

gym/utils/play.py
```diff
@@ -18,6 +21,55 @@
...
+            else:
+                assert False, (
+                    self.env.spec.id
+                    + " does not have explicit key to action mapping, "
+                    + "please specify one manually"
+                )
+        relevant_keys = set(sum(map(list, keys_to_action.keys()), []))
+        return relevant_keys
+
+    def _get_video_size(self, zoom: float = None) -> Tuple[int, int]:
```
RedTachyon, line 48: Optional[float]

gym/utils/play.py
```diff
@@ -18,6 +21,55 @@
...
+        if zoom is not None:
+            video_size = int(video_size[0] * zoom), int(video_size[1] * zoom)
+
+        return video_size
+
+    def process_event(self, event: Event) -> None:
+        if event.type == pygame.KEYDOWN:
+            if event.key in self.relevant_keys:
+                self.pressed_keys.append(event.key)
+            elif event.key == 27:
```
RedTachyon, line 61: What is 27? Ideally replace this with some constant that already might exist in pygame, or define it yourself, or at the very least add a comment

gym/utils/play.py
```diff
@@ -18,6 +21,55 @@
...
+            if event.key in self.relevant_keys:
+                self.pressed_keys.remove(event.key)
+        elif event.type == pygame.QUIT:
+            self.running = False
+        elif event.type == VIDEORESIZE:
+            self.video_size = event.size
+            self.screen = pygame.display.set_mode(self.video_size)
+
+
 def display_arr(screen, arr, video_size, transpose):
```
RedTachyon, line 73: Can you add type hints here and to `play()`?

gym/utils/play.py
```diff
@@ -83,63 +135,30 @@ def callback(obs_t, obs_tp1, action, rew, done, info):
...
-    while running:
-        if env_done:
-            env_done = False
+    while game.running:
+        if done:
+            done = False
             obs = env.reset()
         else:
-            action = keys_to_action.get(tuple(sorted(pressed_keys)), 0)
+            action = keys_to_action.get(tuple(sorted(game.pressed_keys)), 0)
```
RedTachyon, line 148: Can we always safely assume that `0` is a good default action? (if I'm understanding this correctly)
Maybe this should be part of the specification?

In particular this seems like it'd crash with any continuous-action environments?

tests/utils/test_play.py
```diff
@@ -0,0 +1,161 @@
...
+@pytest.fixture(autouse=True)
+def close_pygame():
+    yield
+    pygame.quit()
+
+
+def test_play_relevant_keys():
+    env = DummyPlayEnv()
+    game = PlayableGame(env, dummy_keys_to_action())
+    assert game.relevant_keys == {97, 100}
```
RedTachyon, line 72: No magic numbers please

tests/utils/test_play.py
```diff
@@ -0,0 +1,161 @@
...
+    env = DummyPlayEnv()
+    game = PlayableGame(env, dummy_keys_to_action())
+    event = Event(pygame.KEYDOWN, {"key": RELEVANT_KEY})
+    game.process_event(event)
+    event = Event(pygame.KEYUP, {"key": RELEVANT_KEY})
+    game.process_event(event)
+    assert game.pressed_keys == []
+
+
+def test_play_loop():
```
RedTachyon, line 148: Is it possible to do the same test with some actual environment? Say, CartPole, define the game, input a bunch of random actions in a loop just to see that it doesn't crash.

For a more "advanced" version, if possible, you could instantiate one PlayableGame of an env with a fixed seed, and one normal env with the same fixed seed. You input the same sequence of actions through the game interface, and through the regular `env.step`, and check that the outcome is the same.

RedTachyon (commented):
Overall looks pretty good, left a bunch of comments. Have you tried actually playing this with some environments? As a "full" test which can't really be easily automated, but it's helpful to do it as a human.

gianlucadecola:
> Overall looks pretty good, left a bunch of comments. Have you tried actually playing this with some environments? As a "full" test which can't really be easily automated, but it's helpful to do it as a human.

Hey! Thanks for the reply, yes I had some fun trying to balance the cartpole myself :smile: I'm going to look into the comments :+1:

tests/utils/test_play.py
gianlucadecola: Yes, following what suggested by @pseudo-rnd-thoughts I'm adding a test with an actual environment. It's a little bit tricky to manage the Keydown and Keyup event with pygame  but I'm on the way

gym/utils/play.py
gianlucadecola: Yes it is actually the default action if no relevant keys are pressed; I have the same doubt regarding ```0```, I will explore it

gym/utils/play.py
gianlucadecola: The whole ```_get_relevant_keys()``` was extracted "as is" from the old ```play()```. Didn't want to change it before having all tests in place but this is safe enough to replace with a pygame constants  :sweat_smile: (it should be the exit key)

gym/utils/play.py
gianlucadecola: I'm going to explore it more in depth, this logic was already there in the old ```play()```, it seems also to me that  ```set(keys_to_actions.keys())``` may be fine

gym/utils/play.py
gianlucadecola: It's ```Dict[Tuple[int], int]``` where the key is ```(ord(key),)```
https://github.com/openai/gym/blob/5ae6bf9634c73ed09d73ceeb41e3491d7e791cce/gym/envs/classic_control/mountain_car.py#L240-L242  

reference of the actions:
https://github.com/openai/gym/blob/5ae6bf9634c73ed09d73ceeb41e3491d7e791cce/gym/envs/classic_control/mountain_car.py#L50-L54

https://github.com/openai/gym/blob/9a1d4b67684785c03c187c0a7e8ab217b8f9a28f/gym/utils/play.py#L125-L135

gianlucadecola:
Pushed a bunch of updates; still need to assess some minor things like ```set(sum(list(key) for key in keys_to_action), []))``` and explore this: 

> Can we always safely assume that `0` is a good default action? (if I'm understanding this correctly) Maybe this should be part of the specification?
> 
> In particular this seems like it'd crash with any continuous-action environments?

gym/utils/play.py
gianlucadecola: @RedTachyon About this I think we can safely say that, since we can not map a continuos-action environment to discrete keyboard presses it is ok to leave ```0``` as default because ```play()``` will crash before since there is no ```keys_to_action```. (Maybe we could add an explicit message for continuos-action environment?)
An example of a keys to action mapping it is indeed this:
https://github.com/openai/gym/blob/5ae6bf9634c73ed09d73ceeb41e3491d7e791cce/gym/envs/classic_control/mountain_car.py#L240-L242

gianlucadecola:
@RedTachyon I think everything is in place now

pseudo-rnd-thoughts:
Thanks @gianlucadecola I think the play function changes are complete. 
Do you want to add a PlayPlot docstring, type hinting and test in this PR or another?

gianlucadecola:
I think we can merge this and then I'll submit another PR assessing only the PlayPlot part as soon as I can manage to work on it :+1:

pseudo-rnd-thoughts:
@jkterry1 Looks good to merge

PR #2746, opened 2022-04-11
Add gravity and wind as kwarg in Lunar Lander

jjshoots:
# Description

This change is for LunarLander.
This adds adjustable gravity and wind to the env.
More changes to this env and perhaps other envs are possible in due time.

To set gravity value to say, 5, simply pass `gravity=-5`.

To enable wind, pass the wind kwarg `enable_wind=True`.
By default, the wind function is ` C tanh(sin(2 k x) + sin(pi k x)`, which is proven to be non-periodic for all `k` (see [this](https://math.stackexchange.com/questions/1776688/why-sinx-sin-pi-x-is-not-periodic)).
In implementation, I've set `k = 0.01`, `C=15`.
This is just a simpler version instead of using a Dryden wind field or Perlin noise sampling.

## Type of change

# Checklist:

### DIFF
--- a/gym/envs/box2d/lunar_lander.py
+++ b/gym/envs/box2d/lunar_lander.py
@@ -353,7 +353,7 @@ class LunarLander(gym.Env, EzPickle):
             else:
                 m_power = 1.0
             # 4 is move a bit downwards, +-2 for randomness
-            ox = +tip[0] * (4 / SCALE + 2 * dispersion[0]) + side[0] * dispersion[0]
+            ox = +tip[0] * (4 / SCALE + 2 * dispersion[0]) + side[0] * dispersion[1]
             oy = -tip[1] * (4 / SCALE + 2 * dispersion[0]) - side[1] * dispersion[1]
             impulse_pos = (self.lander.position[0] + ox, self.lander.position[1] + oy)
             p = self._create_particle(
--- a/gym/envs/box2d/lunar_lander.py
+++ b/gym/envs/box2d/lunar_lander.py
@@ -351,10 +351,13 @@ class LunarLander(gym.Env, EzPickle):
         ):
             # the function used for wind is tanh(sin(2 k x) + sin(pi k x)),
             # which is proven to never be periodic, k = 0.01
-            wind_mag = math.tanh(
-                math.sin(0.02 * self.wind_idx)
-                + (math.sin(math.pi * 0.01 * self.wind_idx))
-            ) * WIND_POWER
+            wind_mag = (
+                math.tanh(
+                    math.sin(0.02 * self.wind_idx)
+                    + (math.sin(math.pi * 0.01 * self.wind_idx))
+                )
+                * WIND_POWER
+            )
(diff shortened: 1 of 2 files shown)


balisujohn:
This is interesting; I could definitely see this kind of parameterization being useful for sim2sim research.

gym/envs/box2d/lunar_lander.py
```diff
@@ -618,6 +651,11 @@ def heuristic(env, s):
 
 
 def demo_heuristic_lander(env, seed=None, render=False):
+
+    # wind power must be reduced for heuristic landing
+    global WIND_POWER
```
RedTachyon, line 656: This is a big no-no, all caps suggests that this should be a global constant, so changing this in the code is a bad pattern.

If the wind power is something that users are meant to change, maybe make it a constructor argument?

jjshoots: Hi Ariel, the wind power isn't something that user's are meant to change.

The only reason it's here is that the heuristic landing won't work with the default wind setting, and this causes it to fail tests. So for heuristic landings, I've reduced the wind power to be 1/15th of the original magnitude (so that the wind code is still tested in tests, but still solvable with heuristic landing.

I'm aware this is not the most optimal solution, I'll think of something soon.

jjshoots: I've made it a kwarg in the latest commit.

jjshoots: @jkterry1 done!

RedTachyon:
So actually, what is the rationale for keeping the wind strength a constant? It might be a good option to showcase curriculum learning by changing its value throughout the training (I'm assuming learning is easier if there is no wind, and more difficult if it's stronger)

It might even be an opportunity to showcase the new `reset` optional arguments and make its magnitude settable at reset.

@jkterry1 what do you think?

gym/envs/box2d/lunar_lander.py
```diff
@@ -123,7 +123,13 @@ class LunarLander(gym.Env, EzPickle):
     `continuous=True` argument like below:
     ```python
     import gym
-    env = gym.make("LunarLander-v2", continuous=True)
+    env = gym.make(
+        "LunarLander-v2",
+        continuous: bool = False,
+        gravity: float = -10.0,
+        enable_wind: bool = False,
```
pzhokhov, line 130: should the env be renamed into "Lander"? LunarLander makes it sound like the landing is happening on the moon, in which case, accounting for wind does not seem necessary?

pzhokhov, line 130: should the env be renamed into "Lander"? LunarLander makes it sound like the landing is happening on the moon, in which case, accounting for wind does not seem necessary?

pseudo-rnd-thoughts: While that could be more accurate to change the name, for posterity's sake (and breaking a load of code) then we won't change the name

Issue #2738, opened 2022-04-06
[Proposal] Simplify `gym.make` implementation


RedTachyon:
### Proposal 

Rewrite the entire `registration.py` file so that the top-level `make` function actually does its thing, instead of calling `EnvRegistry` which calls `EnvSpec` through an `EnvSpecTree`

### Motivation

`gym.make` is crucial to the practical use of Gym, but its implementation is pretty chaotic and overcomplicated, which makes any modifications difficult, and bugs hard to spot. When you call it, it instantly just calls `registry.make` on a global object, which does some of the logic, and then `spec.make` which has the rest of the logic. It feels object-oriented to a fault.

### Implementation details

1. The top-level function `make` should perform most/all of the environment creation logic, with parts delegated to separate functions
2. `EnvSpec` becomes **just** a dataclass, without extra logic
3. I don't think `EnvSpecTree` is actually necessary, since we're not really working with multi-layer trees here. Imo a dictionary is sufficient to represent all envs, and the extra layers can just be flattened - and that's perfectly fine. I don't see any actual use cases for the nested structure other than maybe asymptotically slightly faster lookups of *incorrect* env names
4. `EnvRegistry` gets turned into a dictionary (the same one as mentioned above), with no logic of its own
5. `register` gets changed similarly to `make`, and it essentially just adds an entry to the global registry (dict)

Pinging in particular @JesseFarebro since you added EnvSpecTree in #2481 and the whole plugin system.

### Rough edges

The plugin system is a total undocumented mystery to me. I have no idea if it will work if I just copy-paste it. 

Also, I think backwards-compatibility of the undocumented low-level implementation details shouldn't be a priority. This reminds me why I'm not a fan of python and always exporting everything, but I *hope* nobody's accessing the `registry` directly, for example. It might be best to keep this for 1.0 release, not sure though.

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/openai/gym/issues) in the repo (**required**)

JesseFarebro:
This is long overdue. At the time of #2481 I didn't know how much of an appetite there was for breaking changes as there's quite a bit of code that uses the environment registry object. If you have any questions about how the entry-point plugin system works just let me know.

PR #2748, opened 2022-04-13
Rewriting of the registration mechanism

RedTachyon:
# Description

This is a total rewrite of `gym.envs.regitration` as proposed in #2738.

It technically doesn't really add any new features beyond fixing a few things that used not to work, but they weren't necessarily documented, so it's hard to call it a bug fix either.

Instead, this significantly simplifies the implementation of registration so that we can actually start adding features to it

The current version implements all significant* behaviors from the old registration mechanism. The intended mechanism is as follows:

1. The registry is a global dictionary sitting in `gym.envs.registration`
2. When registering an environment, you create `EnvSpec` and add it to the registry
3. When creating an environment, you get the appropriate `EnvSpec` from the registry
4. If the passed environment name is invalid, try to give an extra hint on what went wrong ("You tried to create "cartploe", did you mean "CartPole"?)
5. If the passed environment name does not specify the version, create the newest version
6. There is a support for namespaces in the format `namespace/EnvName-vX` (this already existed, but was undocumented, so I'll add it to the docs somewhere later)

One thing I have yet to add is adding an optional minimal env checker on `make` to check if the environment uses the up-to-date API.

*Some low-level details (like the EnvSpecTree) are removed. Also, the error messages may be a bit different, but following the same spirit, and passing the previous tests for where errors and warnings are expected. This is with the exception of one test which was just bugged.

## Highlights

To justify this change, here's a comparison of how `make` changed

In the old implementation, you have `gym.make` which takes an environment name and any kwargs, and instantly just calls `registry.make`. This finds the right `EnvSpec`, and then calls `spec.make`. The actual env creation logic lives in `spec.make`.

Now, it's all handled by `gym.make` -- it finds the right `spec` in the `registry`, initializes it, applies whatever wrappers necessary, and returns the environment.

This is significant when trying to change something about `gym.make`. Previously, you'd have to propagate any changes across the three functions. Now you just change the one function responsible for that. 

An extra bug that this accidentally solves is something that lived in the `register` function. In the old code, all of its logic was in a `try-except-finally` block, where the actual registration happened in `finally`. So even if you completely malformed your environment registration, it will throw an exception and then... still register it. 

Notably, `EnvSpecTree` has been removed and replaced with a dictionary and a few functions. This might cause a worse asymptotical lookup complexity if someone has thousands of environments and namespaces registered, and tries to create an incorrectly named environment, but that's about it (if you have the right environment name, the lookup is still just a dictionary lookup)

## Note:
This is a very core part of the code, so will require significant review. I'm going for full backwards compatibility in terms of `make`, `register` and `spec`.

# Checklist:

### DIFF
--- a/gym/envs/registration.py
+++ b/gym/envs/registration.py
@@ -96,7 +96,7 @@ class EnvSpec:
 
     id_requested: InitVar[str]
     entry_point: Optional[Union[Callable, str]] = field(default=None)
-    reward_threshold: Optional[int] = field(default=None)
+    reward_threshold: Optional[float] = field(default=None)
     nondeterministic: bool = field(default=False)
     max_episode_steps: Optional[int] = field(default=None)
     order_enforce: bool = field(default=True)
--- a/gym/envs/registration.py
+++ b/gym/envs/registration.py
@@ -9,8 +9,8 @@ import re
 import sys
 from dataclasses import dataclass, field
 from typing import (
+    Any,
     Callable,
-    Literal,
     Optional,
     Sequence,
     SupportsFloat,
@@ -29,6 +29,15 @@ if sys.version_info < (3, 10):
     import importlib_metadata as metadata  # type: ignore
 else:
     import importlib.metadata as metadata
+
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+
+    class Literal(str):
+        def __class_getitem__(cls, item):
+            return Any
+
 
 from gym import Env, error, logger
 
--- a/gym/envs/registration.py
+++ b/gym/envs/registration.py
@@ -1,9 +1,17 @@
+from __future__ import annotations
+
 import contextlib
+import copy
+import difflib
 import importlib
+import importlib.util
 import re
 import sys
 from dataclasses import dataclass, field
 from typing import Callable, Optional, Tuple, Type, Union
+
+from gym.envs.__relocated__ import internal_env_relocation_map
+from gym.wrappers import AutoResetWrapper, OrderEnforcing, TimeLimit
 
 if sys.version_info < (3, 10):
     import importlib_metadata as metadata  # type: ignore
@@ -28,10 +36,10 @@ def parse_env_id(id: str) -> Tuple[Optional[str], str, Optional[int]]:
     """Parse environment ID string format.
 
     This format is true today, but it's *not* an official spec.
-    [username/](env-name)-v(version)    env-name is group 1, version is group 2
+    [namespace/](env-name)-v(version)    env-name is group 1, version is group 2
 
     2016-10-31: We're experimentally expanding the environment ID format
-    to include an optional username.
+    to include an optional namespace.
     """
     match = ENV_ID_RE.fullmatch(id)
     if not match:
@@ -44,6 +52,18 @@ def parse_env_id(id: str) -> Tuple[Optional[str], str, Optional[int]]:
         version = int(version)
 
     return namespace, name, version
+
+
+def get_env_id(ns: Optional[str], name: str, version: Optional[int]):
+    """Get the full env ID given a name and (optional) version and namespace.
+    Inverse of parse_env_id."""
+
+    full_name = name
+    if version is not None:
+        full_name += f"-v{version}"
+    if ns is not None:
+        full_name = ns + "/" + full_name
+    return full_name
 
 
 @dataclass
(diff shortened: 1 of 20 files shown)


pseudo-rnd-thoughts:
Thanks for the changes, I can't see any obvious issues with the code though I haven't done a thorough review
You say that you want to preserve full backward compatibility of the core feature but some of the low-level features have been removed. 
Can you think of a reasonable case where removing the low-level features could break someone's code? I.e., should this be released as part of 1.0 or can this be pre 1.0?

RedTachyon:
I don't think there are *reasonable* cases where this would break someone code. The removed features are undocumented implementation details and weren't intended for end-user usage as far as I know.

But you [never know](https://xkcd.com/1172/)

tests/envs/test_registration.py
```diff
@@ -188,112 +193,108 @@ def test_spec_with_kwargs():
...
-    registry = registration.EnvRegistry()
-    registry.register(id="Test-v0", entry_point=None)
-    registry.register(id="Test-v15", entry_point=None)
-    registry.register(id="Test-v9", entry_point=None)
-    registry.register(id="Other-v100", entry_point=None)
+    register(id="Test1-v0", entry_point=None)
+    register(id="Test1-v15", entry_point=None)
+    register(id="Test1-v9", entry_point=None)
+    register(id="Other1-v100", entry_point=None)
     try:
```
ikamensh, line 200: is this construct a poor man's version of `with pytest.raises(error.DeprecatedEnv)` ?

gym/envs/registration.py
```diff
@@ -680,63 +310,168 @@ def make(id: Literal[
...
 
 
-def spec(id: str) -> EnvSpec:
-    return registry.spec(id)
+# Global registry of environments. Meant to be accessed through `register` and `make`
+registry: dict[str, EnvSpec] = dict()
+current_namespace: Optional[str] = None
+
+
+def check_spec_register(spec: EnvSpec):
```
ikamensh, line 327: This public function lacks a docstring. What it means to "check" spec register?

gym/envs/registration.py
```diff
@@ -680,63 +310,168 @@ def make(id: Literal[
...
-    with registry.namespace(ns):
-        yield
+    global current_namespace
+    old_namespace = current_namespace
+    current_namespace = ns
+    yield
+    current_namespace = old_namespace
+
+
+def register(id: str, **kwargs):
```
ikamensh, line 389: public function w/o docstring

gym/envs/registration.py
```diff
@@ -680,63 +310,168 @@ def make(id: Literal[
...
+    spec = EnvSpec(id=full_id, **kwargs)
+    check_spec_register(spec)
+    if spec.id in registry:
+        raise error.RegistrationError(
+            f"Attempted to register {spec.id} but it was already registered"
+        )
+    registry[spec.id] = spec
+
+
+def make(
```
ikamensh, line 401: public function w/o docstring

gym/envs/registration.py
```diff
@@ -680,63 +310,168 @@ def make(id: Literal[
...
+    elif spec_.max_episode_steps is not None:
+        env = TimeLimit(env, spec_.max_episode_steps)
+
+    if autoreset:
+        env = AutoResetWrapper(env)
+
+    return env
+
+
+def spec(env_id: str) -> EnvSpec:
```
ikamensh, line 469: public function w/o docstring

tests/envs/test_registration.py
```diff
@@ -188,112 +193,108 @@ def test_spec_with_kwargs():
...
-    spec_tree[myenv] = spec
-    assert spec_tree.tree.keys() == {None}
-    assert myenv in spec_tree.tree[None].keys()
-    assert spec_tree.tree[None][myenv].keys() == {None}
-    assert spec_tree.tree[None][myenv][None] == spec
-    assert spec_tree.__repr__() == "├──Test: [ v1 ]\n" + f"└──{myenv}: [  ]\n"
+    spec("test/Test3")
+
+
+# def test_env_spec_tree():
```
ikamensh, line 253: Since `EnvSpecTree` class is deleted, it's probably best to delete this test code as well, instead of commenting out.

ikamensh (commented):
Left some generic software quality comments. Re: docstrings, even a one-liner re-affirming what you would guess from the name can help reader be confident they didn't misunderstand the purpose of a method.

In general I'm also a fan of adding one-liner comments to unit test to say what it's intended to verify, but this is not a core contribution of this PR anyway.

tests/envs/test_registration.py
RedTachyon: I think so, I wanted to make as few changes to the tests as possible, but might as well make some updates to them as well

gym/envs/registration.py
RedTachyon: Good catch, I'll add that

RedTachyon: This should probably be renamed. The intention is that it performs a bunch of checks to make sure that the registration will be valid. It also shouldn't be public

gym/envs/registration.py
RedTachyon: Good catch, I'll add that (and the others)

younik (approved):
Well done! LGTM

gym/envs/registration.py
```diff
@@ -7,34 +7,29 @@
...
-from collections import defaultdict
-from collections.abc import MutableMapping
-from dataclasses import InitVar, dataclass, field
-
-import numpy as np
-
-from gym import Env, error, logger
-from gym.envs.__relocated__ import internal_env_relocation_map
-
 if sys.version_info >= (3, 8):
```
pseudo-rnd-thoughts, line 33: As we build for python 3.7 and greater then this is not needed anymore

gym/envs/registration.py
```diff
@@ -78,558 +75,191 @@ def parse_env_id(id: str) -> Tuple[Optional[str], str, Optional[int]]:
...
-            env = cls(**_kwargs)
-
-        # Make the environment aware of which spec it came from.
-        spec = copy.deepcopy(self)
-        spec.kwargs = _kwargs
-        env.unwrapped.spec = spec
-
-        if self.order_enforce:
-            from gym.wrappers.order_enforcing import OrderEnforcing
+def check_namespace_exists(ns: Optional[str]):
```
pseudo-rnd-thoughts, line 114: docstring just for the future if someone is trying to understand the function

gym/envs/registration.py
```diff
@@ -78,558 +75,191 @@ def parse_env_id(id: str) -> Tuple[Optional[str], str, Optional[int]]:
...
-        assert env.spec is not None, "expected spec to be set to the unwrapped env."
-        if env.spec.max_episode_steps is not None:
-            from gym.wrappers.time_limit import TimeLimit
+    raise error.NamespaceNotFound(f"Namespace {ns} not found. {suggestion_msg}")
 
-            env = TimeLimit(env, max_episode_steps=env.spec.max_episode_steps)
 
-        if self.autoreset:
-            from gym.wrappers.autoreset import AutoResetWrapper
+def check_name_exists(ns: Optional[str], name: str):
```
pseudo-rnd-thoughts, line 135: docstring

gym/envs/registration.py
```diff
@@ -78,558 +75,191 @@ def parse_env_id(id: str) -> Tuple[Optional[str], str, Optional[int]]:
...
-            message = f"Environment `{name}` doesn't exist"
-            if namespace is not None:
-                message += f" in namespace `{namespace}`"
-            message += "."
-            suggestions = difflib.get_close_matches(name, self.names(namespace), n=1)
-            if suggestions:
-                message += f" Did you mean: `{suggestions[0]}`?"
-        # Throw the error
-        raise error.NameNotFound(message)
+def check_version_exists(ns: Optional[str], name: str, version: Optional[int]):
```
pseudo-rnd-thoughts, line 168: docstring

gym/envs/registration.py
```diff
@@ -78,558 +75,191 @@ def parse_env_id(id: str) -> Tuple[Optional[str], str, Optional[int]]:
...
+        # Python 3.8 doesn't support plugin.module, plugin.attr
+        # So we'll have to try and parse this ourselves
+        try:
+            module, attr = plugin.module, plugin.attr  # type: ignore  ## error: Cannot access member "attr" for type "EntryPoint"
+        except AttributeError:
+            if ":" in plugin.value:
+                module, attr = plugin.value.split(":", maxsplit=1)
+            else:
+                module, attr = plugin.value, None
+        except:
```
pseudo-rnd-thoughts, line 231: Would we want to raise a warning here to say there was an issue loading a plugin, otherwise a plugin may not be loaded for unknown reasons?

gym/envs/registration.py
```diff
@@ -78,558 +75,191 @@ def parse_env_id(id: str) -> Tuple[Optional[str], str, Optional[int]]:
...
-                logger.warn(
-                    f"Custom namespace `{spec.namespace}` is being overridden "
-                    f"by namespace `{self._ns}`. If you are developing a "
-                    "plugin you shouldn't specify a namespace in `register` "
-                    "calls. The namespace is specified through the "
-                    "entry point package metadata."
-                )
-            # Replace namespace
-            spec.namespace = self._ns
+def find_newest_version(ns: Optional[str], name: str) -> Optional[int]:
```
pseudo-rnd-thoughts, line 210: docstring

gym/envs/registration.py
```diff
@@ -680,63 +310,190 @@ def make(id: Literal[
...
+                "plugin you shouldn't specify a namespace in `register` "
+                "calls. The namespace is specified through the "
+                "entry point package metadata."
+            )
+
+    latest_versioned_spec = max(
+        (
+            spec_
+            for spec_ in registry.values()
+            if spec_.name == spec.name and spec_.version is not None
```
pseudo-rnd-thoughts, line 343: Do we need to check that the `spec_.namespace == spec.namespace` as there may be two environments, one with a namespace and the other without. i.e. atari where we have `ALE/Pong-v5` and `Pong-v4` but if the versions were the same then could this be an issue?

gym/envs/registration.py
pseudo-rnd-thoughts: Very minor point, would `find_highest_version` be a more accurate name? Newest is possibly confusing as you could think the latest spec to be added

tests/envs/test_registration.py
```diff
@@ -2,9 +2,9 @@
 
 import gym
 from gym import envs, error
-from gym.envs import registration
+from gym.envs import register, registration, registry, spec
```
pseudo-rnd-thoughts, line 5: This shouldn't matter in implementation but in this test you sometimes do `envs.register` and sometimes `register` which got me confused. Could you only one of the method to make it consistent as I was getting confused thinking that they were different

arjun-kg:
Registering the exact same ID again now raises an error as opposed to overriding with a warning before. Is this intentional? (Would this be in issue for backward compatibility or would noone ever intentionally do this?)

Also, couple of things not related to the PR changes, but part of registration - 
1. The malformed env error is not human readable - `Currently all IDs must be of the form re.compile('^(?:(?P<namespace>[\\w:-]+)\\/)?(?:(?P<name>[\\w:.-]+?))(?:-v(?P<version>\\d+))?$')`.
2. The parser allows the id even if it's something like `Env-v1.3`, or `Env-v1-v1.2` putting everything into name with `version=None` . This might be an issue when someone registers envs as - `Env-v1.1` , `Env-v1.2`, `Env-v1.3` these all get registered as different names with no version without warning.

gym/envs/registration.py
RedTachyon: I'd need to be sure about the consequences of that. This code is just copy-paste of the old implementation, so I'll keep this outside the scope of this PR

gym/envs/registration.py
RedTachyon: Renamed

gym/envs/registration.py
RedTachyon: Good catch, fixed

tests/envs/test_registration.py
RedTachyon: Fixed

tests/envs/test_registration.py
RedTachyon: I'll keep this out of the scope of this PR for now, this was taken from the original code, and this PR is meant to primarily improve a specific aspect of registration, and keep the functionality otherwise. This can be updated later

RedTachyon:
@arjun-kg Thanks, I changed back the error-warning thing during registration. I changed the error message because that's fairly straight-forward and doesn't actually change functionality. Updating the parser would be a whole other mess, so keeping this outside the scope of this

RedTachyon:
I think I addressed everything, a few things that I'm deliberately keeping out of scope for now (so that the basic functionality can be merged faster) are:

- The parser regex might be a bit iffy with weirdly written versions
- Plugin system might require an extra warning when failing to import a plugin
- Some tests use a weird construct to emulate `with pytest.raises`, so they might need a revamp
- Add an extra env checker when doing `gym.make` (this would be a completely new feature, so I'd keep it for another PR - for now this just replicates the existing features)

gym/envs/registration.py
pseudo-rnd-thoughts: I tested for python 3.7 and without the `__future__` then the import fails however with the `__future__` I didnt have an issue

Markus28:
@RedTachyon what about the comment about `check_name_exists`? Why don't we need to filter the specs by namespace?

RedTachyon:
@Markus28 I don't think I see the comment?

Markus28:
@RedTachyon Hmm idk, maybe something weird is going on with GitHub, I can't link the comments and it doesn't show up in the review section. These were my comments, I am mostly concerned about the second one:
image
image

RedTachyon:
@Markus28 that's like the second time I'm seeing the same issue in a week (from another person), you entered these comments as a review and need to click a green button somewhere on top of the page to submit the review for them to actually show up

Markus28 (commented):
Thanks, forgot that

gym/envs/registration.py
```diff
@@ -78,558 +75,191 @@ def parse_env_id(id: str) -> Tuple[Optional[str], str, Optional[int]]:
...
+        return
 
-        return env
+    if namespace is None and name in internal_env_relocation_map:
+        relocated_namespace, relocated_package = internal_env_relocation_map[name]
+        message = f"The environment `{name}` has been moved out of Gym to the package `{relocated_package}`."
 
+        # Check if the package is installed
+        # If not instruct the user to install the package and then how to instantiate the env
+        if importlib.util.find_spec(relocated_package) is None:
```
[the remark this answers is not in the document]
RedTachyon: So I think this was copy-pasted verbatim from the [previous version of the code](https://github.com/openai/gym/blob/c6deb81ad3b6da7adbbffd4a2df2002a53861d8f/gym/envs/registration.py#L288), it's possible it doesn't work, but that's out of scope here

gym/envs/registration.py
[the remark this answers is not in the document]
RedTachyon: Good catch, I'll add the `if spec_.namespace == ns` condition

gym/envs/registration.py
```diff
@@ -680,63 +310,190 @@ def make(id: Literal[
...
-        finally:
-            if attr is None:
-                raise error.Error(
-                    f"Gym environment plugin `{module}` must specify a function to execute, not a root module"
-                )
+def register(id: str, **kwargs):
+    """
+    Register an environment with gym. The `id` parameter corresponds to the name of the environment,
+    with the syntax as follows:
+    `(namespace)/(env_name)-(version)`
```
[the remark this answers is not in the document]
RedTachyon: - Updated
- Good point, this actually breaks `with namespace`, I'll fix it
- Yep, I'll fix that too

gym/envs/registration.py
```diff
@@ -680,63 +310,190 @@ def make(id: Literal[
...
+            spec_
+            for spec_ in registry.values()
+            if spec_.namespace == spec.namespace
+            and spec_.name == spec.name
+            and spec_.version is None
+        ),
+        None,
+    )
+
+    if unversioned_spec and spec.version is not None:
```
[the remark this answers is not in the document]
RedTachyon: Fixed

PR #2749, opened 2022-04-13
Add mild domain randomization to Car Racing Env

jjshoots:
# Description

This adds a mild domain randomization trait to the Car Racing Env.

Simply pass `domain_randomize=True`, and the road and background colours will be randomized.

## Type of change

### Before 
before1

### After
after1
after2
after3

# Checklist:

### DIFF
--- a/gym/envs/box2d/car_racing.py
+++ b/gym/envs/box2d/car_racing.py
@@ -147,7 +147,12 @@ class CarRacing(gym.Env, EzPickle):
         "render_fps": FPS,
     }
 
-    def __init__(self, verbose=1, lap_complete_percent=0.95, hardcore=False):
+    def __init__(
+        self,
+        verbose: bool = True,
+        lap_complete_percent: float = 0.95,
+        hardcore: bool = False,
+    ):
         EzPickle.__init__(self)
         self.hardcore = hardcore
         self._init_colors()
@@ -301,7 +306,7 @@ class CarRacing(gym.Env, EzPickle):
             elif pass_through_start and i1 == -1:
                 i1 = i
                 break
-        if self.verbose == 1:
+        if self.verbose:
             print("Track generation: %i..%i -> %i-tiles track" % (i1, i2, i2 - i1))
         assert i1 != -1
         assert i2 != -1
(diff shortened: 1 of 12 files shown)


pseudo-rnd-thoughts:
Could you add a test for `hardcore=True`?

jjshoots:
> 

Yea I couldn't find any tests for CarRacing, and I agree with you that there should be. I'll get to implementing one then.

gym/envs/box2d/car_racing.py
```diff
@@ -183,6 +193,22 @@ def _destroy(self):
...
+            self.norm_road_color = self.np_random.uniform(0, 0.8, size=3)
+
+            self.bg_color = self.np_random.uniform(0, 210, size=3)
+
+            self.grass_color = np.copy(self.bg_color)
+            idx = self.np_random.integers(3)
+            self.grass_color[idx] += 20
+        else:
+            # default colours
+            self.norm_road_color = np.array([0.4, 0.4, 0.4])
```
RedTachyon, line 208: Do we have to store the colors with two different representations? (one is 0-1, the other is 0-255)

gym/envs/box2d/car_racing.py
```diff
@@ -128,10 +125,15 @@ class CarRacing(gym.Env, EzPickle):
...
-    There are no arguments supported in constructing the environment.
+    `lap_complete_percent` dictates the percentage of tiles that must be visited by
+    the agent before a lap is considered complete.
+
+    Passing `hardcore=True` enabled the domain randomized variant of the environment.
+    In this scenario, the background and track colours are different on every reset.
 
     ### Version History
-    - v0: Current version
+    - v1: Current version (0.23.1)
```
RedTachyon, line 135: I guess v0 can be labelled as the original version; v1 will come live with 0.24.0 at the earliest, so probably better to put that (seeing as 0.23.1 is already out)

RedTachyon (commented):
Overall looks good, just left one nitpicky comment for code quality

Actually, one thought. Would it make sense to name this something else than `hardcore`? I know there are some envs that already follow this naming pattern, but it's extraordinarily undescriptive.

Also - we might want to register an extra env to make it easier to create this version, something like "CarRacingHardcore" or whatever Hardcore might get renamed to.

gym/envs/box2d/car_racing.py
jjshoots: Gotcha, fixed it

jjshoots:
@RedTachyon I've changed `hardcore` to `domain_randomize`, hopefully that's more descriptive, also registered it.

Issue #2718, opened 2022-03-25
ImportError: cannot import name 'Monitor' from 'gym.wrappers'


lu0909:
The error message is as follows:
Traceback (most recent call last):
  File "F:\xx\rsa-rl\code\rsa-rl-tutorials-master\rsa-rl-tutorials-master\examples\ksp-agent.py", line 8, in <module>
    from rsarl.envs import DeepRMSAEnv
  File "<frozen zipimport>", line 259, in load_module
  File "C:\Users\xx\.conda\envs\pytorch2\lib\site-packages\rsarl-1.0.0-py3.9.egg\rsarl\__init__.py", line 2, in <module>
  File "<frozen zipimport>", line 259, in load_module
  File "C:\Users\xx\.conda\envs\pytorch2\lib\site-packages\rsarl-1.0.0-py3.9.egg\rsarl\utils\__init__.py", line 3, in <module>
  File "<frozen zipimport>", line 259, in load_module
  File "C:\Users\xx\.conda\envs\pytorch2\lib\site-packages\rsarl-1.0.0-py3.9.egg\rsarl\utils\random_seed.py", line 2, in <module>
  File "<frozen zipimport>", line 259, in load_module
  File "C:\Users\xx\.conda\envs\pytorch2\lib\site-packages\pfrl-0.3.0-py3.9.egg\pfrl\__init__.py", line 20, in <module>
  File "<frozen zipimport>", line 259, in load_module
  File "C:\Users\xx\.conda\envs\pytorch2\lib\site-packages\pfrl-0.3.0-py3.9.egg\pfrl\wrappers\__init__.py", line 4, in <module>
  File "<frozen zipimport>", line 259, in load_module
  File "C:\Users\xx\.conda\envs\pytorch2\lib\site-packages\pfrl-0.3.0-py3.9.egg\pfrl\wrappers\monitor.py", line 4, in <module>
ImportError: cannot import name 'Monitor' from 'gym.wrappers' (C:\Users\xx\.conda\envs\pytorch2\lib\site-packages\gym-0.23.1-py3.9.egg\gym\wrappers\__init__.py)

**System Info**
Describe the characteristic of your environment:
 *Gym was installed using pip in anaconda prompt.
 *  OS/version is windows10. 
 * Python version is python3.9.

**Additional context**

### Checklist

- [ ] I have checked that there is no similar [issue](https://github.com/openai/gym/issues) in the repo (**required**)

balisujohn:
[...]

Issue #2709, opened 2022-03-21
[bug] pip install gym[atari] - insufficient dependencies


ikamensh:
**Describe the bug**
installing gym with atari extras doesn't install `pygame` but it's still imported (erroring out) during render() call of atari env.

**Code example**
`pip install gym[atari]`
`python -c "import gym; env=gym.make('Pong-v4'); env.render()"`

Output:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "python3.8/site-packages/gym/core.py", line 286, in render
    return self.env.render(mode, **kwargs)
  File "python3.8/site-packages/gym/core.py", line 286, in render
    return self.env.render(mode, **kwargs)
  File "python3.8/site-packages/gym/envs/atari/environment.py", line 264, in render
    from gym.envs.classic_control import rendering
  File "python3.8/site-packages/gym/envs/classic_control/__init__.py", line 1, in <module>
    from gym.envs.classic_control.cartpole import CartPoleEnv
  File "python3.8/site-packages/gym/envs/classic_control/cartpole.py", line 10, in <module>
    import pygame
ModuleNotFoundError: No module named 'pygame'
```

**System Info**
pip freeze                                                                                                 (matari) 11:09:59
ale-py==0.7.4
AutoROM==0.4.2
AutoROM.accept-rom-license==0.4.2
gym==0.23.1
gym-notices==0.0.6
idna==3.3
numpy==1.21.2
requests==2.26.0
torch==1.11.0

python -V                                                                                                      (matari) 11:10:22
Python 3.8.5

** Suggested solution **

Either change import structure or include pygame into dependencies associated with "atari" extra in setup.py

ikamensh:
Reported it to ALE package https://github.com/openai/gym/issues/2709 , probably more relevant there.

PR #2752, opened 2022-04-14
New Step API with terminated, truncated bools instead of done

arjun-kg:
# Description
`step` method is changed to return five items instead of four. 

Old API - `done=True` if episode ends in any way.

New API -
`terminated=True` if environment terminates (eg. due to task completion, failure etc.)
`truncated=True` if episode truncates due to a time limit or a reason that is not defined as part of the task MDP

Link to docs - https://github.com/Farama-Foundation/gym-docs/pull/115 (To be updated with latest changes) 

### Changes
1. All existing environment implementations are changed to new API without direct support for old API. However gym.make for any environment will default to old API through a compatibility wrapper. 
2. Vector env implementations are changed to new API, with backward compatibility for old API, defaulting to old API. New API can set by a newly added argument `new_step_api=True` in constructor. 
3. All wrapper implementations are changed to new API, and have backward compatibility and default to old API (can be switched to new API with `new_step_api=True`). 
4. Some changes in phrasing - `terminal_reward`, `terminal_observation` etc. is replaced with `final_reward`, `final_observation` etc. The intention is to reserve the 'termination' word for only if `terminated=True`. (for some motivation, Sutton and Barto uses terminal states to specifically refer to special states whose values are 0, states at the end of the MDP. This is not true for a truncation where the value of the final state need not be 0. So the current usage of `terminal_obs` etc. would be incorrect if we adopt this definition)
6. All tests are continued to be performed for old API (since the default is old API for now). A single exception for when the test env is unwrapped and so the compatibility wrapper doesn't apply. Also, special tests are added just for testing new API. 
7. `new_step_api` argument is used in different places. It's meaning is taken to be "whether this function / class should **output** step values in new API or not". Eg. `self.new_step_api` in a wrapper signifies whether the wrapper's step method outputs items in new API (the wrapper itself might have been written in new or old API, but through compatibility code it will output according to `self.new_step_api`) 
8. `play.py` alone is retained in old API due to the difficulty in having it be compatible for both APIs simultaneously, and being slightly lower priority. 

#### StepAPICompatibility Wrapper
1. This wrapper is added to support conversion from old to new API and vice versa.
2. Takes `new_step_api` argument in `__init__`. False (old API) by default.
3. Wrapper applied at make with  `new_step_api=False` by default. It can be changed during make like `gym.make("CartPole-v1", new_step_api=True)`. The order of wrappers applied at make is as follows - core env -> PassiveEnvChecker -> StepAPICompatibility -> other wrappers

#### step_api_compatibility function
This function is similar to the wrapper, it is used for backward compatibility in wrappers, vector envs. It is used at interfaces between env / wrapper / vector / outside code. Example usage, 

```python
# wrapper's step method
def step(self, action):

    # here self.env.step is made to return in new API, since the wrapper is written in new API
    obs, rew, terminated, truncated, info = step_api_compatibility(self.env.step(action), new_step_api=True) 

    if terminated or truncated:
        print("Episode end")
    ### more wrapper code

    # here the wrapper is made to return in API specified by self.new_step_api, that is set to False by default, and can be changed according to the situation
    return step_api_compatibility((obs, rew, terminated, truncated, info), new_step_api=self.new_step_api) 
```

#### TimeLimit 
1. In the current implementation of the timelimit wrapper, existence of `'TimeLimit.truncated'` key in `info` means that truncation has occurred. The boolean value it is set to refers to whether the core environment has already ended. So, `info['TimeLimit.truncated']=False`, means the core environment has already terminated. We can infer `terminated=True`, `truncated=True` from this case. 
2. To change old API to new, the compatibility function first checks info. If there is nothing in info, it returns `terminated=done` and `truncated=False` as there is no better information available. If TimeLimit info is available, it accordingly sets the two bools.

#### Backward Compatibility
The PR attempts to achieve almost complete backward compatibility. However, there are cases which haven't been included. Environments directly imported eg. `from gym.envs.classic_control import CartPoleEnv` would not be backward compatible as these are rewritten in new API. `StepAPICompatibility` wrapper would need to be used manually in this case. Envs made through `gym.make` all default to old API. Vector and wrappers also default to old API. These should all continue to work without problems. But due to the scale of the change, bugs are expected. 

#### Warning Details
Warnings are raised at the following locations:
1. `gym.Wrapper` constructor - warning raised if `self.new_step_api==False`.  This means any wrapper that does not explicitly pass `new_step_api=True` into super() will raise the warning since `self.new_step_api=False` by default. This is taken care of by wrappers written inside gym. Third party wrappers will face a problem in a specific situation - if the wrapper is not impacted by step API. eg. a wrapper subclassing ActionWrapper. This would work without any change for both APIs, however to avoid the warning, they still need to pass `new_step_api=True` into super(). The thinking is - "If your wrapper supports new step API, you need to pass `new_step_api=True` to avoid the warning". 
2.  `PassiveEnvChecker`, `passive_env_step_check` function - if step return has 4 items a warning is raised. This happens only once since this function is only run once after env initialization. Since `PassiveEnvChecker` is wrapped first before step compatibility in make, this will raise a warning based on the core env implementation's API. 
3. `gym.VectorEnv` constructor - warning raised if `self.new_step_api==False`. 
4. `StepAPICompatibility` wrapper constructor - the wrapper that is applied by default at make. If `new_step_api=False`, a warning is raised. This is independent of whether the core env is implemented in new or old api and only depends on the `new_step_api` argument. 

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist:

- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `pre-commit run --all-files` 
- [ ] I have made corresponding changes to the documentation (need to update with latest changes)
- [x] My changes generate no new warnings (only intentional warnings)
- [ ] I have added tests that prove my fix is effective or that my feature works (added two tests but maybe needs more to be comprehensive)
- [x] Locally runs with atari / pybullet envs


gym/core.py
```diff
@@ -325,8 +338,13 @@ def reset(self, **kwargs):
...
     def step(self, action):
-        observation, reward, done, info = self.env.step(action)
-        return self.observation(observation), reward, done, info
+        step_returns = self.env.step(action)
+        if len(step_returns) == 5:
+            observation, reward, terminated, truncated, info = step_returns
+            return self.observation(observation), reward, terminated, truncated, info
+        else:
+            observation, reward, done, info = step_returns
+            return self.observation(observation), reward, done, info
```
XuehaiPan, line 347:
```suggestion
        observation, *rests = self.env.step(action)
        return (self.observation(observation), *rests)
```

gym/core.py
```diff
@@ -338,8 +356,13 @@ def reset(self, **kwargs):
...
     def step(self, action):
-        observation, reward, done, info = self.env.step(action)
-        return observation, self.reward(reward), done, info
+        step_returns = self.env.step(action)
+        if len(step_returns) == 5:
+            observation, reward, terminated, truncated, info = step_returns
+            return observation, self.reward(reward), terminated, truncated, info
+        else:
+            observation, reward, done, info = step_returns
+            return observation, self.reward(reward), done, info
```
XuehaiPan, line 365:
```suggestion
        observation, reward, *rests = self.env.step(action)
        return (observation, self.reward(reward), *rests)
```

pseudo-rnd-thoughts:
Im thinking about the backward compatibility of this PR, in particular for online tutorials and uni courses, etc 
As `return_two_dones=True` is the default for all environments, will this not break a huge amount of code without a warning being raised as it is the expected value now. 
Is it not better to have `return_two_dones=False` for all environments until 1.0 with the warnings being raised explaining to people to change across and that all wrappers have been modified to expect this now?

arjun-kg:
@pseudo-rnd-thoughts  Yes it will break code. But my point to making new API the default is just to make the code a lot simpler, with old API still supported for some (but not all) features. Most wrappers and vector code only support new API. It would get way too convoluted for all that to support both APIs. We'd need to support any code that could've arbitrarily applied any wrapper or vector code. 

And for the code breaking, the fix is to apply the compatibility wrappers manually which is a little better than immediately having to rewrite the broken code but yes, definitely not ideal. Maybe it's worth the complicated code to retain complete compatibility, at this point I'm not sure which way is better and wouldn't mind either way based on a consensus.

gym/wrappers/step_compatibility.py
```diff
@@ -0,0 +1,69 @@
...
+            terminated = True
+            truncated = True
+
+        return obs, rew, terminated, truncated, info
+
+    def _step_returns_new_to_old(self, step_returns):
+        assert len(step_returns) == 5
+        logger.warn(
+            "Using a wrapper to transform new step API (which returns two booleans terminated, truncated) into old (returns one boolean done). "
+            "This wrapper will be removed in the future. "
```
XuehaiPan, line 59:
```suggestion
            "This wrapper will be removed in Gym 1.0. "
```

@arjun-kg 
I think @pseudo-rnd-thoughts's point is warn for the exact version rather than _"in the future"_.

pseudo-rnd-thoughts:
@arjun-kg So I understand your perspective that we should try to not support one done returns and I agree with the idea. 

But this is a massively backward compatibility breaking change that I think we should avoid until gym 1.0
Personally, I think that the default should be one done return *with* a warning being raised that all wrappers have been modified to expect environments with two done returns. Plus details that this will be changed to two done returns at gym 1.0 plus the compatibility wrappers to change environments from one to two and vice versa. 

The additional issue I see with having two done returns as default is that as *no* warning is raised to alert users to the change then code will start failing with no obvious reason without them looking at the change logs which I suspect a majority of people are not going to do. Therefore, we are going to get a massive increase in issues around this PR.

gym/wrappers/step_compatibility.py
pseudo-rnd-thoughts: Sort of, it was more around the default two done parameter at gym.make()

arjun-kg:
@pseudo-rnd-thoughts Oh that makes sense. We could have old api as default with warning saying wrappers and vector api have been upgraded and will fail unless new api is used, and maybe guide them to a doc to see how to switch api (or use compatibility wrapper) for whatever is not available in the old api. So yes like you said, it won't be an error without warning.

pseudo-rnd-thoughts:
Thanks, could you add a test that checks that without `return_two_dones=True` then the warning is raised.

arjun-kg:
@pseudo-rnd-thoughts I've made the changes. All the tests assume new api, so argument `return_two_dones=True` is passed for all env making in tests.

pseudo-rnd-thoughts:
@arjun-kg Thanks for doing that, the workflows haven't happened but I imagine that we are going to get a massive number of deprecation warnings. 
If true, would you these be able to silence these warnings in the CI, you can pass a parameter to pytest to ignore certain warnings or add a `@pytest.mark.silencewarnings` (I believe) to all of the test functions. The first opinion is probably best.

gym/core.py
```diff
@@ -61,12 +61,17 @@ def np_random(self, value: RandomNumberGenerator):
         self._np_random = value
 
     @abstractmethod
-    def step(self, action: ActType) -> Tuple[ObsType, float, bool, dict]:
+    def step(
+        self, action: ActType
+    ) -> Union[
+        Tuple[ObsType, float, bool, bool, dict], Tuple[ObsType, float, bool, dict]
```
RedTachyon, line 67: I don't know if I like this approach to backwards compatibility. If this is the official state of (for example) 0.24.0, then you can't reliably write an algorithm that will work for all valid 0.24.0 environments. I think we should just say that an environment should have the signature of (ObsType, float, bool, bool, dict), and then provide a wrapper-like compatibility layer that can convert an old-style environment to a new-style environment.

gym/core.py
```diff
@@ -76,13 +81,17 @@ def step(self, action: ActType) -> Tuple[ObsType, float, bool, dict]:
         Returns:
             observation (object): agent's observation of the current environment. This will be an element of the environment's :attr:`observation_space`. This may, for instance, be a numpy array containing the positions and velocities of certain objects.
             reward (float) : amount of reward returned after previous action
-            done (bool): whether the episode has ended, in which case further :meth:`step` calls will return undefined results. A done signal may be emitted for different reasons: Maybe the task underlying the environment was solved successfully, a certain timelimit was exceeded, or the physics simulation has entered an invalid state. ``info`` may contain additional information regarding the reason for a ``done`` signal.
+            terminated (bool): whether the episode has ended due to a termination, in which case further step() calls will return undefined results
```
RedTachyon, line 84: I would rephrase "termination" to something like "reaching a terminal state", or otherwise to indicate that it's about the intrinsic properties of the environment

gym/envs/registration.py
```diff
@@ -96,6 +96,7 @@ class EnvSpec:
     max_episode_steps: Optional[int] = field(default=None)
     order_enforce: bool = field(default=True)
     autoreset: bool = field(default=False)
+    return_two_dones: bool = field(default=False)
```
RedTachyon, line 99: I don't think this will actually be useful. It only really has any value if people are explicitly setting it, which is a new (and very temporary) feature, and we explicitly deprecate one of the settings the moment it's introduced. This goes back to the general comment that I think the library should only support one step semantic, and compatibilities can be done through a wrapper-like thing

gym/wrappers/step_compatibility.py
```diff
@@ -0,0 +1,76 @@
+import gym
+from gym import logger
+
+
+class StepCompatibility(gym.Wrapper):
```
RedTachyon, line 5: This desperately needs some comments and docstrings

gym/wrappers/step_compatibility.py
```diff
@@ -0,0 +1,76 @@
...
+    def __init__(self, env, return_two_dones=False):
+        super().__init__(env)
+        self._return_two_dones = return_two_dones
+        if not self._return_two_dones:
+            logger.deprecation(
+                "[StepAPI] Initializing environment in old step API which returns one bool instead of two. "
```
RedTachyon, line 11: The [StepAPI] thing implies there's a syntax like that in gym overall, which is not the case as far as I know, so I think it can be removed

RedTachyon (commented):
My single biggest concern is the approach taken to backwards compatibility, which makes a lot of the code needlessly complex. I'm of the moderately strong opinion that all the code inside of gym after this PR is merged, should explicitly only support the new step semantic. 

Since we definitely want environments written with `obs, reward, done, info` to still work, we can have a wrapper doing something like

```python
class OldToNew:
    ...
    def step(self, action: ActType) -> Tuple[ObsType, float, bool, bool, dict]:
        outputs = self.env.step(action)
        if len(outputs) == 4:
            obs, reward, done, info = outputs
            return obs, reward, done, False, info
        elif len(outputs) == 5:
            # We applied the wrapper needlessly
            return outputs
        else:
            raise ValueError("You messed something up real hard")
            
            
class NewToOld:
    ...
    def step(self, action: ActType) -> Tuple[ObsType, float, bool, bool, dict]:
        outputs = self.env.step(action)
        if len(outputs) == 4:
            # We applied the wrapper needlessly
            return outputs
        elif len(outputs) == 5:
            obs, reward, terminated, truncated, info = outputs
            return obs, reward, terminated or truncated, info
        else:
            raise ValueError("You messed something up real hard")
```

Then we apply this as appropriate, giving the users some control over what will be applied, and all the code should be compatible both ways as long as environments weren't trying to do their own truncations and just used TimeLimitWrapper instead.

A bunch of individual code comments are about this exact issue before I realized that's like the only big thing I have to comment about here.

pseudo-rnd-thoughts:
@RedTachyon I believe the `StepCompatibility` wrapper in `gym/wrappers/step_compatibility` and `gym/vector/step_compatibility` are exactly what you describe as `OldtoNew` and `NewtoOld`
Currently, all environment created using `.make` use the backward compatibility wrapper with one done but raise a warning to users about the two done API and the changes made. 
To use the two done API, user just add the `return_two_done=True` as a kwargs for `.make`

RedTachyon:
Good point. The thing is that all the internal code supports both old and new APIs, and I think we should stick to just one, and old-style environments will be "projected" into the new API and processed as such. I don't think there should be code in gym (outside of the explicit compatibility layer) which checks `if len(outputs) == 4: ... else: ...`

gym/envs/registration.py
arjun-kg: Do you mean a wrapper that's always applied at make is preferable to an argument? Or no wrapper at make? My point was to have this to allow backward compatibility and, if users so chose to use the new api, they could just switch it to True at make.

arjun-kg:
> I don't think there should be code in gym (outside of the explicit compatibility layer, and maybe some tests) which checks if len(outputs) == 4: ... else: ...

Right now, only core and TimeLimit has this kind of check. The only other place where there are details about compatibility is in registration and vector init, applying the compatibility wrapper, and giving the user an argument to toggle between new and old api at make. Everything else is switched cleanly to the new API.

gym/envs/registration.py
RedTachyon: I think I just find it confusing as to what this is supposed to do. Does this indicate what the environment intrinsically returns? Does it decide whether a wrapper should be applied?

I'd prefer a wrapper that's always applied, unless explicitly turned off, which defaults to a no-op if the environment already follows the new API

RedTachyon:
I'd remove it from core at least in that case. Probably also from TimeLimit, and then the compatibility wrapper would be applied as the very first wrapper, so that TimeLimit only has to ever operate on new-API envs

RedTachyon:
Or can you maybe explain what is the exact workflow in these situations? The name `return_two_renders` is very unclear to me.
1. I'm calling `gym.make("AntBulletEnv-v0")` on an env written with the old API, and want it to follow the new API
2. I'm calling the same thing, want it to follow the old API (is this possible?)
3. I'm creating a brand well-written new env, I want it to follow the new API
4. I'm creating the same thing, I want it to follow the old API (is this possible?)

arjun-kg:
@RedTachyon 
So the compatibility wrapper now is always applied and switches to old api no matter what the style of the core env. `return_two_dones` essentially means whether the wrapped env is in new API or not (irrespective of the input env). The compatibility wrapper is always applied at make. and `return_two_dones=False` by default so no matter what env is sent in, it switches to old API by default unless `return_two_dones` is specified.  For your questions, all four scenarios are possible - 

1. `gym.make("AntBulletEnv-v0", return_two_dones=True)` - the old api is explicitly converted to new api through the wrapper
2. `gym.make("AntBulletEnv-v0")`  - the wrapper is present but passes through without changes. This also ensures old api envs are backward compatible. 
3. `gym.make("BrandNewAPIEnv-v0", return_two_dones=True)` - the wrapper is present but passes through without changes
4. `gym.make("BrandNewAPIEnv-v0")` - the new api is explicitly converted to old api through the wrapper. This ensures all the envs in gym that are in new api (eg. CartPole) are backward compatible. Codes already using CartPole will switch to the old API with a warning. Ofc the minor inconvenience is that, for people who completely switch to new API, they will still have to send the `return_two_dones=True` argument till 1.0. 

The reason I had TimeLimit be compatible with both APIs was for it to work smoothly in all the above scenarios. The above envs would first be wrapped by `StepCompatibility` wrapper, then wrapped by order enforcing and then timelimit and work smoothly irrespective of how the environments were transformed. 

Would it be clearer if the `return_two_dones` was replaced by `new_step_api` argument?

lebrice:
Am I correct in assuming that an episode cannot possibly be both terminated and truncated?
If so, then aren't you using two bools to describe only 3 possible states?

Assuming this is correct, I'd suggest another alternative, which I believe might be more backward-compatible:
- What about returning something other than `True` or `False` for `done` in case where it has been truncated? (e.g., the string "truncated", or some Enum value?)

For example:
```python
RUNNING = 0
DONE = 1
TRUNCATED = 2
```
This would slightly alter the Env API:
```python
class Env(Generic[Observation, Action]):
    def step(self, action: Action) -> tuple[Observation, Action, int, dict]: ...
```
Also note, weirdly enough `issubclass(bool, int)` is always true in Python, so this would technically be fully backward-compatible with the current API. 

I believe this has the following benefits:
- All the code on GitHub that depends on `if done: (...)` wouldn't be affected, since `bool(2) == bool(1) == True`
- the number of outputs of `step` would stay the same
- wrappers could be updated incrementally, instead of having to update all the wrappers at the same time.

This might require a tiny bit of refactoring in the VectorEnv api, but not much: just changing the dtype of the `done` array from `bool` to something like `uint8`.


Also, another question: Is the TimeLimit wrapper the only one that would produce this? Or are all envs expected to return this second boolean in case they were truncated?

lebrice:
Actually, now that I think about it, keeping the bools is also a funky, but totally valid option (since bool subclases int!). This might make this even more backward-compatible:
```python
RUNNING = False
DONE = True
TRUNCATED = 2
```

arjun-kg:
@lebrice  thank you for your comment. So what you suggest was a big part of the discussion (see https://github.com/openai/gym/issues/2510#issuecomment-1044546882). I think the general idea is that two bools with explicit meaning is more clear than `0, 1, 2`, especially for beginner users. For eg. for bootstrapping we simply look at `terminated` bool and don't worry about anything else. And it's also a question of future clarity and ease of use vs backward compatibility. 

> Am I correct in assuming that an episode cannot possibly be both terminated and truncated?

I believe this is possible, for eg. the robot reaches a terminal state, and that coincides with a timelimit on the exact same step. Although, arguably we can just call this terminated and we don't need the information that the episode has also truncated on the same step (maybe we might need it for some practical purpose?). But as it stands, all four states are legal. 

> Also, another question: Is the TimeLimit wrapper the only one that would produce this? Or are all envs expected to return this second boolean in case they were truncated?

So, returning two bools is the step signature here, so all envs are expected to return the second bool. This is less messy than returning different number of items when it goes through the TimeLimit wrapper. This also allows for defining truncation conditions from inside the core environment definition if someone wishes to.

araffin:
> Am I correct in assuming that an episode cannot possibly be both terminated and truncated?
If so, then aren't you using two bools to describe only 3 possible states?

> So what you suggest was a big part of the discussion (see https://github.com/openai/gym/issues/2510#issuecomment-1044546882). I think the general idea is that two bools with explicit meaning is more clear than 0, 1, 2, especially for beginner users.

Those were already discussed in https://github.com/openai/gym/issues/2510#issuecomment-1068026884, and I would disagree on the fact that is more clear: it was not clear between people implementing the feature https://github.com/openai/gym/issues/2510#issuecomment-1091758558, so I guess that would be worse for beginners...

> This is less messy than returning different number of items when it goes through the TimeLimit wrapper. 

The number of items is the same, it is more the presence or the absence of `truncated` in the info dict that tells the difference (we could also set the key at every step if that's the concern). 
Again, TimeLimit is the one responsible for the truncation in more than 90% of the cases (currently 100% for built-in envs, car racing needs to be updated), and is the supported option by different libraries: https://github.com/openai/gym/issues/2510#issuecomment-1100861903
Using two booleans, you will always return `truncated=False` when writing a custom env (as done in this PR), which is not really beginner friendly.

>  This also allows for defining truncation conditions from inside the core environment definition if someone wishes to.

Again, this is not true, you can always set a key in the info dict (I agree that the current name of the key should be improved for clarity).

Before merging this PR, please pause and ponder, as this is a massive breaking (changing the core signature of Gym) that does not seem necessary (the problem it tries to address can be solved without breaking the API): https://github.com/openai/gym/issues/2510#issuecomment-1100861903

jkterry1:
@araffin My reasoning for wanting to do this has been fairly scattered given the number of PRs and issues and meetings that have surrounded this change, but it is as follows:

- Most environments in RL involve truncation of some sort, and likely will continue to do so indefinitely
- You cannot correctly reimplement DQN or other things in an environment that truncated without incorporating knowledge of if a state is truncated in your learning code, leading to massive issues with paper reproductions. This is a major reason why Google Brain researchers often refuse to use Gym.
- While this is provided in infos, this was not consistently done in third party environments or used in learning code. Many users don't even know of the existence of the difference in my experience. It is my view that such an important feature for correct implementation has to explicitly be part of the main API in some manner in order for this to be adequate. My concern however is that this makes a crucial feature nonexplicit, as you have to read the docs to know it exists, what it means, or what to correctly handle it. As you know from your GitHub issues and the general reproducibility crisis in RL, people don't generally do this.

In essence, I am of the opinion that even well documented nonexplicit support for something this important is not enough. 


As far as the options for making this explicit go, there are two main ones:

1. Xuehai's proposal for changing dones to [0,1,2] (which if I understand correctly you prefer)
2. This way of having two booleans

Xuehai's method certainly has easier backwards compatibility, but would require all the users to read the documentation in detail for it to do any actual good, which is my main objection. I also have consistently differed in your view on the merits of breaking changes in general --- I feel they're worth it in the pursuit of a stable API at 1.0 given that Gym will be used far more in the next 5 years than it ever has been before.

I also think that the specific argument you're making about this approach being unclear because it was unclear to the maintainers is unfair. The issue there was that we changed the plan from `done, truncated` to `terminated, truncated`, and this wasn't properly communicated to all parties involved. I do agree that `done, truncated` is ambiguous, but I don't see how this could apply to `terminated, truncated`, hence the change. This approach should also be identically clear compared to the [0,1,2] approach, since it's using the same encoding.

I also think that the backwards compatibility issues here will be mitigated by this change being off by default with a warning for users to upgrade their code to turn it on for some number of months, as well as a forthcoming compatibility wrapper that will let old environments and learning code work with it.

However, if you disagree with any of these points I'd be curious to hear your concerns. The one concern I will admit to having is making absolutely beginners to RL learn what truncation means in order to write code that runs, however tutorials could just encourage users to add a single line that `or`s the two values and ignore the distinction?

lebrice:
Are there other potential ways to have an episode be truncated than by using the TimeLimit wrapper or calling `reset` manually before the episode has actually ended?

lebrice:
> Xuehai's method certainly has easier backwards compatibility, but would require all the users to read the documentation in detail for it to do any actual good, which is my main objection

If you don't mind me giving my 2 cents @jkterry1, I disagree:

- If, as a user, you care about the difference between truncation vs termination, then learning that done=2 means truncation (or learning that `gym.TRUNCATED == 2`) is very easy.

- On the other hand, requiring every RL practionner, student, and teacher out there to, at some point, update their examples, dig up their old code, just to insert two booleans that are used to describe 3 possible states, definitely seems like a worse alternative to me.

> You cannot correctly reimplement DQN or other things in an environment that truncated without incorporating knowledge of if a state is truncated in your learning code, leading to massive issues with paper reproductions. This is a major reason why Google Brain researchers often refuse to use Gym

This proposed change would quite literally break **all the gym-based DQN implementations on GitHub**. Are you certain that  that wouldn't hurt reproducibility more than the problem it is trying to fix?

jkterry1:
"Are there other potential ways to have an episode be truncated than by using the TimeLimit wrapper or calling reset manually before the episode has actually ended?"
- Not really, though you can use your own TimeLimit wrapper, for instance the wrapper will be moved into C logic inside the Arcade Learning Environment in the near future. 

Regarding the rest of your points (if I understand correctly, your only problem with the two boolean approach is backwards compatibility not API design):
- The old code you would have to update for this would already have to be updated given the previous and planned changes in Gym, and this isn't even a major or hard update to make compared to other previous or uncontested upcoming breaking changes. The goal is to get Gym to a state that's suitable to be stable semi-permanently.
- Like I mentioned, a compatibility wrapper will be introduced for all old environment and learning code

pseudo-rnd-thoughts:
Looking at the deepmind environment spec, they have a worse approach that is less easy to understand, IMO, compared to this solution. They have a step type (First, Mid, Last) and a discount factor allowing them to have a discount > 0 for truncated time steps which to me is a more confusing approach to thinking about what is theoretically happening compared to this PR. 
https://github.com/deepmind/dm_env/blob/master/docs/index.md

Personally, I think with some high-quality documentation to explain how to implement this with examples for training and running + upgraded stable baselines and other projects to show people this should not be a major issue, should ensure that RL algorithms are being implemented correctly. 
Personally, I was not aware of this bug in my own code until I saw this PR and was not taught about it in my degree. 
Though of course, people could run on an old version of gym if necessary.

araffin:
@jkterry1 

> Most environments in RL involve truncation of some sort, and likely will continue to do so indefinitely

yes

> You cannot correctly reimplement DQN or other things in an environment that truncated without incorporating knowledge of if a state is truncated in your learning code

DQN is not the main issue but yes (most impact I've seen so far is PPO/SAC on continuous control tasks)

> While this is provided in infos, this was not consistently done in third party environments or used in learning code. 

yes, and there is an easy fix to that: enforcing the presence of that key in the info dict (can be done the same way the "reset needed" is done), throwing a warning and then an error if not present, and showing its use in all basic examples of gym.

> as you have to read the docs to know it exists

well, you have to read the doc anyway to learn how to use `env.step()` (as mentioned above, we would include it in all examples).

>  well documented nonexplicit support

If you enforce its presence and show its usage in all examples, I call that explicit. And even with the proposed approach, the user still need to read the doc to know what to do with that additional boolean. 

> I do agree that done, truncated is ambiguous, but I don't see how this could apply to terminated, truncated, hence the change.
> and this wasn't properly communicated to all parties involved.

I've already answered that here: https://github.com/openai/gym/issues/2510#issuecomment-1091758558
More importantly, I asked the question at first because this was not clear to me and I wanted to check I wasn't alone (be it `done` vs `truncated` or `terminated` vs `truncated`, the name change doesn't make a difference to me).

> Not really, though you can use your own TimeLimit wrapper,

This is why I'm proposing a more general solution, because having `truncated=False` all the time in the env may be confusing (truncation is handled by `TimeLimit` wrapper in the vast majority of cases) , that should also make more sense: adding termination reason in the info dict as best practice.

>  for this would already have to be updated given the previous and planned changes in Gym, and this isn't even a major or hard update to make compared to other previous or uncontested upcoming breaking changes.

"we are already making breaking changes" should not be an argument for more breaking changes...
I'm afraid that we cannot find bigger breaking change than this one, in the core API of Gym.

Even for smaller breaking changes, it takes time to update not only code but documentation and tutorials too.
Each breaking change also causes confusion and frustration for users, so they should be done only when mandatory:
- https://github.com/DLR-RM/stable-baselines3/issues/875
- https://github.com/ray-project/ray/issues/23286
- https://github.com/openai/gym/issues/2780
- https://github.com/openai/gym/issues/2718
- https://github.com/openai/gym/issues/2674
- https://github.com/openai/gym/issues/2781
- https://github.com/openai/gym/issues/2709

pseudo-rnd-thoughts:
@araffin Thanks for your comments

If I understand your argument then
1. Yes this problem is important as it effect agent performance (in particular PPO and SAC you note)
2. We don't need to change the specification because people read the `env.step()` and therefore know about the truncation problem and modify their training scripts appropriately. 
3. Terminated and Truncated are not clear in what they mean therefore will confuse users. 
4. We should not make large breaking changes to the core API of gym as this can cause confusion and difficulties for years to come. 

I hope I have correctly understood and outlined your arguments, if not could you correct me.

My response is
1. This is not only a performance issue but an implementation issue for 99% reinforcement learning algorithms and could further affect performance. This is an issue in academia, in particular, for replicating other work accurately is critical to the health of the field.
2. For me, I have worked in RL for 5 years and never heard of this issue before. Talking to four of my PhD friends, all active researchers in RL, none of them has heard of this issue as well. Just to check, I found the first four 'dqn tutorials' from google [pytorch](https://pytorch.org/tutorials/intermediate/reinforcement_q_learning.html), [tensorflow](https://www.tensorflow.org/agents/tutorials/1_dqn_tutorial), [towardsdatascience](https://towardsdatascience.com/deep-q-learning-tutorial-mindqn-2a4c855abffc), [seba1511](http://seba1511.net/tutorials/intermediate/reinforcement_q_learning.html). Looking at all of them, the only one that implemented this correct is TensorFlow but because they use `suite_gym` not `gym` so a discount factor is returned not done. In the pytorch tutorial, they comment out the `info` parameter from `env.step` so not even the top dqn tutorial on the internet has the correct implementation!! I could repeat this with uni courses, GitHub projects, etc but I think this shows the general point that a majority of implements are not correct even if the documentation talks about it. Therefore, for me, this is a critical issue that needs to be fixed and the previous way of just using documentation has not fixed it. 
3. You are probably right that this will cause a lot of confusion for people originally, but we will be making a lengthy blog post explaining the theoretical and practical changes this makes. Which should help clear up the issue and allow people to comment. However, for me, point 2 is such a big issue that **needs** to be addressed then, we need to change something
4. You have linked a number of areas where the changes have caused confusion and broken some projects and so are right to worry about this breaking / causing confusion. With this change, it will not be so instant with deprecation warnings being raised for a long time. The warning will explain to users the change and how to update to the new API but all environments will be created through `gym.make` will have a wrapper to make them compatible with the old step API code. The gym wrappers will be modified to use the new API however users have add compatibility wrappers in their code to make this work.

I would be interested in your response to this, in particular, point two as this is the primary motivation for this change I believe.  Can you find a way to change all the projects, tutorials and RL explanations across the internet that incorrectly talk about this without making this change? 

Plus people who don't want to use the change, then they could use the compatible wrappers or just an old version of gym.

avnishn:
I see the issue here with returning done signals that correspond to time limit terminations. In fact once when making a SAC implementation this issue plagued me for 6 months. 

I'm in favor of @XuehaiPan's proposal -- there's no need to expand the number of returns here -- using 0, 1, 2 will at least provide some sort of backwards compatibility to the code. 

On a more meta note though, I'd recommend making the new behavior off by default (I see that others agree with me on that here) but doing the deprecation notice over the course of 2 minor releases, providing a link to a migration guide of sorts in that deprecation notice. The deprecation notice can be upon import. 

The reason I say to do it over 2 minor releases is that it gives downstream libraries a chance to catch that gym will be making breaking changes. We (The ray and RLlib team) have had difficulty keeping up with the gym 0.20-0.23 releases and their changes.

CC @sven1977 @gjoliver

araffin:
> I would be interested in your response to this, in particular, point two as this is the primary motivation for this change I believe
> the previous way of just using documentation has not fixed it.

The main issue is that there was no previous way, the feature was never documented (see https://github.com/openai/gym/issues/2510#issuecomment-989734977 and https://github.com/openai/gym/issues/2510#issuecomment-1100622816).
And I said before, even if you provide an additional boolean to the user, you still need documentation to know how to use it properly.

(w.r.t. SB3, I created issues and talked about it one year ago (RLVS21) and only recently added it explicitly to our doc https://github.com/DLR-RM/stable-baselines3/pull/896)

> For me, I have worked in RL for 5 years and never heard of this issue before.

I would say that's true for most people that started working on RL in the past years.
However, repos like TD3 original case/Spinning Up/SAC original codebase/[pytorch-a2c-ppo-acktr-gail](https://github.com/ikostrikov/pytorch-a2c-ppo-acktr-gail) all treat the timeout termination properly but most of them fail to document it.
(as a side story, you can also read about the tale of the "al" variable in https://github.com/aravindr93/mjrl/issues/13)

And at the same time, for others, it seems obvious, for instance one reviewer of the 2017 time limit in RL paper (see https://openreview.net/forum?id=HyDAQl-AW) describes the problem as "well known" and the other says "topics are actually not new and are well covered by the existing RL formalisms. "

> . Can you find a way to change all the projects, tutorials and RL explanations across the internet that incorrectly talk about this without making this change?

By making this change, you will mostly just break those projects/tutorials that won't probably be fixed.

As a maintainer of SB3, I can only agree with RLlib developers, ensuring smooth change for the user/answering user issues take time and energy. For instance, the PR that adds gym 0.23 support (https://github.com/DLR-RM/stable-baselines3/pull/780) took longer than expected because several issues were discovered by our tests (not the first time :/), we need to update not only code but our doc/tutorials and ensure backward compat for the user. Despite all those efforts, we still got issues from users that got their code broken, and those users are only the tip of the iceberg (very few people will actually open issue when they encounter one).

gjoliver:
Having more capable API is a nice thing, but the reality is that compatibility must be maintained for much longer time somewhere along the tech stack.
Often times, it means other library maintainers have to step up and support multiple versions of APIs in their code, which means huge amount of duplicated effort.
I like the info dict alternative the best.

pseudo-rnd-thoughts:
There seems to be wholehearted agreement that more documentation is required to explain this problem and how to solve it for both training, replay buffer, etc. @araffin Thanks, I have not realised this was not documented before in gym.
This will be certainly done, no matter if this PR is merged or not. 

@avnishn If this PR is merged, this feature will be turned off by default (with a warning explaining the change in the future) with all gym wrappers being compatible with the old and new step API. 

@avnishn There seem to be three options to me for fixing the problem of people ignoring the time limit truncation
1. Documentation update with no code change with the truncation specified in the info parameter (this is the option @araffin believes is the best) 
2. Documentation update + change the done to an integer (0, 1 and 2) as @XuehaiPan proposed. My problem is that this change will be invisible to most users as they will cast the output to bool making a breaking change that I would predict would make minimal difference in improving correct code implementation. Secondly, I would predict this would cause a number of minor bugs if a user does something like `target_q_value =  gamma * (reward + (1 - done) * max(target_network(next_state)))` then the done is no longer 0 but -1 in some cases. Therefore, to me, I don't think this is a viable solution as it is both a breaking change that does not break the API enough for most users to care or change or notice. Ironically, possibly introducing more bugs and reproducibility issues. @avnishn Have I missed something?
3. Documentation change + change the step function to return 5 values (termination and truncation) compared to the four values currently. This is the most breaking change and can be done over the next 6+ months with the gym API supporting both the old and new step API. The advantage is that we can encourage/force a large number of the currently incorrectly implemented code to fix for the new API. For correctly implemented code, I suspect the change should be relatively simple with just changing `if done and 'TimeLimit.truncation' in info` to `if truncation`. I know in reality due to the number of packages and implemented algorithms it will be more complex than this but in my option, it should help improve reproducibility. However, this will also require more documentation to explain how to correctly use the new return value. 

@avnishn @araffin @gjoliver Thanks for your comments
We will work to update the documentation now, before making a decision on if to merging this

XuehaiPan:
> https://github.com/openai/gym/pull/2752#issuecomment-1114033233 This is why I'm proposing a more general solution, because having `truncated=False` all the time in the env may be confusing (truncation is handled by `TimeLimit` wrapper in the vast majority of cases) , that should also make more sense: adding termination reason in the info dict as best practice.

@araffin Some RL framework also support episode truncation for extremely long episodes. For example, RLlib support `horizon` in train/evaluation configuration. The episode will be truncated when hits the horizon.

https://github.com/ray-project/ray/blob/de0c6f61326c5bcf60bee77eda9dc7ca7284f5a2/rllib/evaluation/sampler.py#L828-L864

------

> https://github.com/openai/gym/pull/2752#issuecomment-1117201642 Documentation update + change the done to an integer (0, 1 and 2) as @XuehaiPan proposed. My problem is that this change will be invisible to most users as they will cast the output to bool making a breaking change that I would predict would make minimal difference in improving correct code implementation. Secondly, I would predict this would cause a number of minor bugs if a user does something like `target_q_value = gamma * (reward + (1 - done) * max(target_network(next_state)))` then the done is no longer 0 but -1 in some cases. Therefore, to me, I don't think this is a viable solution as it is both a breaking change that does not break the API enough for most users to care or change or notice. Ironically, possibly introducing more bugs and reproducibility issues. @avnishn Have I missed something?

@pseudo-rnd-thoughts

> 1. My problem is that this change will be invisible to most users as they will cast the output to bool making a breaking change that I would predict would make minimal difference in improving correct code implementation.

Agreed. So if we have to making breaking changes anyway when we reach Gym 1.0. I think this core API change (two booleans) is a chance pushing the downstream codebase to have the correct Q-iteration / value-iteration implementation. It's more explicit than a silent migration using integers.

Changing `bool` to `int` maybe backward compatible, but it may introduce more potential bugs. For example, bit operations on integers (e.g. `y = r + gamma * (~d) * q'`).

```python
In [1]: import numpy as np

In [2]: bools = np.array([False, True], dtype=np.bool8)

In [3]: ints = np.array([0, 1, 2], dtype=np.int64)

In [4]: bools
Out[4]: array([False,  True])

In [5]: ints
Out[5]: array([0, 1, 2])

In [6]: 1 - bools
Out[6]: array([1, 0])

In [7]: 1 - ints
Out[7]: array([ 1,  0, -1])

In [8]: abs(1 - ints)
Out[8]: array([1, 0, 1])

In [9]: ~bools
Out[9]: array([ True, False])

In [10]: ~ints
Out[10]: array([-1, -2, -3])
```

> 2. I would predict this would cause a number of minor bugs if a user does something like `target_q_value = gamma * (reward + (1 - done) * max(target_network(next_state)))` then the done is no longer 0 but -1 in some cases. 
- In PR #2608

If we change the `done` to 3 state integers (0, 1, 2), the correct implementation for Q learning is:

```python
target_q_value = reward + gamma * abs(1 - done) * max(target_network(next_state))
```

To be charity:

```python
effective_gamma = gamma * abs(1 - done)
```

We should add `abs()` here.

araffin:
> Thanks, I have not realised this was not documented before in gym.
This will be certainly done, no matter if this PR is merged or not.

:+1: 

> before making a decision on if to merging this

in order of preference, I obviously have 1) key in the info dict 2) three states done 3) two boolean

> For correctly implemented code, I suspect the change should be relatively simple with just changing if done and 'TimeLimit.truncation' in info to if truncation

Depending on what is allowed, you may have to update all the evaluation loop too (the `while not done:` loops).

pseudo-rnd-thoughts:
@araffin This is the current documentation, could you make comments over there on changes or on discord if you are part of the gym server then my name is PseudoRnd
https://github.com/Farama-Foundation/gym-docs/pull/128

gym/core.py
```diff
@@ -242,18 +251,20 @@ class Wrapper(Env[ObsType, ActType]):
         Don't forget to call ``super().__init__(env)`` if the subclass overrides :meth:`__init__`.
     """
 
-    def __init__(self, env: Env):
+    def __init__(self, env: Env, new_step_api: bool = False):
```
pseudo-rnd-thoughts, line 254: Personally, Im really unsure of this approach to adding new_step_api in wrappers. Even if we go with this approach, I don't like this code style as every wrappers needs to add this parameter. 
I can't remember our proposed method but I think that the whole API should support the new API with the compatibility wrapper to used to convert between the two.

gym/utils/play.py
```diff
@@ -1,4 +1,7 @@
 """Utilities of visualising an environment."""
+
+# TODO: Convert to new step API in 1.0
```
pseudo-rnd-thoughts, line 3: Is this TODO still to do

gym/utils/step_api_compatibility.py
```diff
@@ -0,0 +1,180 @@
+"""Contains methods for step compatibility, from old-to-new and new-to-old API, to be removed in 1.0."""
```
pseudo-rnd-thoughts, line 1: I was considering this a wrapper, one for `NewStepAPI` and `OldStepAPI` instead of these two functions and modifying the `Wrapper` class

pyproject.toml
```diff
@@ -42,3 +42,6 @@ reportUntypedFunctionDecorator = "none"
 reportMissingTypeStubs = false
 reportUnboundVariable = "warning"
 reportGeneralTypeIssues ="none"
+
+[tool.pytest.ini_options]
+filterwarnings = ['ignore:.*step API.*:DeprecationWarning'] # to be removed when old step API is removed
```
pseudo-rnd-thoughts, line 47: Add TODO to make it easier to find later

gym/wrappers/autoreset.py
```diff
@@ -31,22 +42,26 @@ def step(self, action):
...
+        )
 
-        if done:
+        if terminated or truncated:
 
             new_obs, new_info = self.env.reset(return_info=True)
             assert (
-                "terminal_observation" not in new_info
-            ), 'info dict cannot contain key "terminal_observation" '
+                "closing_observation" not in new_info
```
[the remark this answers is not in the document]
arjun-kg: Oh that sounds better yes

gym/core.py
arjun-kg: They don't need to override the `new_step_api` attribute if they're using the old style, but they do need if it will be in the new style (cuz we're doing opt-in). I understand there is some ugliness, but I don't know how to retain 100% backward compatibility while keeping it clean

gym/utils/play.py
arjun-kg: It was too much of a pain to make this compatible with both APIs, so just added this to TODO when we fully switch to new API (I will add a warning though)

gym/utils/step_api_compatibility.py
arjun-kg: I don't get you :thinking:

arjun-kg:
@pseudo-rnd-thoughts thank you very much for looking through the code! Couldn't have been easy :P 

> Personally, I think the whole of gym should only support the new API with only the make and the compatibility wrapper that users can add where needed.
This would mean that the conversion only happened when explicitly wrapped by the user.

So this is kinda similar to the original implementation (in the sense the wrappers and vector envs didn't have any frills and were switched cleanly to new api), and you also echo Ariel's thoughts in his earlier review, but we also wanted 100% backward compatibility, I don't think this would work then?

gym/core.py
RedTachyon: I kinda prefer the style suggested here, though the current version might be more simple to understand, so I think I'm fine with either

RedTachyon:
So my understanding of the current approach is as follows:
1. We technically support both old and new APIs
2. By default, everything happens with the old API
3. We have two magic functions, merged into one supermagic function (`step_api_compatibility`), which convert returns between the APIs
4. We have a wrapper which deliberately also performs this conversion
5. All wrappers must now know what API they follow. Existing external wrappers will default to following the old API
6. Wrappers will also, quietly, perform API conversions to make sure everything gets patched through properly
7. Similarly, vector envs are created with a specific API in mind, and will go through conversions so that it fits, no matter what their underlying envs do

Does that sound about right?

Overall I'll give it a **conditional approval** (besides a few smaller comments that should show up in a few moments), with the condition being:
**A good majority of these changes has to be undone in 1.0**, when everything will implicitly follow the new API, and conversions will be done by users, at their risk. I do think we should keep the conversion functions, but they shouldn't be called anywhere inside of the gym code.

My reaction to most of these changes was a quick trip from "that's a bad idea", through "that's gonna be annoying to maintain and debug" to "I guess it's necessary for now?", which is quite a fun road. So while I know that nothing is as permanent as temporary fixes, this *has* to be temporary.

And to be clear - I don't even mean this as criticism of the code or the PR - the code does what it's supposed to do, I just don't love what it does. I'm not even sure if there's a better way, I just have a vague impression that there should be one, and it would take quite some time to figure it out.


If for some reason we expect this code and this approach to survive beyond that, then I'd strongly suggest trying to find that elusive "better way" before we get to that point. It should be something that's relatively self-contained, so that the actual gym code is written with one API in mind, and compatibility concerns are isolated into a few files, and that's it. Right now the compatibility stuff is unfortunately everywhere.

arjun-kg:
@RedTachyon  Thank you very much for your review! 

> Does that sound about right?

Yup that's right, for all the points. 

>  **A good majority of these changes has to be undone in 1.0**, when everything will implicitly follow the new API, and conversions will be done by users, at their risk. I do think we should keep the conversion functions, but they shouldn't be called anywhere inside of the gym code.

Yes, that is the idea. In hindsight, I agree with keeping the conversion functions, considering the fact that this is going to change gym API that has been used for atleast the past five years.  

> And to be clear - I don't even mean this as criticism of the code or the PR - the code does what it's supposed to do, I just don't love what it does. 

I get what you mean. There's also the fact that all the tests only test for the old API. The compatibility function is simple enough so there ideally shouldn't be issues, but I do feel uneasy that every wrapper is not tested for new API as well. And in an attempt to be backward compatible, this code makes it a little annoying to actually use the new API (having to define the `new_step_api` attribute in new wrappers, taking care of compatibility everywhere since everything would default to old API). But again, I don't know if there is a better way. 

> If for some reason we expect this code and this approach to survive beyond that, then I'd strongly suggest trying to find that elusive "better way" before we get to that point. 

Not sure what you mean by 'beyond that', but the reason for this mess is backward compatibility. Things will get much more clean once we decide to not support the old API, and just guide users to use compatibility functions.

gym/envs/registration.py
```diff
@@ -594,16 +602,26 @@ def make(
...
+        env = TimeLimit(env, max_episode_steps, new_step_api)
     elif spec_.max_episode_steps is not None:
-        env = TimeLimit(env, spec_.max_episode_steps)
+        env = TimeLimit(env, spec_.max_episode_steps, new_step_api)
 
     if autoreset:
-        env = AutoResetWrapper(env)
+        env = AutoResetWrapper(env, new_step_api)
+
+    if not disable_env_checker:
```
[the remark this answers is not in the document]
arjun-kg: Oh some screw up while dealing with merge conflicts. But that piece of code is gone now.

gym/utils/step_api_compatibility.py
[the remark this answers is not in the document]
arjun-kg: Makes sense. We can atleast remove the warning that it'll be removed, and decide later what exactly we want to do.

gym/core.py
arjun-kg: I would prefer to keep this explicit, to show that this abstract class handles both APIs, especially since it's in the core file.

gym/core.py
arjun-kg: Same comment, even though it adds a few lines, I wanted to keep this explicit.

araffin:
> My reaction to most of these changes was a quick trip from "that's a bad idea", through "that's gonna be annoying to maintain and debug" to "I guess it's necessary for now?"

It's probably useless to repeat it, but this is not needed...

I am also concerned that despite the worries from two major users of Gym, RLLib (https://github.com/openai/gym/pull/2752#issuecomment-1116312780 w/ @sven1977 @gjoliver and @avnishn ) and SB3 (see https://github.com/openai/gym/pull/2752#issuecomment-1114033233, in the meantime more issues came), the additional bool seems to be the chosen solution.

In the last months, the strategy of Gym of "break first and fix later" instead of "pause and ponder" (i.e., wait until the feature/fix is fullly ready before releasing it, which may mean wait some months) has been particularly hard to follow (SB3 gym 0.24 is still on hold despite efforts from all parts... https://github.com/DLR-RM/stable-baselines3/pull/780, RLLib team is sharing that concern too), and the lack of transparency did not help (discussions done offline/on discord and no summary of it on github).

So, if I were you, I would do such core change only if I'm 100% sure of what I'm doing and if there is no better way for the users.

pseudo-rnd-thoughts:
@araffin I hear what you are saying, would you want to have a discord call to discuss it with @arjun-kg and myself? I think this should help speed up communication and ensure that everyone understands each other. 
If so, could you add yourself to the discord server https://discord.gg/nHg2JRN489 and add me, pseudo-rnd

araffin:
> you add yourself to the discord server

done ;)

>  and the lack of transparency did not help (discussions done offline/on discord and no summary of it on github).

>  would you want to have a discord call 


that's exactly what I was talking about... the minutes/summary of the call should be written here afterward for transparency.

araffin:
> the minutes/summary of the call should be written here afterward for transparency.

### Summary of the discord call

statu quo from both sides, arguments of the discussions were repeated:
Antonin thinks doing such breaking changes should be avoiding to avoid confusion/breaking code (many issues will be opened because of it) and because it is not needed, while the rest of the gym dev teams think such breaking change is needed to make termination due to truncation/timelimit explicit.

As a result, SB3 will only support newest version of Gym via a specific branch (https://github.com/DLR-RM/stable-baselines3/pull/780) until Gym is stable again.

RedTachyon:
I'm doing a relatively thorough review, which will take Time™, so I'll just progressively add things that I notice (if I notice anything big after this one, of course)

For reference, I'm testing it in [this colab](https://colab.research.google.com/drive/16TdzbKkGCRusKGSNSTUrnpRmqWsFoKBn?usp=sharing)
I'll probably keep modifying it, but I'll try to keep the previous parts the same.

One thing I dislike in the current version is how many warnings appear. If I use the old API with cartpole, I get two deprecation warnings about converting back and forth. What's more, I get it every time I run a step in a new cell. So while a loop doesn't get us spammed with warnings, it's a huge pain for interactive use. This has to be changed imo, both of the points mentioned (i.e. it should be only one warning, and only once). 

On the same note, in the notebook there's a small toy environment implemented to emulate an end-user's experience, written in an old variant and a new variant. Both work with both APIs, so that's good! But what happens is that whenever there's an old-style API involved (either in the env implementation, or the core API), you again get that warning at every step. So again, this should be only a single warning. 

But also I don't necessarily think this should be happening during `env.step()` calls. If we're converting the env (which is a deprecated functionality, at least by default), the warning should ideally appear at initialization, since that's when the "bad stuff" is happening (i.e. not using the new API). If you `make` your env with the old API and then call `env.step`, the cat's out of the box, there's nothing you can do at this stage to change your ways and eliminate the warnings (beyond silencing them, obviously, which is not the best solution)


---

As a separate thing, I don't know if you already did something like this, but I highly recommend interactively just testing out the library using this PR's branch, checking how various usecases behave. This is something that tests might not pick up, but is important to the user experience, which [unfortunately is important](https://www.smbc-comics.com/comic/2012-05-01). Since you're the most familiar with the code here, you might have some suspicions as to what could be a rough edge, what could be weird, so just be an end user

image

arjun-kg:
> One thing I dislike in the current version is how many warnings appear.

A simple fix might be to move the warnings to `__init__` for wrappers, and vectors where it warns if `new_step_api=False`. This would include warning after `gym.make` if the env step returns in old API. This would however not warn when the implementation itself is in old step API (and the output of step is in new API through transformation), or if someone uses the transformation functions explicitly. But it might be reasonable to assume the users know what's going on if they decided to set `new_step_api=True` or explicitly use the transformation functions. What do you think? 

> but I highly recommend interactively just testing out the library using this PR's branch, checking how various usecases behave.

I did this a little bit on my local machine, checked atari / pybullet for eg. But yes can do more. The rough edges may be some simultaneous usage of two types of info dicts in vector envs, transformations between the two info types, while also between the two step APIs. Nothing else comes to mind but I might discover problems as I use the code yeah.

RedTachyon:
Just two questions because I think I misunderstood one thing during a call. If there's a third-party wrapper written in the old API (i.e. no mentions of new_step_api whatsoever), and it is applied to an old-API environment, everything will work alright? (except for a warning)

If a third-party wrapper is written in old API, can we do:
```python
env = gym.make(something, new_step_api=True)
env = OldStyleWrapper(env)
```
? Is there an automatic way to make it work? Or does it require an update to the wrapper?

If the question to the first question is "yes", then after the merge conflicts are resolved, I think this can be merged. I still have... *many* problems, but they all seem to be necessary for now, so let's hope we can rip it out soon-ish for 1.0

(the second question doesn't affect the approval, it's just for me to understand where we're standing because there *will* be issues raised about it)

arjun-kg:
@RedTachyon 

> If there's a third-party wrapper written in the old API (i.e. no mentions of new_step_api whatsoever), and it is applied to an old-API environment, everything will work alright?

Yes. This was one of the reasons behind introducing `new_step_api` attribute in the base wrapper and having it default to False, to have this kind of backward compatibility.

> ? Is there an automatic way to make it work? Or does it require an update to the wrapper?

No, this will not work (unless the wrapper doesn't need to unpack step returns). There's no 'automatic way' to do this. But the change to the wrapper is not too bad to get it to work: 

```python
class OldStyleWrapper:
   ...
   def step(self, action):
        o, r, d, i = step_api_compatibility(self.env.step(action), new_step_api=False)
        ...
        return o, r, d, i
```

pseudo-rnd-thoughts:
LGTM

PR #2756, opened 2022-04-15
Optional pygame import for box2d with try except

pseudo-rnd-thoughts:
Currently the CI is seg faulting without reporting an error (I know, not good), see https://github.com/openai/gym/pull/2750/checks in the run tests. 
The cause of it was 07fd227a4fc0ecbd4fdac5563c0f48deeb5688ed in car racing somehow effecting async vector environments due to pygame.quit() . 
I was not able to isolate the issue to a single cause so have instead rewritten the fix to use try except rather than local imports which seems to have solved the seg fault issue and the original issue.

I have modified all of the box2d environment as pygame was made optional in all of them and to prevent the seg fault happening in the future. 

@RedTachyon @gianlucadecola I believe you were both involved in the optional pygame PRs, was there a reason you chose local imports over try except? 
If there isn't an issue with this approach, could you review this PR?

### DIFF
--- a/gym/envs/box2d/bipedal_walker.py
+++ b/gym/envs/box2d/bipedal_walker.py
@@ -19,6 +19,12 @@ import gym
 from gym import error, spaces
 from gym.utils import EzPickle, colorize, seeding
 
+try:
+    import pygame
+    from pygame import gfxdraw
+except ImportError:
+    pygame, gfxdraw = None, None
+
 FPS = 50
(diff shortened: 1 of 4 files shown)


pseudo-rnd-thoughts:
Looking at #2671, it has implemented the optional pygame however the action's don't show seg fault so Im going to close this PR as it may be not needed

Issue #2751, opened 2022-04-14
How do I solve this 'FrozenLakeEnv' object has no attribute 'lastaction'


Anna4142:
### Question

If you're a beginner and have basic questions, please ask on [r/reinforcementlearning](https://www.reddit.com/r/reinforcementlearning/) or in the [RL Discord](https://discord.com/invite/xhfNqQv) (if you're new please use the beginners channel). Basic questions that are not bugs or feature requests will be closed without reply, because GitHub issues are not an appropriate venue for these.

Advanced/nontrivial questions, especially in areas where documentation is lacking, are very much welcome.

pseudo-rnd-thoughts:
Could you provide some example code? 
Are you sure that you have reset the environment?

Looking at the code, I'm guessing that you have tried rendering without resetting

jkterry1:
@younik if this is a reproducible bug, you're the one who's been working with this code

pseudo-rnd-thoughts:
Im fairly sure their code is 
```
import gym

env = gym.make('FrozenLake-v1')
env.render()

> AttributeError: 'FrozenLakeEnv' object has no attribute 'lastaction'
```
We can add PR to add a check for render that reset has been called before render or move the variables into the constructor

younik:
> @younik if this is a reproducible bug, you're the one who's been working with this code

Yes, it is reproducible as @pseudo-rnd-thoughts showed. This is a problem in all environments: if you don't call `reset()` before `render()`, you will get a non-explanatory error.

@jkterry1 should be possible to drop the constraint of calling reset after the creation? It is common to all environments, so it could be done internally, isn't it?

jkterry1:
I spoke to younik about this on a call, we have to have reset first

PR #2757, opened 2022-04-15
Add check before rendering Frozen Lake

pseudo-rnd-thoughts:
Issue #2751 reported that `lastaction` was not an attribute of Frozen Lake in the following code
```
import gym

env = gym.make("FrozenLake-v1")
env.render()
```

This PR adds a check within `render()` that `lastaction` is an attribute otherwise raises an AttributeError explaining the problem

### DIFF
--- a/gym/envs/toy_text/frozen_lake.py
+++ b/gym/envs/toy_text/frozen_lake.py
@@ -234,6 +234,9 @@ class FrozenLakeEnv(Env):
             return int(self.s), {"prob": 1}
 
     def render(self, mode="human"):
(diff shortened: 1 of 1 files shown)


Markus28:
Not directly related to this PR, but it might be sensible to add something similar to the `OrderEnforcing` wrapper?

pseudo-rnd-thoughts:
@Markus28 That is a good idea but I realised that this wont be an issue soon as the render function will be deprecated / removed  soon in #2671 
Therefore, I think this PR should be closed

PR #2595, opened 2022-02-04
Use mujoco bindings instead of mujoco_py

rodrigodelazcano:
Changes made:

- Create Viewer() class to render window in "human" mode with dm_viewer and glfw
- Modified the default viewer_setup() method for all mujoco_environments (only for v3 envs)

Issue #1541, opened 2019-06-20
Contact forces are zero in Ant-v2/v3


pratikac:
The contact forces are all zero in the the MuJoCo Ant-v2/v3 environments. If one runs

```
import gym
import numpy as np

e = gym.make('Ant-v3')
for _ in range(100):
    e.reset()
    for i in range(1000):
        x=e.step(e.action_space.sample())[0]
        print(np.linalg.norm(x[27:]))
```

the last 84 dimensions in the state out of 111 dimensions are always zero. These dimensions are the contact forces in ant_v3.py (https://github.com/openai/gym/blob/master/gym/envs/mujoco/ant_v3.py#L124). Since the ant is in contact with the ground, I believe there should be some contact forces that are non-zero. Can you please clarify if this is the expected behavior?

AlexanderGri:
The same goes for the Humanoid environment - contact forces are zero in the case of MuJoCo 2.0 . The environment uses ([v2](https://github.com/openai/gym/blob/master/gym/envs/mujoco/humanoid.py#L32) and [v3](https://github.com/openai/gym/blob/master/gym/envs/mujoco/humanoid_v3.py#L63)) the `cfrc_ext` field of `PyMjData` which is a proxy to the `cfrc_ext` field of MuJoCo C structure `_mjData` . But the [changelog](https://www.roboti.us/index.html#changelist) to MuJoCo 2.0 contains the following line:

> The function mj_rnePostConstraint **is no longer called by default**. Instead, it is called only when the model contains force-related sensors which need the results of this function. The user **can** still **call it directly** in order **to compute** the mjData fields cacc, **cfrc_ext**, cfrc_int if desired.

Due to this Humanoid and Ant reward structure differs between MuJoCo 2.0 and <2.0 . Is it intentional?

pshvechikov:
Also, found this very frustrating. 
Algorithm benchmarks differ between versions of MuJoCo. 
This can misguide researchers and invalidate comparison to previously published papers. 
So, this seems like a severe bug. 
Please, @christopherhesse @pzhokhov, led some light on the issue, as soon as it is possible for you. 
Thanks in advance.

christopherhesse:
Thanks for reporting this bug!  It looks like this could have been introduced by https://github.com/openai/gym/pull/1401

Can you guys verify that using `mujoco-py==1.50.1.68` fixes this issue?

AlexanderGri:
Thanks for your response!

[...]

PR #2762, opened 2022-04-19
Change mujoco_py bindings for mujoco Deepmind bindings

rodrigodelazcano:
# Description

This PR is the continuation of https://github.com/openai/gym/pull/2595. The PR updates the python bindings for the mujoco environments. The new v4 versions of the mujoco bindings now use the new mujoco python bindings from DeepMind  https://pypi.org/project/mujoco/. 

Fixes # (issue)
This PR also fixes the contact force issue in the Ant env (only v4) at  https://github.com/openai/gym/issues/1541
## Type of change

### v2-v4 benchmark by @vwxyzjn 
@vwxyzjn provided benchmarks between old mujoco environments (v2) and the newer (v4). Results are provided here https://github.com/openai/gym/pull/2595#issuecomment-1099152505

# Checklist:


setup.py
```diff
@@ -14,19 +14,25 @@
     "accept-rom-license": ["autorom[accept-rom-license]~=0.4.2"],
     "box2d": ["box2d-py==2.3.5", "pygame==2.1.0"],
     "classic_control": ["pygame==2.1.0"],
-    "mujoco": ["mujoco_py>=1.50, <2.0"],
+    "mujoco_py": ["mujoco_py<2.2,>=2.1"],
```
saran-t, line 17: I have mild concerns around `gym` declaring a dependency on both `mujoco` and `mujoco-py`. In principle, this isn't a problem as long as only one gets imported, but if someone tries to import both it's likely that they'll run into inscrutable errors. Would you consider declaring `mujoco-py` as an optional dependency (`extras_require`)?

saran-t: Ignore me: this is _already_ in `extras_require`.

gym/envs/mujoco/mujoco_env.py
```diff
@@ -136,9 +173,21 @@ def do_simulation(self, ctrl, n_frames):
...
+            if self._mujoco_bindings.__name__ == "mujoco_py":
+                self.sim.step()
+            else:
+                self._mujoco_bindings.mj_step(self.model, self.data)
+
+        # As of MuJoCo 2.0, force-related quantities like cacc are not computed
+        # unless there's a force sensor in the model.
+        # See https://github.com/openai/gym/issues/1541
+        if self._mujoco_bindings.__name__ != "mujoco_py":
+            self._mujoco_bindings.mj_rnePostConstraint(self.model, self.data)
```
vwxyzjn, line 190: This is not commented as in https://github.com/openai/gym/pull/2595/commits/ab380345dc67e5c8b767bd2f2dcea5722161c412, which causes regression in Ant-v4. See https://wandb.ai/costa-huang/cleanRL/reports/-4-19-MuJoCo-v4-vs-v2-CleanRL-s-PPO--VmlldzoxODYzODM0, https://github.com/vwxyzjn/validate-new-gym-mujoco-envs

5AF41CE3-F3EB-4040-99CF-3FC722F30565

gym/envs/mujoco/half_cheetah_v3.py
```diff
@@ -10,125 +10,6 @@
...
-
-    * v3: support for gym.make kwargs such as xml_file, ctrl_cost_weight, reset_noise_scale etc. rgb rendering comes from tracking camera (so agent does not run away from screen)
-    * v2: All continuous control environments now use mujoco_py >= 1.50
-    * v1: max_time_steps raised to 1000 for robot based tasks. Added reward_threshold to environments.
-    * v0: Initial versions release (1.0.0)
-    """
-
     def __init__(
         self,
         xml_file="half_cheetah.xml",
```
pseudo-rnd-thoughts, line 15: This doesn't matter much but in other mujoco environments the `xml_file` is not passed as a parameter to the environment. Is there a reason for this environment?

gym/envs/mujoco/mujoco_rendering.py
```diff
@@ -0,0 +1,556 @@
...
+                raise RuntimeError(
+                    "Environment variable {} must be one of {!r}: got {!r}.".format(
+                        "MUJOCO_GL", _ALL_RENDERERS.keys(), backend
+                    )
+                )
+
+        else:
+            for name, import_func in _ALL_RENDERERS.items():
+                try:
+                    self.opengl_context = _ALL_RENDERERS["osmesa"](width, height)
```
kngwyu, line 252: Shouldn't this use `name` instead of `osmesa`?
Or
```suggestion
                    self.opengl_context = import_func(width, height)
```

gym/envs/mujoco/walker2d_v4.py
```diff
@@ -0,0 +1,278 @@
+from turtle import distance
```
Trinkle23897, line 1: This is an unused import for walker2d.

gym/envs/mujoco/ant_v4.py
```diff
@@ -0,0 +1,298 @@
...
+
+        mujoco_env.MujocoEnv.__init__(self, xml_file, 5)
+
+    @property
+    def healthy_reward(self):
+        return (
+            float(self.is_healthy or self._terminate_when_unhealthy)
+            * self._healthy_reward
+        )
+
```
Trinkle23897, line 198: I think this has a bug:

- when `terminate_when_unhealthy` is true (default setting), `self.healthy_reward` should be conditioned on `is_healthy`; however, the above code is not (always `self._healthy_reward`)
- when `terminate_when_unhealthy` is false, `self.healthy_reward` should always be `self._healthy_reward` (done is always False); however, the above code is not (conditioned on `self.is_healthy`)

```suggestion
    @property
    def healthy_reward(self):
        return (
            float(self.is_healthy or not self._terminate_when_unhealthy)
            * self._healthy_reward
        )

```

gym/envs/mujoco/mujoco_env.py
duburcqa: Computing wrong quantities on purpose to avoid regression of performance makes no sense to me. It seems way more appropriate to just remove the contact forces from the observation space and reward computation, instead of carrying out computations pretending it is doing what it is supposed to do.

kngwyu: I agree with @duburcqa
Doesn't this change make the benchmark compatible with MuJoCo < 2.0? Or do all those three (mujoco-py with MuJoCo < 2.0, mujoco-py with MuJoCo >= 2.0, and mujoco) produce different results?

duburcqa: Theoretically, for this change in particular it is the same as mujoco < 2.0, but there are probably other internal changes in mujoco binary so I doubt the result would be exactly the same as before.

gym/envs/mujoco/mujoco_rendering.py
rodrigodelazcano: Yes, you are right. Thank you for pointing this out.

gym/envs/mujoco/walker2d_v4.py
rodrigodelazcano: I'll remove this. Thanks!

setup.py
```diff
@@ -14,19 +14,25 @@
...
+    "mujoco": ["mujoco==2.1.5", "imageio>=2.14.1"],
     "toy_text": ["pygame==2.1.0", "scipy>=1.4.1"],
     "other": ["lz4>=3.1.0", "opencv-python>=3.0"],
 }
 
 # Meta dependency groups.
-nomujoco_blacklist = {"mujoco", "accept-rom-license", "atari"}
+nomujoco_blacklist = {"mujoco_py", "mujoco", "accept-rom-license", "atari"}
```
pseudo-rnd-thoughts, line 24: Are we able to raise a warning if users pick this option to say that this is not being maintained anymore

rfernand2:
Thanks for all the work on this!  Will this be in gym 0.24.0 and is there an ETA for that release?

rodrigodelazcano:
> Thanks for all the work on this! Will this be in gym 0.24.0 and is there an ETA for that release?

:) @jkterry1 can probably address this question better.

rodrigodelazcano:
This PR is ready to be merged. Last additional change:

- Ant-v4 has an option to use contact forces in its observation space. To use this feature set the `use_contact_forces` argument to `True`. Note that this option has been added because an ablation test with PPO was performed (https://github.com/vwxyzjn/validate-new-gym-mujoco-envs) and shows that there is a regression on learning when using contact forces. Results shown on image bellow.
image

saran-t:
I know that people spent a great deal of effort validating the v4 envs and this may well be a bit late now, but please consider pinning to `mujoco==2.2.0` instead of `2.1.5` for your v0.24.0 release, since that's the first open sourced version of MuJoCo. It'll allow people to actually look at source code should any future discrepancies arise.

pseudo-rnd-thoughts:
@saran-t Has there been any significant changes since `2.1.5` for `2.2.0` that would change the training results you think otherwise we should be able to do that

Trinkle23897:
@pseudo-rnd-thoughts No. I ran the alignment test and it seems that `2.1.5` can align with `2.2.0` per step.

PR #2764, opened 2022-04-20
Vector Api for website

RushivArora:
Rephrased some methods in the docstrings for vector api:
- Removed all phrases that state "sub-environment"
- Added a sentence to clarify it is multiple copies of the same environment


pseudo-rnd-thoughts:
@jkterry1 Looks good to merge

Issue #2763, opened 2022-04-19
[Bug Report] Lunar Lander windy demo heuristic fails 


pseudo-rnd-thoughts:
In running the CI, I had the lunar lander using the windy parameter failed for me
https://github.com/openai/gym/runs/6086223589?check_suite_focus=true

We should fix this as otherwise other CIs will continue to fail for no reason to do with the PR

pseudo-rnd-thoughts:
Closed with fixed by #2766 through removing the test.

PR #2766, opened 2022-04-20
reduce wind power further for heuristic landing

jjshoots:
This addresses #2763


pseudo-rnd-thoughts:
Could you add an issue for readding windy to lunar lander? And I can close https://github.com/openai/gym/issues/2763

PR #2770, opened 2022-04-22
Avoid using deperecated distutils package

kngwyu:
# Description

Avoid using [`distutils`](https://docs.python.org/3/library/distutils.html) module that is deprecated with removal planned for Python 3.12.
Following recommendations in [PEP 632](https://peps.python.org/pep-0632/), `distutils.spawn.find_executable` is replaced by `shutil.which`, and `distutils.version` is replaced by `packaing.version`.
The later change requires [`packaging`](https://packaging.pypa.io/en/latest/) as an additional dependency. If you don't like that, I can revise this PR to manually compare versions.

## Type of change

### Screenshots

Before: image

# Checklist:

### DIFF
--- a/gym/wrappers/monitoring/video_recorder.py
+++ b/gym/wrappers/monitoring/video_recorder.py
@@ -417,7 +417,7 @@ class ImageEncoder:
             )
 
         try:
-            self.proc.stdin.write(frame.tostring())
+            self.proc.stdin.write(frame.tobytes())
         except Exception as e:
             stdout, stderr = self.proc.communicate()
             logger.error("VideoRecorder encoder failed: %s", stderr)
--- a/gym/wrappers/monitoring/video_recorder.py
+++ b/gym/wrappers/monitoring/video_recorder.py
@@ -9,6 +9,7 @@ from io import StringIO
 
 import numpy as np
 import packaging.version
+
 from gym import error, logger
 
 
--- a/gym/wrappers/monitoring/video_recorder.py
+++ b/gym/wrappers/monitoring/video_recorder.py
@@ -8,7 +8,6 @@ import tempfile
 from io import StringIO
 
 import numpy as np
-import packaging.version
 
 from gym import error, logger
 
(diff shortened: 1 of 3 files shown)


vwxyzjn:
This looks pretty straightforward to me. Please fix the pre-commit and pin the `packaging` version and everything else LGTM.

kngwyu:
Thank you guys for the reviews. I fixed the isort error and require `packaging >= 21.0`.
From [CHANGELOG](https://packaging.pypa.io/en/latest/changelog.html), any version of packaging should be fine, but I selected the latest version without legacy Python support.

XuehaiPan:
> The later change requires [packaging](https://packaging.pypa.io/en/latest/) as an additional dependency. If you don't like that, I can revise this PR to manually compare versions.

`packaging` is not always usable.

pkg_resources

Instead, we should use `pkg_resources.parse_version`, which comes with `setuptools`. We can assume the user always have `pip` (will install `setuptools`), right?

kngwyu:
Yeah actually `pkg_resources` ships a vendored packaging and it's also usable by `pkg_resources.packaging`, but I don't know the recommended way to use it. The document only recommends to use packaging https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html.

> We can assume the users always have pip (will install setuptools), right?

Maybe? But I don't know if there is a good way to ensure API compatibility of pkg_resources.

XuehaiPan:
Actually, `gym` requires `numpy >= 1.18.0`. The check for `numpy >= 1.9.0` is redundant.

https://github.com/openai/gym/blob/dcae553204957a75db5e4169e33547ccac599609/setup.py#L48-L53

https://github.com/openai/gym/blob/dcae553204957a75db5e4169e33547ccac599609/gym/wrappers/monitoring/video_recorder.py#L421-L426

kngwyu:
Ah that's good to know. I'll remove the corresponding code.

gym/wrappers/monitoring/video_recorder.py
```diff
@@ -418,12 +417,7 @@ def capture_frame(self, frame):
             )
 
         try:
-            if distutils.version.LooseVersion(
-                np.__version__
-            ) >= distutils.version.LooseVersion("1.9.0"):
-                self.proc.stdin.write(frame.tobytes())
-            else:
-                self.proc.stdin.write(frame.tostring())
+            self.proc.stdin.write(frame.tostring())
```
XuehaiPan, line 420:
```suggestion
            self.proc.stdin.write(frame.tobytes())
```

kngwyu: Thank you!

Issue #2657, opened 2022-03-01
[Proposal] New `info` API for vectorized environments


vwxyzjn:
### Motivation

The current vector env API would return an `info` dict per environment. So if  if I have 5 envs, then the dict could look like
```[{}, {}, {}, {}, {"truncated": True}]```

While this API is intuitive, it's not a friendly interface to high-throughput environments such as [brax](https://github.com/google/brax) (also see https://github.com/google/brax/issues/164) and [envpool](https://github.com/sail-sg/envpool). In these envs, they prefer the following paradigm: instead of 5 dictionaries, it's just a single dictionary with an array of length 5.

```{"truncated":[False, False, False, False, True]}```

## Pros and Cons

Brax and envpool's approach feels more ergonomic for high-throughput envs, whereas Gym's approach is maybe more efficient for "sparse" info keys that don't appear often (I could be wrong with this). 

### Proposal 

I propose to change the `info` API for vectorized environments to adapt brax's current design for `info`.  This might also have implications for wrappers and other APIs. For example, the current vectorized environment API would return `info` like this (with the `RecordEpisodeStatistics` used):

```python
[
    {
        "episode": {
            "r": 75.0,
            "l": 75,
            "t": 36.770559,
        },
        "terminal_observation": array([0.2097086, -0.5355723, -0.21343598, -0.11173592], dtype=float32),
    },
    {},
    {},
    {},
    {},
]
```

I am not sure what's the best way to handle this... Maybe something like the following?

```python
{
    "episode": {
        "r": [75.0, None, None, None, None],
        "l": [75, None, None, None, None],
        "t": [36.770559, None, None, None, None],
    },
    "terminal_observation": [
        array([0.2097086, -0.5355723, -0.21343598, -0.11173592], dtype=float32),
        None,
        None,
        None,
        None,
    ]
}
```

### Alternative 1

Alternatively, it's possible to do this in a very hacky way: brax or env pool could always override the first sub-env's info dict to always provide their desired info so something like:

```[{"truncated":[False, False, False, False, True]}, {}, {}, {}, {}]```

This way, we don't have to change the current API, however, it's extremely hacky and unintuitive.

### Alternative 2

Alternatively, maybe `info` should be reserved for something sparse like `episode` and `terminal_observation`. And for more frequent infos they should be accessed via attributes (i.e., `envs.get_attr(name)` (enabled with https://github.com/openai/gym/pull/1600). So something like

```python
root_joints = brax_envs.get_attr("root_joints")
[...]

PR #2773, opened 2022-04-23
New `info` API for vectorized environments #2657

gianlucadecola:
This is a draft referring to issue #2657 in order to gain some feedback before moving on.
For the new `info` API I'm working to make the code more modular and backward-compatible with the old ```info``` in order to make switching to a different ```info``` format in the future eventually easier.

Default behaviour is still the same so
```envs = gym.vector.make("CartPole-v1", asynchronous=False, num_envs=3)```
will return `info` in the format:
```
[{'terminal_observation': array([ 0.1307018,  1.7346063, -0.2267962, -2.7928443], dtype=float32)},
 {},
 {}]
```

If we want to switch to another format  
```envs = gym.vector.make("CartPole-v1", asynchronous=False, num_envs=3, info_format="brax")```
we get
```
{'terminal_observation': [None,
                          array([-0.09317806, -1.5342456 ,  0.22191852,  2.548524  ], dtype=float32),
                          None]}
```

If we then want to format info for example like proposed by @pseudo-rnd-thoughts 
```
{
   "key": [
         (data, environment number) 
         ...
   ], 
   ...
}
```
All we need to do is registering a new ```VecEnvInfoStrategy```

The same approach applies to wrappers like ```RecordEpisodeStatistics``` which get the ```info_strategy``` by the wrapped environments and process the info accordingly.

Current draft works on SyncVectorEnv only. If we agree on this approach I will continue implementing it with Async and other wrappers

### CODE REVIEW

**Comment by gianlucadecola** [comment]
I think all changes are in place now; the new `info` API works for both async and sync vector envs; the changes are backward compatible with the previous gym version and, in the future, if we want to add different info formatting all we need to do is adding a new ProcessingStrategy

File: gym/vector/utils/info_strategies.py [hunk 1]
```diff
@@ -0,0 +1,99 @@
...
+        for k in info.keys():
+            info_array = self.info.get(k, [None for _ in range(self.num_envs)])
+            info_array[env_num] = info[k]
+            self.info[k] = info_array
+
+    def get_info(self) -> dict:
+        return self.info
+
+
+class StrategiesEnum(Enum):
```
pseudo-rnd-thoughts, line 84: `InfoStrategiesEnum` is more descriptive

gym/vector/utils/info_strategies.py
```diff
@@ -0,0 +1,99 @@
...
+    classic: str = "classic"
+    brax: str = "brax"
+
+
+def get_info_strategy(info_format: str) -> VecEnvInfoStrategy:
+    strategies = {
```
pseudo-rnd-thoughts, line 90: Should these be global variables as if people wants their own custom Info Strategy, I dont think it is possible currently.

pseudo-rnd-thoughts:
Thanks for the PR, it looks good to me. I will get someone else to review as well. 
Would you be able to add documentation for this PR please, https://github.com/Farama-Foundation/gym-docs

gym/vector/utils/info_strategies.py
gianlucadecola: Good catch! I didn't think about it. :+1:  I can switch back to the plain old dictionary as it was before and do the validation inside the env or, maybe something like that? 
```python
class InfoStrategyFactory():
    strategies = {
        InfoStrategiesEnum.classic.value: ClassicVecEnvInfoStrategy,
        InfoStrategiesEnum.brax.value: BraxVecEnvInfoStrategy,
    }

    def get_info_strategy(info_format: str) -> VecEnvInfoStrategy:
        if info_format not in InfoStrategyFactory.strategies:
            raise InvalidInfoFormat(
                "%s is not an available format for info, please choose one between %s"
                % (info_format, list(InfoStrategyFactory.strategies.keys()))
            )
        return InfoStrategyFactory.strategies[info_format]
```

With this approach the user can add a new strategy with:
```python
from gym.vector.utils import InfoStrategyFactory
InfoStrategyFactory.strategies['my_strategy'] = MyNewStrategy
envs = gym.vector.make("CartPole-v1", asynchronous=False, num_envs=3, info_format="my_strategy")
```

Then we can also add a setter method in the class with validation to check if the added strategy matches the InfoStrategy interface  :thinking: 

```python
from gym.vector.utils import InfoStrategyFactory
InfoStrategyFactory.add_strategy('my_strategy', MyNewStrategy)
envs = gym.vector.make("CartPole-v1", asynchronous=False, num_envs=3, info_format="my_strategy")
```

gianlucadecola:
Add some fixes and a merge conflict 

> Thanks for the PR, it looks good to me. I will get someone else to review as well. Would you be able to add documentation for this PR please, https://github.com/Farama-Foundation/gym-docs

Yup, will update the docs :+1:

gym/vector/utils/info_strategies.py
pseudo-rnd-thoughts: Cool, Im good with that solution, we just need to explain it in the docs

gym/vector/__init__.py
```diff
@@ -10,7 +10,9 @@
 __all__ = ["AsyncVectorEnv", "SyncVectorEnv", "VectorEnv", "VectorEnvWrapper", "make"]
 
 
-def make(id, num_envs=1, asynchronous=True, wrappers=None, **kwargs):
+def make(
+    id, num_envs=1, asynchronous=True, wrappers=None, info_format="classic", **kwargs
```
pseudo-rnd-thoughts, line 14: Could you add `info_format` to the docstring

gym/vector/utils/info_strategies.py
```diff
@@ -0,0 +1,99 @@
...
+        }
+    """
+
+    def __init__(self, num_envs: int):
+        self.num_envs = num_envs
+        self.info = {}
+
+    def add_info(self, info: dict, env_num: int):
+        for k in info.keys():
+            info_array = self.info.get(k, [None for _ in range(self.num_envs)])
```
pseudo-rnd-thoughts, line 76: For optimisation, should this not be a numpy array? This adds some complications for the dtype of the array, but I think we should be able to check if the data is a numpy type and use that as the dtype otherise dtype=object

gym/vector/utils/info_strategies.py
pseudo-rnd-thoughts: Could you add a test for `add_strategy`

gym/wrappers/record_episode_statistics.py
```diff
@@ -4,6 +4,71 @@
...
+        return tuple(self.info)
+
+
+class BraxVecEnvStatsInfoStrategy:
+    def __init__(self, num_envs: int):
+        self.num_envs = num_envs
+        self.info = {}
+
+    def add_info(self, info: dict, env_num: int):
+        self.info = {**self.info, **info}
```
pseudo-rnd-thoughts, line 45: Could there be a problem if `self.info` and `info` contain the same key?

gym/wrappers/record_episode_statistics.py
```diff
@@ -4,6 +4,71 @@
...
+            )
+            info_array[env_num] = episode_info[k]
+            self.info["episode"][k] = info_array
+
+    def get_info(self):
+        return self.info
+
+
+def get_statistic_info_strategy(wrapped_env_strategy: str):
+    strategies = {
```
pseudo-rnd-thoughts, line 62: A comment like above on the info strategies, is there a way of adding statistic info strategies

tests/vector/test_info_format_strategies.py
```diff
@@ -0,0 +1,55 @@
...
+)
+
+
+@pytest.mark.parametrize(("num_envs"), [3])
+def test_classic_vec_env_info_strategy(num_envs):
+    infos = ClassicVecEnvInfoStrategy(num_envs)
+    for i in range(num_envs):
+        info = {"example_info": i}
+        infos.add_info(info, i)
+
```
pseudo-rnd-thoughts, line 17: Could you add an extra line only a single environment info is provided. i.e. `infos.add_info({'env 1 info': 1}, 1)`

tests/vector/test_info_format_strategies.py
```diff
@@ -0,0 +1,55 @@
...
+
+@pytest.mark.parametrize(("num_envs"), [5])
+def test_brax_vec_env_info_strategy_with_nones(num_envs):
+    infos = BraxVecEnvInfoStrategy(num_envs)
+    for i in range(num_envs):
+        if i % 2 == 0:
+            info = {"example_info": i}
+            infos.add_info(info, i)
+
```
pseudo-rnd-thoughts, line 40: If we implement numpy arrays, could this test for multiple different dtypes, floats, ints, strings, lists and dicts should be all of the comment ones

tests/wrappers/test_record_video.py
```diff
@@ -98,3 +98,17 @@ def test_record_video_within_vector():
     mp4_files = [file for file in os.listdir("videos") if file.endswith(".mp4")]
     assert len(mp4_files) == 2
     shutil.rmtree("videos")
+
+
+def test_record_video_within_vector_brax_info():
```
pseudo-rnd-thoughts, line 103: Is there a check that the brax info is correctly saved?

pseudo-rnd-thoughts:
@gianlucadecola Sorry, I did a bit more complete review and found a couple more things that could be updated.

gianlucadecola:
> @gianlucadecola Sorry, I did a bit more complete review and found a couple more things that could be updated.

No problem :+1: I'll take a look

gym/vector/utils/info_strategies.py
gianlucadecola: It can be an idea; how do you think it should be initialized? np.nan for every env?

gym/wrappers/record_episode_statistics.py
gianlucadecola: refactored as the InfoStrategyFactory :+1:

tests/vector/test_info_format_strategies.py
gianlucadecola: This condition should not happen imho; in a vectorized environment the standard behavior was looping through the envs and appending the info to a list. The ClassicInfoStrategy indeed implements `env_num: int` as parameter just to adhere the interface but it actually don't use it:  

for a quick reference:

``` python
observations, infos = [], []
for i, (env, action) in enumerate(zip(self.envs, self._actions)):
  observation, self._rewards[i], self._dones[i], info = env.step(action)
  if self._dones[i]:
      info["terminal_observation"] = observation
      observation = env.reset()
  observations.append(observation)
  infos.append(info)
```

so for example by doing
```infos.add_info({'env 1 info': 1}, 1)``` 
on this 
``` [{"example_info": 0}, {"example_info": 1}, {"example_info": 2}]``` 

we will obtain
 ``` [{"example_info": 0}, {"example_info": 1}, {"example_info": 2}, {'env 1 info': 1}]```

File: gym/vector/utils/info_strategies.py (hunk 4 above)
Comment (pseudo-rnd-thoughts) [reply gym/vector/utils/info_strategies.py:76]: I dont think we should use `np.nan` unless necessary. To me it is a special value that shouldn't be used unless it is an actual `nan`
Personally, I was originally thinking `np.zeros` as the dtype should automatically know the `zero` dtype. However for bool, then this get a bit complex as it is quite reasonable to get `False` as info data. We could have a special case just for bools. 
What are your ideas?

File: gym/wrappers/record_episode_statistics.py (hunk 5 above)
Comment (gianlucadecola) [reply gym/wrappers/record_episode_statistics.py:45]: It will update the dictionary with the value of `info`; 
Since ```RecordEpisdeStatistics``` only adds:
```
"episode": {
      "r": episode_return,
      "l": episode_length,
      "t": round(time.perf_counter() - self.t0, 6),
  }
```
I think we are okay; otherwise we can use a more verbose way of updating the dictionary and raising a warning if there is a key collision (If there is an environment that return an "episode" key in in the `step` )

File: tests/vector/test_info_format_strategies.py (hunk 7 above)
Comment (pseudo-rnd-thoughts) [reply tests/vector/test_info_format_strategies.py:17]: Ohh I get what you mean, ignore my comment. I forgot about the info data being updated after it being passed to the info strategy. Do we want a check in the `add_info` that the data does not already exist or do we assume people will not be stupid? (like me)?

File: gym/wrappers/record_episode_statistics.py (hunk 5 above)
Comment (pseudo-rnd-thoughts) [reply gym/wrappers/record_episode_statistics.py:45]: I didn't realise that is what it returns. Ok, if we know there won't be a key collision then there isn't an issue. Are we sure this can't be an issue in the other places as well?

File: gym/vector/utils/info_strategies.py (hunk 4 above)
Comment (gianlucadecola) [reply gym/vector/utils/info_strategies.py:76]: I've thought about it; this is actually quite inefficient
```python
[None for _ in range(self.num_envs)]
```
and should be at least changed to:
```python
[None] * self.num_envs
```
Which has a creation time comparable to a np.array.
However, I also expect that users want to do some kind of processing on the info and will probably convert the list themselves to a numpy array. With this in mind, I think we are moving towards this approach:
https://github.com/openai/gym/issues/2657#issuecomment-1055597901
We can leverage the new modularity of the `info` and add this as a new strategy with np.arrays initialized as np.zeros (with proper management of the bool case). 
I'm just thinking at the benefits of keeping them both at this point and the only benefit I see is that `None` is clearer than 0's or False (https://github.com/openai/gym/issues/2657#issuecomment-1104474552)

gym/wrappers/record_episode_statistics.py
gianlucadecola: I've done some search around the codebase and I didn't find anything; there can still be the case of a custom environment that returns info with the "episode" key, in this case I think it is good to raise at least a warning telling the user that the wrapper is overriding an existing key. (this will happen silently also now on the current version on master)

gym/vector/utils/info_strategies.py
```diff
@@ -59,21 +62,37 @@ class BraxVecEnvInfoStrategy(VecEnvInfoStrategy):
...
     """
 
     def __init__(self, num_envs: int):
         self.num_envs = num_envs
         self.info = {}
 
+    def _init_info_array(self, num_envs: int, key: str, dtype: type) -> np.ndarray:
+        if dtype == bool:
+            warnings.warn(
+                "found boolean dtype in info for key `%s`, empty values are set to `False`"
```
pseudo-rnd-thoughts, line 80: Smart format string

gym/vector/utils/info_strategies.py
```diff
@@ -59,21 +62,37 @@ class BraxVecEnvInfoStrategy(VecEnvInfoStrategy):
 
     Example with 3 environments in which only the last has the `terminal_observation` info:
         {
```
pseudo-rnd-thoughts, line 64: Could you make this an actual docstring code example. See TransformObservation wrapper for an example, you should just need to put a seperating line, >>> for the first line and ... for the rest

gym/vector/utils/info_strategies.py
```diff
@@ -87,15 +106,22 @@ class InfoStrategiesEnum(Enum):
...
+    _strategies = {
         InfoStrategiesEnum.classic.value: ClassicVecEnvInfoStrategy,
         InfoStrategiesEnum.brax.value: BraxVecEnvInfoStrategy,
     }
 
     def get_info_strategy(info_format: str) -> VecEnvInfoStrategy:
-        if info_format not in InfoStrategyFactory.strategies:
+        if info_format not in InfoStrategyFactory._strategies:
             raise InvalidInfoFormat(
                 "%s is not an available format for info, please choose one between %s"
```
pseudo-rnd-thoughts, line 117: Smart format string

gianlucadecola:
So I'm converting this as a draft since it will require a different approach based on: https://github.com/google/brax/issues/194  
The new default behavior will be the Brax format so this PR will introduce a breaking change. Instead of having different strategies a wrapper will be added to convert back the info to the "classic" format.

gianlucadecola:
Added tests, fixed some conflicts and update docstring to match current docstyle. It should be ready for review @pseudo-rnd-thoughts

gianlucadecola:
Did the renaming and incorporate the info_processor into VecEnv

gym/wrappers/record_episode_statistics.py
```diff
@@ -7,6 +7,89 @@
...
+            env_num (int): env number.
+        """
+        self.info = {**self.info, **infos}
+
+    def get_info(self):
+        """Return info."""
+        return self.info
+
+
+class BraxVecEnvStatsInfo:
```
pseudo-rnd-thoughts, line 44: Update name

gym/vector/sync_vector_env.py
```diff
@@ -115,7 +124,7 @@ def reset_wait(
             else:
                 observation, data = env.reset(**kwargs)
```
pseudo-rnd-thoughts, line 125: Update to `info` from `data`

gym/vector/vector_env.py
```diff
@@ -211,6 +213,27 @@ def seed(self, seed=None):
             "Please use `env.reset(seed=seed) instead in VectorEnvs."
         )
 
+    def _add_info(self, infos: dict, info: dict, env_num: int):
```
pseudo-rnd-thoughts, line 216: While a docstring here is not necessary, this is a complex and important functions. Could you give it a docstring to explain what you are doing? And a return type

Technically, we don't need the return type as infos is a reference here so it will be updated outside the function. However, this is not a critical problem and may make more sense to return it

gym/vector/vector_env.py
```diff
@@ -211,6 +213,27 @@ def seed(self, seed=None):
...
+                info_array, array_mask = self._init_info_array(type(info[k]))
+            else:
+                info_array, array_mask = infos[k], infos[f"_{k}"]
+
+            info_array[env_num], array_mask[env_num] = info[k], True
+            infos[k], infos[f"_{k}"] = info_array, array_mask
+        return infos
+
+    def _init_info_array(self, dtype: type) -> np.ndarray:
+        if dtype not in [int, float, bool]:
```
pseudo-rnd-thoughts, line 228: Does this work for numpy type, `int64`, etc? 
Could you add this to the docstring to explain the optimised dtypes

gym/vector/vector_env.py
```diff
@@ -211,6 +213,27 @@ def seed(self, seed=None):
...
+            dtype = object
```
pseudo-rnd-thoughts, line 229: This is unnecessary, just add it to the next line

gym/wrappers/record_episode_statistics.py
```diff
@@ -7,11 +7,98 @@
 import gym
 
 
+class ClassicStatsInfo:
```
pseudo-rnd-thoughts, line 10: Would it be possible to inherit from `VecVenStatsInfo` then only override the `add_episode_statistics`? This seem like the only difference

Update name to `VecEnvClassicStatsInfo`

gym/wrappers/record_episode_statistics.py
```diff
@@ -56,35 +159,35 @@ def reset(self, **kwargs):
 
     def step(self, action):
         """Steps through the environment, recording the episode statistics."""
+        infos_processor = self.stats_info_processor(self.num_envs)
```
pseudo-rnd-thoughts, line 162: Do we need to initialise this with every step?

gym/wrappers/vec_info_to_classic.py
```diff
@@ -0,0 +1,85 @@
...
+    Example::
+
+    >>> # actual
+    ...  {
+    ...      k: np.array[0., 0., 0.5, 0.3],
+    ...      _k: np.array[False, False, True, True]
+    ...  }
+    ...
+    ... # classic
+    ... [{}, {}, {k: 0.5}, {k: 0.3}]
```
pseudo-rnd-thoughts, line 29: `>>>`

gym/wrappers/vec_info_to_classic.py
```diff
@@ -0,0 +1,85 @@
...
+    environments. If using other wrappers that perform
+    operation on info like `RecordEpisodeStatistics` this
+    need to be the outermost wrapper.
+
+    i.e. ClassicVectorInfo(RecordEpisodeStatistics(envs))
+
+    Example::
+
+    >>> # actual
+    ...  {
```
pseudo-rnd-thoughts, line 23: >>>

pseudo-rnd-thoughts (changes requested):

gym/wrappers/record_episode_statistics.py
gianlucadecola: So, I'm initializing now in the `__init__` and in the `step` just resetting `info_processor.info` to an empty dict. Seems better?

gym/wrappers/record_episode_statistics.py
gianlucadecola: I made `VecEnvStatsInfo` inherit from `StatsInfo` which can be seen more as the "base" class for non vectorized environments.
Renamed to `StatsInfo`  instead of `VecEnvClassicStatsInfo` because it is used when the wrapped env is not vectorized. I think `Classic` word can be removed here because we do not have this differentiation in non vectorized envs.

gym/vector/vector_env.py
gianlucadecola: Something like this `np.float64(1)` will fall into the `object` category. Do you suggest adding all the NumPy numeric dtypes to the list? Maybe in a more clean way instead of a long list of numpy dtypes + native dtypes

pseudo-rnd-thoughts: Yes, I think we should certainly support numpy dtype. For numpy arrays, I think we are going to have to support them through objects not as actual multi-dimensional arrays

gym/vector/async_vector_env.py
```diff
@@ -406,10 +411,20 @@ def step_wait(self, timeout=None):
                 f"The call to `step_wait` has timed out after {timeout} second(s)."
             )
 
-        results, successes = zip(*[pipe.recv() for pipe in self.parent_pipes])
+        observations_list, rewards, dones, infos = [], [], [], {}
+        successes = []
+        for i, pipe in enumerate(self.parent_pipes):
+            result, success = pipe.recv()
+            obs, rew, done, info = result
+            infos = self._add_info(infos, info, i)
```
pseudo-rnd-thoughts, line 419: Could you move infos to after dones such that we address each result in order

gym/wrappers/record_episode_statistics.py
```diff
@@ -7,11 +7,77 @@
...
+            env_num (int): env number.
+        """
+        self.info = {**self.info, **infos}
+
+    def get_info(self):
+        """Return info."""
+        return self.info
+
+
+class VecEnvStatsInfo(StatsInfo):
```
pseudo-rnd-thoughts, line 45: Change to `VectorEnvStatsInfo`

gym/wrappers/record_episode_statistics.py
```diff
@@ -46,6 +124,10 @@ def __init__(self, env: gym.Env, deque_size: int = 100):
         self.return_queue = deque(maxlen=deque_size)
         self.length_queue = deque(maxlen=deque_size)
         self.is_vector_env = getattr(env, "is_vector_env", False)
+        if self.is_vector_env:
```
pseudo-rnd-thoughts, line 127: Is this compatible when the `VectorListInfo` is applied to the vector envs?

gym/wrappers/vec_info_to_classic.py
```diff
@@ -0,0 +1,84 @@
...
+    operation on info like `RecordEpisodeStatistics` this
+    need to be the outermost wrapper.
+
+    i.e. ClassicVectorInfo(RecordEpisodeStatistics(envs))
+
+    Example::
+
+    >>> # actual
+    >>> {
+    ...      k: np.array[0., 0., 0.5, 0.3],
```
pseudo-rnd-thoughts, line 24: "k" and "_k"

gym/wrappers/vec_info_to_classic.py
```diff
@@ -0,0 +1,84 @@
...
+    ...      _k: np.array[False, False, True, True]
+    ...  }
+    >>> # classic
+    >>> [{}, {}, {k: 0.5}, {k: 0.3}]
```
pseudo-rnd-thoughts, line 28: "k" and "_k"

tests/wrappers/test_record_episode_statistics.py
```diff
@@ -50,8 +50,11 @@ def test_record_episode_statistics_with_vectorenv(num_envs, asynchronous):
...
     for _ in range(max_episode_step + 1):
         _, _, dones, infos = envs.step(envs.action_space.sample())
-        for idx, info in enumerate(infos):
-            if dones[idx]:
-                assert "episode" in info
-                assert all([item in info["episode"] for item in ["r", "l", "t"]])
-                break
+        if any(dones):
+            assert "episode" in infos
+            assert "_episode" in infos
```
pseudo-rnd-thoughts, line 55: Could we also assert that the `info["_episode"] == dones`

pseudo-rnd-thoughts (changes requested):

gym/wrappers/record_episode_statistics.py
gianlucadecola: You mean a scenario like this 
`RecordEpisodeStatistics(VectorListInfo(gym.vector.make('CartPole-v1', num_envs=3)))`
 instead of this
`VectorListInfo(RecordEpisodeStatistics(gym.vector.make('CartPole-v1', num_envs=3)))`
?
This won't work, I've stated in docstring of `VectorListInfo` that it need to be the outermost wrapper if there are other wrappers performing operations on info. Other than that, if there is a way (which atm I'm not aware of) of checking all the wrappers chaining I could raise a warning or an exception if `VectorListInfo` is not the outermost

pseudo-rnd-thoughts: Yes, we could add a test that checks the `env` (and all sub wrappers) are not instances of `VectorListInfo`. There is similar code in the `tests/wrappers/utils.py`. 
Hopefully, this will stop an issue in the future

gym/wrappers/record_episode_statistics.py
pseudo-rnd-thoughts: Yes, amazing, thanks

gym/wrappers/record_episode_statistics.py
```diff
@@ -7,11 +7,77 @@
 import gym
 
 
+class EnvStatsInfo:
```
RedTachyon, line 10: I'm not sure I like the way these two classes are structured in general. The only functionality which is not a one-liner is `VectorEnvStatsInfo.add_episode_statistics`. Wouldn't it be possible to just maintain the `self.info` dictionary inside of `RecordEpisodeStatistics`, and then in the code it'd be updated with the one-liners I mentioned, and `add_episode_statistics` which would be a separate function?

gym/vector/vector_env.py
```diff
@@ -211,6 +217,42 @@ def seed(self, seed=None):
...
+            else:
+                info_array, array_mask = infos[k], infos[f"_{k}"]
+
+            info_array[env_num], array_mask[env_num] = info[k], True
+            infos[k], infos[f"_{k}"] = info_array, array_mask
+        return infos
+
+    def _init_info_arrays(self, dtype: type) -> np.ndarray:
+        if dtype in [int, float, bool] or issubclass(dtype, np.number):
+            array = np.zeros(self.num_envs, dtype=dtype)
```
RedTachyon, line 249: Can we add some explicit handling for numpy arrays too? I think the current implementation can cause some problems through a mixup between e.g. `np.float32` and `np.float64`, or it will even default to the `object` type. And I can easily see passing an array in the info dict as a desired functionality

gym/vector/vector_env.py
```diff
@@ -211,6 +217,42 @@ def seed(self, seed=None):
...
+        else:
+            array = np.zeros(self.num_envs, dtype=object)
```
RedTachyon, line 251: Was there a discussion somewhere already whether it should be initialized as `0` as opposed to `nan`? Intuitively I prefer `nan` because a `0` can have an actual meaning. Especially since for `object` we're using None. And in the end you should use the mask to index it away anyways.

gym/wrappers/vector_list_info.py
```diff
@@ -0,0 +1,84 @@
...
+        observation, reward, done, infos = self.env.step(action)
+        classic_info = self._convert_info_to_classic(infos)
+
+        return observation, reward, done, classic_info
+
+    def reset(self, **kwargs):
+        """Resets the environment using kwargs."""
+        if not kwargs.get("return_info"):
+            obs = self.env.reset(**kwargs)
+            return obs
```
RedTachyon, line 54: `return self.env.reset(**kwargs)`

gym/wrappers/vector_list_info.py
```diff
@@ -0,0 +1,84 @@
...
+        classic_info = self._process_episode_statistics(infos, classic_info)
+        for k in infos:
+            if k.startswith("_"):
+                continue
+            for i, has_info in enumerate(infos[f"_{k}"]):
+                if has_info:
+                    classic_info[i][k] = infos[k][i]
+        return classic_info
+
+    def _process_episode_statistics(self, infos, classic_info):
```
RedTachyon, line 71: Type annotations, docstring

tests/vector/test_vector_env_info.py
```diff
@@ -0,0 +1,30 @@
...
+
+ENV_ID = "CartPole-v1"
+NUM_ENVS = 3
+ENV_STEPS = 50
+
+
+@pytest.mark.parametrize("asynchronous", [True, False])
+def test_vector_env_info(asynchronous):
+    env = gym.vector.make(ENV_ID, num_envs=NUM_ENVS, asynchronous=asynchronous)
+    env.reset()
```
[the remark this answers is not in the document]
RedTachyon: (end everywhere else where this pattern appears)

tests/wrappers/test_vector_list_info.py
```diff
@@ -0,0 +1,55 @@
...
+    VectorListInfo(vector_env)
+
+    with pytest.raises(AssertionError):
+        VectorListInfo(env)
+
+
+def test_info_to_list():
+    env_to_wrap = gym.vector.make(ENV_ID, num_envs=NUM_ENVS)
+    wrapped_env = VectorListInfo(env_to_wrap)
+    _, info = wrapped_env.reset(return_info=True)
```
RedTachyon, line 24: (also seeding in these tests)

RedTachyon (commented):
I left a bunch of comments, it's on the right track - most comments are stylistic and nitpicks, with I think one or two about functionality.

gym/vector/vector_env.py
gianlucadecola: Here: https://github.com/openai/gym/pull/2773#discussion_r859582951
Basically we decided for 0 since the `_key` can be used for masking whether is a 0 coming from not having data or a 0 with an actual meaning. There was also some discussion in the issue regarding the initialization: https://github.com/openai/gym/issues/2657#issuecomment-1055597901

tests/vector/test_vector_env_info.py
```diff
@@ -0,0 +1,30 @@
...
+    for _ in range(ENV_STEPS):
+        action = env.action_space.sample()
+        _, _, dones, infos = env.step(action)
+        if any(dones):
```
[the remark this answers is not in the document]
gianlucadecola: Correct. Will add other test cases :+1:

gym/vector/vector_env.py
gianlucadecola: No problem for me, atm if an array is added to the info dict the resulting output will fall in the `object` dtype resulting in this:
```python
"terminal_observation": [
        array([0.2097086, -0.5355723, -0.21343598, -0.11173592], dtype=float32),
        None,
        None
    ]
```
Something I can think of is making the output in this form:
```python
"terminal_observation": [
        array(
            [[0.2097086, -0.5355723, -0.21343598, -0.11173592],
            [0., 0., 0., 0.],
            [0., 0., 0., 0.]],
           dtype=float32),
    ]
```
I'm not sure this adhere to the format of google-brax tho

gianlucadecola:
Hey, thanks a lot for the review,  added a bunch of comments, will work on that soon  :+1:

gym/vector/vector_env.py
gianlucadecola: Some more context also here: https://github.com/google/brax/issues/194

gym/wrappers/vector_list_info.py
```diff
@@ -0,0 +1,84 @@
...
+        episode_statistics = infos.pop("episode", False)
+        if not episode_statistics:
+            return classic_info
+
+        episode_statistics_mask = infos.pop("_episode")
+        for i, has_info in enumerate(episode_statistics_mask):
+            if has_info:
+                classic_info[i]["episode"] = {}
```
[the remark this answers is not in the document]
gianlucadecola: I've added details in the docstring

gym/wrappers/record_episode_statistics.py
gianlucadecola: So I've simplified things a lot, got rid of the classes, and just left the function to process episode_statistics. Previously the idea was to have the same interface for both vectorized and non vectorized envs and then choose the processing strategy depending on the type of env (this was the reason of some unused args in the non vectorized envs class)

gianlucadecola:
@RedTachyon I've done some changes, when you have time could you please take a look? Particularly at:
https://github.com/openai/gym/pull/2773#discussion_r877098631
https://github.com/openai/gym/pull/2773#discussion_r877113582

PR #2774, opened 2022-04-23
Remove orphan comment

ikamensh:
The comment used to be due to the line that is not there anymore, like:
```python
from gym.envs.mujoco.mujoco_env import MujocoEnv

# ^^^^^ so that user gets the correct error
# message if mujoco is not installed correctly
```

Now the comment is useless.

### DIFF
--- a/gym/envs/mujoco/__init__.py
+++ b/gym/envs/mujoco/__init__.py
@@ -1,5 +1,3 @@
-# ^^^^^ so that user gets the correct error
-# message if mujoco is not installed correctly
 from gym.envs.mujoco.ant import AntEnv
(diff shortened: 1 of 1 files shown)


ikamensh:
I've realised the import that this comment referred to was not deleted, but rather moved by isort in PR #2602. So maybe the comment should be moved to the import instead of deleting it. Opinions?

pseudo-rnd-thoughts:
If the code that the comment refers to still exists then it would be helpful that the comment to br next to the code.
If the comment is unneeded then it would be good to remove to prevent confusion

pseudo-rnd-thoughts:
@jkterry1 LGTM

PR #2775, opened 2022-04-24
Fix docstring format on seed arg to reset

efokschaner:
I have a copy of this string in another project and it failed the darglint docstring style check.
This fixes that, if it's something you care to fix.

## Type of change

Docstring-only fix

This PR is not locally tested, lmk if it requires local testing before merge.


pseudo-rnd-thoughts:
Thanks @efokschaner for the PR
@jkterry1 LGTM

Issue #2781, opened 2022-04-27
[Bug Report] pygame is required but is not listed in package setup requirements


piteren:
**Describe the bug**
gym/envs/classic_control/cartpole.py imports pygame but it is not listed in package setup requirements

**Code example**
gym.make('CartPole-v1')

**System Info**
gym installed with pip
NAME="Ubuntu"
VERSION="20.04.4 LTS (Focal Fossa)"
Python 3.7.13

- [x] I have checked that there is no similar [issue](https://github.com/openai/gym/issues) in the repo (**required**)

balisujohn:
What version of gym do you have installed?

piteren:
with:
pip install gym
I have installed:
gym                          0.23.1

balisujohn:
pygame is listed in [requirements.txt](https://github.com/openai/gym/blob/master/requirements.txt) in the version of gym you're using. I'm not super knowledgeable about pip packaging; is there a particular config file you're referring to when you say pygame is missing from the package setup requirements?

piteren:
Yes, it is listed in requirements.txt, but as far as I know it should be also listed in setup.py in install_requires [line 48] section for package proper installation.

balisujohn:
try

`pip install gym[box2d]`

or 

`pip install gym[all]`

piteren:
I have just installed pygame separately with
`pip install pygame`
and it works for me.

pseudo-rnd-thoughts:
@piteren Thanks for the comment, this is a fix we have made in the current master branch however it is not in the current release. 
We hope it will be released soon

bluenote10:
Nonetheless this is a regression in a sense, because it makes the basic [tutorial](https://gym.openai.com/docs/) invalid.

Following the suggested steps of the tutorial is an error now:

```sh
$ virtualenv /tmp/tmp_venv -p /usr/bin/python3.8
$ pip install gym
$ python -c "import gym; gym.make('CartPole-v0')"
```

This used to work fine with gym version 0.21.0. In fact, it looks like in 0.21 it wasn't necessary to have pygame installed to use the cartpole env. It is a bit unfortunate that the env now requires such a "dirty" / heavy dependency as pygame :confused: I'd prefer to use the cartpole env without pygame like it was in 0.21.

pseudo-rnd-thoughts:
@bluenote10 Thanks for the comment, I understand your annoyance. 
[...]

PR #2782, opened 2022-04-27
Place try except around all imports of pygame 

pseudo-rnd-thoughts:
Currently, if a user has not installed pygame or another of the optional modules then an error is thrown without a reason
This PR wraps all of these imports with a try except that will raise the import error with a helpful message saying that the appropriate `pip install gym[...]` needs to be run. 

Inspired by https://github.com/openai/gym/issues/2781

### DIFF
--- a/gym/envs/box2d/bipedal_walker.py
+++ b/gym/envs/box2d/bipedal_walker.py
@@ -3,20 +3,25 @@ __credits__ = ["Andrea PIERRÉ"]
 import math
 from typing import Optional
 
-import Box2D
 import numpy as np
-from Box2D.b2 import (
-    circleShape,
-    contactListener,
-    edgeShape,
-    fixtureDef,
-    polygonShape,
-    revoluteJointDef,
-)
 
 import gym
 from gym import error, spaces
 from gym.utils import EzPickle
+
+try:
+    import Box2D
+    from Box2D.b2 import (
+        circleShape,
+        contactListener,
+        edgeShape,
+        fixtureDef,
+        polygonShape,
+        revoluteJointDef,
+    )
+except ImportError:
+    raise ImportError("Box2D is not installed, run `pip install gym[box2d]`")
+
 
 FPS = 50
 SCALE = 30.0  # affects how fast-paced the game is, forces should be adjusted as well
@@ -586,8 +591,11 @@ class BipedalWalker(gym.Env, EzPickle):
         return np.array(state, dtype=np.float32), reward, done, {}
 
     def render(self, mode: str = "human"):
-        import pygame
-        from pygame import gfxdraw
+        try:
+            import pygame
+            from pygame import gfxdraw
+        except ImportError:
+            raise ImportError("Pygame is not installed, run `pip install gym[box2d]`")
 
         if self.screen is None:
             pygame.init()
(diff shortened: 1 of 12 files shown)


gym/envs/box2d/bipedal_walker.py
```diff
@@ -586,8 +591,11 @@ def step(self, action: np.ndarray):
         return np.array(state, dtype=np.float32), reward, done, {}
 
     def render(self, mode: str = "human"):
-        import pygame
-        from pygame import gfxdraw
+        try:
+            import pygame
+            from pygame import gfxdraw
+        except ImportError:
+            raise ImportError("Pygame is not installed, run `pip install gym[box2d]`")
```
balisujohn, line 598: Capitalization of first letter of error message inconsistent with other error messages.

gym/envs/toy_text/taxi.py
```diff
@@ -237,7 +237,12 @@ def render(self, mode="human"):
             return self._render_gui(mode)
 
     def _render_gui(self, mode):
-        import pygame  # dependency to pygame only if rendering with human
```
balisujohn, line 240: maybe this comment shouldn't be deleted, unless it's out of date

balisujohn (changes requested):
This seems like an overall useful PR to me. I think it's always better to fail with more descriptive error messages, and I think it is a nice touch how it even tells you which specialized install to do to get the correct dependencies for the particular environment you're using.

Also something interesting is that gym does have a DependencyNotInstalled error that is used by the Mujoco envs

 https://github.com/openai/gym/blob/bf093c689045953e3e1c0c846aeb5d2e2e5e4798/gym/error.py#L80

https://github.com/openai/gym/blob/bf093c689045953e3e1c0c846aeb5d2e2e5e4798/gym/envs/mujoco/mujoco_env.py#L14

I would say that on balance I would recommend using the DependencyNotInstalled error (should require a minimal change) instead of re-raising an import error, but I'm curious to hear others' opinions.

gym/envs/box2d/bipedal_walker.py
pseudo-rnd-thoughts: Thanks for the spot, I will revert to the rest of the code base

gym/envs/toy_text/taxi.py
pseudo-rnd-thoughts: Thanks, I deleted it because I thought it was kinda obvious but I imagine in a year or twos time it may not so I will change back

pseudo-rnd-thoughts:
Thanks for the review
I didn't know about the error so thanks for bring it up, I will change all of the errors to be that as I agree it is a more descriptive error message.

gym/envs/box2d/bipedal_walker.py
balisujohn: I just meant within the PR, since some error messages within the PR were capitalized and some weren't. I'm not sure what the standard is across gym.

pseudo-rnd-thoughts: Ohh, I realised that when implementing the fix
Personally, I dont mind, just agree that it should be consistent
I changed it to lower case on a coin flip. Can change it to upper case if anyone has any major issues with that

balisujohn:
LGTM :^)

PR #2783, opened 2022-04-29
Adding docstring and tests to `PlayPlot`

gianlucadecola:
In a previous PR (https://github.com/openai/gym/pull/2743) I was working on adding tests to `gym.utils.play`. 
`PlayPlot` was left outside; I've now added docstring, type hints and tests.
Tests are actually skipped in the current CI configuration since matplotlib is not a requirement.

The only thing that bothers me is that the window created for plotting hangs on the screen until all tests are finished

### DIFF
--- a/gym/utils/play.py
+++ b/gym/utils/play.py
@@ -1,4 +1,4 @@
-from typing import Callable, Dict, Optional, Tuple
+from typing import Callable, Dict, List, Optional, Tuple
 
 import pygame
 from numpy.typing import NDArray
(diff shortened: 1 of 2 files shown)


jkterry1:
I'm temped to say this function should be removed and not patched?

gianlucadecola:
> I'm temped to say this function should be removed and not patched?

Seems more like an example usage of the callback than a function strictly related to gym indeed. Maybe we can move out of gym and put it as an example in the docs? Just to have some reference on how to use the callback?

jkterry1:
Sounds good to me

PR #2784, opened 2022-04-29
Docstrings for wrappers

Markus28:
# Description
I added docstrings (corresponding to the table in the docs) for a few wrappers. 

Since I was already adding typing stuff for the docstrings, I added type hints for three constructors. We might want to move that to a different PR though, especially because I have no idea how type hints work.

## Changes
- New docstrings
- Type hints for GrayScaleObservation, ResizeObservation, TimeLimit


pseudo-rnd-thoughts:
@jkterry1 LGTM
@Markus28 Thanks for doing this, Im doing this for all of the docstring currently missing and this is very helpful

PR #2785, opened 2022-04-30
Add Pydocstyle to CI

pseudo-rnd-thoughts:
Currently gym has a variety of docstrings style across the project
This PR adds pydocstyle to the pre-commit CI to ensure that functions have a docstring

Pydocstyle has three docstyles: google, numpy and pep257, and @RedTachyon and I agree that the google docstyle is the best of those. 

Due to the number of docstrings, we have excluded all folders except the root folder and gym/ root folder (I accidentally added test_autoreset and autoreset wrapper) 
The aim is to add more PRs for each of the gym folders with docstrings

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -1,9 +1,10 @@
+"""Core API for Environment, Wrapper, ActionWrapper and ObservationWrapper."""
 from __future__ import annotations
 
 from abc import abstractmethod
-from typing import Generic, Optional, SupportsFloat, Tuple, TypeVar, Union
+from typing import Generic, Optional, SupportsFloat, TypeVar, Union
 
-from gym import spaces
+from gym import Space
 from gym.logger import deprecation
 from gym.utils import seeding
 from gym.utils.seeding import RandomNumberGenerator
@@ -13,9 +14,10 @@ ActType = TypeVar("ActType")
 
 
 class Env(Generic[ObsType, ActType]):
-    """The main OpenAI Gym class. It encapsulates an environment with
-    arbitrary behind-the-scenes dynamics. An environment can be
-    partially or fully observed.
+    """The main OpenAI Gym class.
+
+    It encapsulates an environment with arbitrary behind-the-scenes dynamics.
+    An environment can be partially or fully observed.
 
     The main API methods that users of this class need to know are:
 
@@ -42,8 +44,8 @@ class Env(Generic[ObsType, ActType]):
     spec = None
 
     # Set these in ALL subclasses
-    action_space: spaces.Space[ActType]
-    observation_space: spaces.Space[ObsType]
+    action_space: Space[ActType]
+    observation_space: Space[ObsType]
 
     # Created
     _np_random: RandomNumberGenerator | None = None
@@ -60,11 +62,10 @@ class Env(Generic[ObsType, ActType]):
         self._np_random = value
 
     @abstractmethod
-    def step(self, action: ActType) -> Tuple[ObsType, float, bool, dict]:
-        """Run one timestep of the environment's dynamics. When end of
-        episode is reached, you are responsible for calling :meth:`reset`
-        to reset this environment's state.
+    def step(self, action: ActType) -> tuple[ObsType, float, bool, dict]:
+        """Run one timestep of the environment's dynamics.
 
+        When end of episode is reached, you are responsible for calling :meth:`reset` to reset this environment's state.
         Accepts an action and returns a tuple (observation, reward, done, info).
 
         Args:
@@ -93,8 +94,7 @@ class Env(Generic[ObsType, ActType]):
         return_info: bool = False,
         options: Optional[dict] = None,
     ) -> Union[ObsType, tuple[ObsType, dict]]:
-        """Resets the environment to an initial state and returns an initial
-        observation.
+        """Resets the environment to an initial state and returns an initial observation.
 
         This method should also reset the environment's random number
         generator(s) if ``seed`` is an integer or if the environment has not
@@ -200,6 +200,7 @@ class Env(Generic[ObsType, ActType]):
         return self
 
     def __str__(self):
+        """Returns a string for the environment with the spec id if specified."""
         if self.spec is None:
             return f"<{type(self).__name__} instance>"
         else:
@@ -230,94 +231,117 @@ class Wrapper(Env[ObsType, ActType]):
     """
 
     def __init__(self, env: Env):
+        """Wraps the environment to allow a modular transformation."""
         self.env = env
 
-        self._action_space: spaces.Space | None = None
-        self._observation_space: spaces.Space | None = None
+        self._action_space: Space | None = None
+        self._observation_space: Space | None = None
         self._reward_range: tuple[SupportsFloat, SupportsFloat] | None = None
         self._metadata: dict | None = None
 
     def __getattr__(self, name):
+        """Gets an attribute if the name does not start with _."""
         if name.startswith("_"):
             raise AttributeError(f"accessing private attribute '{name}' is prohibited")
         return getattr(self.env, name)
 
     @property
     def spec(self):
+        """Returns the environment spec."""
         return self.env.spec
 
     @classmethod
     def class_name(cls):
+        """Returns the class name."""
         return cls.__name__
 
     @property
-    def action_space(self) -> spaces.Space[ActType]:
+    def action_space(self) -> Space[ActType]:
+        """Returns the action space."""
         if self._action_space is None:
             return self.env.action_space
         return self._action_space
 
     @action_space.setter
-    def action_space(self, space):
+    def action_space(self, space: Space):
+        """Sets the action space."""
         self._action_space = space
 
     @property
-    def observation_space(self) -> spaces.Space:
+    def observation_space(self) -> Space:
+        """Returns the observation space."""
         if self._observation_space is None:
             return self.env.observation_space
         return self._observation_space
 
     @observation_space.setter
-    def observation_space(self, space):
+    def observation_space(self, space: Space):
+        """Sets the observation shape."""
         self._observation_space = space
 
     @property
     def reward_range(self) -> tuple[SupportsFloat, SupportsFloat]:
+        """Return the reward range."""
         if self._reward_range is None:
             return self.env.reward_range
         return self._reward_range
 
     @reward_range.setter
-    def reward_range(self, value):
+    def reward_range(self, value: tuple[SupportsFloat, SupportsFloat]):
+        """Sets the reward range."""
         self._reward_range = value
 
     @property
     def metadata(self) -> dict:
+        """Returns the environment metadata."""
         if self._metadata is None:
             return self.env.metadata
         return self._metadata
 
     @metadata.setter
     def metadata(self, value):
+        """Sets the metadata."""
         self._metadata = value
 
-    def step(self, action: ActType) -> Tuple[ObsType, float, bool, dict]:
+    def step(self, action: ActType) -> tuple[ObsType, float, bool, dict]:
+        """Steps through the environment with action."""
         return self.env.step(action)
 
     def reset(self, **kwargs) -> Union[ObsType, tuple[ObsType, dict]]:
+        """Resets the environment with kwargs."""
         return self.env.reset(**kwargs)
 
     def render(self, **kwargs):
+        """Renders the environment with kwargs."""
         return self.env.render(**kwargs)
 
     def close(self):
+        """Closes the environment."""
         return self.env.close()
 
     def seed(self, seed=None):
+        """Seeds the environment."""
         return self.env.seed(seed)
 
     def __str__(self):
+        """Returns the wrapper name and the unwrapped environment string."""
         return f"<{type(self).__name__}{self.env}>"
 
     def __repr__(self):
+        """Returns string of the wrapper."""
         return str(self)
 
     @property
     def unwrapped(self) -> Env:
+        """Unwraps all wrappers to the environment."""
         return self.env.unwrapped
 
 
 class ObservationWrapper(Wrapper):
+    """A wrapper that can modify the returning observations from a step."""
+
     def reset(self, **kwargs):
+        """Resets the environment with the modified observations."""
         if kwargs.get("return_info", False):
             obs, info = self.env.reset(**kwargs)
             return self.observation(obs), info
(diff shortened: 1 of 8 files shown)


gym/error.py
```diff
@@ -168,83 +123,69 @@ def __init__(
...
 
 
 class WrapAfterConfigureError(Error):
-    pass
+    """Unused error."""
 
 
 class RetriesExceededError(Error):
-    pass
+    """unused error."""
```
Markus28, line 162: Capitalization

pseudo-rnd-thoughts: For all of the unused error, I was tempted to delete them all but just wanted to confirm their won't be an issue with this.

pseudo-rnd-thoughts: https://github.com/openai/gym/blame/master/gym/error.py
The error looks like they were added in the project's initial commits, wow

Im going to add a deprecation warning to them for gym 1.0 I think
This is the simplest way of removing without causing problems to other projects

CONTRIBUTING.md
```diff
@@ -38,3 +38,13 @@ The CI will run several checks on the new code pushed to the Gym repository. The
...
+
+## Docstrings
+Pydocstyle has been added to the pre-commit process such that all new functions follow the (google docstring style)[https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html].
+All new functions require either a short docstring, a single line explaining the purpose of a function
+or a multiline docstring that documents each argument and the return type (if there is one) of the function.
+In addition, new file and class require top docstrings that should outline the purpose of the file/class.
+For classes, code block examples can be provided in the top docstring and not the constructor arguments.
+
+To check your docstrings are correct, run `pre-commit run --al-files` or `pydocstyle --source --explain --convention=google`.
+If all docstrings that fail, the source and reason for the failure is provided.
```
Markus28, line 50: I think if should be something like for, right?

gym/core.py
```diff
@@ -238,125 +240,152 @@ def __init__(self, env: Env):
...
-        return self.env.reset(**kwargs)
+    """A wrapper that can modify the action before stepping."""
 
     def step(self, action):
+        """Runs the environment step using the modified action."""
         return self.env.step(self.action(action))
 
     @abstractmethod
     def action(self, action):
+        """Modifies the action during the action."""
```
Markus28, line 385: That sentence sounds a little weird

tests/wrappers/test_autoreset.py
```diff
@@ -10,33 +12,31 @@
...
 
 class DummyResetEnv(gym.Env):
-    """
-    A dummy environment which returns ascending numbers starting
-    at 0 when self.step() is called. After the third call to self.step()
-    done is true. Info dicts are also returned containing the same number
-    returned as an observation, accessible via the key "count".
-    This environment is provided for the purpose of testing the
-    autoreset wrapper.
+    """A dummy environment which returns ascending numbers starting at 0 when self.step() is called.
```
Markus28, line 15: Add code tags for `self.step()`

tests/wrappers/test_autoreset.py
```diff
@@ -10,33 +12,31 @@
...
+
+    After the second call to self.step() done is true.
```
Markus28, line 17: Add code tag

tests/wrappers/test_autoreset.py
```diff
@@ -10,33 +12,31 @@
...
-        self.observation_space = gym.spaces.Box(
-            low=np.array([-1.0]), high=np.array([1.0])
+            low=np.array([0]), high=np.array([2]), dtype=np.int64
         )
+        self.observation_space = gym.spaces.Discrete(2)
         self.count = 0
 
-    def step(self, action):
+    def step(self, action: int):
+        """Steps the DummyEnv with the incremented step, reward and done if self.count > 1 and updated info."""
```
Markus28, line 33: Code tag

tests/wrappers/test_autoreset.py
```diff
@@ -46,86 +46,86 @@ def reset(
...
 
 @pytest.mark.parametrize("spec", spec_list, ids=[spec.id for spec in spec_list])
 def test_make_autoreset_true(spec):
-    """
-    Note: This test assumes that the outermost wrapper is AutoResetWrapper
-    so if that is being changed in the future, this test will break and need
-    to be updated.
+    """Tests gym.make with autoreset=True, and check that the reset actually happens.
+
+    Note: This test assumes that the outermost wrapper is AutoResetWrapper so if that
```
Markus28, line 68: Use .. note::

tests/wrappers/test_autoreset.py
```diff
@@ -46,86 +46,86 @@ def reset(
...
+     is being changed in the future, this test will break and need to be updated.
     Note: This test assumes that all first-party environments will terminate in a finite
```
Markus28, line 70: Use .. note::

Markus28 (commented):
- I think the attributes metadata, reward_range, spec, action_space, observation_space should be documented in the `gym.Env` top docstring
- We should think about the issue that was raised by Sander regarding tuples for the `gym.Env` `step` method
- According to https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html, properties should only be documented in the getter method. I think you are also documenting  the setters. Maybe pydocstyle complains if you don't do that idk
- Do properties have to be documented as attributes as well in order to be used for autodocs? If that's the case, np_random should definitely documented in the `gym.Env` top docstring.
- https://github.com/openai/gym/blob/d7b39e5e908fa6640132be3355ba879cc03c74b4/gym/core.py#L140 and https://github.com/openai/gym/blob and https://github.com/openai/gym/blob/d7b39e5e908fa6640132be3355ba879cc03c74b4/gym/core.py#L36 and https://github.com/openai/gym/blob/d7b39e5e908fa6640132be3355ba879cc03c74b4/gym/core.py#L174 replace Note by .. note::
- The docstring of `AutoResetWrapper` was missing some inline code tags

Some other remarks that may be outside the scope of this PR:
- For `ActionWrapper`, `ObservationWrapper`, `RewardWrapper`, we could use the documentation from the website (that probably should have been part of my PR but I forgot). In that case it might also make sense to use autodocs for the website. 
- I have noticed that I have been using incorrect codetags for my wrapper PR

gym/error.py
pseudo-rnd-thoughts: Done in eaf3a4c6e2dde0a128761ed50b4d9b884f575d8b

CONTRIBUTING.md
pseudo-rnd-thoughts: As VSCode and the Sphinx google code example only uses `Note:`, Im going to keep it as 'Note:' until we move to the website and it becomes an issue

pseudo-rnd-thoughts:
@Markus28 I have made a lot of changes to core.py adding :meth: and :attr: where appropriate, could you review it again

gym/core.py
```diff
@@ -142,64 +156,66 @@ def render(self, mode="human"):
               the list of supported modes. It's recommended to call super()
```
Markus28, line 156: Wrong indent
image

gym/core.py
```diff
@@ -142,64 +156,66 @@ def render(self, mode="human"):
...
+        Environments will automatically :meth:`close()` themselves when
         garbage collected or when the program exits.
         """
         pass
 
     def seed(self, seed=None):
-        """Sets the seed for this env's random number generator(s).
+        """Deprecated function that sets the seed for the environment's random number generator(s).
+
+        Use :meth:`env.reset(seed=seed)` as the new API for setting the seed of the environment.
```
Markus28, line 187: :meth:`env.reset(seed=seed)` does not show up correctly on the website
image

gym/core.py
```diff
@@ -60,28 +64,32 @@ def np_random(self, value: RandomNumberGenerator):
...
-            reward (float) : amount of reward returned after previous action
-            done (bool): whether the episode has ended, in which case further :meth:`step` calls will return undefined results. A done signal may be emitted for different reasons: Maybe the task underlying the environment was solved successfully, a certain timelimit was exceeded, or the physics simulation has entered an invalid state. ``info`` may contain additional information regarding the reason for a ``done`` signal.
-            info (dict): contains auxiliary diagnostic information (helpful for debugging, learning, and logging). This might, for instance, contain:
-
-                - metrics that describe the agent's performance or
-                - state variables that are hidden from observations or
-                - information that distinguishes truncation and termination or
-                - individual reward terms that are combined to produce the total reward
+            `(observation, reward, done, info)`: A tuple containing the environment's current `observation`,
```
Markus28, line 91: Fix formatting please

image


Also, it seems that a list in Returns actually works fine with the website:
image
So I would change it back

gym/core.py
```diff
@@ -93,28 +101,34 @@ def reset(
...
-
-        This method should also reset the environment's random number
-        generator(s) if ``seed`` is an integer or if the environment has not
-        yet initialized a random number generator. If the environment already
-        has a random number generator and :meth:`reset` is called with ``seed=None``,
-        the RNG should not be reset.
-        Moreover, :meth:`reset` should (in the typical use case) be called with an
+        """Resets the environment to an initial state and returns the initial observation.
+
+        This method can reset the environment's random number generator(s) if :param:`seed` is an integer or
```
Markus28, line 106: :param:`seed` not working
image

gym/core.py
```diff
@@ -93,28 +101,34 @@ def reset(
...
         integer seed right after initialization and then never again.
 
         Args:
-            seed (int or None):
-                The seed that is used to initialize the environment's PRNG. If the environment does not already have a PRNG and ``seed=None`` (the default option) is passed, a seed will be chosen from some source of entropy (e.g. timestamp or /dev/urandom).
-                However, if the environment already has a PRNG and ``seed=None`` is passed, the PRNG will *not* be reset. If you pass an integer, the PRNG will be reset even if it already exists. Usually, you want to pass an integer *right after the environment has been initialized and then never again*. Please refer to the minimal example above to see this paradigm in action.
-            return_info (bool): If true, return additional information along with initial observation. This info should be analogous to the info returned in :meth:`step`
-            options (dict or None): Additional information to specify how the environment is reset (optional, depending on the specific environment)
+            seed (int or None): The seed that is used to initialize the environment's PRNG.
```
Markus28, line 113: Do we use this style or `Optional[int]`?

gym/core.py
```diff
@@ -142,64 +156,66 @@ def render(self, mode="human"):
...
 
-        Environments will automatically close() themselves when
+        Environments will automatically :meth:`close()` themselves when
         garbage collected or when the program exits.
         """
         pass
 
     def seed(self, seed=None):
-        """Sets the seed for this env's random number generator(s).
+        """Deprecated function that sets the seed for the environment's random number generator(s).
```
Markus28, line 185: This docstring explains should document the args imo

gym/core.py
```diff
@@ -142,64 +156,66 @@ def render(self, mode="human"):
...
+            >>> class MyEnv(Env):
+            ...    metadata = {'render_modes': ['human', 'rgb_array']}
+            ...
+            ...    def render(self, mode='human'):
+            ...        if mode == 'rgb_array':
+            ...            return np.array(...) # return RGB frame suitable for video
+            ...        elif mode == 'human':
+            ...            ... # pop up a window and render
+            ...        else:
+            ...            super(MyEnv, self).render(mode=mode) # just raise an exception
```
Markus28, line 169: Why is this in interactive interpreter mode? I see how that makes sense for one-line stuff but you don't define entire functions in interactive mode

gym/core.py
```diff
@@ -142,64 +156,66 @@ def render(self, mode="human"):
...
+        Use :meth:`env.reset(seed=seed)` as the new API for setting the seed of the environment.
 
         Note:
             Some environments use multiple pseudorandom number generators.
             We want to capture all such seeds used in order to ensure that
             there aren't accidental correlations between multiple generators.
 
         Returns:
-            list<bigint>: Returns the list of seeds used in this env's random
+            list<bigint>: Returns the list of seeds used in this environment's random
```
Markus28, line 195: Same issue regarding the style of types. Alternatively `List[int]`

Markus28 (commented):
I think it looks mostly good, apart from a few formatting issues that become apparent on the website. Imo there should be more precise instructions in CONTRIBUTING.md regarding the use of restructured text, documentation of `__init__` kwargs in top-level docstrings, convention for type annotations in docstrings etc.
It seems to me that `Note:` is formatted exactly like `.. note::` in Sphinx, so we can leave it.
Also I did not see the issue that was mentioned by Sander with functions that return tuples.

gym/core.py
pseudo-rnd-thoughts: Personally, I have been making all code examples in an interactive interpreter style. 
Otherwise, Im unsure when to have `>>>` and `...`. How would you do it? Im happy to change it to that

gym/core.py
pseudo-rnd-thoughts: It seems that `:param:` doesn't exist https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#cross-referencing-python-objects

gym/core.py
Markus28: I like to use interactive style if it is a one-liner that produces some output that is of interest. E.g. if we sample from a `Space`: It can be done in a single line and the output is actually important for understanding what the method does. 

For blocks that don't produce any output, I don't see any advantage in using interactive style (except that it magically enables syntax highlighting in PyCharm for some reason). But that's just my taste

Markus28 (approved):
Generally, this looks good to me. One minor comment regarding the notation of the interval (-inf, inf)

gym/core.py
pseudo-rnd-thoughts: Fair, that makes sense. I think that main reason I did this was for pycharm

gym/core.py
```diff
@@ -37,7 +37,7 @@ class Env(Generic[ObsType, ActType]):
     - :attr:`metadata` - The metadata of the environment, i.e. render modes
     - :attr:`np_random` - The random number generator for the environment
 
-    Note: a default reward range set to [-\inf,+\inf] already exists. Set it if you want a narrower range.
+    Note: a default reward range set to :math:`[-\infty,+\infty]` already exists. Set it if you want a narrower range.
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: Thanks, fixed

PR #2790, opened 2022-05-01
Remove PlayPlot from gym.utils.play

gianlucadecola:
As a continuation of https://github.com/openai/gym/pull/2783 I'm removing the PlayPlot class. I will follow with a docs update to keep a reference on how to set up a callback in the play function. 
@jkterry1 any suggestion on where to put it in the docs?

### DIFF
--- a/gym/utils/play.py
+++ b/gym/utils/play.py
@@ -110,18 +110,6 @@ def play(
     verifying that the frame-level preprocessing does not render the game
     unplayable.
 
-    If you wish to plot real time statistics as you play, you can use
-    gym.utils.play.PlayPlot. Here's a sample code for plotting the reward
-    for last 5 second of gameplay.
-
-        def callback(obs_t, obs_tp1, action, rew, done, info):
-            return [rew,]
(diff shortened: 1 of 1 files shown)


pseudo-rnd-thoughts:
Personally, I think it should be a deprecated object in gym, warning that it will be removed at 1.0, with the class then removed at 1.0 to prevent the object just disappearing for some people who use it

gianlucadecola:
By looking at the `__init__.py` of gym.utils there is the following docstring
```python
"""A set of common utilities used within the environments. These are
not intended as API functions, and will not remain stable over time.
"""
``` 
So maybe we can remove it without a warning?

pseudo-rnd-thoughts:
Hilariously that comments has been made 6 years ago and has not been modified since
https://github.com/openai/gym/blame/master/gym/utils/__init__.py

So I think that we shouldn't change it, particularly as a majority of people will not be checking the comment.

gianlucadecola:
Seems legit, I'm closing this PR and making one with a deprecation warning

Issue #2786, opened 2022-04-30
module 'gym.envs.box2d' has no attribute 'CarRacing'


zhouliang-yu:
**Describe the bug**
Hi! 
When I try to run the code 
```python
env_id = "CarRacing-v1"  
env = gym.make(env_id)
```
there is a bug saying that "module 'gym.envs.box2d' has no attribute 'CarRacing'"

but before that, I did install the Box2D by pip install

**System Info**
The code is run on colab
 * support version
python3.7
gym                           0.17.3

- [ ] I have checked that there is a similar [issue](https://github.com/openai/gym/issues/1989) in the repo (**gym'**)
but still cannot solve my problem

pseudo-rnd-thoughts:
Did you run `pip install gym[box2d]`?

edilgin:
I am also getting the same error. Also got an error when `pip install gym[box2d]` it says legacy-install-failure and I couldn't see any solution to that.

zhouliang-yu:
> Did you run `pip install gym[box2d]`?

yes but still not work..

pseudo-rnd-thoughts:
@zhouliang-yu You say you are using 0.17.3, however I can't find a release for that. 

I have created a next python environment and run
`pip install gym[box2d]`
```
import gym

gym.make('CarRacing-v1')
```

@zhouliang-yu and @edilgin Could you run, this should test if the issue is with gym.make or something else
```
import Box2d
import pygame
from gym.envs.box2d import CarRacing
env = CarRacing()
env.reset()
```

zhouliang-yu:
yes

> @zhouliang-yu You say you are using 0.17.3, however I can't find a release for that.
> 
> I have created a next python environment and run with no issue so this is strange `pip install gym[box2d]`
> 
> ```
> import gym
> 
> gym.make('CarRacing-v1')
> ```
> 
> @zhouliang-yu and @edilgin Could you run, this should test if the issue is with gym.make or something else
> 
> ```
> import Box2d
> import pygame
> from gym.envs.box2d import CarRacing
> env = CarRacing()
> env.reset()
> ```

yes problem solved

Issue #2733, opened 2022-04-06
[Bug Report] module 'gym.envs.box2d' has no attribute 'LunarLander'


ahoang512:
If you are submitting a bug report, please fill in the following details and use the tag [bug].

**Describe the bug**
module 'gym.envs.box2d' has no attribute 'LunarLander'

**Code example**
env = gym.make("LunarLander-v2")

**System Info**
Working in a Jupyter notebook on a Mac

**Additional context**

### Checklist

- [ ] I have checked that there is no similar [issue](https://github.com/openai/gym/issues) in the repo (**required**)

ahoang512:
duplicate of https://github.com/openai/gym/issues/1423

pseudo-rnd-thoughts:
That is not a very helpful error message, we should look at improving that

jkterry1:
@pseudo-rnd-thoughts do you want to try creating a PR?

pseudo-rnd-thoughts:
Would it be possible to add this as part of https://github.com/openai/gym/issues/2738?
@RedTachyon would you happy to do this as part of rewriting make? Or should I try and do it as a separate PR?

PR #2791, opened 2022-05-01
Change gym environment spec entry points

pseudo-rnd-thoughts:
Inspired by https://github.com/openai/gym/issues/2786 and https://github.com/openai/gym/issues/2733
When importing from `__init__.py` and the true import statement also fails as in the cases above.
Then the python right is not very help just saying that the import fails

This PR changes the entry points of all gym environments - mujoco to their true environment point
Therefore, if a user has not imported all of the necessary parts for the environment then a more helpful error will occur
Example
```
import gym
gym.make('CarRacing-v1')

# new PR 
gym.error.DependencyNotInstalled: box2D is not installed, run `pip install gym[box2d]`
# old 
ImportError: cannot import name 'CarRacing' from 'gym.envs.box2d' (/gym/gym/envs/box2d/__init__.py)
```

This should be completely backward compatible

### DIFF
--- a/gym/envs/__init__.py
+++ b/gym/envs/__init__.py
@@ -10,41 +10,41 @@ _load_env_plugins()
 
 register(
     id="CartPole-v0",
-    entry_point="gym.envs.classic_control:CartPoleEnv",
+    entry_point="gym.envs.classic_control.cartpole:CartPoleEnv",
     max_episode_steps=200,
     reward_threshold=195.0,
 )
 
 register(
     id="CartPole-v1",
-    entry_point="gym.envs.classic_control:CartPoleEnv",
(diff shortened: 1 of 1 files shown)


RedTachyon:
Do you know what exactly is the reason that the old error looks the way it does? The PR looks, but it'd be good to be certain what is the underlying reason for this behavior so that we know what we're changing.

pseudo-rnd-thoughts:
It is a bit confusing to explain as there are two imports happening.
For example, if `env 1` actually just hides the true import for `envs.env_1` then when the second import happens and fails, import 1 does not return the true import error but just `env 1` could not be imported, `ImportError`

Therefore by moving the entry points to the "true" import location then we shouldn't get this issue

RedTachyon:
Makes sense, can be merged imo

PR #2792, opened 2022-05-01
Add deprecation warning on PlayPlot.

gianlucadecola:
As a continuation of https://github.com/openai/gym/pull/2790
PlayPlot will be removed and it will be now marked as deprecated

### DIFF
--- a/gym/utils/play.py
+++ b/gym/utils/play.py
@@ -6,6 +6,7 @@ from pygame import Surface
 from pygame.event import Event
 
 from gym import Env, logger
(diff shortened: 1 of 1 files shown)


pseudo-rnd-thoughts:
Could you specify at 1.0 instead of future otherwise LGTM

jkterry1:
The specific plans for the naming of 1.0 are rather fluid at the moment, let's go with "near future"

gianlucadecola:
Updated

Issue #2797, opened 2022-05-02
[Bug Report] Misleading type hint for `MultiDiscrete`


Markus28:
Type hints suggest that `MultiDiscrete` spaces must be flat: https://github.com/openai/gym/blob/b4f40f5ab621cafa778af393f540e51588fe2b8d/gym/spaces/multi_discrete.py#L33

That's actually not true, something like

```
MultiDiscrete([[1,2], [3, 4]])
```

is allowed. Whether it is a good idea is another question entirely.

Markus28:
Also, why do we need a dtype?

pseudo-rnd-thoughts:
One of the aims in the future is to have the whole project type checked, at least the critical parts like space. 

From memory, nvec that is passed to a numpy array so you are right, the type is greater than what the docstring actually says.

Markus28:
I will change the type to `Union[np.ndarray, list[int]]`. 

I feel like it is reasonable to require a numpy array if someone wants multiple axes. I have no idea how to encode an arbitrarily nested list as a type hint. Even if it was possible, it wouldn't guarantee flawless conversion to a numpy array.

But I'm still puzzled by the argument `dtype`. Why would we want to allow someone to pass `float` to a `MultiDiscrete` space? (That is actually possible and results in really dumb behavior)

pseudo-rnd-thoughts:
To make an arbitrarily nested list, I suspected we would have to construct a recursive type hint but that doesn't seem good.
For now, I think it is fine
Passing a numpy array is probably the way to get people to use arbitrarily sized arrays so I dont think we should worry about it with the new typing.

For the dtype, I agree, this doesn't seem to make much sense but someone may care about using `uint8`, `int16`, `int32` or `int64` plus for backward compatibility then I think we have to keep the parameter
But we could add a check that the dtype passed supports ints and raise an error or warning if not??

PR #2798, opened 2022-05-02
Pydocstyle spaces docstring

Markus28:
Follow-up to https://github.com/openai/gym/pull/2785

Reformatting of docstrings to comply with google style. Some docstrings were added, rewritten or expanded.

Will resolve https://github.com/openai/gym/issues/2797

### DIFF
--- a/gym/spaces/__init__.py
+++ b/gym/spaces/__init__.py
@@ -3,9 +3,8 @@
 Spaces describe mathematical sets and are used in Gym to specify valid actions and observations.
 Every Gym environment must have the attributes ``action_space`` and ``observation_space``.
 If, for instance, three possible actions (0,1,2) can be performed in your environment and observations
-are vectors in the two-dimensional unit cube, the environment code may contain the following two lines:
+are vectors in the two-dimensional unit cube, the environment code may contain the following two lines::
 
-::
     self.action_space = spaces.Discrete(3)
     self.observation_space = spaces.Box(0, 1, shape=(2,))
 """
--- a/gym/spaces/__init__.py
+++ b/gym/spaces/__init__.py
@@ -1,3 +1,14 @@
+"""This module implements various spaces.
+
+Spaces describe mathematical sets and are used in Gym to specify valid actions and observations.
+Every Gym environment must have the attributes `action_space` and `observation_space`.
+If, for instance, three possible actions (0,1,2) can be performed in your environment and observations
+are vectors in the two-dimensional unit cube, the environment code may contain the following two lines:
+
+::
+    self.action_space = spaces.Discrete(3)
+    self.observation_space = spaces.Box(0, 1, shape=(2,))
+"""
 from gym.spaces.box import Box
 from gym.spaces.dict import Dict
 from gym.spaces.discrete import Discrete
--- a/gym/spaces/__init__.py
+++ b/gym/spaces/__init__.py
@@ -1,7 +1,7 @@
 """This module implements various spaces.
 
 Spaces describe mathematical sets and are used in Gym to specify valid actions and observations.
-Every Gym environment must have the attributes `action_space` and `observation_space`.
+Every Gym environment must have the attributes ``action_space`` and ``observation_space``.
 If, for instance, three possible actions (0,1,2) can be performed in your environment and observations
 are vectors in the two-dimensional unit cube, the environment code may contain the following two lines:
 
--- a/gym/spaces/dict.py
+++ b/gym/spaces/dict.py
@@ -41,7 +41,7 @@ class Dict(Space[TypingDict[str, Space]], Mapping):
         ...             }
         ...         ),
         ...     }
-        ...)
+        ... )
 
     It can be convenient to use ``Dict`` spaces if you want to make complex observations or actions more human-readable.
     Usually, it will be not be possible to use elements of this space directly in learning code. However, you can easily
--- a/gym/spaces/dict.py
+++ b/gym/spaces/dict.py
@@ -13,9 +13,9 @@ from gym.utils import seeding
 
 
 class Dict(Space[TypingDict[str, Space]], Mapping):
-    """A dictionary of simpler spaces.
+    """A dictionary of :class:`Space` instances.
 
-    Elements of this space are (ordered) dictionaries of elements from the simpler spaces.
+    Elements of this space are (ordered) dictionaries of elements from the constituent spaces.
 
     Example usage::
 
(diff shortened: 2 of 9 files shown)


gym/spaces/__init__.py
```diff
@@ -1,3 +1,14 @@
+"""This module implements various spaces.
+
+Spaces describe mathematical sets and are used in Gym to specify valid actions and observations.
+Every Gym environment must have the attributes ``action_space`` and ``observation_space``.
+If, for instance, three possible actions (0,1,2) can be performed in your environment and observations
+are vectors in the two-dimensional unit cube, the environment code may contain the following two lines:
+
+::
+    self.action_space = spaces.Discrete(3)
```
pseudo-rnd-thoughts, line 9: For code, I think you need to use `>>>` or `...` before the line

gym/spaces/__init__.py
```diff
@@ -1,3 +1,14 @@
+"""This module implements various spaces.
+
+Spaces describe mathematical sets and are used in Gym to specify valid actions and observations.
+Every Gym environment must have the attributes ``action_space`` and ``observation_space``.
+If, for instance, three possible actions (0,1,2) can be performed in your environment and observations
+are vectors in the two-dimensional unit cube, the environment code may contain the following two lines:
+
+::
```
pseudo-rnd-thoughts, line 8: I think this should be `Example::`

gym/spaces/dict.py
```diff
@@ -49,6 +55,23 @@ def __init__(
         seed: Optional[dict | int | seeding.RandomNumberGenerator] = None,
         **spaces_kwargs: Space,
     ):
+        """Constructor of `Dict` space.
+
+        This space can be instantiated in one of two ways: Either you pass a dictionary
+        of spaces to `__init__` via the ``spaces`` argument, or you pass the spaces as separate
+        keyword arguments (where you will need to avoid the keys ``spaces`` and ``seed``)
+
+        Example::
```
pseudo-rnd-thoughts, line 64: I believe a line is required between example and the code block for the docstring to format correctly

gym/spaces/__init__.py
Markus28: Hm PyCharm does indeed not do syntax highlighting without those. However, I was under the impression that any indented text after :: will be considered code. I don't want to use `>>>` and `...` here because that bit of code doesn't make much sense in an interactive interpreter. Maybe we can wait to see how it plays with Sphinx and fix it in a second pass if necessary.

gym/spaces/__init__.py
Markus28: I actually wanted to do `... the following two lines::`. I think that should work as well and I fixed it.

gym/spaces/discrete.py
```diff
@@ -27,16 +26,30 @@ def __init__(
         seed: Optional[int | seeding.RandomNumberGenerator] = None,
         start: int = 0,
     ):
+        r"""Constructor of ``Discrete`` space.
+
+        This will construct the space :math:`\\{\text{start}, ..., \text{start} + n - 1\\}`.
+
+        Args:
+            n (int): The number of elements of this space
+            start (int): The smallest element of this space.
```
pseudo-rnd-thoughts, line 35: Args are not in the same order as parameters

gym/spaces/multi_discrete.py
```diff
@@ -30,12 +31,29 @@ class MultiDiscrete(Space[np.ndarray]):
...
+
+            >> d = MultiDiscrete(np.array([[1, 2], [3, 4]]))
+            >> d.sample()
+            array([[0, 0],
+                   [2, 3]])
+
+        Args:
+            nvec: vector of counts of each categorical variable. This will usually be a list of integers. However,
+                you may also pass a more complicated numpy array if you'd like the space to have several axes.
+            dtype: TODO
```
pseudo-rnd-thoughts, line 55: TODO to do

gym/spaces/multi_discrete.py
```diff
@@ -44,29 +62,35 @@ def __init__(
 
     @property
     def shape(self) -> tuple[int, ...]:
-        """Has stricter type than gym.Space - never None."""
+        """Has stricter type than ``gym.Space` - never None."""
```
pseudo-rnd-thoughts, line 65: You need an extra backtick

gym/spaces/space.py
```diff
@@ -10,18 +11,27 @@
...
-
-    WARNING - Custom observation & action spaces can inherit from the `Space`
-    class. However, most use-cases should be covered by the existing space
-    classes (e.g. `Box`, `Discrete`, etc...), and container classes (`Tuple` &
-    `Dict`). Note that parametrized probability distributions (through the
-    `sample()` method), and batching functions (in `gym.vector.VectorEnv`), are
-    only well-defined for instances of spaces provided in gym by default.
-    Moreover, some implementations of Reinforcement Learning algorithms might
-    not handle custom spaces properly. Use custom spaces with care.
+    """Any space inherits from this class.
```
pseudo-rnd-thoughts, line 14: Personally, I think the original first line was a more helpful description of what the space class does

gym/spaces/space.py
```diff
@@ -10,18 +11,27 @@
...
+
+    Spaces are crucially used in Gym to define the format of valid actions and observations.
+    They serve various purposes:
+
+    * They clearly define how to interact with environments, i.e. they specify what actions need to look like and what observations will look like
```
pseudo-rnd-thoughts, line 19: I believe to do bullet points in sphinx you need `-` not `*`

gym/spaces/tuple.py
```diff
@@ -9,19 +10,30 @@
 
 
 class Tuple(Space[tuple], Sequence):
-    """
-    A tuple (i.e., product) of simpler spaces
+    """A tuple (more precisely the cartesian product) of simpler spaces.
```
pseudo-rnd-thoughts, line 13: "simpler spaces" is not necessarily true, as it is possible to have a Tuple of Tuple of Discrete Spaces

gym/spaces/space.py
Markus28: Hmm I agree that it is much more concrete. I had some concerns I felt like the first sentence should point out that `Space` is essentially just an abstract superclass. I replaced it with 

> Superclass that is used to define observation and action spaces.

gym/spaces/tuple.py
Markus28: I agree and this term also shows up in older docstrings. What about "constituent spaces"?

pseudo-rnd-thoughts: That would work or `Subspaces` as Tuples variable is called `subspaces`

pseudo-rnd-thoughts:
Looks good to merge

PR #2800, opened 2022-05-04
BugFix - VecEnvWrapper doesn't forward call, set_attr to self.env

arjun-kg:
# Description 

1. `VecEnvWrapper` only forwards some select methods to self.env. I added those methods which cause issues now with the current implementations of `SyncVectorEnv` and `AsyncVectorEnv`. (example code below). If for eg. `reset` is implemented in a new vector env, this would later need to also be added to `VecEnvWrapper`. 
2. Added a corresponding test. 

```python
import gym 
from gym.vector import VectorEnvWrapper

class DummyVecEnvWrapper(VectorEnvWrapper):
    def __init__(self, env):
        super().__init__(env)

if __name__=='__main__':
    env = gym.vector.make("CartPole-v1", num_envs=3)
    wrapped = DummyVecEnvWrapper(gym.vector.make("CartPole-v1", num_envs=3))

    # works fine
    env.call("gravity")
    env.set_attr("gravity", [20, 20, 20])

    # Error - goes to VectorEnv where these methods raise not implemented errors
    wrapped.call("gravity")
    wrapped.set_attr("gravity", [20, 20, 20])

```

## Type of change

# Checklist:

### DIFF
--- a/gym/vector/vector_env.py
+++ b/gym/vector/vector_env.py
@@ -262,6 +262,12 @@ class VectorEnvWrapper(VectorEnv):
     def seed(self, seed=None):
         return self.env.seed(seed)
 
+    def call(self, name, *args, **kwargs):
+        return self.env.call(name, *args, **kwargs)
+
+    def set_attr(self, name, values):
+        return self.env.set_attr(name, values)
+
     # implicitly forward all other methods and attributes to self.env
     def __getattr__(self, name):
         if name.startswith("_"):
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts:
@arjun-kg This looks good to me, could you add a test for `call`

arjun-kg:
@pseudo-rnd-thoughts a test is already added - for `set_attr`, `call` and `get_attr`

pseudo-rnd-thoughts:
Ohh ignore me, I didn't see it. Could you add docstrings to the Vector Wrapper and tests for their purpose

pseudo-rnd-thoughts:
LGTM

Issue #2767, opened 2022-04-20
[Proposal] Tests for all envs


jjshoots:
### Proposal 

There should be more comprehensive tests for all envs. Probably the easiest test would be similar to that in PettingZoo where every state of a rollout is asserted to be contained in the observation space.

There should also be tests that assert the behaviour of environments where this can be tested. One example is the heuristic landing for Lunar Lander (which currently only works for the non-winded case).

pseudo-rnd-thoughts:
## Proposed changes 
* Update the`spec_list` to remove atari and provide a full spec list and a no-mujoco spec list
* Remove all atari tests - Currently atari is not being installed so there is no point in the tests being there. https://github.com/openai/gym/issues/2734#issuecomment-1109729283
* Upgrade the action dim test - Check that environments do not allow out of bounds actions, i.e. extra dims, outside observation space
* Upgrade the environment determinism - Improve the number of time steps that we environment step equivalence. This will also test the environments can step / rollout
* Simplify the observation shape check - This could be part of the determinism check to help improve performance and reduce test time
* Test render - Current that all of the render modes for environments return the expected output
* Test kwargs for classic control, box2d, and toy text, some of these environments have kwargs, hardcore, randomize, etc. Currently there are no tests that the kwargs effect the environments "correctly" however this is difficult to test. There is a frozenlake DFS map generation test already for this
*  Mujoco - There is a mujoco v2 to v3 tests however it is unclear if this is unneeded anymore with v4 coming out. Possibly add new v4 tests
* Registration - These tests are actually in a good condition, I believe so these will be kept

PR #2802, opened 2022-05-05
Upgrade test_action_dim. Refer to  #2767

gianlucadecola:
For one of the proposals of  #2767
This PR aim at obtaining more comprehensive tests to ensure the correct action is performed in the environment.
While adding tests I found that correct action check is performed in different ways across environments; sometimes a check is done before step computations, other times, `step` is left to fail due to subsequent invalid operations caused by a wrong action.

I've tried to add more consistency across the action validation.

For Discrete environments I'm raising an exception, however, continuous action spaces are typically allowed to take out of bound action and perform action clipping inside (also wrong numpy shape input are accepted). Since I didn't want to break backward compatibility I've decided to just raise a warning in these situations.


tests/envs/test_action_dim_check.py
```diff
@@ -17,3 +19,63 @@ def test_serialize_deserialize(environment_id):
 
     with pytest.raises(ValueError, match="Action dimension mismatch"):
         env.step(0.1)
+
+
+@pytest.mark.parametrize(
+    "environment_id",
```
pseudo-rnd-thoughts, line 25: Could you use the `spec_list` and filter for all environments that have a discrete action space?

tests/envs/test_action_dim_check.py
```diff
@@ -17,3 +19,63 @@ def test_serialize_deserialize(environment_id):
...
+    (
+        "MountainCarContinuous-v0",
+        "LunarLanderContinuous-v2",
+        "CarRacing-v1",
+        "BipedalWalker-v3",
+        "Pendulum-v1",
+    ),
+)
+def test_box_actions_out_of_bound(environment_id):
```
pseudo-rnd-thoughts, line 59: A repeat of before with continuous action spaces

pseudo-rnd-thoughts (commented):
So my idea behind this PR was more on the tests side to check that all environments would work as intended but the idea of validation environments inputs is interesting. 
I guess this is a more generalised version of environments checking their input instead of each environment checking it individually. 

If this is the approach we want to take, then could we generalise the function to work with every known Space? The contains function is believe is implemented by every Space so it shouldn't be too bad.

gianlucadecola:
Maybe we should defer this to another PR since it is also a design change and leave here only the tests;
the drawback will be that for discrete actions, instead of checking for a specific Exception, if an out-of-bound action is performed I need to check for a generic Exception since it is not consistent across all envs.

For Box actions spaces I think I could instead check that the output of an "out-of-bound" is clipped to the bound of the action_space.
What do you think?


> I guess this is a more generalised version of environments checking their input instead of each environment checking it individually.

Yes, while looking at the code when writing tests I thought that more consistency across error handling for out-of-action-space action would be nice

pseudo-rnd-thoughts:
@RedTachyon What are your thoughts on this PR? 
I think that it is reasonable to raise an error if an action outside of the action space is used with `env.step` for all space type. However, with the other breaking changes that we are making, I don't want to make too many that are unnecessary. This should help spot errors that users don't check for so Im unsure on the balance what to do.

RedTachyon:
This seems to be like 3 PRs combined into one, no? I see (1) automatic action validation for built-in envs (which is something to consider, but far from an obvious feature for me personally), (2) modifications to car racing (which at a glance seem reasonable), and (3) actual tests?

I'd very strongly prefer those to be split up, with a space to discuss each of them separately. The order of priority imo would be (3) [tests for current code] -> (2) [refactoring without changing functionality]  -> (1) [new functionality/deprecating certain functionalities]


As a quick counterpoint about (1), it often might be good enough to just clip the actions to the correct range, which I think already happens in some environments. I don't know if that's a definitive argument against this change, but it's *an* argument, and something like that needs a stand-alone discussion.

gianlucadecola:
> This seems to be like 3 PRs combined into one, no? I see (1) automatic action validation for built-in envs (which is something to consider, but far from an obvious feature for me personally), (2) modifications to car racing (which at a glance seem reasonable), and (3) actual tests?

Definitely right, gone a little too far with refactoring

> I'd very strongly prefer those to be split up, with a space to discuss each of them separately. The order of priority imo would be (3) [tests for current code] -> (2) [refactoring without changing functionality] -> (1) [new functionality/deprecating certain functionalities]
> 
> As a quick counterpoint about (1), it often might be good enough to just clip the actions to the correct range, which I think already happens in some environments. I don't know if that's a definitive argument against this change, but it's _an_ argument, and something like that needs a stand-alone discussion.

I'm going to revert back those changes and leave this PR only for actual tests :+1:

gianlucadecola:
It should be ok now :)

tests/envs/test_action_dim_check.py
```diff
@@ -17,3 +33,59 @@ def test_serialize_deserialize(environment_id):
 
     with pytest.raises(ValueError, match="Action dimension mismatch"):
         env.step(0.1)
+
+
+@pytest.mark.parametrize("env", make_envs_by_action_space_type(spec_list, Discrete))
+def test_discrete_actions_out_of_bound(env):
+    # Car racing currently allow `out-of-bound` actions
```
pseudo-rnd-thoughts, line 40: Why do we allow out of bounds actions for car racing?

tests/envs/test_action_dim_check.py
```diff
@@ -17,3 +33,59 @@ def test_serialize_deserialize(environment_id):
...
+    for i, (is_upper_bound, is_lower_bound) in enumerate(
+        zip(env.action_space.bounded_above, env.action_space.bounded_below)
+    ):
+        if is_upper_bound:
+            obs, _, _, _ = env.step(upper_bounds)
+            upper_bounds[i] += np.cast[dtype](OOB_VALUE)
+            oob_obs, _, _, _ = oob_env.step(upper_bounds)
+
+            assert np.alltrue(obs == oob_obs)
+            return
```
pseudo-rnd-thoughts, line 83: If the assert is successful why are we terminating the test here? Should we not test the lower bound as well, if the agent has it?

tests/envs/test_action_dim_check.py
gianlucadecola: It seems to me that it checks only wether the action matches `steer`, `gas`, `brake` so if an out-of-bound action is performed the state of the car does not change.
https://github.com/openai/gym/blob/31e6f23e67a670e7276582710aa92534790bf92d/gym/envs/box2d/car_racing.py#L445-L454  
I think @jjshoots could give some hints, maybe I'm missing some behavior

tests/envs/test_action_dim_check.py
gianlucadecola: It will indeed be a more comprehensive test, I will add it

tests/envs/test_action_dim_check.py
```diff
@@ -17,3 +33,59 @@ def test_serialize_deserialize(environment_id):
...
+    upper_bounds = env.action_space.high
+    lower_bounds = env.action_space.low
+
+    for i, (is_upper_bound, is_lower_bound) in enumerate(
+        zip(env.action_space.bounded_above, env.action_space.bounded_below)
+    ):
+        if is_upper_bound:
+            obs, _, _, _ = env.step(upper_bounds)
+            oob_action = upper_bounds.copy()
+            oob_action[i] += np.cast[dtype](OOB_VALUE)
```
pseudo-rnd-thoughts, line 80: could you assert that the OOB value is actually out of bounds for the environment first

gianlucadecola: Good, actually find a bug while adding this check :+1:

tests/envs/test_action_dim_check.py
```diff
@@ -17,3 +33,59 @@ def test_serialize_deserialize(environment_id):
...
+    # Car racing currently allow `out-of-bound` actions
+    if "CarRacing" in env.spec.id:
+        return
+
+    env.reset()
+
+    action_space = env.action_space
+    upper_bound = action_space.start + action_space.n - 1
+
+    with pytest.raises(Exception):
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: Yes, you are right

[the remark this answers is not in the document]
gianlucadecola: Some have this checks:
https://github.com/openai/gym/blob/745e7059e7754e26cfcaa82d0d9a0defec3a08c4/gym/envs/box2d/lunar_lander.py#L425-L427
Some let the step fail with different errors (Acrobot will fail here with an IndexError):
https://github.com/openai/gym/blob/745e7059e7754e26cfcaa82d0d9a0defec3a08c4/gym/envs/classic_control/acrobot.py#L190
Taxy will for example fail here with a KerError:
https://github.com/openai/gym/blob/745e7059e7754e26cfcaa82d0d9a0defec3a08c4/gym/envs/toy_text/taxi.py#L210-L211
I think it is reasonable to add more consistency across error handling but I think it may be another PR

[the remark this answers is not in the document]
pseudo-rnd-thoughts: Ok, lets do that as part of another PR

PR #2805, opened 2022-05-06
Render order enforcing

pseudo-rnd-thoughts:
Add render order enforcement as most gym environments cannot be visualised before render
Inspired by https://github.com/openai/gym/issues/2751

As some users may want this as a feature of their environments, I have added a setting where if the users add `disable_render_order_enforcing` to the base environment then this will not be checked

Added tests to verify these features are working
Deleted old test that did not seem to actually test anything that `OrderEnforcing` actually implemented. 

This has also changed the error from `AssertationError` to `ResetNeeded` error


gym/wrappers/order_enforcing.py
```diff
@@ -9,10 +9,22 @@ def __init__(self, env):
...
-        return observation, reward, done, info
+        assert self._has_reset, "Cannot call env.step() before calling env.reset()"
+        return self.env.step(action)
 
     def reset(self, **kwargs):
         self._has_reset = True
         return self.env.reset(**kwargs)
+
+    def render(self, **kwargs):
+        if hasattr(self.unwrapped, "disable_render_order_enforcing"):
```
RedTachyon, line 20: Do we need this? That seems to be a completely new pattern, and not a very useful one in my opinion. Can't we just make render order enforcing an automatic feature of `OrderEnforcing`, and if someone wants to disable it, they can disable the wrapper

gym/wrappers/pixel_observation.py
```diff
@@ -80,6 +80,7 @@ def __init__(
 
         # Extend observation space with pixels.
 
+        self.env.reset()
```
RedTachyon, line 83: I guess this is necessary to get the shape of the image? Is there any reasonable way to skip the reset? (my intuition is that there isn't, it's fine if that's just the answer)

tests/wrappers/utils.py
```diff
@@ -0,0 +1,11 @@
+from typing import Union
+
+import gym
+
+
+def has_wrapper(wrapped_env: Union[gym.Wrapper, gym.Env], wrapper_type: type):
```
RedTachyon, line 6: The Union is redundant, `gym.Wrapper` subclasses `gym.Env`

RedTachyon: And a return type hint would be nice just out of principle

RedTachyon (commented):
Looks mostly good, just not sure about the approach to disabling render order enforcing, imo it can either be just integrated with regular order enforcing, or included as some setting in that wrapper specifically.

pseudo-rnd-thoughts:
I prefer your solution of using a render parameter will modify it to that

gym/wrappers/pixel_observation.py
pseudo-rnd-thoughts: The assumption is that the environment is generated using `gym.make` therefore, we will need to reset before rendering. Also this was failing the tests without it.

tests/wrappers/utils.py
pseudo-rnd-thoughts: Fixed in 02c7957d4b2851c99fb06c7afbfee055eccdf09c

gym/wrappers/order_enforcing.py
pseudo-rnd-thoughts: Changed to a wrapper parameter in 6c19af22551ff187053205cd7776dea6b8eaca29

gym/utils/env_checker.py
```diff
@@ -384,11 +387,21 @@ def check_env(env: gym.Env, warn: bool = True, skip_render_check: bool = True) -
...
+        except ResetNeeded as e:
+            assert str(e) == "Cannot call env.step() before calling env.reset()"
+        try:
+            env.render()
+        except ResetNeeded as e:
+            assert (
+                str(e) == "Cannot call env.render() before calling env.reset(), "
+                "if this is a intended action, set `disable_render_order_enforcing` "
+                "on the OrderEnforcer wrapper = False."
+            )
```
Markus28, line 404: Do we assume that the environment instance is completely fresh when it is passed to this method? In that case we would also want to assert that an exception is actually thrown. That is not being done atm.

pseudo-rnd-thoughts: That is a good point, I will change to `pytest.raises`

Markus28: Yeah I agree. That should be done in a unit test of the OrderEnforcing wrapper

pseudo-rnd-thoughts: Cool, I have removed it from the env checking

tests/wrappers/test_order_enforcing.py
```diff
@@ -2,39 +2,42 @@
...
 from gym.envs.classic_control import CartPoleEnv
+from gym.error import ResetNeeded
 from gym.wrappers import OrderEnforcing
 from tests.envs.spec_list import spec_list
 from tests.wrappers.utils import has_wrapper
 
 
 @pytest.mark.parametrize("spec", spec_list, ids=[spec.id for spec in spec_list])
 def test_gym_make_order_enforcing(spec):
+    """checks that gym.make wrappers the environment with the OrderEnforcing wrapper."""
```
RedTachyon, line 13: (style) Checks that [...]

gym/wrappers/order_enforcing.py
```diff
@@ -34,15 +42,10 @@ def reset(self, **kwargs):
...
-            if not self.unwrapped.disable_render_order_enforcing:
-                assert (
-                    self._has_reset
-                ), "Cannot call env.render() before calling env.reset()"
-        else:
-            assert self._has_reset, (
-                "Cannot call env.render() before calling env.reset(), if this is a intended property, "
-                "set `disable_render_order_enforcing=True` on the base environment (env.unwrapped)."
+        """Renders the environment with `kwargs`."""
+        if self._disable_render_order_enforcing is False and not self._has_reset:
```
RedTachyon, line 46: `if not self._disable_render_order_enforcing and not self._has_reset`
Or even
`if not (self._disable... or self._has_reset)`

RedTachyon (approved):
Two stylistic nitpicks, otherwise good to go

pseudo-rnd-thoughts:
Thanks for fixed

PR #2807, opened 2022-05-07
Add a minimal env checker on `gym.make`

RedTachyon:
The long awaited env checker that will automatically run on `gym.make` has arrived and is... pretty underwhelming.

Basically all it does is check whether there is an observation/action space, and whether the env accepts all the new `env.reset` arguments.

What I strongly do not want to do here is actually call any methods like `reset` or `step` which would be necessary for a thorough API verification. This would have unintended consequences while creating the env, and that's a bad thing. It should be 100% unintrusive.

For this reason, I can't add e.g. a verification of the updated `step` signature because of Python's insane type system (or lack thereof). Afaik there is no way of checking the return type of a function without actually running it (interpreting the code in our brains counts as running it).

I can add whatever other checks someone can come up with, but options are rather limited.

One thing I'm not sure about is whether a failed check should result in an exception or a warning - it can be explicitly disabled anyways, and if someone insists on using an incorrectly implemented env, they'd definitely disable it even if it's just a warning.


tests/test_core.py
```diff
@@ -19,8 +21,16 @@ class UnittestEnv(core.Env):
...
     action_space = spaces.Discrete(3)
 
-    def reset(self, *, seed: Optional[int] = None, options: Optional[dict] = None):
+    def reset(
+        self,
+        *,
+        seed: Optional[int] = None,
+        return_info: bool = False,
+        options: Optional[dict] = None
+    ):
```
pseudo-rnd-thoughts, line 30: Return type?

pseudo-rnd-thoughts:
It looks good to me
Could you provide an example where calling reset or step is bad? I recognise that this could slow down several environment with initialisation but `.make` is not called particularly regularly and needs to be fast

tests/test_core.py
RedTachyon: What do you mean?

XuehaiPan: Missing type hint for the return values? (`Union[ObsType, Tuple[ObsType, dict]]`)

RedTachyon: That's automatically inherited from `gym.Env`, and individual envs don't currently have return annotations, so I don't think it's necessary

RedTachyon:
I think the simple example will be in case you want to call `reset` with some specific arguments for the first time, which might happen to be impactful, e.g. cause some irreversible changes to the environment. 

I have an almost-example of this in a Unity env I'm working with, where on environment creation I can do some extra setup, which has to be done **before** the first reset. The moment the environment is reset, this setup is set in stone and cannot be modified any further, and attempting to do so would result in a crash or a lack of functionality.

In my actual code I mostly circumvent this issue by doing all of that setup and the first reset in the initializer.

It could also cause subtle bugs and annoyances for replicability - by resetting the env, you're stepping the RNG, which is not what was intended by the env.

pseudo-rnd-thoughts:
Fair, that sounds very reasonable. An alternative solution, could we add a wrapper that does nothing just checks what is being passed through, allowing us to raise an issue if different from expected. This could run on only the first time the `reset` or `step` function is called

XuehaiPan:
> An alternative solution, could we add a wrapper that does nothing just checks what is being passed through, allowing us to raise an issue if different from expected.

We can put these checks in `OrderEnforcing`.

pseudo-rnd-thoughts:
Personally, I disagree as the purpose of the wrapper would be completely different from the `OrderEnforcing` wrapper. 
This could be a separate wrapper that is automatically applied through `gym.make` called `EnvCheckerWrapper`

pseudo-rnd-thoughts:
Looks good to merge

gym/utils/env_checker.py
```diff
@@ -426,3 +426,42 @@ def check_env(env: gym.Env, warn: bool = True, skip_render_check: bool = True) -
...
+        "return_info" in signature.parameters or "kwargs" in signature.parameters
+    ), "The `reset` method does not provide the `return_info` keyword argument. You can disable this message by calling `gym.make(..., check_env=False)`"
+
+    assert (
+        "seed" in signature.parameters or "kwargs" in signature.parameters
+    ), "The environment cannot be reset with a random seed. This behavior will be deprecated in the future. You can disable this message by calling `gym.make(..., check_env=False)`"
+
+    assert (
+        "return_info" in signature.parameters or "kwargs" in signature.parameters
+    ), "The `reset` method does not provide the `return_info` keyword argument. You can disable this message by calling `gym.make(..., check_env=False)`"
```
Markus28, line 446: I think this is identical to lines 436-438

Markus28 (commented):
Maybe we should also check for
- metadata
- The options keyword argument in reset

RedTachyon:
This has been updated to use the generic env checker instead after some discussions with Jordan

pseudo-rnd-thoughts:
@RedTachyon This looks good, this PR hasn't made any modifications to the `env_checker` function, are there any issues with it before merging?

gym/envs/registration.py
```diff
@@ -516,6 +518,14 @@ def make(
     if autoreset:
         env = AutoResetWrapper(env)
 
+    if not disable_env_checker:
```
araffin, line 521: what is the reason for having that as opt-out instead of opt-in?

user will see a double reset from his/her side and step into the env, which might be quite confusing (for instance using real robot) and/or costly.

RedTachyon: FWIW I initially wanted to stick to a non-invasive checker that only looks at function signatures, but the usefulness of that was pretty limited. 

The main purpose of having it opt-out is sort of educational - to let people know if the environment isn't correctly implemented, basically saying "Hey, it looks like it works alright, but here's a few missing elements, so you might want to fix it". If it were opt-in, then it would be largely useless, since you might as well run the checker manually

araffin: btw, the PR still mentions "What I strongly do not want to do here is actually call any methods like reset or step which would be necessary for a thorough API verification. This would have unintended consequences while creating the env, and that's a bad thing. It should be 100% unintrusive."

but you went the other way at the end?

RedTachyon: > but you went the other way at the end?

I didn't retroactively change the original PR message, so essentially that was just my thought process at the time. Essentially I had (and still have) the same points you're stating here.

In the meantime, there were some "offline" discussions where the conclusion was to use the full env checker and make it opt-out (based on what I described earlier). Which for me is a bit "**ugh**, fine", as in I didn't veto it, but I don't love it.

araffin: as a compromise would it be possible to let the user define a global parameter/env variable that would disable that new feature?
something like `gym.disable_auto_check()` or `os.environ["DISABLE_AUTO_CHECK"] = 1`

because with gym 0.24, I now have to patch all pieces of code to `gym.make(..., disable_env_checker=False)` to ensure old behavior (including at test time).

btw, this behavior will probably break video recording too...

jkterry1: So the reason we added this as default on was ultimately that even before the API changes, virtually no environments code complied with the Gym API, which is bad for standardization, the field, and so on. The intention of this change was to make this known is a very obvious manner so that things are actually fixed.

You're correct that for environments where resetting is undesirable they'll have to add a flag, but this is a minority of the plethora of environments. As far as a global parameter/variable or breaking video recording, let me get back to you

RedTachyon: I like the idea of a global switch, either environment variable or a global setting in gym, I'll have to think a bit which one would be better.

PR #2808, opened 2022-05-08
[Proposal] Add turbulance disturbance to LunarLander

virgilt:
# Description

Added a torque function for added turbulence to the lander. The torque follows the same magnitude as the linear wind function.

Fixes # (2799)

## Type of change

# Checklist:


virgilt:
There is a mistake in this pull request that I need to fix.

agentydragon:
For this kind of change, you should increase the version number of the environment. And probably would also be better to make the turbulence optional.

pseudo-rnd-thoughts:
This functionality I believe is turned off by default so it shouldn't affect any previous code but you are probably right that increasing the version number would be helpful to log the change in environment functionality.

virgilt:
@jjshoots I ran the provided tests and the lander is working. Also reran `pre-commit run --all-files`.

virgilt:
Here is the test that I ran:
```

import gym
env = gym.make("LunarLander-v2", enable_wind=True)
observation, info = env.reset(seed=42, return_info=True)
for _ in range(1000):
   env.render()

   if done:
      observation, info = env.reset(return_info=True)
env.close()
```

jjshoots:
> This functionality I believe is turned off by default so it shouldn't affect any previous code but you are probably right that increasing the version number would be helpful to log the change in environment functionality.

This is just my opinion but I don't think incrementing version numbers is necessary here right? If it doesn't change/break the functionality of the current code and still works with everyone else's use of LunarLander-v2, incrementing version numbers doesn't really do anything for the project imho (and may cause more people's implementations to break because of version number changes.

pseudo-rnd-thoughts:
@virgilt At the very least, could you update the version history in the documentation with the change in a similar way to car racing

pseudo-rnd-thoughts:
@virgilt a gif would be great just to see what environment looks like with these changes

pseudo-rnd-thoughts:
Sorry, in addition, could you add a test that this feature works. @jjshoots any good ideas of testing functionality other than, with different wind powers, the observations are different. 
@virgilt Could this push the robot out of the observation space? If so, could you clip the wind to the observation space or abapt the observation space to the wind power.

jjshoots:
@pseudo-rnd-thoughts The observations are actually the same with and without wind, since wind is an unobservable thing in the env (making the env a POMDP instead of just an MDP as it was previously).

The wind can push the lander outside of the space, yes, but that also means the environment ends. Right now we're testing to make sure that the wind doesn't result in an unlearnable env because of the wind being more powerful than what the motors can meaningfully reject.

virgilt:
@pseudo-rnd-thoughts I added version history within lunar_lander.py ...Do I still need to add a gif, or is that no longer needed because of @jjshoots explanation?

@jjshoots I added torque_power to kwargs and an assert.

pseudo-rnd-thoughts:
Ok, we cannot see this effect then don't worry about the gif @virgilt.

virgilt:
@jjshoots what do you think about removing the upper bound on wind_power and torque_power, and just putting a recommended range within the docs?

jjshoots:
@virgilt This is fair too, but I think @pseudo-rnd-thoughts is a better candidate to handle this question.

The context is pretty much that wind and turbulence need to be within a particular range for the environment to be solvable. Right now it's handled with an assert that keeps the value in that range, but Travis is suggesting that we remove the assert and put the recommended range in the docs.

jjshoots:
Here's a gif of a somewhat trained agent in the environment with wind enabled. In some instances the lander actually crashes, but I think this is something that can be improved with more learning/tweaking hyperparams.

Peek 2022-05-13 21-10

jjshoots:
@virgilt can you change `torque_power` to perhaps `turbulence_power`, this makes things slightly more descriptive. Other than that, LGTM!

virgilt:
@jjshoots I change the variable name in all places.

virgilt:
> @virgilt This is fair too, but I think @pseudo-rnd-thoughts is a better candidate to handle this question.
> 
> The context is pretty much that wind and turbulence need to be within a particular range for the environment to be solvable. Right now it's handled with an assert that keeps the value in that range, but Travis is suggesting that we remove the assert and put the recommended range in the docs.

@jjshoots, these changes were suggested by @pseudo-rnd-thoughts. We wanted your opinion on it.

jjshoots:
If that's the case, then I'm voting for removing the asserts, changing them to warnings as well as putting it in the docs :)

pseudo-rnd-thoughts:
In addition, could you update the environment docstring
Currently, it is "`wind_power` dictates the maximum magnitude of wind."
Could you it talk about torque power and the recommended range of both parameters

virgilt:
I’ll try to have to this wrapped up either tonight or tomorrow.

virgilt:
@jjshoots @pseudo-rnd-thoughts am I only changing the asserts for wind_power and turbulence_power? Or am I also changing gravity?

virgilt:
@jjshoots @pseudo-rnd-thoughts I added warnings, removed assert, and updated docs with definition/recommendations for wind_power and turbulence_power.

gym/envs/box2d/lunar_lander.py
```diff
@@ -153,10 +155,11 @@ class LunarLander(gym.Env, EzPickle):
     `k` is set to 0.01.
     `C` is sampled randomly between -9999 and 9999.
 
-    `wind_power` dictates the maximum magnitude of wind.
+    `wind_power` dictates the maximum magnitude of linear wind applied to the craft. The recommended value for `wind_power` is between 0.0 and 20.0.
+    `turbulence_power` dictates the maximum magnitude of rotational wind applied to the craft. The recommended value for `turbulence_power` is between 0.0 and 2.0.
 
     ### Version History
-    - v2: Count energy spent
+    - v2: Count energy spent and in 0.24, added turbulance with wind power and turbulence_power parameters
```
jjshoots, line 162: This is a minor nitpick, but what is `and in 0.24`?

jjshoots:
@pseudo-rnd-thoughts Other than the minor question on the version documentation, LGTM.

gym/envs/box2d/lunar_lander.py
virgilt: It's the gym version. Should I change it to v0.24 for less confusion?

pseudo-rnd-thoughts: To make it easiest to understand, yes, but no critical

PR #2814, opened 2022-05-11
Upgrade test_determinism. Refer to #2767

gianlucadecola:
As a part of https://github.com/openai/gym/issues/2767

> Upgrade the environment determinism - Improve the number of time steps that we environment step equivalence. This will also test the environments can step / rollout

Upgraded steps from 4 to 50, manage done status with reset

### DIFF
--- a/tests/envs/test_determinism.py
+++ b/tests/envs/test_determinism.py
@@ -7,18 +7,18 @@ from tests.envs.spec_list import spec_list
 @pytest.mark.parametrize("spec", spec_list, ids=[spec.id for spec in spec_list])
 def test_env(spec):
     """Run a rollout with two environments and assert equality.
-    
+
     This test run a rollout of NUM_STEPS steps with two environments
     initialized with the same seed and assert that:
-    
+
     - observation after first reset are the same
     - same actions are sampled by the two envs
     - observations are contained in the observation space
     - obs, rew, done and info are equals between the two envs
 
     Args:
-        spec (EnvSpec): Environment specification  
-    
+        spec (EnvSpec): Environment specification
+
     """
     # Note that this precludes running this test in multiple
     # threads. However, we probably already can't do multithreading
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts:
This looks good, could you do the following
1. Could you add a test to assert the observations are with the observation space
2. The `test_random_rollout` appears to have the same function as this test so I think it can be removed
3. Add a docstring 

Personally, I think that the test should generate the environments and set seeds then run the for loop checking the step values within the for loop rather than saving all of the values to a list then checking as it is currently.   
This should reduce the amount of code in the function while allowing us to check the action, observation, reward, info at the same time (not in two separate parts as it is currently)

gianlucadecola:
Made the changes :+1:

Markus28:
Regarding the docstring: I don't think it really complies with the new convention we are using. You can check CONTRIBUTING.md to find details on how to format docstrings. The first line should be a very short sentence describing what the function is supposed to do (i.e. high level idea), ending with a period.

gianlucadecola:
> 

Thanks for the hint! It should be ok now, pydocstyle now complains only about the module level docstring and assert_equals function, I can update them in this PR if needed

Markus28:
Awesome, thank you! I guess you only really need to document the stuff you changed. But if you feel like it, you're welcome to complete the other docstrings in test_determinism.py as well (I don't think anyone is working on the tests atm) :)

tests/envs/test_determinism.py
```diff
@@ -6,57 +6,78 @@
...
-        assert_equals(o1, o2, f"[{i}] ")
-        assert r1 == r2, f"[{i}] r1: {r1}, r2: {r2}"
-        assert d1 == d2, f"[{i}] d1: {d1}, d2: {d2}"
+        assert rew1 == rew2, f"[{i}] r1: {rew1}, r2: {rew2}"
+        assert done1 == done2, f"[{i}] d1: {done1}, d2: {done2}"
 
         # Go returns a Pachi game board in info, which doesn't
         # properly check equality. For now, we hack around this by
         # just skipping Go.
         if spec.id not in ["Go9x9-v0", "Go19x19-v0"]:
```
pseudo-rnd-thoughts, line 72: To my knowledge, gym does not have Go in it, so this line can be removed.

pseudo-rnd-thoughts (approved):
Looks good to me (minus the Go comment), ready to merge

Issue #2809, opened 2022-05-09
[Bug Report] Import of custom environments is broken in 0.22


vadim0x60:
**Describe the bug**
In gym 0.21 there is a useful feature for loading custom environments. One can call

```
import gym
gym.make('module:Env')
```

And gym will import the `module` before trying to make `Env`. In gym 0.22 it got broken. 

**Code example**

Make a file named `test.py`:

```
print('File loaded!')

import gym

class DummyEnv(gym.Env):
    action_space = gym.spaces.Discrete(2)
    observation_space = gym.spaces.Discrete(2)

    def reset(self):
        obs = observation_space.sample()
        return obs

    def step(self, action):
        obs = observation_space.sample()
        r = 0
        done = False
        return obs, r, done, {}

gym.register('Dummy-v0', entry_point='test:DummyEnv')
```

then, in python console or in another file, run `gym.make('test:Dummy-v0')`. If you have gym 0.21.0 installed, you get

```
File loaded!
<OrderEnforcing<DummyEnv<Dummy-v0>>>
```

With gym 0.22.0 it's

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/320012346/.cache/pypoetry/virtualenvs/cleanrl-9SW_WqcQ-py3.9/lib/python3.9/site-packages/gym/envs/registration.py", line 676, in make
    return registry.make(id, **kwargs)
  File "/home/320012346/.cache/pypoetry/virtualenvs/cleanrl-9SW_WqcQ-py3.9/lib/python3.9/site-packages/gym/envs/registration.py", line 490, in make
    versions = self.env_specs.versions(namespace, name)
  File "/home/320012346/.cache/pypoetry/virtualenvs/cleanrl-9SW_WqcQ-py3.9/lib/python3.9/site-packages/gym/envs/registration.py", line 220, in versions
    self._assert_name_exists(namespace, name)
  File "/home/320012346/.cache/pypoetry/virtualenvs/cleanrl-9SW_WqcQ-py3.9/lib/python3.9/site-packages/gym/envs/registration.py", line 297, in _assert_name_exists
    raise error.NameNotFound(message)
gym.error.NameNotFound: Environment `test:Dummy` doesn't exist.
```

**System Info**

Python 3.9.12, installing gym 0.21.0 and 0.22.0 via [poetry](https://python-poetry.org/) on Ubuntu 18.04.6 LTS, 4.15.0-176-generic

### Checklist

[...]

PR #2815, opened 2022-05-12
Add support for including module with gym.make

arjun-kg:
# Description

Resolves (https://github.com/openai/gym/issues/2809)

Adding (back) support for loading modules during make for eg. `gym.make('module:env-v0')`. See the issue for more details on use-case and motivation. Corresponding docs here -  https://github.com/Farama-Foundation/gym-docs/pull/130

Involves a tiny change in registration code with the `id` being split to check for a module, and module being imported if not None. This adds the custom env to registry (the expectation is that registration is done in the imported module) and everything works like normal after that. 

One drawback of this is that now the `id` variable could include module in make but not in registration so there could be confusion? I still use the original id for parsing with `parse_env_id` and don't touch anything else. I don't really see downsides to this. The alternative, if I wanted to keep meaning of `id` consistent, the id should now be split into `module, namespace, name, version` and this change reflected everywhere including the re pattern, but that felt like overkill since this isn't a mainstream feature. 

## Type of change

# Checklist:


gym/envs/registration.py
```diff
@@ -460,6 +460,9 @@ def make(
     if isinstance(id, EnvSpec):
         spec_ = id
     else:
+        module, id = (None, id) if ":" not in id else id.split(":")
+        if module is not None:
```
RedTachyon, line 464: A specific error message would be nice here if this fails, we should maybe explicitly say that "You tried to create an environment via importing a package, and that import didn't work"

tests/envs/dummy_env.py
```diff
@@ -0,0 +1,17 @@
...
+        self.obs_space = gym.spaces.Discrete(2)
+
+    def reset(self):
+        return self.obs_space.sample()
+
+    def step(self, action):
+        return self.obs_space.sample(), 0, False, {}
+
+
+gym.register("Dummy-v0", entry_point="tests.envs.dummy_env:DummyEnv")
```
RedTachyon, line 17: Do we need the full path here? (`tests.envs.dummy_env`) Or can we just do `dummy_env:DummyEnv`?

tests/envs/test_registration.py
```diff
@@ -284,3 +284,8 @@ def test_namespace():
 
     del registry["MyDefaultNamespace/MyDefaultEnvironment-v0"]
     del registry["MyDefaultEnvironment-v1"]
+
+
+def test_import_module_during_make():
+    # Test custom environment which is registered at make
+    gym.make("tests.envs.dummy_env:Dummy-v0")
```
RedTachyon, line 291: Same comment about the full path

tests/envs/dummy_env.py
```diff
@@ -0,0 +1,17 @@
+import gym
+
+
+class DummyEnv(gym.Env):
+    def __init__(self):
+        super().__init__()
+        self.act_space = gym.spaces.Discrete(2)
+        self.obs_space = gym.spaces.Discrete(2)
+
+    def reset(self):
```
RedTachyon, line 10: For consistency, can you update the signature here to follow the new conventions?

tests/envs/dummy_env.py
```diff
@@ -0,0 +1,17 @@
+import gym
+
+
+class DummyEnv(gym.Env):
```
RedTachyon, line 4: A docstring/comment would be useful to state in which test this is used

RedTachyon: And maybe even renaming Dummy to something more descriptive for the specific test - it's actually crucially important that we don't accidentally create a DummyEnv in the main test file

RedTachyon (commented):
Implementation is basically what I've been thinking about, so I left a few auxiliary comments.

tests/envs/dummy_env.py
arjun-kg: I assume you mean returning info with reset?

tests/envs/dummy_env.py
pseudo-rnd-thoughts: @arjun-kg Could you test this locally with the example from #2809 `pybullet_envs:HopperBulletEnv-v0`? Does this work with it?

RedTachyon: @pseudo-rnd-thoughts We actually discussed this exact thing on Discord, and it turns out that pybullet tried to be too clever for its own good, so its registration method fails with the new registration setup. I'm currently working on adding a compatibility layer that would be only slightly cursed, but would deprecate their approach in a more "chill" way instead of breaking everything.

tests/envs/dummy_env.py
RedTachyon: Yep, and the arguments

tests/envs/register_during_make_env.py
```diff
@@ -0,0 +1,22 @@
+import gym
+
+
+class RegisterDuringMakeEnv(gym.Env):
+    """Used in `test_registration.py` to check if `env.make` can import and register an env"""
+
+    def __init__(self):
+        super().__init__()
+        self.act_space = gym.spaces.Discrete(2)
+        self.obs_space = gym.spaces.Discrete(2)
```
Markus28, line 10: Why not `self.action_space` and `self.observation_space`?

arjun-kg: Hmm just instinctively did it, I've seen a trend of using shorthand type names when the variables don't really matter. Like, `o, r, d, i = env.step(action)` in tests when we're not really going to care about the variables themselves and writing the whole thing can get verbose. But in hindsight having proper names keeps things uniform, I will modify it.

Markus28: Since it's part of the official API, it really should be `self.action_space`. I think there is currently a PR that would raise an exception if the environment is instantiated via `make` without `self.action_space`.

pseudo-rnd-thoughts:
@RedTachyon Is this ready to merge?

RedTachyon:
Looks good now

Issue #2817, opened 2022-05-12
[Bug Report] PyBullet Envs don't work with master


arjun-kg:
Opening this issue in case other people run into this before it can be fixed. 

Pybullet has this piece of code before they register all their envs when `pybullet_envs` is imported - 
```python
def register(id, *args, **kvargs):
  if id in registry.env_specs:
    return
  else:
    return gym.envs.registration.register(id, *args, **kvargs)
```
In the current registration mechanism, `registry` is a dict and so `registry.env_specs` is invalid. This will be fixed before the next release.

PR #2818, opened 2022-05-12
Add a compatibility layer for the new registration mechanism

RedTachyon:
As I now found out, some envs like pybullet were too clever for their own good, and added extra logic instead of just calling `gym.register(...)`, so they will break with the new registration mechanism (which was meant to be backwards compatible for reasonable uses - I didn't expect someone would inspect the registry directly, since all that is handled by gym anyways)

This PR adds a thin layer around the registry, which is a dict, and it will remain a dict with a few extra methods. This should give you a warning if you try to do something weird, and not change the dict behavior in any way. Ideally, this will be removed in 1.0

I'm not sure if we actually need this (because maybe we can let pybullet people know so that they can update their code), but I'm leaning towards keeping the interoperability for now.

### DIFF
--- a/gym/envs/registration.py
+++ b/gym/envs/registration.py
@@ -340,6 +340,8 @@ def make(id: EnvSpec, **kwargs) -> Env: ...
 class EnvRegistry(dict):
     """A glorified dictionary for compatibility reasons"""
 
+    # TODO: remove this at 1.0
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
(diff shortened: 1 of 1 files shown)


RedTachyon:
(see #2817)

pseudo-rnd-thoughts:
Could you expand the `EnvRegistry` docstring to explain why this exists
Otherwise, looks good to merge

RedTachyon:
Done. Also I tested locally with a pybullet env and it seems to work now.

PR #2819, opened 2022-05-13
Improve ``play``

Markus28:
This PR is supposed to make the `play` function more convenient to work with

### Changes

Changes to `key_to_action`:
- No longer require the key-combinations to be ordered when passing them to the function
- Allow characters instead of corresponding `ord("<key>")`
- Allow strings representing key-combinations keys instead of tuples

In total these changes allow `{"wa": action}` instead of `{sorted(ord("w"), ord("a")): action}` while remaining backwards compatible.

Additional changes:

- Set the default `fps` to `None` and try to infer framerate from metadata (previously default was 30)
- Previously, `play` only worked with discrete action spaces and used action `0` when an unknown key combination or no key was pressed. I introduced a keyword argument for a NOOP action for this

All of this should be mostly backward compatible (the framerate will change if `fps` was not specified and "render_fps" in metadata is not 30). 

### Example:

```
import gym
from gym.utils.play import play
import numpy as np


play(gym.make("CarRacing-v1"), keys_to_action={"w": np.array([0, 0.7, 0]),
                                               "a": np.array([-1, 0, 0]),
                                               "s": np.array([0, 0, 1]),
                                               "d": np.array([1, 0, 0]),
                                               "wa": np.array([-1, 0.7, 0]),
                                               "dw": np.array([1, 0.7, 0]),
                                               "ds": np.array([1, 0, 1]),
                                               "as": np.array([-1, 0, 1]),
                                               }, noop=np.array([0,0,0]))

```

### Docstrings
If this is merged before the PR for utils docstrings, I will add docstrings there. Otherwise, this PR will be updated.

### DIFF
--- a/gym/utils/play.py
+++ b/gym/utils/play.py
@@ -102,7 +102,7 @@ def play(
     fps: Optional[int] = None,
     zoom: Optional[float] = None,
     callback: Optional[Callable] = None,
-    keys_to_action: Optional[Dict[Union[Tuple[int], str], ActType]] = None,
+    keys_to_action: Optional[Dict[Union[Tuple[Union[str, int]], str], ActType]] = None,
     seed: Optional[int] = None,
     noop: ActType = 0,
 ):
--- a/gym/utils/play.py
+++ b/gym/utils/play.py
@@ -1,4 +1,4 @@
-from typing import Callable, Dict, Optional, Tuple
+from typing import Callable, Dict, Optional, Tuple, Union
 
 import pygame
 from numpy.typing import NDArray
@@ -20,6 +20,8 @@ except ImportError as e:
 from collections import deque
 
 from pygame.locals import VIDEORESIZE
+
+from gym.core import ActType
 
 
 class MissingKeysToAction(Exception):
@@ -97,11 +99,12 @@ def display_arr(
 def play(
     env: Env,
     transpose: Optional[bool] = True,
-    fps: Optional[int] = 30,
+    fps: Optional[int] = None,
     zoom: Optional[float] = None,
     callback: Optional[Callable] = None,
-    keys_to_action: Optional[Dict[Tuple[int], int]] = None,
+    keys_to_action: Optional[Dict[Union[Tuple[int], str], ActType]] = None,
     seed: Optional[int] = None,
+    noop: ActType = 0,
 ):
     """Allows one to play the game using keyboard.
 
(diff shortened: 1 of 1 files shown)


pseudo-rnd-thoughts:
Let making docstring for environments, some have the `get_keys_to_action`. 
Would you update them with the letter version as that is easier to understand. 
Personally, I think it makes more sense for this data to be in the metadata or as a separate function. Thoughts?

Markus28:
I tend to agree. Is the metadata required to be a class attribute (that's how it is implemented most of the time) or do we also allow it to be an instance attribute? In the former case, using metadata might be problematic because the keys_to_action mapping can depend on constructor arguments (e.g. whether a discretized version of the action space is used). 


I *thought* that keys_to_action was part of metadata so maybe someone changed it at some point for a good reason?

gym/utils/play.py
```diff
@@ -97,11 +99,12 @@ def display_arr(
 def play(
     env: Env,
     transpose: Optional[bool] = True,
-    fps: Optional[int] = 30,
+    fps: Optional[int] = None,
     zoom: Optional[float] = None,
     callback: Optional[Callable] = None,
-    keys_to_action: Optional[Dict[Tuple[int], int]] = None,
+    keys_to_action: Optional[Dict[Union[Tuple[int], str], ActType]] = None,
```
pseudo-rnd-thoughts, line 105: I realised that a `Tuple[str]` is unnecessary but it feels weird compared to passing a `Tuple[int]` and took me a while to work out why we don't need it. Could you add a comment to the docstring on this

gym/utils/play.py
```diff
@@ -161,7 +164,19 @@ def callback(obs_t, obs_tp1, action, rew, done, info):
...
+    for key_combination, action in keys_to_action.items():
+        key_code = tuple(
+            sorted(ord(key) if isinstance(key, str) else key for key in key_combination)
+        )
+        key_code_to_action[key_code] = action
+
+    game = PlayableGame(env, key_code_to_action, zoom)
+
+    if fps is None:
+        fps = env.metadata.get("render_fps", 30)
```
pseudo-rnd-thoughts, line 179: I think some environments don't have "render_fps", wrapper `monitoring.video_recorder` provides some code on this

pseudo-rnd-thoughts (commented):
You are probably right that the constructor arguments could change it so lets not modify it. Otherwise looks good

gym/utils/play.py
Markus28: True, but I think it's fine as long as I use `get`, right? I think that's also how `monitoring.video_recorder` does it :)

pseudo-rnd-thoughts: That is true, plus we say in the documentation that the default is 30.

pseudo-rnd-thoughts:
Would you still be able to update environments with integer key values to their string versions?

pseudo-rnd-thoughts:
Then looks good to merge

Markus28:
Yeah I will take a look at that. From what I have seen so far, at least one environment uses the arrow keys, so we still need tuples of integers there

pseudo-rnd-thoughts:
Looks good to merge

PR #2822, opened 2022-05-16
Fix edge case for out of bounds

andrewtanJS:
# Description
This deals with the edge case where an object's corner may not be in the screen, but the object still overlaps and should be drawn.

Additionally, more tests were performed to ensure that both the performance as well as consistency to the previous version was maintained. The resulting observations from `step()` are unaffected between changes.
## Type of change

### Screenshots
There is still significant performance increase from not drawing out of bounds objects, and loosening the condition does not adversely affect these changes by a significant margin.
Before clipping:
image

After:
image

# Checklist:


pseudo-rnd-thoughts:
LGTM

PR #2829, opened 2022-05-23
mountain_car reward set to 0 if reaching goal

mreitschuster:
# Description

mountain_car reward was always -1

description states it should  be 0 if goal is reached.:
 The goal is to reach the flag placed on top of the right hill as quickly as possible, as such the agent is penalised with a reward of -1 for each timestep it isn't at the goal and is not penalised (reward = 0) for when it reaches the goal.

now code behaves as description defines.

# Checklist:

- [X] I have run the [`pre-commit` checks](https://pre-commit.com/) with `pre-commit run --all-files` (see `CONTRIBUTING.md` instructions to set it up)

### DIFF
--- a/gym/envs/classic_control/mountain_car.py
+++ b/gym/envs/classic_control/mountain_car.py
@@ -127,6 +127,8 @@ class MountainCarEnv(gym.Env):
 
         done = bool(position >= self.goal_position and velocity >= self.goal_velocity)
         reward = -1.0
+        if done:
+            reward = 0
 
         self.state = (position, velocity)
         return np.array(self.state, dtype=np.float32), reward, done, {}


pseudo-rnd-thoughts:
Thanks for spotting that bug, looks good to merge

mreitschuster:
sorry, I am not familiar with doing a PR in github to another project. Is my understanding correct that I do not have any action left, but it basically now depends on someone (with the appropriate authority according to policy) need to finish the workflows & approve?

pseudo-rnd-thoughts:
@mreitschuster Yeah, @jkterry1 will approve the workloads and merge it. Thanks for your PR
If you want to get involved in the development of gym, then join our discord https://discord.gg/nHg2JRN489 and message me PseudoRnd

RedTachyon:
If we want to merge this, we'd need to bump the version, no? It's not a huge difference, but it will change your results when evaluating any policy.

pseudo-rnd-thoughts:
@RedTachyon you are probably right, should we modify the docstring then because it would be easier to keep the version number the same and all policies would act the same? And change the docstring to reflect the implementation (and note in the docstring that the reward was modified)

jkterry1:
Closing in favor of mark's new PR (which he'll link here)

PR #2831, opened 2022-05-23
Fix scale for Car Racing

### LINKED ISSUE #2830: [Bug Report] CarRacing environment - Change of SCALE (car_racing.py) results to shifted initial position
release 0.23.1

**Describe the bug**
Trying to run the car_racing.py with different values on track scale parameter SCALE leads to shifted starting position of the car.
While the camera starts at the expected starting position on the track, the car initial position is shifted.
 
The only parameters modified is ZOOM, reduced to 0.2 in order to make the bug visible, and SCALE in order to achieve larger tracks.

default value of SCALE: 6.0
example value of SCALE: 10.0

Below there are two examples that show the car position for 
default value of SCALE 

281891809_420748906224647_7156620405392163796_n

and example value of SCALE.
281920040_717363976074899_6156146512703631858_n

The initial angle position of the car is specified here:
`self.car = Car(self.world, *self.track[0][1:4])`

and it is related to first tile of the track that is generated but the visual result is different.

**System Info**
 * Anaconda environment
 * MacOS Monterey 12.3
 * Python 3.9.12

**Desired solution**
The desired behaviour would be that regardless the SCALE, the initial position of the car should match the first track tile.

I think in the previous version of CarRacing env, with pyglet, this was not an issue.

### Checklist

andrewtanJS:
# Description
This shifts the relative position of the rendering coordinates by `PLAYFIELD`, so that `car_dynamics` need not take it into consideration when calculating the coords to render.

Fixes #2830

## Type of change

### Screenshots
Before:
image

After:
image

After (SCALE=10), note that the change in the car size is as expected
image

# Checklist:

### DIFF
--- a/gym/envs/box2d/car_dynamics.py
+++ b/gym/envs/box2d/car_dynamics.py
@@ -280,9 +280,7 @@ class Car:
             for f in obj.fixtures:
                 trans = f.body.transform
                 path = [trans * v for v in f.shape.vertices]
-                path = [
-                    (coords[0], coords[1]) for coords in path
-                ]
+                path = [(coords[0], coords[1]) for coords in path]
                 path = [pygame.math.Vector2(c).rotate_rad(angle) for c in path]
                 path = [
                     (
@@ -317,10 +315,7 @@ class Car:
                 ]
                 white_poly = [trans * v for v in white_poly]
 
-                white_poly = [
-                    (coords[0], coords[1])
-                    for coords in white_poly
-                ]
+                white_poly = [(coords[0], coords[1]) for coords in white_poly]
                 white_poly = [
                     pygame.math.Vector2(c).rotate_rad(angle) for c in white_poly
                 ]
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts:
@andrewtanJS Is this a fix for when the change from pyglet to pygame or is this a fix of a long-running bug and therefore we need to bump the version?

andrewtanJS:
@pseudo-rnd-thoughts This was a bug from the pyglet to pygame change, so no version bump is necessary.

pseudo-rnd-thoughts:
It looks good to merge. Quick question, is it possible for someone to change SCALE and now get a variable that does not exist or make a change that does nothing?

andrewtanJS:
As long as `SCALE` is a positive float, the car should be scaled by the corresponding amount. Otherwise, the environment would correctly throw an error.

pseudo-rnd-thoughts:
I more meant the car dynamics scale and playfield is now removed but someone could have modified them before and have an impact?

andrewtanJS:
Oh I see. The code itself doesn't have any other references to the car_dynamics scale and playfield beyond what is removed in the PR, but it is possible for someone to have modified it in their own use, and changing the car_dynamics scale would get a variable that does not now exist.

In any case, this change does bring it back to be consistent with before the update to pygame, as `SCALE` did not previously exist within car_dynamics.

pseudo-rnd-thoughts:
Ok, then I think this is good to merge

PR #2837, opened 2022-05-24
Removed downloading mujoco in py.Dockerfile

pseudo-rnd-thoughts:
As we have moved to the deepmind mujoco bindings then there is no need to download mujoco anymore


pseudo-rnd-thoughts:
We are still testing with `mujoco-py` so this PR is not possible for now

PR #2841, opened 2022-05-26
Test BipedalWalker hardcore kwargs. Refer to #2767

gianlucadecola:
As a continuation of  #2767 I'm adding a test for the argument `hardcore` of BipedalWalker. The `hardcore=False` version create only grass terrain while the `hardcore=True` version create also pit, stumps and ladders.


pseudo-rnd-thoughts:
LGTM

PR #2842, opened 2022-05-27
box.contains constraint relaxation

Mohan-Zhang-u:
# Description

This change relaxed a constraint used by box.contains.

According to https://numpy.org/doc/stable/reference/generated/numpy.can_cast.html, np.can_cast by default sets casting to 'safe'. However, this setting may cause potential problems in practice. Consider code snippets below:

```
from gym import spaces
import numpy as np
observation_space = spaces.Box(low=np.array([0., 0.], dtype=np.float32), high=np.array([1., 1.], dtype=np.float32), shape=(2,))
observation = np.array([.5, .5], dtype=np.float64)
observation_space.contains(observation)
```

Before the proposed change, the output would be False, simply because np.can_cast(np.float64, np.float32, casting='safe') returns False. After the proposed change, the output would be True, because np.can_cast(np.float64, np.float32, casting='same_kind') returns True.

I believe that casting='same_kind', which allows only safe casts or casts within a kind, should produce intended behavior suggested by function name 'contains', as observation = np.array([.5, .5], dtype=np.float64) should be considered inside observation_space = spaces.Box(low=np.array([0., 0.], dtype=np.float32), high=np.array([1., 1.], dtype=np.float32), shape=(2,)). 

The fix provides a more general behavior.

Fixes # (issue)

## Type of change

### Screenshots

# Checklist:

### DIFF
--- a/gym/spaces/box.py
+++ b/gym/spaces/box.py
@@ -198,7 +198,7 @@ class Box(Space[np.ndarray]):
             x = np.asarray(x, dtype=self.dtype)
 
         return bool(
-            np.can_cast(x.dtype, self.dtype)
+            np.can_cast(x.dtype, self.dtype, casting='same_kind')
             and x.shape == self.shape
             and np.all(x >= self.low)
             and np.all(x <= self.high)


pseudo-rnd-thoughts:
Thanks for your PR, while this initially sounds like a good plan. Im not sure this implementation is the way to go
The problem is that in some cases, it may be true that a `float64` number is not contained within a `float32`, this is due to the valid ranges of the data types. While this problem is very unlikely in real RL environments, this is still a real problem

Secondly, the observation space should be modified to reflect the actual observations or vice versa. 
In what case are you working with where this is a problem?

pseudo-rnd-thoughts:
I understand but why can you only not update the observation shape to either `float64` or change your observation to use `float32`.

jkterry1:
Hey, I discussed this with the team, and I think this would result in _very_ unexpected behaviors when passing things like a fp16 dtype to a fp64 GPU

PR #2850, opened 2022-05-29
Remove a redundant warning in registration

### LINKED ISSUE #2849: [Question] How to suppress  Custom namespace `ALE` is being overridden by namespace `ALE` warning?
### Question

Installing gym and then ale-py (easiest is to do that in a colab), `import gym` throws a warning:

```
 Custom namespace `ALE` is being overridden by namespace `ALE`
```
how to get rid of it?

(does not seem normal that basic install of gym throws a warning)
gym 0.24
ale-py 0.7.5

RedTachyon:
I accidentally doubled one check for the namespace in the new registration mechanism, which caused some erroneous warnings. I removed one instance of this check, and added a sanity check in the remaining one (if we're overriding a namespace with the same namespace, we don't really care)

Fixes #2849

### DIFF
--- a/gym/envs/registration.py
+++ b/gym/envs/registration.py
@@ -417,17 +417,7 @@ current_namespace: Optional[str] = None
 
 def _check_spec_register(spec: EnvSpec):
     """Checks whether the spec is valid to be registered. Helper function for `register`."""
-    global registry, current_namespace
-    if current_namespace is not None:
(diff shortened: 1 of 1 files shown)


pseudo-rnd-thoughts:
This looks good however could you move this to the stable branch as all critical bug fixes are happening there

PR #2853, opened 2022-05-30
Add TensorFlow and Torch Conversion Wrappers

justinDeu:
# Description

Adds wrapper environments for PyTorch and TensorFlow to interface with an underlying JaxEnvironment. The wrappers allow for PyTorch or TensorFlow Tensors to be used for the actions and observations will be provided as Tensors as well. The Wrapper environments handle converting between the Tensor type and Jax arrays as needed. 

This is needed for when gym moves to using Jax-enabled environments. 

## Type of change

- [ X ] New feature (non-breaking change which adds functionality)

# Checklist:

# Notes:

* Vector environments are currently not fully supported. When the VectorEnvs concatenate the observations they force the Jax arrays to be numpy arrays. The tests and rough code are in place within the wrappers to handle this, but it was decided not to fully support vector environments at this time because more changes will be needed in the future, specifically something along the lines of a `BraxVectorEnv`.
* The `action_space` and `observation_space` properties and setters were not overridden because it may break other features within the libraries. This means users must convert from PyTorch/TensorFlow tensors to numpy arrays to do checks against the spaces such as `contains` or `in` which are very common. Currently, I think gym assumes spaces to be numpy arrays. Should this be the pattern when using type-conversion wrappers such as these, or should the user be able to interact with the spaces in the wrapped type? For the latter option, the `gym.spaces` interface may need to be refactored to support typed spaces.


gym/dev_wrappers/tf_wrapper.py
```diff
@@ -0,0 +1,48 @@
...
+            obs = [self._jax_to_tf(o) for o in obs]
+        else:
+            obs = self._jax_to_tf(obs)
+
+        return obs, reward, done, info
+
+    def reset(self, **kwargs) -> Union[tf.Tensor, tuple[tf.Tensor, dict]]:
+        return_info = kwargs.get("return_info", False)
+        if return_info:
+            obs, info = super().reset(**kwargs)
```
justinDeu, line 29: Is it preferred to interact with the wrapped environment this way or with `self.env.reset(...)`? Are they functionally the same or is there ever a scenario where they will behave differently? The same goes for calls to `step`.

I noticed in the Brax repo for their wrappers they use this method, but in other wrappers within gym the `self.env` method is used.

gym/dev_wrappers/tf_wrapper.py
```diff
@@ -0,0 +1,48 @@
...
+        if return_info:
+            return obs, info
+        else:
+            return obs
+
+    def _tf_to_jax(self, arr: tf.Tensor):
+        return jax.dlpack.from_dlpack(tf.experimental.dlpack.to_dlpack(arr))
+
+    def _jax_to_tf(self, arr):
+        return tf.experimental.dlpack.from_dlpack(jax.dlpack.to_dlpack(arr))
```
justinDeu, line 48: Would it be best to pull these out as separate utility functions that can be used elsewhere immediately or should they remain here for the time being?

tests/dev_wrappers/test_tf_wrapper.py
```diff
@@ -0,0 +1,83 @@
...
+        seed: Optional[int] = None,
+        return_info: Optional[bool] = False,
+        options: Optional[dict] = None
+    ):
+        super().reset(seed=seed)
+        self.t = self.return_reward_idx
+        if not return_info:
+            return jnp.array([self.t])
+        else:
+            return jnp.array([self.t]), {}
```
justinDeu, line 40: This is repeated in both test files. I noticed other tests all define their own dummy environments as well. Should these be abstracted to their own modules that can be imported across multiple test files?

gym/dev_wrappers/torch_wrapper.py
```diff
@@ -0,0 +1,53 @@
+from typing import Optional, Tuple, Union
+
+import torch
+from brax.io.torch import jax_to_torch, torch_to_jax
+
+from gym import Env, Wrapper
+
+
+class TorchWrapper(Wrapper):
```
pseudo-rnd-thoughts, line 9: Minor point that I may be wrong on, `to_torch_v0`, most of our wrappers don't say wrapper in their class, that is inferred. Plus the underscore style is a change we are going for in the new wrappers as we allow the version number in case we mess up and need to change something critical.  
Same for the tensorflow wrapper, `to_tensorflow_v0`. 
I feel there could be a better name, I can't think of one now

tests/dev_wrappers/test_tf_wrapper.py
pseudo-rnd-thoughts: Im planning on doing that in a separate PR soon so don't worry about it.

tests/dev_wrappers/test_tf_wrapper.py
```diff
@@ -0,0 +1,83 @@
...
+
+    return thunk
+
+
+def test_tf_wrapper_reset_info():
+    env = DummyJaxEnv(return_reward_idx=0)
+    env = TensorFlowWrapper(env)
+    obs = env.reset()
+    assert isinstance(obs, tf.Tensor)
+    del obs
```
pseudo-rnd-thoughts, line 56: There shouldn't be much need to delete variables here I believe

tests/dev_wrappers/test_tf_wrapper.py
```diff
@@ -0,0 +1,83 @@
...
+def make_env(return_reward_idx):
+    def thunk():
+        env = DummyJaxEnv(return_reward_idx)
+        return env
+
+    return thunk
+
+
+def test_tf_wrapper_reset_info():
+    env = DummyJaxEnv(return_reward_idx=0)
```
pseudo-rnd-thoughts, line 52: Could you assert that the env observations are jax and that the jax and tensorflow observations are equal

tests/dev_wrappers/test_torch_wrapper.py
```diff
@@ -0,0 +1,92 @@
...
+def make_env(return_reward_idx):
+    def thunk():
+        env = DummyJaxEnv(return_reward_idx)
+        return env
+
+    return thunk
+
+
+def test_torch_wrapper_reset_info():
+    env = DummyJaxEnv(return_reward_idx=0)
```
pseudo-rnd-thoughts, line 52: repeat of the previous point on asserting the equality of jax and pytorch environments

pseudo-rnd-thoughts (changes requested):
Generally the whole thing looks good. I have requested a couple of changes and you will need to run pre-commit

gym/dev_wrappers/tf_wrapper.py
```diff
@@ -0,0 +1,48 @@
...
+
+        if self.is_vector_env:
+            obs = [self._jax_to_tf(o) for o in obs]
+        else:
+            obs = self._jax_to_tf(obs)
+
+        return obs, reward, done, info
+
+    def reset(self, **kwargs) -> Union[tf.Tensor, tuple[tf.Tensor, dict]]:
+        return_info = kwargs.get("return_info", False)
```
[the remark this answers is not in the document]
justinDeu: refactored to this format but with `kwargs.get("return_info", False)` as the conditional check to correctly handle when it is called as `reset(return_info=False)`

gym/dev_wrappers/tf_wrapper.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def reset(self, **kwargs) -> Union[tf.Tensor, tuple[tf.Tensor, dict]]:
+        result = self.env.reset(**kwargs)
+        if kwargs.get("return_info", False):
+            return self._jax_to_tf(result[0]), result[1]
+        else:
+            return self._jax_to_tf(result)
+
+    def _tf_to_jax(self, arr: tf.Tensor):
+        # TODO: Verify this captures vector envs and converts everything from tensorflow to jax
```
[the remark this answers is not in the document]
justinDeu: added this functionality to both wrappers but am not certain of the best way to detect if the observation is a dict or not to know which function to call. I added a way in the torch wrapper if that is the approach to take, but there may be something better that I am missing.

gym/dev_wrappers/torch_wrapper.py
```diff
@@ -17,6 +22,10 @@ def __init__(self, env: Union[Wrapper, Env], device: Optional[torch.device] = No
         for an underlying Jax environment then convert the return observations from Jax arrays
         back to torch tensors.
 
+        Functionality for converting between torch and jax types originally copied from
+        https://github.com/google/brax/blob/9d6b7ced2a13da0d074b5e9fbd3aad8311e26997/brax/io/torch.py
+        Under the Apache 2.0 license. Copyright is held by the authors
+
```
justinDeu, line 28: I changed the text of the copyright statement to represent that the file wasn't copied but the functionality in the conversion methods specifically. I think this is where you intended for the statement to go?

pseudo-rnd-thoughts:
@justinDeu Sorry for never getting back to you, we need to get some previous work finished before this could be merged but that is now finished. 
Could you resolve your merge conflicts and I will have a look at the code again

gym/dev_wrappers/torch_wrapper.py
```diff
@@ -1,14 +1,19 @@
-from typing import Optional, Tuple, Union
```
pseudo-rnd-thoughts: Could you put the copyright here in a top docstring, the text below is good. This is good  https://github.com/openai/gym/blob/master/gym/utils/env_checker.py

gym/dev_wrappers/torch_wrapper.py
pseudo-rnd-thoughts: See comment above

gym/dev_wrappers/tf_wrapper.py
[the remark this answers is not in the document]
pseudo-rnd-thoughts: We should be able to do something like
```python
if isinstance(arr, dict):
     return {key: _tf_to_jax(value) for key, value in arr.items()}
```

justinDeu:
For the name of the wrappers, does the current style of `JaxToTfV0` and `JaxToTorchV0` when naming the environments does this seem too verbose and awkward? It is very clear to the user what is happening, but would something simpler such as `ToTfV0` and `ToTorchV0` suffice?

gym/dev_wrappers/torch_wrapper.py
justinDeu: Could you check on the changes I made to address this? I moved the entire comment minus the args to a docstring at the top of the file but that may be too much? Maybe move the description of the class back into the class' docstring?

justinDeu:
Ideally there should probably be separate tests for all the `x_to_y` functions or will the uses within the wrapper environments suffice? The dict conversions aren't used within the environment currently. Should I remove that from the torch environment? Previously, we said that there would be a separate wrapper that should generically handle the vector envs.

pseudo-rnd-thoughts:
Why have you lost of the `@functools.singledispatch` from the functions. This is important for users and handling multiple input types. Could you add this to conversion functions, `torch_to_jax` https://github.com/google/brax/blob/main/brax/io/torch.py

justinDeu:
Passes all of the pre-commit checks except pyright which is giving this weird error: `error: "to_dlpack" is not exported from module "gym.dev_wrappers.torch_wrapper.dlpack" (reportPrivateImportUsage)`

It seems like it could be something with the way the dlpack from torch is imported but I couldn't figure it out

gym/dev_wrappers/tf_wrapper.py
```diff
@@ -0,0 +1,99 @@
...
+    return tf.experimental.dlpack.from_dlpack(jax_dlpack.to_dlpack(value))
+
+
+@jax_to_tf.register(abc.Mapping)
+def _jax_dict_to_tf(
+    value: Dict[str, Union[DeviceArray, Any]]
+) -> Dict[str, Union[tf.Tensor, Any]]:
+    """Converts a dictionary of Jax DeviceArrays to TensorFlow tensors."""
+    return type(value)(**{k: jax_to_tf(v) for k, v in value.items()})
+
```
pseudo-rnd-thoughts, line 58: Could you add a tuple version

pseudo-rnd-thoughts (changes requested):
Thanks for the changes
Could you add `Tuple` versions of all the conversion functions
Could you add a test for each conversion that `XX_to_jax(jax_to_XX(val)) == val` for a jax numpy floats, ints, tuples and dicts. 

Otherwise, ready to merge

tests/dev_wrappers/test_tf_wrapper.py
```diff
@@ -6,6 +6,7 @@
 import tensorflow as tf
 
 import gym
+from gym.dev_wrappers.tf_wrapper import jax_to_tf, tf_to_jax
```
justinDeu, line 9: To access the conversion functions, this is the import needed. Eventually, they'd get moved out of `dev_wrappers` so it would make a little more sense right? Maybe there is some better way to provide access to them?

tests/dev_wrappers/test_torch_wrapper.py
```diff
@@ -94,3 +95,27 @@ def test_torch_wrapper_vector_env():
...
+        (3.0, 4.0, 5.0),
+        [3.0, 4.0, 5.0],
+        {
+            "a": 6.0,
+            "b": 7,
+        },
+    ],
+)
+def test_torch_conversion(value):
+    assert torch_to_jax(jax_to_torch(value)) == value
```
justinDeu, line 114: I added the tests prior to making changes to handle `Tuple`s and the tests just passed. I believe `Tuple`s are being caught under the `abc.Mapping` conversion.

gym/dev_wrappers/torch_wrapper.py
```diff
@@ -0,0 +1,121 @@
...
+    tensor = jax_dlpack.from_dlpack(tensor)
+    return tensor
+
+
+@torch_to_jax.register(abc.Mapping)
+def _torch_dict_to_jax(
+    value: Dict[str, Union[torch.Tensor, Any]]
+) -> Dict[str, Union[DeviceArray, Any]]:
+    """Converts a dictionary of PyTorch Tensors into a Dictionary of Jax DeviceArrays."""
+    return type(value)(**{k: torch_to_jax(v) for k, v in value.items()})
```
justinDeu, line 49: This and the subsequent reversal function for each wrapper type is what is capturing the `Tuple`s. The typing may need to be updated to reflect that or an explicit `Tuple` function could be added?

tests/dev_wrappers/test_tf_wrapper.py
pseudo-rnd-thoughts: I think we can add `from gym.dev_wrappers import tf_wrappers` to the wrapper `__init__.py` of `wrappers` which should allow `from gym.wrappers.tf_wrappers import jax_to_tf`

tests/dev_wrappers/test_torch_wrapper.py
pseudo-rnd-thoughts: Yeah agreed

justinDeu:
Currently, `spaces.Discrete` returns an `int` which is causing some issues with the conversion functions. The lowest conversion registered for the types are their respective container types: `np.ndarray`, `torch.Tensor`, and `tf.Tensor`. In order to preserve types and not fall back on a default, values with unregistered types are not allowed to be converted meaning the value returned from `spaces.Discrete` cannot be converted. 

The issue comes up right now when running the TensorFlow tests with the `numpy_to_jax` function. If a conversion for `int` were added, it may be difficult to identify in the roundtrip when to go back to a numeric or container type depending on how Jax stores the values. It appears that Jax stores them in DeviceArrays with shape `()`. 

`spaces.Discrete` in the future could return `np.int32` but a similar notion does not exist in PyTorch at least; I did not check TensorFlow. PyTorch uses tensors with no shape similar to Jax. It may be better for `spaces.Discrete` to instead return a no-shape `np.array` in the future making it more generic. This should mean the conversion functions do not need to handle only numeric types and could do something along the lines of:

```python
def _numpy_to_jax(value: np.array):
    if value.shape == ():
         # handle the numeric case, may not even need special handling
    else:
         # do it as is for arrays
```

justinDeu:
The issue I am facing now is telling the difference between and a numeric type and a no-shape array when round-tripping. Both will convert to a `Jax.DeviceArray(value, shape=())`. Going back from Jax to the original type, I haven't come up with a way of knowing whether to return a number or no-shape array to preserve the types when round-tripping.

```
2                                   --->     Jax.DeviceArray(2, shape=())       --->           2
np.array(2, shape=())   --->     Jax.DeviceArray(2, shape=())       --->           np.array(2, shape=())
```

PR #2864, opened 2022-06-05
Updates the environment checker

pseudo-rnd-thoughts:
We are still interested in an environment checker being run on `gym.make` however running `env_checker` as previously implemented resulted in `step` and `resest` being called.
However, this was found to cause a significant number of issue 

Therefore, this PR makes the following upgrades
* The "env_checker" is a "invasive" checker that calls `step`, `reset` and `render` to check that the environments follow the gym API specification
* Add a "passive" env checker that does not change the `step` or `reset` functions but checks the returns data is correct. In addition, passive observation and action space checkers are added and run on initialisation. 
* Add a wrapper for the passive env checker that is added to the environment on `gym.make` (otherwise `disable_env_checker=True` is passed. This check will run on only the first `reset`, `step` and `render` function checking their function signatures, and that data returned. As this wrapper needs direct access to the environment, is it the first wrapper applied to the environment in `gym.make`. 
* Various upgrades have been made to the environment checker such that checking all observation and action spaces, but not custom spaces sadly. I have attempted to make the code more modular allowing new tests to be added in the future.

### DIFF
--- a/gym/utils/env_checker.py
+++ b/gym/utils/env_checker.py
@@ -199,8 +199,10 @@ def check_render(env: gym.Env, headless: bool = False):
         env.close()
 
 
-def check_env(env: gym.Env, skip_render_check: bool = True):
+def check_env(env: gym.Env, warn: bool = None, skip_render_check: bool = True):
     """Check that an environment follows Gym API.
+
+    This is an invasive function that calls the environment's reset and step.
 
     This is particularly useful when using a custom environment.
     Please take a look at https://www.gymlibrary.ml/content/environment_creation/
@@ -208,8 +210,12 @@ def check_env(env: gym.Env, skip_render_check: bool = True):
 
     Args:
         env: The Gym environment that will be checked
+        warn: Ignored
         skip_render_check: Whether to skip the checks for the render method. True by default (useful for the CI)
     """
+    if warn is not None:
+        logger.warn("`check_env` warn parameter is now ignored.")
+
     assert isinstance(
         env, gym.Env
     ), "Your environment must inherit from the gym.Env class https://www.gymlibrary.ml/content/environment_creation/"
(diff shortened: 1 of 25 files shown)


gym/wrappers/env_checker.py
```diff
@@ -0,0 +1,57 @@
...
+from gym.utils.passive_env_checker import (
+    check_action_space,
+    check_observation_space,
+    passive_env_render_check,
+    passive_env_reset_check,
+    passive_env_step_check,
+)
+
+
+class PassiveEnvChecker(gym.Wrapper):
```
younik, line 15: Why define the Wrapper here and the method implementation in another file?

gym/utils/env_checker.py
```diff
@@ -8,477 +8,241 @@
...
         # Don't check render mode that require a
         # graphical interface (useful for CI)
         if headless and "human" in render_modes:
             render_modes.remove("human")
+
         # Check all declared render modes
-        for render_mode in render_modes:
-            env.render(mode=render_mode)
+        for mode in render_modes:
+            env.render(mode=mode)
```
younik, line 202: This will not work anymore with the new Render API.

Initially, I thought to just change the env_checker to receive the environment specification. In this way, internally it can creates as many copies as it needs without touching the copy made for the user. Having the env_spec here will also solve the problem for rendering (and in general for checking the parameters that are passed to`__init__`)

gym/wrappers/env_checker.py
pseudo-rnd-thoughts: Good question, I think that is because the passive environment checker file has more functions than are used in the wrapper and I use the passive env reset and step check in the full invasive environment checker. So these functions would have to be in a separate place

gym/utils/env_checker.py
pseudo-rnd-thoughts: I don't total get what you mean, but as this PR is going to be in the next sub-release, would you be able to make the changes in your render PR?

PR #2865, opened 2022-06-05
Rewrite setup

pseudo-rnd-thoughts:
Looking at setup.py I found it difficult to know what modules were used in testing etc 
In this PR, we modify setup.py such that
1. requirements are pulled from the requirements.txt allowing requirements to kept up-to-date
2. add meta dependency group for testing to replace "noatari" and "nomujoco" as these are undocumented and should not be used by users
3. updates the test_requirements.txt on setup to use the installed testing modules
4. updated the setup parameters for actual testing modules and requirements
5. updates the docker files for testing


kir0ul:
Just a thought: maybe at some point `setup.py` should be completely migrated to `pyproject.toml` as it's considered legacy nowadays: https://pip.pypa.io/en/stable/reference/build-system/setup-py/

pseudo-rnd-thoughts:
I didn't even realise that was possible. Is there any interesting and important advantages to it?

PR #2867, opened 2022-06-05
Rewrite env tests

pseudo-rnd-thoughts:
This PR rewrites the environment tests to make them more cohesive 
In addition, this should allow more tests to be added easily

Changes:
* Disables the environment checker for all tests due to speed and unneeded environment warnings
* Spec list - Replaced this with utils file to contain all utils functions. Simplified the spec list generation to just check if the environment can be created or not. This should make the function much simpler and raises a warning for all of the environments that cannot be initialised plus work with other modules if they have registered their environments. Added multiple variables for different testing types: all, mujoco, gym, initialised environments
* Test environment implementation - Previously there was several files each with a single or several very small tests all checking various environment implementation details. This bring all of these tests into a single file that can be easily added to
* Test environments - Simplified the environment checker to just run the `env_checker` was all environment check code should be in the function not in the test (this is linked to the upgrade env checker PR #2864). Added the test_determinism, check 
* Removed atari env spec tests - As atari is not installed in testing then there is no need for these environments anymore
* Test mujoco - Simplifies the v2/v3 conversation to use parameterize and added v4 tests (commented out due to issues, aim to add more in the future)

For the next generation wrappers, I wish to add a single testing environments to rule all the current implementation. However that will be in the future.


tests/envs/utils.py
```diff
@@ -0,0 +1,78 @@
...
+        return None
+
+
+# Tries to make all environment to previous recreation if not necessary.
+all_testing_initialised_envs = list(
+    filter(None, [try_make_env(env_spec) for env_spec in gym.envs.registry.values()])
+)
+
+# All testing, mujoco and gym environment specs
+all_testing_env_specs = [env.spec for env in all_testing_initialised_envs]
```
arjun-kg, line 26: So you first get all the env specs -> initialize the envs -> discard erroneous initializations -> get the specs again from remaining envs -> use them in tests where they are initialized again. 

There's some duplication here, instead maybe you could directly pass the initialized filtered envs to the tests?

tests/envs/utils.py
```diff
@@ -0,0 +1,78 @@
...
+]
+gym_testing_env_specs = [
+    env_spec
+    for env_spec in all_testing_env_specs
+    if any(
+        f"gym.{ep}" in env_spec.entry_point
+        for ep in ["box2d", "classic_control", "toy_text"]
+    )
+]
+# TODO, add minimum testing env spec in testing
```
arjun-kg, line 40: How will this be used? Can it be added in a later PR if this isn't used anywhere now?

arjun-kg (commented):
Looks good overall. I've also added some minor individual comments.

tests/envs/utils.py
pseudo-rnd-thoughts: Yes, my plan was that to be a future pr. 
The idea was with more environments being added in the future, phys3d, phys2d, etc
This is going to continue slow down CI but I don't believe that every test needs to check over all environments
Therefore, we can have a reduce list of representative environments, i.e., a box2d environment, a classic control environment, etc
This would be useful for wrappers where we shouldn't need to check the wrapper works with every environments, we just need to check that the wrapper works with each type of observation, action space

tests/envs/utils.py
pseudo-rnd-thoughts: Yes, that is basically how it works. 
I have tried to use the pre-initialised environments but will check again

tests/envs/test_mujoco.py
```diff
@@ -0,0 +1,71 @@
...
+    np.testing.assert_allclose(old_reset_obs, new_reset_obs)
+
+    for i in range(num_actions):
+        action = old_env.action_space.sample()
+        old_obs, old_reward, old_done, old_info = old_env.step(action)
+        new_obs, new_reward, new_done, new_info = new_env.step(action)
+
+        np.testing.assert_allclose(old_obs, new_obs, atol=EPS)
+        np.testing.assert_allclose(old_reward, new_reward, atol=EPS)
+        np.testing.assert_allclose(old_done, new_done, atol=EPS)
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: Done

tests/envs/utils.py
```diff
@@ -0,0 +1,78 @@
...
+
+import numpy as np
+
+import gym
+from gym import logger
+from gym.envs.registration import EnvSpec
+
+
+def try_make_env(env_spec: EnvSpec) -> Optional[gym.Env]:
+    """Tries to make the environment showing if it is possible. Warning no wrappers, including time limit."""
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: Clarified

tests/envs/utils.py
```diff
@@ -0,0 +1,78 @@
...
+    try:
+        return env_spec.make(disable_env_checker=True).unwrapped
+    except ImportError as e:
+        logger.warn(f"Not testing {env_spec.id} due to error: {e}")
+        return None
+
+
+# Tries to make all environment to previous recreation if not necessary.
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: Yes, basically, we just try to make all of the environment to test them
This makes sure that all of them will be tested

tests/envs/test_mujoco.py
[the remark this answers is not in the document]
pseudo-rnd-thoughts: Done

tests/envs/utils.py
arjun-kg: Although if we ignore the duplication, functionally this should be fine. There would probably be unnecessary complications in using the same pre-initialized envs everywhere (which if we want to avoid we have to create deep copies, and even then there might be issues). So, thinking again, maybe it's fine unless there's a more elegant solution

arjun-kg:
This looks good to merge.

PR #2869, opened 2022-06-06
Add Graph to Spaces

jjshoots:
# Description

This adds graph spaces to Gym.

Essentially, this PR adds support of Gym spaces to also include graph representations.
Graphs are represented using a `namedtuple`.
Graph spaces by default accept either `Discrete` or `Box` spaces as `node` or `edge` features, while the `node_list` is an `n x 2` integer only vector representing the list of nodes that each edge is connected to.

The space does not make any restrictions on amount of edges between nodes, nor does it restrict graphs from having self-connections, and further does not make any assumptions or restrictions on whether edges are unidirectional or bidirectional. 

## Type of change

# Checklist:

### DIFF
--- a/gym/spaces/graph.py
+++ b/gym/spaces/graph.py
@@ -16,15 +16,15 @@ class GraphObj:
         self,
         nodes: np.ndarray,
         edges: Optional[np.ndarray] = None,
-        node_list: Optional[np.ndarray] = None,
+        edge_links: Optional[np.ndarray] = None,
     ):
 
         self.nodes = nodes
         self.edges = edges
-        self.node_list = node_list
+        self.edge_links = edge_links
 
     def __repr__(self) -> str:
-        return f"GraphObj(nodes: \n{self.nodes}, \n\nedges: \n{self.edges}, \n\nnode_list\n{self.node_list})"
+        return f"GraphObj(nodes: \n{self.nodes}, \n\nedges: \n{self.edges}, \n\nedge_links\n{self.edge_links})"
 
     def __eq__(self, other) -> bool:
         if not isinstance(other, GraphObj):
@@ -37,14 +37,14 @@ class GraphObj:
             elif np.all(self.edges != other.edges):
                 return False
 
-            if other.node_list is None:
+            if other.edge_links is None:
                 return False
-            elif np.all(self.node_list != other.node_list):
+            elif np.all(self.edge_links != other.edge_links):
                 return False
         else:
             if other.edges is not None:
                 return False
-            if other.node_list is not None:
+            if other.edge_links is not None:
                 return False
 
         return True
@@ -52,7 +52,7 @@ class GraphObj:
 
 class Graph(Space):
     """
-    A dictionary representing graph spaces with `node_features`, `edge_features` and `edge_list`.
+    A dictionary representing graph spaces with `node_features`, `edge_features` and `edge_links`.
 
     Example usage::
 
@@ -149,19 +149,19 @@ class Graph(Space):
                 if not self.edge_space.contains(edge):
                     return False
 
-            if len(x.node_list) != len(x.edges):
+            if len(x.edge_links) != len(x.edges):
                 return False
 
-            if x.node_list.shape[-1] != 2:
+            if x.edge_links.shape[-1] != 2:
                 return False
 
-            if not np.issubdtype(x.node_list.dtype, np.integer):
+            if not np.issubdtype(x.edge_links.dtype, np.integer):
                 return False
 
-            if x.node_list.max() >= len(x.edges):
+            if x.edge_links.max() >= len(x.edges):
                 return False
 
-            if x.node_list.min() < 0:
+            if x.edge_links.min() < 0:
                 return False
 
         return True
@@ -185,7 +185,7 @@ class Graph(Space):
             ret["nodes"] = sample.nodes.tolist()
             if sample.edges is not None:
                 ret["edges"] = sample.edges.tolist()
-                ret["node_list"] = sample.node_list.tolist()
+                ret["edge_links"] = sample.edge_links.tolist()
             ret_n.append(ret)
         return ret_n
 
@@ -197,7 +197,7 @@ class Graph(Space):
                 ret_n = GraphObj(
                     np.asarray(sample["nodes"]),
                     np.asarray(sample["edges"]),
-                    np.asarray(sample["node_list"]),
+                    np.asarray(sample["edge_links"]),
                 )
             else:
                 ret_n = GraphObj(
--- a/gym/spaces/graph.py
+++ b/gym/spaces/graph.py
@@ -57,7 +57,7 @@ class Graph(Space):
     def graph_obj(
         nodes: np.ndarray, edges: np.ndarray, edge_links: np.array
     ) -> NamedTuple:
-        r"""Returns a NamedTuple representing a graph object
+        r"""Returns a NamedTuple representing a graph object.
 
         Args:
             nodes (np.ndarray): an (n x ...) sized array representing the features for n nodes.
--- a/gym/spaces/graph.py
+++ b/gym/spaces/graph.py
@@ -12,17 +12,6 @@ from gym.utils import seeding
 class GraphObj:
     """
     A base for constructing information as graphs.
-
-    `nodes` must be a nx... sized vector, where ... denotes the shape of
-    the base shape that each node feature must be.
-
-    `edges` must be either None or a np.ndarray where the first
-    dimension (denoted n) is the number of edges.
-    If edges is None, then edge_links must also be None.
-
-    `edge_links` must be a nx2 sized array of ints, where edge_links.max()
-    is not to be equal or larger than the size of the first dimension of nodes, and
-    edge_links.min() is not to be smaller than 0.
     """
     def __init__(
         self,
@@ -30,7 +19,18 @@ class GraphObj:
         edges: Optional[np.ndarray] = None,
         edge_links: Optional[np.ndarray] = None,
     ):
+        """
+        ``nodes`` must be a nx... sized vector, where ... denotes the shape of
+        the base shape that each node feature must be.
 
+        ``edges`` must be either None or a np.ndarray where the first
+        dimension (denoted n) is the number of edges.
+        If edges is None, then edge_links must also be None.
+
+        ``edge_links`` must be a nx2 sized array of ints, where edge_links.max()
+        is not to be equal or larger than the size of the first dimension of nodes, and
+        edge_links.min() is not to be smaller than 0.
+        """
         self.nodes = nodes
         self.edges = edges
         self.edge_links = edge_links
@@ -82,9 +82,22 @@ class Graph(Space):
     def __init__(
         self,
         node_space: Union[Box, Discrete],
-        edge_space: Union[Box, Discrete],
+        edge_space: Union[None, Box, Discrete],
         seed: Optional[Union[int, seeding.RandomNumberGenerator]] = None,
     ):
+        r"""Constructor of :class:`Graph`.
+
+        The argument ``node_space`` specifies the base space that each node feature will use.
+        This argument must be either a Box or Discrete instance.
+
+        The argument ``edge_space`` specifies the base space that each edge feature will use.
+        This argument must be either a Box or Discrete instance.
+
+        Args:
+            node_space (Union[Box, Discrete]): space of the node features.
+            edge_space (Union[None, Box, Discrete]): space of the node features.
+            seed: Optionally, you can use this argument to seed the RNG that is used to sample from the space.
+        """
         self._np_random = None
         if seed is not None:
             if isinstance(seed, seeding.RandomNumberGenerator):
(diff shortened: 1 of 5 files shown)


gym/spaces/graph.py
```diff
@@ -0,0 +1,260 @@
...
+class GraphObj:
+    r"""A base for constructing information as graphs."""
+
+    def __init__(
+        self,
+        nodes: np.ndarray,
+        edges: Optional[np.ndarray] = None,
+        edge_links: Optional[np.ndarray] = None,
+    ):
+        r"""Constructor for Graph information.
```
pseudo-rnd-thoughts, line 22: Could you use the google docstring style?

gym/spaces/graph.py
```diff
@@ -0,0 +1,260 @@
...
+
+from gym.spaces.box import Box
+from gym.spaces.discrete import Discrete
+from gym.spaces.multi_discrete import MultiDiscrete
+from gym.spaces.space import Space
+from gym.utils import seeding
+
+
+class GraphObj:
+    r"""A base for constructing information as graphs."""
```
pseudo-rnd-thoughts, line 14: Could you improve this docstring with example code

gym/spaces/graph.py
```diff
@@ -0,0 +1,260 @@
...
+
+        The argument ``edge_space`` specifies the base space that each edge feature will use.
+        This argument must be either a Box or Discrete instance.
+
+        Args:
+            node_space (Union[Box, Discrete]): space of the node features.
+            edge_space (Union[None, Box, Discrete]): space of the node features.
+            seed: Optionally, you can use this argument to seed the RNG that is used to sample from the space.
+        """
+        self._np_random = None
```
pseudo-rnd-thoughts, line 106: Why is this necessary, should this not be passed to the `super().__init__` which will handle this?

gym/spaces/graph.py
```diff
@@ -0,0 +1,260 @@
...
+
+        self.node_space = node_space
+        self.edge_space = edge_space
+
+        # graph object creator
+        self.graphObj = GraphObj
+
+        super().__init__(None, None, seed)
+
+    def _generate_sample_space(self, base_space, num) -> Optional[Union[Box, Discrete]]:
```
pseudo-rnd-thoughts, line 129: Add `base_space` and `num` type?

gym/spaces/graph.py
```diff
@@ -0,0 +1,260 @@
...
+        sampled_edges = self._sample_sample_space(edge_sample_space)
+
+        sampled_node_list = None
+        if sampled_edges is not None:
+            if num_edges > 0:
+                sampled_node_list = self.np_random.integers(
+                    low=0, high=num_edges, size=(num_edges, 2)
+                )
+
+        return GraphObj(sampled_nodes, sampled_edges, sampled_node_list)
```
pseudo-rnd-thoughts, line 180: Is this graph object necessary? Could we not return a tuple of nodes, adjacency matrix and sampled node list. 
Here, `sampled_node_list` and in graphobj `edge_links` have completely different variable names, why?

gym/spaces/graph.py
```diff
@@ -0,0 +1,260 @@
...
+            num_edges = self.np_random.integers(num_nodes * num_nodes)
+
+        node_sample_space = self._generate_sample_space(self.node_space, num_nodes)
+        edge_sample_space = self._generate_sample_space(self.edge_space, num_edges)
+
+        sampled_nodes = self._sample_sample_space(node_sample_space)
+        sampled_edges = self._sample_sample_space(edge_sample_space)
+
+        sampled_node_list = None
+        if sampled_edges is not None:
```
pseudo-rnd-thoughts, line 174: simplify the two if statement into one

gym/spaces/graph.py
```diff
@@ -0,0 +1,260 @@
...
+from gym.spaces.discrete import Discrete
+from gym.spaces.multi_discrete import MultiDiscrete
+from gym.spaces.space import Space
+from gym.utils import seeding
+
+
+class GraphObj:
+    r"""A base for constructing information as graphs."""
+
+    def __init__(
```
pseudo-rnd-thoughts, line 16: I haven't done much with graphs in a long time, but can they not be represented by just an adjacency matrix and a node list (I guess, though not necessary)? Have I missed something, but how can be a graph be represented by just a list of nodes?

pseudo-rnd-thoughts (changes requested):
Im the most familiar with graphs for neural networks, is there a prior project you are basing this code off? Or that it would be easily compatible with? 
Sorry, if my questions don't make sense, I just want to understand, what is happening

gym/spaces/graph.py
jjshoots: Sure

gym/spaces/graph.py
jjshoots: Sure

gym/spaces/graph.py
jjshoots: I'm not completely sure about, I kinda just used the Box space as a template for building thid space, but I'll have a look again

gym/spaces/graph.py
jjshoots: Sure

gym/spaces/graph.py
jjshoots: The idea for using tuples can work, but also makes things much harder to enforce and perform checks for. But if that is desired, I can change it to be so.

`sampled_node_list` and `edge_links` is an artifact from some refactoring, thanks for catching this.

gym/spaces/graph.py
jjshoots: Sure

gym/spaces/graph.py
jjshoots: Yes they can, but that comes with 2 immediate downsides - 1. n^2 instead of 2n memory complexity (where n can be m^2 for m nodes), 2. Inability to represent directional edges without having two adjacency matrices (or one with different tokens representing different connections, which further complicates things). Though it is a much rarer use case, an adjacency matrix also can't represent multiple edges between two nodes.

Using a list makes things much easier for computation using GNNs or transformer-like models as well, since most of the time the `edge_links` can just be concatenated on top of the `edges` and directly passed to the model.

In any case, one should be able to construct an adjacency matrix using `edge_links` fairly easily since it's already a list of indices.

jjshoots: In any case, if an adjacency matrix is the preferred way to do things, I propose doing so using a wrapper instead.

gym/spaces/graph.py
jjshoots: It appears I stole template code from an older version of Box, I'll amend things.

gym/spaces/graph.py
```diff
@@ -84,6 +19,8 @@ class Graph(Space):
         self.observation_space = spaces.Graph(node_space=space.Box(low=-100, high=100, shape=(3,)), edge_space=spaces.Discrete(3))
     """
 
+    _graph_obj_ctor = namedtuple("graph_obj", ["nodes", "edges", "edge_links"])
```
pseudo-rnd-thoughts, line 22: Normally, NamedTuples are specified in global scope like a normal class. You could call it `GraphInstance`?

gym/spaces/graph.py
```diff
@@ -114,12 +51,31 @@ def __init__(
...
-        # graph object creator
-        self.graphObj = GraphObj
-
         super().__init__(None, None, seed)
 
-    def _generate_sample_space(self, base_space: Union[None, Box, Discrete], num: int) -> Optional[Union[Box, Discrete]]:
+    @staticmethod
+    def graph_obj(
+        nodes: np.ndarray, edges: np.ndarray, edge_links: np.array
+    ) -> NamedTuple:
```
pseudo-rnd-thoughts, line 59: With the named tuple in the global scope then I don't think this function is needed.

gym/spaces/graph.py
```diff
@@ -147,8 +103,12 @@ def _sample_sample_space(self, sample_space) -> Optional[np.ndarray]:
         else:
             return None
 
-    def sample(self) -> GraphObj:
-        """Returns a random sized graph space with num_nodes between 1 and 10."""
+    def sample(self) -> NamedTuple:
+        """Generates a single sample graph with num_nodes between 1 and 10 sampled from the Graph.
+
+        Returns:
+            A NamedTuple representing a graph with attributes .nodes, .edges, and .edge_links.
```
pseudo-rnd-thoughts, line 110: I think the attributes should be in quotes not as attributes but this doesn't matter much

gym/spaces/graph.py
```diff
@@ -164,17 +124,18 @@ def sample(self) -> GraphObj:
...
+        if sampled_edges is not None and num_edges > 0:
+            sampled_edge_links = self.np_random.integers(
+                low=0, high=num_edges, size=(num_edges, 2)
+            )
 
-        return GraphObj(sampled_nodes, sampled_edges, sampled_edge_links)
+        return Graph.graph_obj(sampled_nodes, sampled_edges, sampled_edge_links)
 
-    def contains(self, x: GraphObj) -> bool:
+    def contains(self, x: NamedTuple) -> bool:
```
pseudo-rnd-thoughts, line 134: The type can be the named tuple object

gym/spaces/graph.py
```diff
@@ -164,17 +124,18 @@ def sample(self) -> GraphObj:
...
         """Return boolean specifying if x is a valid member of this space."""
-        if not isinstance(x, GraphObj):
+        if not isinstance(x, Graph._graph_obj_ctor):
+            print(type(x))
```
pseudo-rnd-thoughts, line 137: I think this is left over from testing

gym/spaces/graph.py
```diff
@@ -0,0 +1,216 @@
...
+            edge_space (Union[None, Box, Discrete]): space of the node features.
+            seed: Optionally, you can use this argument to seed the RNG that is used to sample from the space.
+        """
+        assert isinstance(
+            node_space, (Box, Discrete)
+        ), "Values of the node_space should be instances of Box or Discrete"
+        if edge_space is not None:
+            assert isinstance(
+                edge_space, (Box, Discrete)
+            ), "Values of the edge_space should be instances of Box or Discrete"
```
pseudo-rnd-thoughts, line 64: A repeat of the debug comment above

gym/spaces/graph.py
```diff
@@ -0,0 +1,216 @@
...
+            sampled_edge_links = self.np_random.integers(
+                low=0, high=num_edges, size=(num_edges, 2)
+            )
+
+        return GraphInstance(sampled_nodes, sampled_edges, sampled_edge_links)
+
+    def contains(self, x: GraphInstance) -> bool:
+        """Return boolean specifying if x is a valid member of this space."""
+        if not isinstance(x, GraphInstance):
+            print("fail 1")
```
pseudo-rnd-thoughts, line 132: Remove prints, repeat for all in `contains`

gym/spaces/utils.py
```diff
@@ -118,6 +131,24 @@ def _flatten_dict(space, x) -> np.ndarray:
     return np.concatenate([flatten(s, x[key]) for key, s in space.spaces.items()])
 
 
+@flatten.register(Graph)
+def _flatten_graph(space, x) -> np.ndarray:
+    def _graph_unflatten(space, x):
+        ret = None
+        if space is not None and x is not None:
+            if isinstance(space, Box):
+                ret = x.reshape(x.shape[0], -1)
```
pseudo-rnd-thoughts, line 140: Can we call `flatten(x, space)` for both Box and Discrete here? Just in case the specifications of Box and Discrete change

gym/spaces/utils.py
```diff
@@ -181,14 +212,34 @@ def _unflatten_dict(space: Dict, x: np.ndarray) -> dict:
     )
 
 
+@unflatten.register(Graph)
+def _unflatten_graph(space: Graph, x: GraphInstance) -> GraphInstance:
+    def _graph_unflatten(space, x):
+        ret = None
+        if space is not None and x is not None:
+            if isinstance(space, Box):
+                ret = x.reshape(-1, *space.shape)
```
pseudo-rnd-thoughts, line 221: Same comment just to call `unflatten(space, x)`

pseudo-rnd-thoughts (approved):
Overall looks good to merge with the comments addressed

gym/spaces/utils.py
```diff
@@ -181,14 +219,40 @@ def _unflatten_dict(space: Dict, x: np.ndarray) -> dict:
     )
 
 
+@unflatten.register(Graph)
+def _unflatten_graph(space: Graph, x: GraphInstance) -> GraphInstance:
+    """We're not using `.unflatten()` here because a graph is not a homogeneous space.
```
pseudo-rnd-thoughts, line 224: "We're not using `.unflatten()` for :class:`Box` and :class:`Space`"

gym/spaces/utils.py
```diff
@@ -118,6 +136,26 @@ def _flatten_dict(space, x) -> np.ndarray:
     return np.concatenate([flatten(s, x[key]) for key, s in space.spaces.items()])
 
 
+@flatten.register(Graph)
+def _flatten_graph(space, x) -> np.ndarray:
+    """We're not using `.flatten()` here because a graph is not a homogeneous space, see `.flatten` docstring."""
```
pseudo-rnd-thoughts, line 141: "We're not using `.flatten()` for :class:`Box` and :class:`Space`"

PR #2879, opened 2022-06-10
Env Step Check allow rewards to be np int

### LINKED ISSUE #2878: [Bug Report] Environment Step Check doesn't allow NumPy integer rewards.
**Describe the bug**
Some environments return the rewards in the form of a _NumPy integer_ (ex. `np.int32`).  However, the _passive environment step check_ doesn't allow NumPy integer rewards, as it allows only Python built-in `float` & `int` datatypes and `np.float32`.

**Reproducing Bug**
To reproduce the issue, create an environment where the `step()` would return `rewards` in NumPy integer datatype (ex. `np.int32`).

**System Info**
 * OS: `Windows 10`
 * Python: `3.7.9`
 * Gym: `0.24.1`

**Additional context**
_none_

### Checklist

ttran2:
# Description

Changed so `passive_env_step_check` will also allow NumPy datatypes that are children of `np.integer` (ex. `np.int32`, `np.int64`, ...) and `np.floating` (ex. `np.float32`, `np.float64`, ...).

The reason is that some gym environment returns the reward in various NumPy "float" and "integer" datatypes, which is **not** incorrect behavior, thus allowing them to pass the check.

Fixes #2878

## Type of change

- [ ] ~New feature (non-breaking change which adds functionality)~
- [ ] ~Breaking change (fix or feature that would cause existing functionality to not work as expected)~
- [ ] ~This change requires a documentation update~

### Screenshots

_not applicable_

# Checklist:

### DIFF
--- a/gym/utils/passive_env_checker.py
+++ b/gym/utils/passive_env_checker.py
@@ -276,7 +276,7 @@ def passive_env_step_check(env, action):
         logger.warn("Encountered inf value in observations.")
 
     assert isinstance(
-        reward, (float, int, np.float32)
+        reward, (float, int, np.floating, np.integer)
     ), "The reward returned by `step()` must be a float"
     if np.any(np.isnan(reward)):
         logger.warn("Encountered NaN value in rewards.")


pseudo-rnd-thoughts:
Thanks for the PR, that is a good point. 
Looks good to merge @jkterry1

PR #2884, opened 2022-06-12
fix: add render_mode getter to Wrapper

younik:
Currently, wrapped environments don't forward correctly the attribute `render_mode`.
This PR fixes this; thanks @Markus28 for spotting this


pseudo-rnd-thoughts (approved):
LGTM

Issue #2889, opened 2022-06-14
[Bug] env.render() giving no/wrong results


wookayin:
**Describe the bug**

After the merge of new render API #2671 which is not stable yet, `env.render()` giving no results when it is called more than once. Also the rendering result is not correct in terms of type and shape. Please see the example below.

**Code example**
```python
env = gym.make("HalfCheetah-v4", render_mode='rgb_array')
env.reset()
im = env.render(width=200, height=200)
im2 = env.render()
```

I see three problems here:

* `im` contains a **python list** of [480x480x3] ndarray: `im[0].shape == (480, 480, 3)`. In our early version of gym, `im` itself was a ndarray. I don't think that rendered images contained in a list is correct; and the behavior is different.

* The width/height parameter `env.render(width=200, height=200)` are ignored; the rendering dimensions are wong.

* The next call of `env.render()` will give no results: it returns an empty list, i.e. `im2 == []`

**System Info**

- gym: latest HEAD 053ee809324861dc1bec19feee97b47ba1557a94
- installed from source 
- macOS 
- python 3.10

**Additional context**

/cc @younik who is the author of #2671.

- [x] I have checked that there is no similar [issue](https://github.com/openai/gym/issues) in the repo (**required**)

younik:
Hello,

> im contains a python list of [480x480x3] ndarray: im[0].shape == (480, 480, 3). In our early version of gym, im itself was a ndarray. I don't think that rendered images contained in a list is correct; and the behavior is different.

This is done on purpose; with the new render API, the method `render` returns a collection of frames. You can do some steps without calling render to see this. 

> The width/height parameter env.render(width=200, height=200) are ignored; the rendering dimensions are wong.

When you use the new render API, those arguments should be passed to `__init__`. Tho, you are right, they are still not implemented in `__init__`; I will push a fix soon, thanks for spotting it.
The legacy code still works with dimensions (don't specify `render_mode` to use it).

> The next call of env.render() will give no results: it returns an empty list, i.e. im2 == []
This is the expected behaviour since the `render` call pops the frames from the collection. This is done to allow user to clean the frame collection.

wookayin:
Since this is not a bug as per #2671, I will close this issue. Let me move and continue my argument to #2671 to centralize the discussion.

PR #2891, opened 2022-06-14
Fix: add mujoco render arguments to init

younik:
Allow user to specify render arguments for Mujoco during initialization.

For example:
```
import gym

env = gym.make("Humanoid-v4", render_mode="rgb_array", height=450, width=500)
env.reset()
for _ in range(100):

im = env.render()
env.close()

> im[0].shape
(450, 500, 3)
```

Thanks @wookayin for spotting this (https://github.com/openai/gym/issues/2889)

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -359,6 +359,11 @@ class Wrapper(Env[ObsType, ActType]):
         self._metadata = value
 
     @property
+    def render_mode(self) -> Optional[str]:
+        """Returns the environment render_mode."""
+        return self.env.render_mode
+
+    @property
     def np_random(self) -> RandomNumberGenerator:
         """Returns the environment np_random."""
         return self.env.np_random
--- a/gym/envs/mujoco/ant.py
+++ b/gym/envs/mujoco/ant.py
@@ -1,5 +1,3 @@
-from typing import Optional
-
 import numpy as np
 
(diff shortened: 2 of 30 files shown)


wookayin:
Thanks for initiating the PR. However I am opposed to specifying all the rendering-based arguments to the constructor. Not only the name is confusing, but users would also want to pass those arguments for customizing rendering (also because `.render()` takes kwargs). Use of `render_kwargs` would be another alternative but it doesn't sound pretty. Why don't we simply allow `env.render(..., width=, height=, camera_id=..., etc.)` as before?

younik:
> also because `.render()` takes kwargs

This will not be true anymore with gym 1.0

> Why don't we simply allow `env.render(..., width=, height=, camera_id=..., etc.)` as before?

Because the frames are computed before the `render` call; you should call `render` just to collect frames

> Use of `render_kwargs` would be another alternative but it doesn't sound pretty. 

Do you mean something like this?
```python
render_kwargs = {'height'=450, 'width'=500}
env = gym.make("Humanoid-v4", render_mode="rgb_array", render_kwargs=render_kwargs)
```

I don't have a strong preference for this.

pseudo-rnd-thoughts:
With type hints, looks good to merge

PR #2893, opened 2022-06-14
fix: allow mode as argument

younik:
# Description
After https://github.com/openai/gym/pull/2706, render with Wrappers doesn't work anymore with mode as argument without keyword.
For instance, the following will throw an error:
```
env = gym.make("FrozenLake-v1")
env.reset()
env.render('rgb_array')
```
This should fix this.

## Type of change

# Checklist:

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -387,7 +387,7 @@ class Wrapper(Env[ObsType, ActType]):
         return self.env.reset(**kwargs)
 
     def render(self, *args, **kwargs):
-        """Renders the environment with kwargs."""
+        """Renders the environment."""
         return self.env.render(*args, **kwargs)
 
     def close(self):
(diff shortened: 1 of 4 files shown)


gym/core.py
```diff
@@ -43,7 +43,7 @@ def _deprecate_mode(render_func):  # type: ignore
         def render(
             self: object, *args: Tuple[Any], **kwargs: Dict[str, Any]
         ) -> render_return:
-            if "mode" in kwargs.keys():
+            if "mode" in kwargs.keys() or len(args) > 0:
```
pseudo-rnd-thoughts, line 46: Should any `kwargs` or `args` be passed to render?

pseudo-rnd-thoughts: Why do we not check `if len(kwargs) > 0 or len(args) > 0:`?

younik: Oh, I see what you mean.

For example, right now, no deprecation warning with 
```
env.render(width=400, height=400)
```

This is more related to the new render API; do we want to forbid any type of kwarg for the render method?

younik: For the environment that I worked on, no kwarg to render is allowable. Nevertheless, for other environments that generate rendering at the end, it can be useful to allow it. 
I would let this flexibility

PR #2897, opened 2022-06-17
Call `close()` upon destruction

Markus28:
@pseudo-rnd-thoughts @RedTachyon Does anybody know why we are not doing this already?

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -273,6 +273,10 @@ class Env(Generic[ObsType, ActType]):
         # propagate exception
         return False
 
+    def __del__(self):
+        """The environment's destructor."""
+        self.close()
+
 
 class Wrapper(Env[ObsType, ActType]):
     """Wraps an environment to allow a modular transformation of the :meth:`step` and :meth:`reset` methods.


RedTachyon:
So funny thing, I'm actually doing this in [my own code](https://github.com/RedTachyon/coltra-rl/blob/master/coltra/envs/base_env.py#L249), and it *almost* works. The "almost" being fine for me, but probably not for the official library, because what happens is that in certain cases I get an exception at the very end of the code.

Basically what might happen is that you manually call `env.close()` like a good resource-managing programmer should. This frees up some resources, closes some connection to an external process (e.g. with Unity), all is good.

Then the environment goes out of scope, or gets deleted at the end of the runtime, or maybe a new environment overwrites its only variable. The previous env gets destroyed, so it calls `env.close()` again. It tries to free up resources again, but they don't exist, and if the `close` code doesn't check for that, it might cause an exception.

In my case, I do this because I *really* don't want to ever forget to free up resources to avoid memory leaks, so I'm fine with an ugly error message at the end, and I didn't bother to fix it in a sensible way just yet. It's possible we'd just need something like a `self._closed` field and add a condition.

So overall I think it's a good idea, but needs to be approached carefully, because there are surprising potential pitfalls.

One issue is that among the existing built-in gym envs, we don't really have many situations with a nontrivial `env.close()`, so it will be hard to test for. The closest we have (I think) is Atari, which isn't really built-in anyways, but it does rely on an external process which presumably needs to be closed afterwards.

pseudo-rnd-thoughts:
@RedTachyon Would we want an extra check, like `if self._already_closed` before running it then to avoid issues like you mention?

Markus28:
Thanks @RedTachyon ! I guess we would need some kind of API change if we wanted to implement this, right? Either we make people call `super().close()` and we add a `_already_closed` attribute to `Env`, or we deprecate `close` in favor of `__del__`?
Tbh I don’t really see why we need `close` at all when we could just use garbage collection. People could still manually free resources via `del env`.

pseudo-rnd-thoughts:
Given the discussion on tuesday @Markus28 @RedTachyon What is the happen for this PR? Do we want to try and get this PR done for the next release as it is primarily an API release but it is too complex for us currently?

XuehaiPan:
> I guess we would need some kind of API change if we wanted to implement this, right? Either we make people call `super().close()` and we add a `_already_closed` attribute to `Env` like @pseudo-rnd-thoughts said, or we deprecate `close` in favor of `__del__`?

It is impossible to force all subclasses to call `super().close()` specially for many third-party environments. The only way is to use `metaclass` to modify `close()` when creating the subclass. Users can migrate this transparently.

Markus28:
If we tell people to call the super methods and they don't do it, they either get away with it, which would be fine, or something breaks and they eventually change their code to call the super method. 

Metaclasses have come up in the discussion but it is my understanding that we don't want to go that route in order to keep the code simple to understand.

pseudo-rnd-thoughts:
My stupid though possible plan is in the `reset` function add some magic code that will wrapper the close function if it already exists with a call to the super close function. However this is a bit of a mad idea that we shouldn't do unless we have no good ideas

XuehaiPan:
> My stupid though possible plan is in the reset function add some magic code that will wrapper the close function if it already exists with a call to the super close function.

Why not move this to `OrderEnforcing`? Call `close()` in `OrderEnforcing.__del__()` rather than `Env.__del__()`.

Markus28:
In general, I think I like the idea. However, any wrapper that is applied on top of `OrderEnforcing` would not be closed, right? I guess that might be acceptable as long as we tell everyone to keep using `close`.

pseudo-rnd-thoughts:
@Markus28 Do we have a plan with this? For the new Phys2D environment, Im going to implement this idea which should hopefully replace the numpy environments but I don't think we have a good plan for the rest of the environments

YouJiacheng:
python `__del__` is evil, actually. The time `__del__` called is not guaranteed unless there isn't cyclic reference.
And `__del__` might cause reentrancy which could break users code.
Let envs support context manager protocol is a better idea IMO.

pseudo-rnd-thoughts:
Thanks for the suggestion, interestingly that is already implemented in `gym.Env`. 
I believe we were considering this not for users to call `del env` but just when the object goes out of scope, the environment, in particular, any rendering api is correctly closed.

YouJiacheng:
> when the object goes out of scope

Unfortunately this is not true in general and in practice.
There are lots of cyclic reference in `gym` envs. For example, `self.renderer` hold a `render_frame` function, which is a `partial` of bounded method `self._render`.
In that case, `__del__` will be called after `gc` (note that cpython gc, even full gc, doesn't guarantee that all garbage will be collected) collect the garbage.
even worse, `__del__` won't be called on cpython vm shutdown.

RedTachyon:
Yea, all these caveats are more or less why this is stuck in a limbo for now. Some extra convenience of automatic closing would be nice, but it has to be done in a robust and reliable way, so without that, it might be better to stick with requiring people to call `env.close()` manually.

RedTachyon:
For future reference (in case we start thinking about porting this to gymnasium), I have convinced myself that `__del__` is, indeed, the spawn of Satan himself. Here's a [nice summary](https://www.youtube.com/watch?v=IFjuQmlwXgU). 

Overall I think we can stick with status quo, i.e. users are adults and can be trusted to call `env.close()` when appropriate. Or we can start encouraging them to do `with gym.make("WhateverEnv-v0") as env:` which feels super fancy and proper, and should actually be the best practice. But I'm in no position to moralize about that since I haven't actually used this pattern in any serious situation yet.

jkterry1:

PR #2898, opened 2022-06-17
Pin pytest to 7.0.1

pseudo-rnd-thoughts:
Currently pytest is not pinned to a version, this PR sets the pytest version to the last available version for python 3.6, 7.0.1
https://docs.pytest.org/en/7.1.x/changelog.html#pytest-7-1-0-2022-03-13

This is an issue as the latest version of pytest does not support py 3.6 meaning that the CI will be using different pytest versions for the different versions


RedTachyon:
@pseudo-rnd-thoughts Can you explain why `mock` was removed in this PR?

pseudo-rnd-thoughts:
It was not unused in any of the tests, or at least the CI passes without it. I think the unittest mock is being instead

PR #2899, opened 2022-06-17
Fixed action dim check

pseudo-rnd-thoughts:
I noticed that the action-dims checks where being skipped due to parameterize list being empty
This code should be the same, but the list is no longer empty
I think `pytest.mark.parameterize` doesn't work well with generators


tests/envs/test_action_dim_check.py
```diff
@@ -48,12 +48,16 @@ def test_mujoco_action_dimensions(env_spec: EnvSpec):
     env.close()
 
 
-@pytest.mark.parametrize(
-    "env",
+DISCRETE_ENVS = list(
     filter(
         lambda env: isinstance(env.action_space, spaces.Discrete),
```
RedTachyon, line 53: @pseudo-rnd-thoughts So, uh, this has been merged before I got a chance to look at it, but a more pythonic/readable/better way would probably be to use a list comprehension, so instead of 
`list(filter(lambda env: isinstance(env.action_space, spaces.Discrete), all_testing_initialised_envs))`
it would be
`[env for env in all_testing_initialised_envs if isinstance(env.action_space, spaces.Discrete)]`

pseudo-rnd-thoughts: That is a good point, sorry I sent it to Jordan as well because I thought the code was mostly good. I will see if I can fix this

PR #2900, opened 2022-06-17
Split test registration

pseudo-rnd-thoughts:
When writing new tests for make, I found it difficult to understand the different tests as `test_registration.py` contained tests for three interconnected but separate functions, `make`, `spec` and `register`. 
This PR splits these tests into three files for ease of understanding what is happening
All tests have been copied from `test_registration.py` with a couple of new tests for `make`, time limit, order enforcing and passive environment checker parameters


tests/envs/test_make.py
```diff
@@ -0,0 +1,180 @@
...
+def test_make():
+    env = gym.make("CartPole-v1", disable_env_checker=True)
+    assert env.spec.id == "CartPole-v1"
+    assert isinstance(env.unwrapped, cartpole.CartPoleEnv)
+    env.close()
+
+
+def test_make_deprecated():
+    with pytest.raises(
+        gym.error.Error,
```
RedTachyon, line 45: Are we able to make the error more specific? Or does it just raise a plain `gym.error.Error`?

tests/envs/test_make.py
```diff
@@ -0,0 +1,180 @@
...
+    env.close()
+
+    # Custom max episode steps
+    env = gym.make("CartPole-v1", max_episode_steps=100, disable_env_checker=True)
+    assert has_wrapper(env, TimeLimit)
+    assert env.spec.max_episode_steps == 100
+    env.close()
+
+    # Env spec has no max episode steps
+    assert gym.envs.registry["test.ArgumentEnv-v0"].max_episode_steps is None
```
RedTachyon, line 69: A more robust way to access the env spec might be `gym.spec`, it's designed for this function basically. Just in case the internals of registration change again.

tests/envs/test_make.py
RedTachyon: (same applies to other instances of this pattern)

pseudo-rnd-thoughts: Yes, I can look into adding the match statement

tests/envs/test_make.py
pseudo-rnd-thoughts: Ahh, I hadn't realised that. I will replace it

tests/envs/test_make.py
```diff
@@ -0,0 +1,180 @@
...
+    action_space = gym.spaces.Box(low=0, high=1, shape=(1,))
+
+    def __init__(self, arg1, arg2, arg3):
+        self.arg1 = arg1
+        self.arg2 = arg2
+        self.arg3 = arg3
+
+
+gym.register(
+    id="test.ArgumentEnv-v0",
```
Markus28, line 27: I've never seen this kind of id. Wouldn't we want something like "test/ArgumentEnv-v0"? Probably not in the scope of this PR though.

tests/envs/test_register.py
```diff
@@ -0,0 +1,182 @@
...
+    with pytest.raises(gym.error.RegistrationError):
+        gym.register(versioned_env)
+
+    # Clean everything
+    del gym.envs.registry[unversioned_env]
+
+
+def test_make_latest_versioned_env(register_testing_envs):
+    with pytest.warns(UserWarning, match="Using the latest versioned environment"):
+        env = gym.make("MyAwesomeNamespace/MyAwesomeVersionedEnv")
```
Markus28, line 169: The original code had `disable_env_checker=True`. Why did that go?

Markus28 (approved):
I didn't check the code in detail but all tests are there. When they were modified they were apparently strengthened. I made two comments but otherwise LGTM

tests/envs/test_register.py
pseudo-rnd-thoughts: Fixed

tests/envs/test_make.py
pseudo-rnd-thoughts: That is a good point, @RedTachyon what are your thoughts?

tests/envs/test_make.py
pseudo-rnd-thoughts: Done

PR #2896, opened 2022-06-16
Pitch: Check render_mode in make

younik:
# Description

Currently, every environment needs to check the `render_mode` causing repetitions of code. This PR adds this check inside `make` along with an explanatory error message.

- [x] New feature (non-breaking change which adds functionality)

# Checklist:

- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `pre-commit run --all-files` (see `CONTRIBUTING.md` instructions to set it up)
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally with my changes

PR #2904, opened 2022-06-17
Mujoco metadata

pseudo-rnd-thoughts:
In https://github.com/openai/gym/pull/2896 there is a check that the environment render mode is correct
However, in testing, this was found to be an issue for mujoco as the values are not statically declared but are rather done at runtime in `mujoco_env.py`
Therefore this PR moves the metadata information to each of the individual environments, along with a check that the metadata is correct in `mujoco_env.py`


YouJiacheng:
I think it is better to have `render_mode` declared in base class while `render_fps` updated during runtime to avoid code repetition and risk of inconsistency.
```python
class BaseMujocoEnv(gym.Env):
    metadata = {
        "render_modes": [
            "human",
            "rgb_array",
            "depth_array",
            "single_rgb_array",
            "single_depth_array",
        ]
    }
    def __init__(...):
        ...
        self.metadata = {**BaseMujocoEnv.metadata, "render_fps": int(np.round(1.0 / self.dt))}

```

Issue #2823, opened 2022-05-16
[Proposal] Invalid action masking


vwxyzjn:
### Motivation

It's common for games to have invalid _discrete_ actions (e.g. walking into a wall). The status quo is to create a `gym.spaces.Discrete` action space that contains both valid actions and invalid actions, and if an invalid action is executed, the gym environment/game engine ignores the invalid actions. 

This is less ideal for RL libraries because using action masks (i.e., only sampling valid actions) could significantly speed up training. See [A Closer Look at Invalid Action Masking in Policy Gradient Algorithms](https://arxiv.org/abs/2006.14171).

### Proposal

Let's add a standard API for invalid action masking. This issue kick starts the discussion. The simplest API I can think of is just adding

```python
class Discrete(Space[int]):
    r"""A space consisting of finitely many elements.
    This class represents a finite subset of integers, more specifically a set of the form :math:`\{ a, a+1, \dots, a+n-1 \}`.
    Example::
        >>> Discrete(2)            # {0, 1}
        >>> Discrete(3, start=-1)  # {-1, 0, 1}
    """

    def __init__(
        self,
        n: int,
        seed: Optional[int | seeding.RandomNumberGenerator] = None,
        start: int = 0,
    ):
        # ...

    def action_mask(self) -> Sequence[int]:
        pass
```

### Additional context

* SB3 has implemented invalid action masking via a wrapper (i.e., `env = ActionMasker(env, mask_fn)`). See https://github.com/Stable-Baselines-Team/stable-baselines3-contrib/pull/25 and [docs](https://sb3-contrib.readthedocs.io/en/master/modules/ppo_mask.html#example)
* Griddly does this via a wrapper ([docs](https://griddly.readthedocs.io/en/latest/getting-started/action%20spaces/index.html#sampling-valid-actions))
* Gym-microrts implemented it with a class method (see [code](https://github.com/vwxyzjn/gym-microrts/blob/bfa6e8d922c27d7f9e90f4ff94821c9cbe8c14f8/gym_microrts/envs/vec_env.py#L247))

### Alternative design

We could also make the observations a dicttionary obs space and include a key called action_mask, which is inlined with practice done here https://www.pettingzoo.ml/classic/go. This approach requires no change to the main API but only a recommendation from gym's website.

### Example

`Taxi-v2` (https://www.gymlibrary.ml/environments/toy_text/taxi/) could be a straightforward example. Its action space has 6 discrete actions:

0: move south
1: move north
2: move east
3: move west
[...]

PR #2906, opened 2022-06-17
Added Action masking for Space.sample()

pseudo-rnd-thoughts:
Adds action masking as requested in https://github.com/openai/gym/issues/2823 to allow spaces to mask certain actions. These masks are the positive case where `1` means that it is possible for the action to be taken and `0` for the action to not be possible. For all of the gym environments, this PR adds a parameter in `sample(mask=...)` with the particular type required being dependent on the space. `Box` is a special case where we don't implement masking due to the neural network not being able to provide values for continuous distributions, however, if a good reason is found, this could be added. 
To the gym Taxi environment, we add a new info key "action_mask" which is the recommended method for using the masking for custom environments. 

### Example masks
```python
>>> import numpy as np
>>> from gym import spaces

# Box space doesn't have masks

>>> space = spaces.Discrete(4)
>>> space.sample(mask=np.array([0, 1, 1, 1], dtype=np.int8)) 
2
>>> space.sample(mask=np.array([0, 0, 0, 0], dtype=np.int8)) 
0

>>> space = spaces.MultiDiscrete([4, 2])
>>> space.sample(mask=(np.array([0, 1, 0, 1], dtype=np.int8), np.array([0, 0], dtype=np.int8))) 
[1 0]

>>> space = spaces.MultiDiscrete(np.array([[4, 2], [3, 4]]))
>>> space.sample(mask=((np.array([1, 1, 1, 1], dtype=np.int8), np.array([0, 1], dtype=np.int8)), (np.array([0, 0, 0], dtype=np.int8), np.array([1, 1, 0, 0], dtype=np.int8))))  
[[2 1]
 [0 1]]

>>> space = spaces.MultiBinary([2, 3])
>>> space.sample(mask=np.array([[0, 0, 1], [1, 1, 0]], dtype=np.int8))
[[0 0 0]
 [1 1 0]]

# Composite spaces (Dict, Tuple and Graph)
>>> space = spaces.Dict(a=spaces.Discrete(3), b=spaces.Box(0, 1, (1,)))
>>> space.sample(mask={"a": np.array([0, 1, 1], dtype=np.int8), "b": None}))
OrderedDict([('a', 1), ('b', array([0.6812336], dtype=float32))])

>>> space = spaces.Tuple((spaces.Box(0, 1, (1,)), spaces.Discrete(3)))
>>> space.sample(mask=(None, np.array([0, 0, 0], dtype=np.int8)))  
(array([0.74909943], dtype=float32), 0)

>>> space = spaces.Graph(node_space=spaces.Box(0, 1, (1,)), edge_space=spaces.Discrete(3))
>>> space.sample(mask=(None, np.array([0, 1, 1], dtype=np.int8)), num_nodes=4)) 
GraphInstance(nodes=array([[0.5791068 ], [0.43347424], [0.6848027 ], [0.23124644]], dtype=float32), edges=array([2, 2]), edge_links=array([[1, 0], [2, 1]]))
>>> space.sample(mask=(None, (np.array([1, 1, 1], dtype=np.int8), np.array([1, 0, 0], dtype=np.int8), np.array([0, 1, 0], dtype=np.int8), np.array([0, 1, 1], dtype=np.int8))), num_nodes=4, num_edges=4))  
GraphInstance(nodes=array([[0.21213211], [0.4872798 ], [0.69442934], [0.92085034]], dtype=float32), edges=array([2, 0, 1, 2]), edge_links=array([[0, 2], [0, 0],  [3, 2], [2, 0]]))
```

- [x] Add tests for sample mask
- [x] Add tests for sample with discrete distributions (Discrete, MultiDiscrete, MultiBinary)
- [ ] Add test for Box sample
- [x] Add docstrings 
- [x] Add taxi docstrings
- [x] Fix taxi action mask


gym/spaces/multi_discrete.py
```diff
@@ -63,8 +63,22 @@ def shape(self) -> Tuple[int, ...]:
...
 
-    def sample(self) -> np.ndarray:
+    def sample(self, mask: np.ndarray = None) -> np.ndarray:
         """Generates a single random sample this space."""
+        if mask is not None:
+            assert isinstance(mask, np.ndarray)
+            assert mask.dtype == np.int8
+            assert mask.shape == self.shape
+
+            multi_mask = [np.where(row) for row in mask]
```
vwxyzjn, line 73: This is a good first-pass implementation so I think it's fine here, but do note that its performance might be throttled when we have a large MultiDiscrete action space. See https://github.com/vwxyzjn/gym-microrts/pull/71/files

vwxyzjn (commented):
A great first pass! Thanks for adding this feature!

gym/core.py
```diff
@@ -231,7 +231,7 @@ def seed(self, seed=None):
             there aren't accidental correlations between multiple generators.
 
         Args:
-            seed(Optional int): The seed value for the random number geneartor
+            seed(Optional int): The seed value for the random number generator
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: Yep

gym/spaces/box.py
```diff
@@ -160,6 +160,9 @@ def sample(self) -> np.ndarray:
         Returns:
             A sampled value from the Box
         """
+        if mask is not None:
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: The problem, I was thinking about composite spaces but you are right, changing it to raising an error

gym/spaces/multi_discrete.py
pseudo-rnd-thoughts: I was thinking about that but I couldn't work out a way of doing it efficiently
You can't use `choice` as you want it for each row individually

gym/spaces/discrete.py
```diff
@@ -40,14 +40,23 @@ def __init__(
         self.start = int(start)
         super().__init__((), np.int64, seed)
 
-    def sample(self) -> int:
+    def sample(self, mask: np.ndarray = None) -> int:
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: Done

gym/envs/toy_text/taxi.py
```diff
@@ -214,14 +214,33 @@ def decode(self, i):
...
         i = categorical_sample([t[0] for t in transitions], self.np_random)
         p, s, r, d = transitions[i]
         self.s = s
         self.lastaction = a
         self.renderer.render_step()
-        return (int(s), r, d, {"prob": p})
+
+        taxi_row, taxi_col, pass_loc, dest_idx = self.decode(s)
+        mask = self.valid_mask(taxi_row, taxi_col, pass_loc, dest_idx, 4)
+        return int(s), r, d, {"prob": p, "action_mask": mask}
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: Very good question, from the code it represents the probability of the transition. However this value is permanently 1, so it makes zero sense. I will update the top docstring with a info section

gym/spaces/box.py
```diff
@@ -146,7 +146,7 @@ def is_bounded(self, manner: str = "both") -> bool:
         else:
             raise ValueError("manner is not in {'below', 'above', 'both'}")
 
-    def sample(self) -> np.ndarray:
+    def sample(self, mask: np.ndarray = None) -> np.ndarray:
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: Done

gym/spaces/discrete.py
```diff
@@ -40,14 +40,35 @@ def __init__(
...
 
-    def sample(self) -> int:
+    def sample(self, mask: Optional[np.ndarray] = None) -> int:
         """Generates a single random sample from this space.
 
-        A sample will be chosen uniformly at random.
+        A sample will be chosen uniformly at random with the mask if provided
+
+        Args:
+            mask: An optional mask for if an action can be selected. Expected shape is (n,). If not possible actions, will default to `space.start`
```
Markus28, line 49: We should specify that `mask` should be a binary array of dtype `np.int8`. We should also say whether `mask` specifies the legal or illegal samples. I would avoid references to actions, since this is about spaces in general.

I'm not sure what you mean by "will default to `space.start`, but I guess you mean that the *sample* will default to that, right?

gym/spaces/multi_binary.py
```diff
@@ -51,14 +51,33 @@ def shape(self) -> Tuple[int, ...]:
...
         return self._shape  # type: ignore
 
-    def sample(self) -> np.ndarray:
+    def sample(self, mask: Optional[np.ndarray] = None) -> np.ndarray:
         """Generates a single random sample from this space.
 
         A sample is drawn by independent, fair coin tosses (one toss per binary variable of the space).
 
+        Args:
+            mask: An optional np.ndarray to mask samples, where mask == 0 will have samples == 0
```
Markus28, line 60: Maybe state the shape of the mask

gym/spaces/multi_discrete.py
```diff
@@ -23,8 +24,17 @@ class MultiDiscrete(Space[np.ndarray]):
     2. Button A:   Discrete 2  - NOOP[0], Pressed[1] - params: min: 0, max: 1
     3. Button B:   Discrete 2  - NOOP[0], Pressed[1] - params: min: 0, max: 1
 
-    It can be initialized as ``MultiDiscrete([ 5, 2, 2 ])``
+    It can be initialized as ``MultiDiscrete([ 5, 2, 2 ])`` such that a sample is array([3, 1, 0])
```
Markus28, line 27:
```suggestion
    It can be initialized as ``MultiDiscrete([ 5, 2, 2 ])`` such that a sample might be ``array([3, 1, 0])``.
```

gym/spaces/space.py
```diff
@@ -81,8 +82,17 @@ def shape(self) -> Optional[Tuple[int, ...]]:
...
 
-    def sample(self) -> T_cov:
-        """Randomly sample an element of this space. Can be uniform or non-uniform sampling based on boundedness of space."""
+    def sample(self, mask: Optional[Any] = None) -> T_cov:
+        """Randomly sample an element of this space.
+
+        Can be uniform or non-uniform sampling based on boundedness of space.
+
+        Args:
+            mask: A mask used for sampling, see Space for implementation details.
```
Markus28, line 91: We should state here whether mask specifies legal or illegal samples

Markus28 (commented):
I left some comments.

I have some additional questions regarding the new `Graph` space, not directly related to this PR:
- Why do we only allow `Box` and `Discrete` spaces for node- and edge-features? I guess it's because of `_generate_sample_space` but we could easily do without that method.
- I think we should change the `sample` method to allow more flexibility (e.g. sampling from a G(n,p) or G(n, M) model, etc.)
- In `_generate_sample_space`, why would we expect the `base_space` to be None?

gym/envs/toy_text/taxi.py
```diff
@@ -87,6 +87,20 @@ class TaxiEnv(Env):
     - 2: Y(ellow)
     - 3: B(lue)
 
+    ### Info
+
+    ``step`` and ``reset(return_info=True)`` will return an info dictionary that contains "p" and "action_mask".
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: I believe that single and double backticks are identical for Sphinx though Im struggling to find any documentations on it

gym/envs/toy_text/taxi.py
```diff
@@ -87,6 +87,20 @@ class TaxiEnv(Env):
...
+
+    As Taxi is a stochastic environment for transitions then the "p" key represents the probability of the
+    transition. However, this value is permanently 1.0 for an unknown reason.
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: I was mistaked, the initial state is stochastic but the step states are deterministic

gym/spaces/dict.py
```diff
@@ -137,14 +137,28 @@ def seed(self, seed: Optional[Union[dict, int]] = None) -> list:
 
         return seeds
 
-    def sample(self) -> dict:
+    def sample(self, mask: Optional[TypingDict[str, np.ndarray]] = None) -> dict:
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: Are you correct, I will fix this

gym/spaces/discrete.py
pseudo-rnd-thoughts: Updated, could you confirm

gym/spaces/graph.py
```diff
@@ -92,18 +92,18 @@ def _generate_sample_space(
...
-            return None
-
-    def sample(self) -> NamedTuple:
+    def sample(
+        self, mask: Optional[Tuple[Optional[np.ndarray], Optional[np.ndarray]]] = None
+    ) -> NamedTuple:
         """Generates a single sample graph with num_nodes between 1 and 10 sampled from the Graph.
 
+        Args:
+            mask: An optional tuple for the node space mask and the edge space mask (only valid for Discrete spaces)
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: Agreed, this was unclear. I will update

gym/spaces/space.py
pseudo-rnd-thoughts: Will do

gym/spaces/graph.py
```diff
@@ -70,53 +70,81 @@ def __init__(
...
 
         # we only have edges when we have at least 2 nodes
-        num_edges = 0
-        if num_nodes > 1:
-            # maximal number of edges is (n*n) allowing self connections and two way is allowed
-            num_edges = self.np_random.integers(num_nodes * num_nodes)
+        if num_edges is None:
+            if num_nodes > 1:
+                # maximal number of edges is (n*n) allowing self connections and two-way is allowed
+                num_edges = self.np_random.integers(num_nodes * num_nodes)
```
jjshoots, line 125: This should be `num_nodes * (num_nodes - 1)`.

gym/spaces/graph.py
```diff
@@ -70,53 +70,81 @@ def __init__(
...
+        sampled_node_space = self._generate_sample_space(self.node_space, num_nodes)
+        sampled_edge_space = self._generate_sample_space(self.edge_space, num_edges)
 
-        sampled_nodes = self._sample_sample_space(node_sample_space)
-        sampled_edges = self._sample_sample_space(edge_sample_space)
+        sampled_nodes = (
+            sampled_node_space.sample(node_space_mask)
+            if sampled_node_space is not None
+            else None
+        )
```
jjshoots, line 142: Minor semantics, but on line 112, we asserted that `num_nodes > 0`, so realistically `sampled_node_space` cannot possibly be `None` unless something very bad happened.

gym/spaces/graph.py
pseudo-rnd-thoughts: As `integers` is exclusive (doesn't include the max number) then this should have a max value of  `num_nodes * (num_nodes - 1)`

gym/spaces/graph.py
pseudo-rnd-thoughts: Good point, I will change

gym/spaces/graph.py
```diff
@@ -70,53 +70,81 @@ def __init__(
...
+            node_space_mask, edge_space_mask = None, None
 
         # we only have edges when we have at least 2 nodes
-        num_edges = 0
-        if num_nodes > 1:
-            # maximal number of edges is (n*n) allowing self connections and two way is allowed
-            num_edges = self.np_random.integers(num_nodes * num_nodes)
+        if num_edges is None:
+            if num_nodes > 1:
+                # maximal number of edges is (n*n) allowing self connections and two-way is allowed
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: Good point, I will update

gym/spaces/graph.py
jjshoots: It's `num_nodes*(num_nodes - 1)`, not `(num_nodes * num_nodes) - 1.

The TLD(write) explanation is that we can have every node to every other node connections twice `2 * (num_nodes * (num_nodes - 1) / 2)`.

jjshoots: The comment should be changed to `n (n - 1)` as well. But then again, all of these are just very minor semantics for things to be mathematically in line. It doesn't actually affect code behaviour.

pseudo-rnd-thoughts: Your right, updated

vwxyzjn:
Hi, @PseudoRnd#6426 thanks for this PR. LGTM to be merged as preliminary support for action masking.


That said, making it work with gym-microrts' action space is tricker than I thought. Here is what the gym-microrts' action space roughly looks like

```python
import gym
import numpy as np

height = 16
width = 16
action_space_dims = [6, 4, 4, 4, 4, 7, 7 * 7]
action_plane_space = gym.spaces.MultiDiscrete(action_space_dims)
action_space = np.ones((height, width, len(action_space_dims)))
action_space[:,:,:] = action_space_dims
action_space = gym.spaces.MultiDiscrete(action_space)
print("each unit's action_space has shape", action_plane_space.shape)
print("player's action_space (controlling 256 units at the same time) has shape", action_space.shape)
mask = np.ones((16, 16, sum(action_space_dims)))
print("player's action_space mask has shape", mask.shape)

# each unit's action_space has shape (7,)
# player's action_space (controlling 256 units at the same time) has shape (16, 16, 7)
# player's action_space mask has shape (16, 16, 78)
```

Notice here the (16, 16) is just a batch dimension, which is the reason I could implement mask that way to make things more efficient. This however does mean integrating with gym's action masking implementation is more difficult. To use the current API I think I need to stack a lot of numpy object such as 

```
mask = np.array(
    np.ones(6), np.ones(4), np.ones(4), np.ones(4), np.ones(4),, np.ones(7),, np.ones(49),
    np.ones(6), np.ones(4), np.ones(4), np.ones(4), np.ones(4),, np.ones(7),, np.ones(49),
    np.ones(6), np.ones(4), np.ones(4), np.ones(4), np.ones(4),, np.ones(7),, np.ones(49),
...
)
```

PR #2911, opened 2022-06-19
Inferring Box scalar shape

pseudo-rnd-thoughts:
This PR adds several changes to box
* Changes check that `low` and `high` are instances of `np.ndarray` instead of `np.isscalar` as this function has a number of weird cases that we don't want, e.g. strings
* Adds a check on custom shape that each of the elements are `np.issubdtype(type(dim), np.integer))` that should allow a tuple of `np.int64`. However added conversion after that the shape is a tuple of integers. This is a partial change, however, reflects the expected type of `shape` and don't expect this to affect someone.
* Adds new shape inference from scalar `low` and `high` through checking if the type of `low` and `high` are sub dtypes of `np.integer` or `np.floating`. 
* Updated check on if to expand the `low` and `high` for scalars to remove `isscalar` due to the problems noted above.
* Added new tests for box values with valid and invalid types
* Added new tests for box errors with the expected inputs


RedTachyon:
Overall I like the idea, but are you sure we want the default shape to be `()`? My first intuition is that `(1,)` would be better, though I'm not sure. 

The difference is somewhat subtle, but the main issue I'm finding now is that with zero-dimensional arrays, you can't concatenate them. You can stack them, the output being different (though just by a single empty dimension)

Zero-dim arrays have this weird status kinda between being just a number and being a proper array (even though the type is `np.ndarray`), so imo it might be better to keep the defaults as a proper "full" array. And of course if someone explicitly wants to use a zero-dim array, they can do it manually.

RedTachyon:
(also, whatever the decision will be - tests)

pseudo-rnd-thoughts:
I think you are right on the shape being `(1,)` instead of `()`. I will change it and add tests

gym/spaces/box.py
```diff
@@ -25,6 +25,9 @@ def _short_repr(arr: np.ndarray) -> str:
     return str(arr)
 
 
+BOX_SHAPE_TYPE = Tuple[Union[int, "BOX_SHAPE_TYPE"], ...]
```
RedTachyon, line 28: Why do we need a recursive type? I love them, but (a) not in python, and (b) I don't think we allow shapes like (2, (3, 4), 5)

pseudo-rnd-thoughts: I was getting confused with `MultiDiscrete` where the shape is recursive and weird. Good spot

pseudo-rnd-thoughts: I have cast all of the shape dims to integers so there shouldn't be any need to the asserts anymore

gym/spaces/box.py
```diff
@@ -78,16 +78,19 @@ def __init__(
...
-        elif not np.isscalar(low):
-            shape = low.shape  # type: ignore
-        elif not np.isscalar(high):
-            shape = high.shape  # type: ignore
+            shape = tuple(int(dim) for dim in shape)
+        elif isinstance(low, np.ndarray):
+            shape = low.shape
+        elif isinstance(high, np.ndarray):
+            shape = high.shape
+        elif np.issubdtype(type(low), np.integer) and np.issubdtype(
```
RedTachyon, line 86: Shouldn't we also allow `low` and `high` to be floats/other numeric types? I suspect that in this version, `Box(low=0.5, high=1.2)` would crash, which is not what we want. I think checking `np.isscalar` for this branch was actually appropriate

tests/spaces/test_box.py
```diff
@@ -0,0 +1,20 @@
...
+# Todo, move Box unique tests from test_spaces.py to test_box.py
+
+
+@pytest.mark.parametrize(
+    "box,expected_shape",
+    [
+        (Box(low=np.zeros(2), high=np.ones(2)), (2,)),
+        (Box(low=np.zeros((2, 1)), high=np.ones((2, 1))), (2, 1)),
+        (Box(low=0, high=1, shape=(5, 2)), (5, 2)),
+        (Box(low=0, high=1), (1,)),
```
RedTachyon, line 15: As a consequence of the previous comment, add a test case where bounds are floats (and maybe (int, float), (float, int) combinations? it's a cheap test so go crazy)

gym/spaces/box.py
```diff
@@ -78,16 +78,19 @@ def __init__(
 
         # determine shape if it isn't provided directly
         if shape is not None:
-            shape = tuple(shape)
-        elif not np.isscalar(low):
-            shape = low.shape  # type: ignore
-        elif not np.isscalar(high):
-            shape = high.shape  # type: ignore
+            shape = tuple(int(dim) for dim in shape)
```
RedTachyon, line 81: Why add the explicit type casting? I think this would mainly turn exceptions into weird, 99% undesired behavior which quietly goes through. Currently if some mad lad tries to use a float in a shape, it will crash a little further down the line in `np.full` which I think would be a pretty understandable exception, so we don't need to change anything here imo

gym/spaces/box.py
```diff
@@ -78,16 +78,19 @@ def __init__(
...
+        elif isinstance(high, np.ndarray):
+            shape = high.shape
+        elif np.issubdtype(type(low), np.integer) and np.issubdtype(
+            type(high), np.integer
+        ):
+            shape = (1,)
         else:
             raise ValueError(
-                "shape must be provided or inferred from the shapes of low or high"
+                f"Box shape is inferred from low and high, expect their type to be np.ndarray or np.integer, actual type low: {type(low)}, high: {type(high)}"
```
RedTachyon, line 92: Low and high don't have to be integers, they need to be arrays or individual numbers

tests/spaces/test_box.py
pseudo-rnd-thoughts: Your right, I will add a couple of tests

gym/spaces/box.py
pseudo-rnd-thoughts: Agreed, I have just added an assert that all dims are `np.integers` so the user should get a clean, more understandable error message now

RedTachyon:
Looks good now

PR #2920, opened 2022-06-23
tentative fix for render function showing changing dealer top card

### LINKED ISSUE #2919: [Bug Report] Small bug in blackjack rendering
**Describe the bug**
There is a bug in blackjack rendering where the suit of the displayed card from the dealer is re-randomized on each call to render, and if the dealer's displayed card is a face card, the face card is re-randomized on each call to render. The underlying data structure represents each face card as 10, which means that render has to make up which face card the dealer has drawn. I will be adding an attached PR with a proposed fix, so that the suit/face card remains stable across calls to render. 

**Code example**
````
import time
import gym 

seed = int(time.time())
env = gym.make("Blackjack-v1")
obs = env.reset()
action = 1
for i in range(6):
    env.render()
    input()
    obs, reward, done, info = env.step(action)

````
face-card

non-face-card

**System Info**
this shouldn't be relevant, but python3.8 on Ubuntu 20.04

**Additional context**

### Checklist

### PR DESCRIPTION
**Pull request by balisujohn**
# Description

Changes the _render function and reset function of blackjack.py so that the suit and face card value of the dealer's top card remain stable across calls to render. All side effects moved to reset so render function should not have any side-effects.

Fixes  #2919

## Type of change

### Screenshots

Before
before
After
after

# Checklist:

### DIFF
--- a/gym/envs/toy_text/blackjack.py
+++ b/gym/envs/toy_text/blackjack.py
@@ -175,8 +175,17 @@ class BlackjackEnv(gym.Env):
         self.dealer = draw_hand(self.np_random)
         self.player = draw_hand(self.np_random)
 
-        self.dealer_top_card_suit = None
-        self.dealer_top_card_value_str = None
+        _, dealer_card_value, _ = self._get_obs()
+
+        suits = ["C", "D", "H", "S"]
+        self.dealer_top_card_suit = self.np_random.choice(suits)
+
+        if dealer_card_value == 1:
+            self.dealer_top_card_value_str = "A"
+        elif dealer_card_value == 10:
+            self.dealer_top_card_value_str = self.np_random.choice(["J", "Q", "K"])
+        else:
+            self.dealer_top_card_value_str = str(dealer_card_value)
 
         self.renderer.reset()
         self.renderer.render_step()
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: gym/envs/toy_text/blackjack.py [hunk 1]
```diff
@@ -240,22 +252,15 @@ def get_font(path, size):
...
-            dealer_card_value_str = str(dealer_card_value)
-
         def scale_card_img(card_img):
             return pygame.transform.scale(card_img, (card_img_width, card_img_height))
 
         dealer_card_img = scale_card_img(
             get_image(
-                os.path.join("img", dealer_card_suit + dealer_card_value_str + ".png")
+                os.path.join(
+                    "img",
```
pseudo-rnd-thoughts, line 261: Can this be done with a smart string format

pseudo-rnd-thoughts (approved):
With the string format, otherwise looks good

gym/envs/toy_text/blackjack.py
RedTachyon: `os.path.join` is generally a better option than string interpolation methods when it comes to manipulating paths. In principle we only support Unix, but this would have a high chance of breaking Windows support.

A more elegant way would be using pathlib, but it might be overkill here

pseudo-rnd-thoughts: Agreed, I more meant `"{self.dealer_top_card_suit}{self.dealer_top_card_value_str }.png"`. You are right on `os.path.join`

balisujohn: fixed

PR #2927, opened 2022-06-24
Fix: warning for mode in render with python 3.6

younik:
Previously, the warning for the mode argument of `render` was not showing for a problem with Python 3.6. This should fix this.


gym/core.py
```diff
@@ -65,7 +65,10 @@ def render(
 
 
 decorator = _EnvDecorator
-if sys.version_info[0:2] == (3, 6):  # needed for https://github.com/python/typing/issues/449
+if sys.version_info[0:2] == (
+    3,
+    6,
+):  # needed for https://github.com/python/typing/issues/449
```
pseudo-rnd-thoughts, line 71: If you move the comment to a new line, the if statement can be on a single line

younik: yeah, better

pseudo-rnd-thoughts:
LGTM

PR #2928, opened 2022-06-26
Jax Blackjack Environment

balisujohn:
# Description

This includes a blackjack environment which relies on just in time compilable functions for the step and reset logic, which can also offer the standard gym API. The plan is that this is the first of a series of Jax implementations for all of the toytext environments. This was heavily inspired by https://medium.com/@ngoodger_7766/writing-an-rl-environment-in-jax-9f74338898ba but for now the autoreset capabilities are removed. This may need to be edited for compliance with whatever finalized api we decide on for jax environments. I temporarilly added a main function to jax_blackjack.py so you can test out the environment for yourself easily by calling jax_blackjack.py

## Notes For Reviewers
Here is a non-exhaustive list of areas which I think need reviewer attention:

### `self.state`
Right now all non-rendering related state is contained in `self.state` so for all non-rendering purposes, you can set state by assigning `self.state` to a new state. `self.step` and `self.reset` use `self.state` to give inputs to and to record the output of the side-effect free `self.jit_reset` and `self.jit_step` functions.  

### Pseudorandom

~~Currently, this uses Jax pseudorandom for all non-rendering related side-effects, there is a warning because seed defaults to zero instead of `None`. Also, the render state initialization to decide on dealer top card suit/face card value used in the reset function uses the standard  np_random. Curious if this seems acceptable, or what would be best here.~~

Should behave correctly now.

### Profling
No profiling yet to confirm that this is faster than normal blackjack after JITing. Profiling report will go here, probably will require jit_autoreset to make a fair comparison. 

### Autoreset

We will probably want to implement a JIT autoreset so we can use this with a compiled for loop for collecting rollouts, like is used here: https://medium.com/@ngoodger_7766/writing-an-rl-environment-in-jax-9f74338898ba
If we do that, the last obs will needs to be returned somehow.

## Type of change

# Checklist:

### DIFF
--- a/gym/envs/jax_toy_text/jax_blackjack.py
+++ b/gym/envs/jax_toy_text/jax_blackjack.py
@@ -1,0 +1,452 @@
+
+from functools import partial
+import math
+import os
+import time
+
+import jax.numpy as jnp
+import jax
+from jax import random, jit, pmap, vmap
+import numpy as np
+
+
+import gym
+from gym import spaces
+
+from gym.utils.renderer import Renderer
+from gym.error import DependencyNotInstalled
+
+
+
+deck = jnp.array([1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 10, 10, 10])
+
+
+#converts a tuple of device arrays into a tutple of ints
+def obs_from_device(obs):
+    return tuple([int(jax.device_get(obs[i])) for i in range(len(obs))])
+
+
+
+def cmp(a, b):
+    return (a > b).astype(int) - (a < b).astype(int)
+
+#gets a random card(with replacement)
+def random_card(key):
+    key = random.split(key)[0]
+    choice = random.choice(key,deck,shape=(1,))
+    
+    return choice[0].astype(int), key
+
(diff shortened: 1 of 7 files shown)


gym/envs/jax_toy_text/jax_blackjack.py
```diff
@@ -0,0 +1,463 @@
...
+
+    def reset(
+        self,
+        seed: Optional[int] = 0,
+        return_info: bool = False,
+        options: Optional[dict] = None,
+    ):
+        super().reset(seed=seed)
+
+        key = random.PRNGKey(seed)
```
RedTachyon, line 276: Won't this make it so that `env.reset()` always resets to the same initial state unless you explicitly pass a new random seed every time?

balisujohn: This should behave the same way as Blackjack-v1 now with respect to seed behavior, barring any mistakes.

pseudo-rnd-thoughts (changes requested):
Some general comments about the environment, I have a jax environment checker that should help with performance

PR #2931, opened 2022-06-27
Add option to change colour on carracing through "options" argument,  remove carracingdomainrandomize registration

jjshoots:
## Description

This allows the user to domain randomize the environment on demand.

This PR also removes the CarRacingDomainRandomize from registration.

## Motivation

The idea is to explore representation learning, where for example the agent learns in a particular colour scheme for N episodes, and learns in a separate colour scheme for M episodes.

Under the old code, users could only change the colour scheme every episode, or non at all.

This PR allows the user to specify the env to change colour scheme on demand, or choose to maintain the current colour scheme across episodes.


pseudo-rnd-thoughts:
Could you add a motivation for this PR

jjshoots:
@pseudo-rnd-thoughts Sure, sorry for the exceptioinally short description. I've updated it with the motivation.

PR #2941, opened 2022-06-30
Properly check whether AttributeError is caused by render_mode

Markus28:
@pseudo-rnd-thoughts 

Atm the if statement will be triggered by any attribute error, regardless of which kwarg caused it.

### DIFF
--- a/gym/envs/registration.py
+++ b/gym/envs/registration.py
@@ -620,7 +620,7 @@ def make(
     try:
         env = env_creator(**_kwargs)
     except TypeError as e:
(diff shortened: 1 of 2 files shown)


RedTachyon:
Shouldn't it be `>= 0`?

Markus28:
Yeah it doesn't really matter in practice because the string will be preceded by the signature of the method that caused the error. But I can change it to >=, that's probably more understandable

pseudo-rnd-thoughts:
Could you add a short test because it is a confusing error if it is not for the correct reason

pseudo-rnd-thoughts:
LGTM

PR #2939, opened 2022-06-30
[WIP] Typing fixes

RedTachyon:
# Description

Progressively removing excludes from pyright and making them pass. 

For now it's only toy_text envs, and I narrowed down the `spaces` exclude to the only one that was failing (the new graph space). 

I'm also planning to narrow down the exclusion lists, since the `Graph` space should have never been merged with failing type checks, but it was just blanket excluded.

Fixes # (issue)

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

Issue #2918, opened 2022-06-22
[Bug Report] VSCode PyLance stubs are outdated/incorrect


BlueskyFR:
Hi!

VSCode's PyLance (Python language server) extension contains some builtin stubs, with a "gym-stubs" dir that contains the following line for instance:

```python
metadata: Dict[str, List[str]]
```

It is of course wrong since `metadata` can contain `render_fps` with is of type **int**.

Who/where are those stubs managed? They should be updated since VSCode reports errors in code analysis in the meantime.

BlueskyFR:
Well, maintaining stubs in another non-openai repo seems like a really bad idea, maybe they should just remove them and let you type your stuff

pseudo-rnd-thoughts:
Given that we (the volunteer dev group) have only been doing this for about a year then I suspect that the stub was created when there were no type hints in the project

BlueskyFR:
> Given that we (the volunteer dev group) have only been doing this for about a year then I suspect that the stub was created when there were no type hints in the project

Is Gym a project you contribute to on your free time or are you "assigned" to it? Just curious 😉

pseudo-rnd-thoughts:
Join the discord (https://discord.gg/nHg2JRN489) and we ask for developers to help with stuff when problems occur

RedTachyon:
@BlueskyFR 
> Is Gym a project you contribute to on your free time or are you "assigned" to it? Just curious 😉

Also for posterity, none of us are affiliated with OpenAI, I feel like this might be a misunderstanding. We're not assigned to it because there's nobody who'd be able to assign us anywhere

BlueskyFR:
Thanks! I though OpenAI was actually maintaining this ;)

bschnurr:
https://github.com/openai/gym/blob/0dba07204622b354a1b440ad71fe2660d62674fa/gym/core.py#L106

Not sure what the type of metadata should be, but it should still have a typeAnnoation
something like:
` metadata: Dict[str, List[str|float]] = {"render_modes": []} `

or
`metadata: Dict[str, List[Any]]  = {"render_modes": []} `

bschnurr:
also there is a way to test of type completeness on your public api using pyright
```
pip install gym
pip install pyright

pyright --verifytypes gym
```
Results:

Symbols exported by "gym": 1371
  With known type: 324
  With ambiguous type: 196
  With unknown type: 851
  Functions without docstring: 283
  Functions without default param: 329
  Classes without docstring: 25

[...]

PR #2942, opened 2022-06-30
Full type hinting

pseudo-rnd-thoughts:
A follow up PR to https://github.com/openai/gym/pull/2939 that was inspired by https://github.com/openai/gym/issues/2918 to fully type check the whole gym project (along with tests).
This is removes all files and folders from the exlude folder however more work is required to add the core API to the strict parameter. 
However, this should catch any weird or incorrect type hints but at this time not force PRs to include type hints

In the process, I discovered a couple of strange tests for async vector envs which needs fixing however it was outside the scope of this PR


gym/utils/play.py
```diff
@@ -120,11 +129,11 @@ def display_arr(
 
 def play(
     env: Env,
+    keys_to_action: Dict[Union[Tuple[Union[str, int]], str], ActType],
```
RedTachyon, line 132: This seems like an unnecessary breaking change? (changing the order of arguments)

gym/vector/utils/__init__.py
```diff
@@ -6,7 +6,7 @@
     read_from_shared_memory,
     write_to_shared_memory,
 )
-from gym.vector.utils.spaces import BaseGymSpaces, _BaseGymSpaces, batch_space, iterate
```
RedTachyon, line 9: I kept this export deliberately. It was a bad idea and shouldn't be made private initially, but it's not actively harmful. One less breaking change for people to complain about.

gym/vector/vector_env.py
```diff
@@ -69,7 +69,7 @@ def reset_wait(
         seed: Optional[Union[int, List[int]]] = None,
         return_info: bool = False,
         options: Optional[dict] = None,
-    ):
+    ) -> np.ndarray:
```
RedTachyon, line 72: Don't vector envs have the same (obs, info) semantic as general envs?

gym/wrappers/gray_scale_observation.py
```diff
@@ -32,11 +32,12 @@ def __init__(self, env: gym.Env, keep_dim: bool = False):
...
 
         assert (
-            len(env.observation_space.shape) == 3
+            isinstance(env.observation_space, Box)
+            and len(env.observation_space.shape) == 3
             and env.observation_space.shape[-1] == 3
         )
 
-        obs_shape = self.observation_space.shape[:2]
+        obs_shape = env.observation_space.shape[:2]
```
RedTachyon, line 40: This looks like it could change the behavior of the code, why is this necessary?

gym/vector/vector_env.py
pseudo-rnd-thoughts: I thought it was the cause of some issues but it doesn't raise an error anymore so will revert

gym/vector/utils/__init__.py
pseudo-rnd-thoughts: This is in `vector/utils/__init__.py` so if someone is getting this from here then there is an issue

gym/wrappers/normalize.py
```diff
@@ -81,19 +81,20 @@ def step(self, action):
 
     def reset(self, **kwargs):
         """Resets the environment and normalizes the observation."""
-        return_info = kwargs.get("return_info", False)
-        if return_info:
+        if kwargs.get("return_info", False):
             obs, info = self.env.reset(**kwargs)
+
+            if self.is_vector_env:
```
RedTachyon, line 87: This also looks like a change of behavior, is this intended? Can you explain it?

gym/wrappers/resize_observation.py
```diff
@@ -39,7 +39,16 @@ def __init__(self, env: gym.Env, shape: Union[tuple, int]):
 
         self.shape = tuple(shape)
 
-        obs_shape = self.shape + self.observation_space.shape[2:]
+        assert isinstance(
+            env.observation_space, Box
+        ), f"Expected the observation space to be Box, actual type: {type(env.observation_space)}"
+        assert len(env.observation_space.shape) == 3 and env.observation_space.shape[
```
RedTachyon, line 45: This seems irrelevant to typing, right? Is this actually necessary for this wrapper to work, or is there any conceivable situation where this could break otherwise functional code?

gym/wrappers/gray_scale_observation.py
pseudo-rnd-thoughts: Good point, changed back. The assert statements fixed the problem

gym/utils/play.py
pseudo-rnd-thoughts: I spoke to @Markus28 about it. This was a bug in the previous implementation as `keys_to_action` is a required parameter if you look at the for loop after the docstring.

tests/vector/test_async_vector_env.py
```diff
@@ -28,11 +28,11 @@ def test_create_async_vector_env(shared_memory):
 @pytest.mark.parametrize("shared_memory", [True, False])
 def test_reset_async_vector_env(shared_memory):
     env_fns = [make_env("CartPole-v1", i) for i in range(8)]
-    try:
-        env = AsyncVectorEnv(env_fns, shared_memory=shared_memory)
-        observations = env.reset()
-    finally:
-        env.close()
+
+    env = AsyncVectorEnv(env_fns, shared_memory=shared_memory)
```
RedTachyon, line 32: This is one of the situations where it might be better to do a pyright ignore, or maybe something more clever.

Basically what might (maybe?) happen here is that something crashes inside the environment, the test gets aborted, and then the environment is never properly closed, which can lead to a memory leak, because there are some hanging processes. The `finally` clause prevents that.

It might be possible by using a feature that's existed in gym very quietly, which is using an env as a context manager:
```python
with AsyncVectorEnv(env_fns, shared_memory=shared_memory) as env:
    ... # do whatever
```
I'm not 100% sure, but I think this will make sure to close the environment even if it crashes

RedTachyon: (on second thought, I'm not sure if this is implemented for VectorEnv, but it definitely is for normal envs)

gym/wrappers/normalize.py
pseudo-rnd-thoughts: I believe that it is not a change in behaviour. The problem previous was that `info` could have been unbounded. To avoid the issue, I have all of the return info code in one part of the if statement and without info is the second part. This has the cost of copying the if vector code twice but it shouldn't be a issue

tests/vector/test_spaces.py
```diff
@@ -116,9 +116,11 @@ def test_batch_space_custom_space(space, expected_batch_space_4):
 def test_iterate(space, batch_space):
     items = batch_space.sample()
     iterator = iterate(batch_space, items)
+    final_i = 0
```
RedTachyon, line 119: Just do `i=0` before the loop and get rid of `final_i`

gym/wrappers/resize_observation.py
pseudo-rnd-thoughts: You right I will remove, the assert is not necessary for the PR but I think true however could cause more issues

gym/utils/play.py
Markus28: I think I would actually prefer to leave `key_to_action` optional and try to fetch it from the environment, then throw an exception if that's not possible. That's how it used to work (before I kinda unknowingly broke it with the for loop).

Markus28: Also, I need to update the tests for utils.play, in particular they currently don't seem to cover the case where `env` provides keys_to_action.

tests/vector/test_spaces.py
pseudo-rnd-thoughts: Done

gym/utils/play.py
pseudo-rnd-thoughts: Im going to remove `utils.play` from this PR and let @Markus28 fix it in another PR

tests/envs/utils.py
```diff
@@ -18,22 +18,27 @@ def try_make_env(env_spec: EnvSpec) -> Optional[gym.Env]:
...
     for env_spec in all_testing_env_specs
     if "gym.envs.mujoco" in env_spec.entry_point
 ]
-gym_testing_env_specs = [
+gym_testing_env_specs: List[EnvSpec] = [
     env_spec
     for env_spec in all_testing_env_specs
     if any(
-        f"gym.{ep}" in env_spec.entry_point
+        f"gym.envs.{ep}" in env_spec.entry_point
```
RedTachyon, line 41: Why was this changed? Is this a bugfix unrelated to the typing?

pseudo-rnd-thoughts: It is unrelated to type hinting but it fixes a previous bug

PR #2945, opened 2022-07-01
Try to fetch `keys_to_action` from environment in `play`. Made tests stronger

Markus28:
In `play` we try to fetch `keys_to_action` from the environment, if not possible we throw an exception. Unfortuantely, this is essentially copy-pasted from `PlayableGame`. I did not want to modify `PlayableGame` (e.g. by making `keys_to_action` non-optional) because some users might (?) use this class.

In the tests we now also check that:
- The appropriate exception is thrown by `play` if `keys_to_action` is unavailable
- `play` works if `keys_to_action` is provided by the environment
- `play` works if `keys_to_action` is provided as a dictionary that has characters as keys

### DIFF
--- a/gym/utils/play.py
+++ b/gym/utils/play.py
@@ -201,6 +201,17 @@ def play(
     """
     env.reset(seed=seed)
 
+    if keys_to_action is None:
+        if hasattr(env, "get_keys_to_action"):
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts:
LGTM

PR #2959, opened 2022-07-11
Text space

pseudo-rnd-thoughts:
A completion of https://github.com/openai/gym/pull/2908 to allow it to be merged for `v0.25.0`.
For the complete details of the Text see the above PR.

This PR adds masking to that PR with an optional tuples for the space text length that is optional and a mask a similar specification to the other space's masks.


andrewtanJS:
LGTM, only a minor comment

PR #2962, opened 2022-07-13
Removing return_info argument to env.reset() and deprecated env.seed() function (reset now always returns info)

balisujohn:
# Description

This PR removes the `return_info` argument so `env.reset` always returns a tuple of the form 
```` (obs, info)````where `info` is a dict which contains observation metadata.

Additionally, this PR removes the `env.seed()` function, for both the core env and vector envs.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist:

### CODE REVIEW

File: gym/core.py [hunk 1]
```diff
@@ -89,10 +89,9 @@ class Env(Generic[ObsType, ActType], metaclass=decorator):
 
     - :meth:`step` - Takes a step in the environment using an action returning the next observation, reward,
       if the environment terminated and more information.
-    - :meth:`reset` - Resets the environment to an initial state, returning the initial observation.
+    - :meth:`reset` - Resets the environment to an initial state, returning the initial observation and more information.
```
pseudo-rnd-thoughts, line 92: "observation information"

gym/core.py
```diff
@@ -482,11 +446,8 @@ def observation(self, obs):
 
     def reset(self, **kwargs):
```
pseudo-rnd-thoughts, line 447: could we change this to the actual reset parameters?

gym/utils/env_checker.py
```diff
@@ -281,7 +234,6 @@ def check_env(env: gym.Env, warn: bool = None, skip_render_check: bool = False):
     # ==== Check the reset method ====
     check_reset_seed(env)
     check_reset_options(env)
-    check_reset_info(env)
```
pseudo-rnd-thoughts, line 284: Should we replace this with a signature check?

gym/utils/env_checker.py
```diff
@@ -126,53 +126,6 @@ def check_reset_seed(env: gym.Env):
         )
 
 
-def check_reset_info(env: gym.Env):
```
pseudo-rnd-thoughts, line 129: Would we want to keep these with a signature check?

gym/vector/async_vector_env.py
```diff
@@ -250,15 +218,13 @@ def reset_wait(
...
         timeout: Optional[Union[int, float]] = None,
         seed: Optional[int] = None,
-        return_info: bool = False,
         options: Optional[dict] = None,
     ) -> Union[ObsType, Tuple[ObsType, List[dict]]]:
         """Waits for the calls triggered by :meth:`reset_async` to finish and returns the results.
 
         Args:
             timeout: Number of seconds before the call to `reset_wait` times out. If `None`, the call to `reset_wait` never times out.
             seed: ignored
```
pseudo-rnd-thoughts, line 227: The seed is ignored, how to vector env seed their environments without `seed`

gym/vector/sync_vector_env.py
```diff
@@ -123,26 +105,16 @@ def reset_wait(
...
-                kwargs["return_info"] = return_info
 
-            if not return_info:
-                observation = env.reset(**kwargs)
-                observations.append(observation)
-            else:
-                observation, info = env.reset(**kwargs)
-                observations.append(observation)
-                infos = self._add_info(infos, info, i)
+            observation, info = env.reset(**kwargs)
```
pseudo-rnd-thoughts, line 109: Do we need to use kwargs?

gym/wrappers/atari_preprocessing.py
```diff
@@ -151,11 +151,7 @@ def step(self, action):
     def reset(self, **kwargs):
         """Resets the environment using preprocessing."""
         # NoopReset
-        if kwargs.get("return_info", False):
-            _, reset_info = self.env.reset(**kwargs)
-        else:
-            _ = self.env.reset(**kwargs)
-            reset_info = {}
+        _, reset_info = self.env.reset(**kwargs)
```
pseudo-rnd-thoughts, line 154: could we replace kwarg with the actual args

gym/utils/passive_env_checker.py
```diff
@@ -183,34 +183,23 @@ def env_reset_passive_checker(env, **kwargs):
...
-        obs, info = result
-        assert isinstance(
-            info, dict
-        ), f"The second element returned by `env.reset(return_info=True)` was not a dictionary, actual type: {type(info)}"
-    else:
-        obs = result
+
+    assert isinstance(
+        result, tuple
+    ), "The result returned by `env.reset()` was not a tuple, actual type: <class 'list'>"
```
pseudo-rnd-thoughts, line 196: Why is the actual type fixed?

gym/utils/passive_env_checker.py
```diff
@@ -183,34 +183,23 @@ def env_reset_passive_checker(env, **kwargs):
...
+
+    obs, info = result
+    check_obs(obs, env.observation_space, "reset")
+    assert isinstance(
+        info, dict
+    ), "The second element returned by `env.reset()` was not a dictionary, actual type: <class 'set'>"
```
pseudo-rnd-thoughts, line 202: Why is the actual type fixed?

gym/utils/passive_env_checker.py
```diff
@@ -183,34 +183,23 @@ def env_reset_passive_checker(env, **kwargs):
...
-
-    if "options" not in signature.parameters and "kwargs" not in signature.parameters:
+    if "options" not in signature.parameters or "kwargs" in signature.parameters:
         logger.warn(
             "Future gym versions will require that `Env.reset` can be passed `options` to allow the environment initialisation to be passed additional information."
         )
 
     # Checks the result of env.reset with kwargs
     result = env.reset(**kwargs)
-    if kwargs.get("return_info", False) is True:
```
pseudo-rnd-thoughts, line 201: Personally, I would like the passive environment checker to have as minimal errors as possible. Therefore, keep some of these checks as warnings that the results is tuple and length == 2

pseudo-rnd-thoughts (changes requested):
It mostly looks good, I just have a couple of questions on mainly on the environment checker and passive env checker

gym/core.py
balisujohn: I'm not necessarily against this, but some our wrappers use this strategy to allow the wrapper to be agnostic to changes in the function definitions, such as https://github.com/openai/gym/blob/master/gym/wrappers/order_enforcing.py. Is this suggestion part of a broader goal of moving towards explicit argument type hinting for wrappers?

gym/utils/env_checker.py
balisujohn: Do you mean if "return_info" is in the signature, check to see if the return_info argument is handled correctly? So it would be for legacy third part envs implementing return_info in their signature?

gym/vector/async_vector_env.py
balisujohn: So I looked into this, basically reset_wait doesn't use its arguments, since seeding is performed in reset_async, so I think we can safely remove the arguments in reset_wait for vector envs. I think this should be fixed in a separate PR though.

gym/utils/passive_env_checker.py
balisujohn: good catch

gym/utils/passive_env_checker.py
balisujohn: I made the tuple check into a warning.

gym/utils/passive_env_checker.py
balisujohn: good catch

balisujohn:
Alright so I had a few questions about some of the comments and I made changes to address the other ones. Tests are passing locally so I don't know what the deal with CI/CD is. I think we should remove RandomNumberGenerator in a different PR, since it breaks space seeding tests when done naively.

gym/core.py
pseudo-rnd-thoughts: That is a good. Lets not do it because it allows the wrappers to be partially backward compatible

gym/utils/env_checker.py
pseudo-rnd-thoughts: More like, if return info is in the signature then say that this feature is deprecate and the environment shouldn't use it

gym/utils/env_checker.py
balisujohn: added, but I also brought back check_reset_info. Open to feedback if the current tests seem redundant since there is some redundancy between test_reset_info and test_passive_env_reset_checker

gym/vector/sync_vector_env.py
balisujohn: going to hold off on this unless specifically requested, in accordance with our discussion about kwargs on the other comment.

balisujohn:
Alright passing tests and ready for final review. Definitely we need to be as careful as possible about this PR. It's going to lead to a lot of issues being created since it's a hard deprecation of a core feature. If anyone has ideas for where to add more deprecation warnings, I think that could make this easier on users.

gym/utils/env_checker.py
balisujohn: done, I added a test for the signature check as well

balisujohn:
It occurs to me this will need a counterpart PR in the docs as well

pseudo-rnd-thoughts (changes requested):
A couple more comments for the changes to the environment checker

gym/utils/passive_env_checker.py
```diff
@@ -183,36 +183,24 @@ def env_reset_passive_checker(env, **kwargs):
...
-    if "return_info" not in signature.parameters and not (
-        "kwargs" in signature.parameters
-        and signature.parameters["kwargs"].kind is inspect.Parameter.VAR_KEYWORD
-    ):
-        logger.warn(
-            "Future gym versions will require that `Env.reset` can be passed `return_info` to return information from the environment resetting."
-        )
-
-    if "options" not in signature.parameters and "kwargs" not in signature.parameters:
+    if "options" not in signature.parameters or "kwargs" in signature.parameters:
```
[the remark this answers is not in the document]
balisujohn: I think this is actually an automatic merge error since my original fork for this commit is sort of old. Good catch. I'm reverting it.

gym/utils/env_checker.py
```diff
@@ -73,7 +73,7 @@ def check_reset_seed(env: gym.Env):
         and signature.parameters["kwargs"].kind is inspect.Parameter.VAR_KEYWORD
     ):
         try:
-            obs_1 = env.reset(seed=123)
```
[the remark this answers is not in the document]
balisujohn: So now I call check_reset_return_type before calling check_reset_seed, in check_env, so this should address that possibility.

gym/utils/env_checker.py
```diff
@@ -126,80 +126,66 @@ def check_reset_seed(env: gym.Env):
...
     else:
         raise gym.error.Error(
-            "The `reset` method does not provide a `return_info` or `**kwargs` keyword argument."
+            "The `reset` method does not provide an `options` or `**kwargs` keyword argument."
         )
 
 
-def check_reset_options(env: gym.Env):
-    """Check that the environment can be reset with options.
+def check_reset_info(env: gym.Env):
```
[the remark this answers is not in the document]
balisujohn: So I split check_reset_info out into `check_reset_return_type` and `check_reset_return_info_deprecation`. Some behavior is redundant with the call to `env_reset_passive_checker` , so we may want to stop calling the `env_reset_passive_checker`  in the standard `env_checker` in a later PR. I think the other option would be to instrument `env_reset_passive_checker` with an argument to put it into "assert mode," but I feel like that's out of scope for this PR.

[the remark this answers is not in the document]
pseudo-rnd-thoughts: Good idea for the changes

pseudo-rnd-thoughts (commented):
Looks good to me, do we want to add a check in the environment checker for `seed`. I don't think it should raise an error, just raise a warning

gym/utils/env_checker.py
```diff
@@ -171,6 +171,25 @@ def check_reset_return_info_deprecation(env: gym.Env):
...
+def check_seed_deprecation(env: gym.Env):
+    """Makes sure support for deprecated function `seed` is dropped.
+
+    Args:
+        env: The environment to check
+    Raises:
+        UserWarning
+    """
+    seed_fn = getattr(env, "seed", None)
+    print(seed_fn)
```
pseudo-rnd-thoughts, line 183: Remove prints

tests/utils/test_env_checker.py
```diff
@@ -192,6 +194,31 @@ def test_check_reset_return_info_deprecation(test, func: callable, message: str)
...
+        return
+
+    with pytest.warns(
+        UserWarning, match=f"^\\x1b\\[33mWARN: {re.escape(message)}\\x1b\\[0m$"
+    ):
+        env.seed = seed
+        assert callable(env.seed)
+        check_seed_deprecation(env)
+
+    with warnings.catch_warnings():
```
pseudo-rnd-thoughts, line 214: To check this works as intended we might want to add 
```
with warnings.catch_warnings(record=True) as caught_warnings:
    <test code>

assert len(caught_warnings) == 0

```

pseudo-rnd-thoughts (changes requested):
Other than the print statement, LGTM

gym/utils/env_checker.py
balisujohn: done

tests/utils/test_env_checker.py
balisujohn: sounds good, I changed to this pattern

gym/utils/env_checker.py
```diff
@@ -126,81 +126,92 @@ def check_reset_seed(env: gym.Env):
...
-        except TypeError as e:
-            raise AssertionError(
-                "The environment cannot be reset with options, even though `options` or `**kwargs` appear in the signature. "
-                f"This should never happen, please report this issue. The error was: {e}"
-            )
-    else:
-        raise gym.error.Error(
-            "The `reset` method does not provide an `options` or `**kwargs` keyword argument."
+    if "return_info" in signature.parameters:
+        raise AssertionError(
```
RedTachyon, line 167: I think it's better to make this a warning - it is an important information, but at the same time, in principle, it's not incorrect to implement a custom environment which takes extra optional arguments.

gym/vector/sync_vector_env.py
```diff
@@ -164,8 +136,8 @@ def step_wait(self):
                 info,
             ) = step_api_compatibility(env.step(action), True)
             if self._terminateds[i] or self._truncateds[i]:
+                observation, info = env.reset()
```
RedTachyon, line 139: Is this a mistake? In the old version, reset is after updating the info dict, now it's before, so I think it will record an incorrect final_observation

tests/utils/test_env_checker.py
```diff
@@ -115,95 +116,120 @@ def test_check_reset_seed(test, func: callable, message: str):
             check_reset_seed(GenericTestEnv(reset_fn=func))
 
 
-def _reset_var_keyword_kwargs(self, kwargs):
-    return self.observation_space.sample()
-
-
-def _reset_return_info_type(self, seed=None, return_info=False, options=None):
+def _deprecated_return_info(self, return_info=False):
```
RedTachyon, line 119: What exactly is the purpose of this?

RedTachyon: (some comments/description, and type hints)

balisujohn: The idea is we want to throw a warning if `return_info` is in the signature of an environment's reset function

gym/vector/sync_vector_env.py
balisujohn: Yeah this seems to be either a mistake or a merge error, so I'm reverting it.

tests/utils/test_env_checker.py
RedTachyon: Yea, but these specific functions are extremely odd. Please at least annotate these tests so that it's clear to us 6 months from now what these functions are meant to do exactly, or perhaps restructure the tests so that it's more obvious.

balisujohn: ah, I missed your second comment, yeah I can add some more explanation of the specific reason for the deprecation test, as well as type hints.

gym/utils/env_checker.py
balisujohn: Sounds good, I'm changing this to a UserWarning.

gym/vector/sync_vector_env.py
balisujohn: This seems like a mistake rather than a merge error, and I also found the mistake in the asyn vector env. Tests break when I switch the order of these statements to the correct order, so I need to examine why the tests are wrong.

balisujohn: I addressed this and now tests are passing. One issues with a naive swap of the order is that calling `obs, info = env,reset()` overwrites the `info` object where "final_observation" was saved. So I used a temp variable to write the observation from the last call to `.step()` into the `info` returned by the call to `.reset()`

tests/utils/test_env_checker.py
```diff
@@ -115,95 +118,123 @@ def test_check_reset_seed(test, func: callable, message: str):
...
+    return self.observation_space.sample(), {}
 
 
-def _return_info_obs_outside(self, seed=None, return_info=False, options=None):
-    if return_info:
-        return self.observation_space.sample() + self.observation_space.high, {}
-    else:
-        return self.observation_space.sample()
+def _reset_return_info_type(self, seed=None, options=None):
+    return [1, 2]
```
RedTachyon, line 136: I still don't like that this is a random list of integers. Is this meant to emulate a reset output of an incorrectly implemented env? We need some commentary about this specific sequence of functions.

pseudo-rnd-thoughts: The test can be updated to 
`return [self.observation_space.sample(), {}]`
I think this makes more sense

The point of the function is as a reset function where the type of the return is wrong.

pseudo-rnd-thoughts (approved):
LGTM

wookayin:
Are we sure about this breaking change? Is it a part of 0.26.0 or v1.0? If it is 0.26.0, in my opinion **this should not be the new default yet** as it will break most of the existing code.

It'd be nice if such a breaking change can be introduced ONLY with MAJOR versions. My understanding is that the next release we're expecting is v0.26: https://github.com/openai/gym/pull/3056

wookayin:
Through gym v0.25, it was NEVER warned to users that  `env.reset()` will start returning a Tuple instead of a single observation. There were some warnings about `env.step()` API changes (see `new_step_api`), but we had no compatiblity flag like for `env.reset()`.

It would be REALLY surprising if v0.26 suddenly changes the behavior and there is no way back. This doesn't make much sense, and I strongly suggest the new defaults should be a part of v1.0, and hopefully after some deprecation period.

wookayin:
This has been modified to a full compatibility wrapper in #3066 (EnvCompatibility)

PR #2968, opened 2022-07-15
Add `Sequence` space, update `flatten` functions

Markus28:
We are trying to integrate some of the functionality from https://github.com/entity-neural-network/entity-gym into Gym. We agreed that the best first step would be to implement a `Sequence` space that models arbitrary-length sequences. 

This leads to some problems because we cannot flatten arbitrary-length sequences to a `Box` in a reasonable way. The same problem exists for the `Graph` space. Currently, flattening a graph just means flattening the node and edge feature spaces. This seems quite reasonable to me and I took the same approach for the `Sequence` space. However, some changes were necessary to make flattening work for `Dict` and `Tuple` spaces that also contain `Graph` or `Sequence` spaces.

### TODO
- [x] New tests for flatten functionality, `Sequence` space
- [x] Improve docstrings


gym/spaces/sequence.py
```diff
@@ -0,0 +1,96 @@
...
+            None, None, seed  # type: ignore
+        )  # None for shape and dtype, since it'll require special handling
+
+    @property
+    def is_np_flattenable(self):
+        """Checks whether this space can be flattened to a :class:`spaces.Box`."""
+        return False
+
+    def sample(
+        self, mask: Optional[Tuple[Optional[np.ndarray], Optional[Any]]] = None
```
RedTachyon, line 47: `Optional[Any]` is redundant, it's the same as just writing `Any`.

gym/spaces/sequence.py
```diff
@@ -38,6 +38,11 @@ def __init__(
             None, None, seed  # type: ignore
         )  # None for shape and dtype, since it'll require special handling
 
+    def seed(self, seed: Optional[int] = None) -> list:
+        """Seed the PRNG of this space and the feature space."""
+        super().seed(seed)
```
pseudo-rnd-thoughts, line 43: Normally the seeding for composite spaces includes the super seed list
https://github.com/openai/gym/blob/8b744130bc25d0f4fe2eac8bce8e47527a8eee2c/gym/spaces/dict.py#L103

pseudo-rnd-thoughts (changes requested):
Looks good, a minor point on the seeding
Would you be able to check this against my updated seed testing PR, this should simplify the testing additions
https://github.com/openai/gym/pull/2977

Are we adding the utils.py function support? Otherwise, just raise an error explaining why I think

gym/spaces/sequence.py
Markus28: Are you saying that I should change the typehint to `seed: Any = None`? I can see how that makes sense.

pseudo-rnd-thoughts: I think this is what I mean
```python
seeds = super().seed(seed)
seeds += self.feature_space.seed(seed)
return seeds
```

pseudo-rnd-thoughts (approved):
LGTM, this is a very impressive number of tests. 
Could we confirm with the entity space team that this is good before merging

gym/spaces/sequence.py
Markus28: Ohhh, of course. I'm adding that :)

Markus28:
It might actually make sense to move the spaces and samples from `test_utils.py` to a separate file for better readability

cswinter (approved):
LGTM!

PR #2977, opened 2022-07-17
Split testing spaces

pseudo-rnd-thoughts:
Currently `tests/spaces/test_spaces.py` contains all of the spaces tests which makes it difficult to scale for more spaces

This PR both improves the testing of spaces, adds generic tests for all new spaces and fixes a couple of minor bugs in gym spaces

### Spaces Changes
* Update type hints in `spaces.Box`
* Add checking in `spaces.Dict` for if the same space key is used in spaces and kwargs
* Simplifies `contains` in `spaces.Dict` using positive (not negative) checks
* Update `spaces.Discrete.contains` to use `np.issubdtype` over the `x.dtype.char in np.typecodes["AllIntegers"]` which appears to use an undocumented feature in numpy. 
* Updated `spaces.Discrete.__setstate__` to correctly `np_random` information
* Multi-binary `from_jsonable` arrays now have dtype `np.int8`, previously `np.int32`
* Update Multi-Discrete sample to assume that `nvec` is correct instead of the mask making error messages more clear, the issue with a mask. 
* Set the default minimum length in `spaces.Text`  to 1 to prevent empty strings 
* Fixed `spaces.Text` sampling with a mask
* Added properties for the character set, list and string in `spaces.Text`
* Added `flatdim` functions for `spaces.Graph` spaces which raises an `NotImplemented` error due to the dynamic size 
* Added `flatdim`, `flatten`, `unflatten` and `flatten_space` for `spaces.Text` where the flatten version, the characters are converted to indexes within the space as a Box and padded to the max length. 

### Test changes
* Split `test_spaces.py` into separate files for specialist testing of spaces, i.e. `test_box.py`. As a result, we don't have a single 900+ line testing file
* Added `utils.py` which contains a set of example spaces that is used by `test_spaces.py` and `test_utils.py` for generic testing that the spaces all work. 
* Simplified `test_utils.py` to contain primarily self-consistent tests, i.e. `unflatten_space(flatten_space()) == original space`. This removes particular examples in favour of random samples from spaces.


gym/spaces/box.py
```diff
@@ -56,7 +56,7 @@ def __init__(
         low: Union[SupportsFloat, np.ndarray],
         high: Union[SupportsFloat, np.ndarray],
         shape: Optional[Sequence[int]] = None,
-        dtype: Type = np.float32,
+        dtype: Union[Type, object] = np.float32,
```
RedTachyon, line 59: I'm not so sure about this annotation, since pretty much anything is an `object` anyways (maaaaaaybe except for some extremely exotic metaclass constructs, I'm not sure)

We want it to be either a python type, or a numpy datatype, right? I think numpy has a type annotation for this somewhere in its typing module

gym/spaces/box.py
```diff
@@ -232,7 +232,7 @@ def to_jsonable(self, sample_n):
         """Convert a batch of samples from this space to a JSONable data type."""
         return np.array(sample_n).tolist()
 
-    def from_jsonable(self, sample_n: Sequence[SupportsFloat]) -> List[np.ndarray]:
+    def from_jsonable(self, sample_n: Sequence[Union[float, int]]) -> List[np.ndarray]:
```
RedTachyon, line 235: Why this restriction? I think SupportsFloat is more general?

gym/spaces/dict.py
```diff
@@ -1,15 +1,23 @@
...
+from typing import Sequence as TypingSequence
+from typing import Tuple, Union
 
 import numpy as np
 
 from gym.spaces.space import Space
 from gym.utils import seeding
 
+if sys.version_info >= (3, 8):
+    from typing import OrderedDict as TypingOrderedDict
```
RedTachyon, line 17: Why do we need this? Since 3.7, regular dict is basically an ordered dict. And even then, OrderedDict is a subtype of Dict, so `Union[Dict, OrderedDict]` is the exact same thing as `Dict`

gym/spaces/dict.py
```diff
@@ -75,66 +89,68 @@ def __init__(
...
-            spaces = spaces_kwargs
+        # Convert the spaces into an OrderedDict
         if isinstance(spaces, dict) and not isinstance(spaces, OrderedDict):
             try:
                 spaces = OrderedDict(sorted(spaces.items()))
             except TypeError:  # raise when sort by different types of keys
                 spaces = OrderedDict(spaces.items())
-        if isinstance(spaces, Sequence):
+        elif isinstance(spaces, Sequence):
             spaces = OrderedDict(spaces)
```
RedTachyon, line 99: I see now that in the existing code we're using OrderedDict, which imo we should change to just being a normal dict. It doesn't have to be done in this PR, but we shouldn't get additionally "tied" to this approach

gym/spaces/dict.py
```diff
@@ -75,66 +89,68 @@ def __init__(
...
-            ), "Values of the dict should be instances of gym.Space"
+            ), f"Dict space element is not an instance of Space: key='{key}', space={space}"
+
         super().__init__(
             None, None, seed  # type: ignore
         )  # None for shape and dtype, since it'll require special handling
 
-    def seed(self, seed: Optional[Union[dict, int]] = None) -> list:
+    def seed(
+        self, seed: Optional[Union[TypingDict[str, int], int]] = None
```
RedTachyon, line 127: I don't think `TypingDict[str, int]` is correct, since we can have nested dictionaries. A recursive type would work, but not in python, so I'd stick with the previous version

gym/spaces/dict.py
```diff
@@ -75,66 +89,68 @@ def __init__(
...
-                )
-            except ValueError:
-                subseeds = self.np_random.choice(
-                    np.iinfo(int).max,
-                    size=len(self.spaces),
-                    replace=True,  # we get more than INT_MAX subspaces
-                )
-
+            # replace=False - unique subseed for each subspace, if you have more spaces than np.iinfo(int).max then something is terribly wrong
+            subseeds = self.np_random.choice(
```
RedTachyon, line 140: This will be changed anyways in the other PR, is that right? Why do we need to do anything here?

gym/spaces/dict.py
```diff
@@ -164,14 +180,10 @@ def sample(self, mask: Optional[TypingDict[str, Any]] = None) -> dict:
...
         """Return boolean specifying if x is a valid member of this space."""
-        if not isinstance(x, dict) or len(x) != len(self.spaces):
-            return False
-        for k, space in self.spaces.items():
-            if k not in x:
-                return False
-            if not space.contains(x[k]):
-                return False
-        return True
+        if isinstance(x, dict):
```
RedTachyon, line 183: Flatten the two `if`s?

gym/spaces/multi_binary.py
```diff
@@ -58,7 +58,7 @@ def sample(self, mask: Optional[np.ndarray] = None) -> np.ndarray:
 
         Args:
             mask: An optional np.ndarray to mask samples with expected shape of ``space.shape``.
-                Where mask == 0 then the samples will be 0.
+                Where for each binary action, it expects shape mask in ``dtype=np.int8``.
```
RedTachyon, line 61: Can we keep both pieces of information in the docstring?

gym/spaces/multi_discrete.py
```diff
@@ -8,6 +8,7 @@
 from gym.spaces.space import Space
 from gym.utils import seeding
 
+RECURSIVE_LIST_INT = List[Union["RECURSIVE_LIST_INT", int]]
```
RedTachyon, line 11: Does this actually work well with type checkers? I see there's a similar pattern in the existing code that I never noticed, but last I checked, python didn't really like that.

gym/spaces/dict.py
pseudo-rnd-thoughts: yes, but we still support python 3.6... where this is not true sadly so I think we need to keep this for now or ignore it

gym/spaces/dict.py
pseudo-rnd-thoughts: Agreed, the problem is that python 3.6 dicts are not ordered. We could ignore this or make it a gymnasium change?

gym/spaces/dict.py
```diff
@@ -201,16 +213,16 @@ def to_jsonable(self, sample_n: list) -> dict:
...
+
         n_elements = len(next(iter(dict_of_list.values())))
-        for i in range(n_elements):
-            entry = {}
-            for key, value in dict_of_list.items():
-                entry[key] = value[i]
-            ret.append(entry)
-        return ret
+        result = [
+            OrderedDict({key: value[n] for key, value in dict_of_list.items()})
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: I can't remember the exact reason why I changed, I will see if I can change back to dict otherwise will add the reason

gym/spaces/discrete.py
```diff
@@ -81,18 +81,19 @@ def contains(self, x) -> bool:
...
             return False
+
         return self.start <= as_int < self.start + self.n
 
     def __repr__(self) -> str:
         """Gives a string representation of this space."""
         if self.start != 0:
-            return "Discrete(%d, start=%d)" % (self.n, self.start)
-        return "Discrete(%d)" % self.n
+            return f"Discrete({self.n:d}, start={self.start:d})"
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: I believe it casts it as an integer so we should be able to remove it

gym/spaces/discrete.py
```diff
@@ -120,5 +119,4 @@ def __setstate__(self, state):
         if "start" not in state:
             state["start"] = 0
 
-        # Update our state
-        self.__dict__.update(state)
+        super().__setstate__(state)
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: I believe the super version as a couple of additional changes like changing `np_random` to `_np_random`.
This is also the style that all of the other spaces use so it is a possible bug fix and for consistency

gym/spaces/multi_discrete.py
pseudo-rnd-thoughts: Pycharm doesn't seem to interpret it correctly so Im not sure but this should be the correct typing. 
Any suggestions?

gym/spaces/box.py
pseudo-rnd-thoughts: Given that it should be from json then we know it must be a float or int but you are correct that supports float is a more general type

gym/spaces/box.py
pseudo-rnd-thoughts: The problem is that `np.float32` is not actually a type but an object and `np.result_type()` (in utils.py L322) returns an object

gym/spaces/multi_discrete.py
RedTachyon: My understanding is that (a) it's valid in python, (b) mypy doesn't support it, (c) pyright supports it (hence no errors in CI)

I'd refrain from using this since it will cause issues for at least some users (myself included), at least until recursive types are more widely supported.

The simple alternative is just an unspecified `List`

gym/spaces/dict.py
pseudo-rnd-thoughts: This PR was submitted before #3011 but yes will update to that PR

gym/spaces/dict.py
pseudo-rnd-thoughts: Good point, I didn't consider that case, I will change back

gym/spaces/dict.py
pseudo-rnd-thoughts: Will do

gym/spaces/dict.py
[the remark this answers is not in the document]
pseudo-rnd-thoughts: The problem is that I have added a check that using `Dict.sample()` values roundtripped with `from_json(to_json())` are equal. 
The problem is that `Dict.sample` returns `OrderedDict` and so the type check fails

gym/spaces/multi_binary.py
pseudo-rnd-thoughts: I discovered a bug here as the current sampling allows for the mask shape == space shape and mask value allow `0 -> 0` and `1 -> random 0 or 1` so doesn't allow the mask to force `1` as only value answer. The easiest solution is to add `1 -> 1` and `2 -> random 0 or 1` but this breaks convention. The other option is making the mask shape == space shape + (2,).

gym/spaces/multi_discrete.py
pseudo-rnd-thoughts: Agreed, I will change it to list

gym/spaces/box.py
pseudo-rnd-thoughts: Changed back to `Type`

pseudo-rnd-thoughts:
Close in favour of #3047 and #3048

PR #2973, opened 2022-07-17
Bug fix: Revert an incorrect edition on wrapper.FrameStack

ZhiqingXiao:
# Description

https://github.com/openai/gym/pull/2752 incorrectly removes the obligated positional parameter of `self.observation()`, which makes the wrapper class no longer work. This bug fix corrects it.

@arjun-kg  @pseudo-rnd-thoughts  @jkterry1 

## Type of change

Please delete options that are not relevant.

- [ +] Bug fix (non-breaking change which fixes an issue)

### Screenshots
Please attach before and after screenshots of the change if applicable.

# Checklist:

- [ ] I have run the [`pre-commit` checks](https://pre-commit.com/) with `pre-commit run --all-files` (see `CONTRIBUTING.md` instructions to set it up)
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #2979, opened 2022-07-18
Update frame_stack.py

Sahaj09:
# Description

error being - "TypeError: observation() missing 1 required positional argument: 'observation'"

missing argument for self.observation(). Previously, it was self.observation(None) and in the last commits.

Fixes # (issue)
changed it back to self.observation(None)


pseudo-rnd-thoughts:
Thanks but this is already being fixed in https://github.com/openai/gym/pull/2973

PR #2981, opened 2022-07-21
running lunar-landing env3

carlacodes:
Adding a default wind on environment for the 2d lunar lander env


pseudo-rnd-thoughts:
You should be able to run `gym.make("LunarLander-v2", wind_power=5)` to change the default set wind power if that was the point of this PR

PR #2985, opened 2022-07-21
Lambda reward Wrappers

gianlucadecola:
This PR adds jumpy reward wrappers for compatibility with Jax and NumPy environments.
Observation and action wrapper replacement are also ready but left out from this PR for easier review.

@pseudo-rnd-thoughts

### DIFF
--- a/gym/dev_wrappers/lambda_reward.py
+++ b/gym/dev_wrappers/lambda_reward.py
@@ -66,7 +66,7 @@ class ClipRewardsV0(LambdaRewardV0):
         min_reward: Optional[Union[float, jp.ndarray]] = None,
         max_reward: Optional[Union[float, jp.ndarray]] = None,
     ):
-        """Initialize clip_reward_v0 wrapper.
+        """Initialize ClipRewardsV0 wrapper.
 
         Args:
             env (Env): The environment to apply the wrapper
--- a/gym/dev_wrappers/lambda_reward.py
+++ b/gym/dev_wrappers/lambda_reward.py
@@ -10,13 +10,13 @@ from gym.dev_wrappers import FuncArgType
 from gym.error import InvalidBound
 
 
-class lambda_reward_v0(gym.RewardWrapper):
+class LambdaRewardV0(gym.RewardWrapper):
     """A reward wrapper that allows a custom function to modify the step reward.
 
     Example:
         >>> import gym
         >>> env = gym.make("CartPole-v1")
-        >>> env = lambda_reward_v0(env, lambda r: 2 * r + 1)
+        >>> env = LambdaRewardV0(env, lambda r: 2 * r + 1)
         >>> env.reset()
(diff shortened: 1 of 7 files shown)


gym/dev_wrappers/lambda_reward.py
```diff
@@ -0,0 +1,95 @@
...
+import jax.numpy as jnp
+import jumpy as jp
+import numpy as np
+
+import gym
+from gym.dev_wrappers import FuncArgType
+from gym.error import InvalidBound
+
+
+class lambda_reward_v0(gym.RewardWrapper):
```
RedTachyon, line 13: Please no snake_case class names. I'd even take LambdaReward_v0, or LambdaRewardV0 over this

gianlucadecola: Afaik it was a naming convention for the new dev_wrappers (also here: https://github.com/openai/gym/pull/2853). If we want to switch to standard camel case no problem for me

pseudo-rnd-thoughts: Sorry I forgot to mention that I think we are keeping the normal python style with the addition of `V0` at the end.
@RedTachyon What do you think is better:`LambdaReward_v0` or `LambdaRewardV0`

gym/wrappers/__init__.py
```diff
@@ -1,5 +1,13 @@
 """Module of wrapper classes."""
 from gym import error
+
+try:
+    import jumpy
+except ImportError:
+    pass
+else:
+    from gym.dev_wrappers.lambda_reward import clip_rewards_v0, lambda_reward_v0
```
pseudo-rnd-thoughts, line 9: Could this be part of the import jumpy section?

gym/wrappers/__init__.py
```diff
@@ -1,5 +1,13 @@
 """Module of wrapper classes."""
 from gym import error
+
+try:
+    import jumpy
+
+    from gym.dev_wrappers.lambda_reward import ClipRewardsV0, LambdaRewardV0
+except ImportError:
+    pass
```
pseudo-rnd-thoughts, line 9: What happens when this fails but the user tries to import the class? 
Is the error helpful?

gianlucadecola: Good catch, maybe raising a warning telling the user that all dev_wrappers will not work correctly since jumpy is not available for python3.6 would be better?

pseudo-rnd-thoughts: I think that python 3.6 will be removed for the new wrappers

gianlucadecola: If python3.6 supports removal is programmed anytime soon I can wait and then remove the whole try - except block from all the dev_wrappers PRs, what do you think? Removing the try-except block now will cause python3.6 pipelines to fail because of jumpy

gianlucadecola:

PR #2990, opened 2022-07-24
backward compatibility when metadata doesn't include render mode

pseudo-rnd-thoughts:
In testing `gym-minigrid`, it used `v0.25.0`, however, this raised an error with `render_modes = env_creator.metadata["render_modes"]`  as the environment metadata does not include `"render_modes"`.

Therefore, in this PR, I have changed all of the errors to warnings and removed the respective tests

### DIFF
--- a/gym/envs/registration.py
+++ b/gym/envs/registration.py
@@ -627,6 +627,7 @@ def make(
             assert isinstance(
                 env_creator.metadata, dict
             ), f"Expect the environment creator ({env_creator}) metadata to be dict, actual type: {type(env_creator.metadata)}"
+
             if "render_modes" in env_creator.metadata:
                 render_modes = env_creator.metadata["render_modes"]
 
@@ -651,12 +652,6 @@ def make(
                 # else:
                 #   we don't raise an error as the environment may have forgotten to add the "render_modes"
                 #   if the render mode is not valid for an environment, the environment should raise an error
-            else:
-                logger.warn(
-                    "Applying render_mode to environment without metadata `render_modes`."
-                )
-        else:
-            logger.warn("Applying render_mode to environment with metadata attribute.")
 
     try:
         env = env_creator(**_kwargs)
--- a/gym/envs/registration.py
+++ b/gym/envs/registration.py
@@ -624,45 +624,39 @@ def make(
     # If we have access to metadata we check that "render_mode" is valid and see if the HumanRendering wrapper needs to be applied
     if mode is not None:
         if hasattr(env_creator, "metadata"):
-            if isinstance(env_creator.metadata, dict):
-                if "render_modes" in env_creator.metadata:
-                    render_modes = env_creator.metadata["render_modes"]
+            assert isinstance(
+                env_creator.metadata, dict
+            ), f"Expect the environment creator ({env_creator}) metadata to be dict, actual type: {type(env_creator.metadata)}"
+            if "render_modes" in env_creator.metadata:
+                render_modes = env_creator.metadata["render_modes"]
 
-                    # Apply the `HumanRendering` wrapper, if the mode=="human" but "human" not in render_modes
-                    if (
-                        mode == "human"
-                        and "human" not in render_modes
-                        and (
-                            "single_rgb_array" in render_modes
-                            or "rgb_array" in render_modes
-                        )
-                    ):
-                        logger.warn(
-                            "You are trying to use 'human' rendering for an environment that doesn't natively support it. "
-                            "The HumanRendering wrapper is being applied to your environment."
-                        )
-                        apply_human_rendering = True
-                        if "single_rgb_array" in render_modes:
-                            _kwargs["render_mode"] = "single_rgb_array"
-                        else:
-                            _kwargs["render_mode"] = "rgb_array"
-                    # else:
-                    #   we don't raise an error as the environment may have forgotten to add the "render_modes"
-                    #   if the render mode is not valid for an environment, the environment should raise an error
-                else:
-                    logger.warn(
-                        f"The environment {id} metadata does not include `render_modes`."
+                # Apply the `HumanRendering` wrapper, if the mode=="human" but "human" not in render_modes
+                if (
+                    mode == "human"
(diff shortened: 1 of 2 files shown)


gym/envs/registration.py
```diff
@@ -621,30 +621,48 @@ def make(
...
-            _kwargs["render_mode"] = (
-                "single_rgb_array"
-                if "single_rgb_array" in env_creator.metadata["render_modes"]
-                else "rgb_array"
-            )
-            apply_human_rendering = True
-        elif mode is not None and mode not in render_modes:
-            raise error.Error(
-                f"Invalid render_mode provided: {mode}. Valid render_modes: None, {', '.join(render_modes)}"
+                f"The environment creator for {id} does not contain metadata, this could be due to metadata being set during initialisation. Creator: {env_creator}"
```
Markus28, line 662: IIRC this can also happen for older versions of python if we `make` an environment from a different package (because of something importlib does). Also, it can happen because `env_creator` is a factory function. I don't think we should raise a warning because it might actually trigger them for some of Gym's own environments.

gym/envs/registration.py
```diff
@@ -621,30 +621,48 @@ def make(
     mode = _kwargs.get("render_mode")
     apply_human_rendering = False
 
-    # If we have access to metadata we check that "render_mode" is valid
-    if hasattr(env_creator, "metadata"):
```
Markus28, line 625: Could we just do 

```python
if hasattr(env_creator, "metadata") and "render_modes" in env_creator.metadata:
    ...
elif hasattr(env_creator, "metadata") and "render_modes"  not in env_creator.metadata:  # Longer for readability
    logger.warn("Metadata doesn't specify 'render_modes'.")
```

That would probably avoid all of the nested if statements and cover what's relevant imo (check the other two comments).

Markus28 (commented):
I'm not sure about creating a warning for when metadata is missing (see my comments).

gym/envs/registration.py
```diff
@@ -621,30 +621,48 @@ def make(
...
-        # We might be able to fall back to the HumanRendering wrapper if 'human' rendering is not supported natively
-        if (
-            mode == "human"
-            and "human" not in render_modes
-            and ("single_rgb_array" in render_modes or "rgb_array" in render_modes)
-        ):
+    # If we have access to metadata we check that "render_mode" is valid and see if the HumanRendering wrapper needs to be applied
+    if mode is not None:
+        if hasattr(env_creator, "metadata"):
+            if isinstance(env_creator.metadata, dict):
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: I don't know but if it is not a dictionary then the next line will error and I wanted to avoid that

gym/envs/registration.py
pseudo-rnd-thoughts: Do we ignore it then?

gym/envs/registration.py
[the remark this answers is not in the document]
Markus28: I think that would be the user's problem because `metadata` not being a dictionary is just unreasonable. Imo it falls under EAFP.

gym/envs/registration.py
Markus28: I'm not sure. I am like 80% sure that for some python versions and some gym environments this will be a problem and a warning will be issued, which is kinda undesirable

pseudo-rnd-thoughts:
@Markus28 Based on your comments, I agree so have removed all of the warnings except for the applying human rendering wrapper warnings. While it is possible for the metadata to not be a dictionary, I agree that this is unlikely, so I have replaced the warning with an assert statement

gym/envs/registration.py
```diff
@@ -621,30 +621,37 @@ def make(
...
-                else "rgb_array"
-            )
-            apply_human_rendering = True
-        elif mode is not None and mode not in render_modes:
-            raise error.Error(
-                f"Invalid render_mode provided: {mode}. Valid render_modes: None, {', '.join(render_modes)}"
-            )
+    # If we have access to metadata we check that "render_mode" is valid and see if the HumanRendering wrapper needs to be applied
+    if mode is not None:
+        if hasattr(env_creator, "metadata"):
```
Markus28, line 626: Maybe avoid nesting and use `and`.

gym/envs/registration.py
```diff
@@ -621,30 +621,37 @@ def make(
...
+                        "The HumanRendering wrapper is being applied to your environment."
+                    )
+                    apply_human_rendering = True
+                    if "single_rgb_array" in render_modes:
+                        _kwargs["render_mode"] = "single_rgb_array"
+                    else:
+                        _kwargs["render_mode"] = "rgb_array"
+                # else:
+                #   we don't raise an error as the environment may have forgotten to add the "render_modes"
+                #   if the render mode is not valid for an environment, the environment should raise an error
```
Markus28, line 654: Maybe one of my last comments was somewhat unclear. I think that it is fair to give a warning if `env_creator` has metadata that doesn't contain "render_modes". My point was that we shouldn't worry if `env_creator` has no metadata at all because there may be legitimate reasons for that. Similarly, I think it is fair to give a warning if metadata exists and has "render_modes", but the current mode isn't contained in that list.

Markus28 (commented):
I think this can be merged like this but personally, I would probably add two warnings back. My previous comments may have been a little convoluted, so here's the cases in which, imo, there should be warnings and in which cases not:

- Do *not* issue a warning if `metadata` is missing  (`env_creator` may be a factory function or importlib may do something strange)
- Issue a warning if `metadata` exists but `"render_modes"` is missing
- Issue a warning if `metadata` exists, contains `"render_modes"` and the `HumanRendering` wrapper is applied
- Issue a warning if `metadata` exists, contains `"render_modes"`, the `HumanRendering` wrapper is not applied, and `mode` is not supported according to the metadata

gym/envs/registration.py
pseudo-rnd-thoughts: Done

pseudo-rnd-thoughts:
I agree with your suggestions, all changes made

PR #2993, opened 2022-07-25
Update swimmer_v4.py

lin826:
Fix the type-o of rotors.

# Description

According to the action definition, `rot2` is mapping to the first rotor, and `rot3` is mapping to the second.

## Type of change

# Checklist:


rodrigodelazcano:
Thank you for checking the PR @lin826. The joint `rot` is for the free rotation of the swimmers body with respect to the world, it doesn't have an actuator. See video bellow.

Also, can you change the name of these joints in the `xml` and docs to a more meaningful name, like: `free_body_rot`, `motor1_rot` and `motor2_rot`. (https://github.com/openai/gym/blob/master/gym/envs/mujoco/assets/swimmer.xml)



swimmer_joints

PR #2998, opened 2022-07-26
fix render popping when render is called internally for Wrappers

younik:
Add the utility method `_render` to Wrappers. Calling `_render` save the frames, so they are added again when the user calls `render`.


pseudo-rnd-thoughts:
LGTM

PR #3001, opened 2022-07-28
Avoid metaclass conflicts when inheriting from `gym.Env`

YouJiacheng:
# Description

Using `__init_subclass__` instead of metaclass, suggested by [PEP 487](https://peps.python.org/pep-0487/)(introduced in python3.6).
As a result, downstream package can safely inherit from `gym.Env` and `abc.ABC`/`Protocol`/...(classes with metaclass != `type`), or set their own metaclass without making a intermediate metaclass inherited from custom metaclass and `type(gym.Env)`.

For example, https://github.com/rlworkgroup/metaworld use `abc.ABC`/`abc.ABCMeta`, thus it was broken by the metaclass introduced in `gym-0.25.0`.
https://github.com/rlworkgroup/metaworld/blob/18118a28c06893da0f363786696cc792457b062b/metaworld/envs/mujoco/mujoco_env.py#L31
https://github.com/rlworkgroup/metaworld/blob/18118a28c06893da0f363786696cc792457b062b/metaworld/envs/mujoco/sawyer_xyz/sawyer_xyz_env.py#L14

```python
import abc, gym
class SomeAbstractEnvUseABCMeta(gym.Env, metaclass=abc.ABCMeta): pass
class SomeAbstractEnvUseABC(gym.Env, abc.ABC): pass
``` 
Above code works on `gym<0.25.0`, but crashs on `gym==0.25.0`
image

## Type of change

If user-code/downstream library doesn't depend on `type(gym.Env)`, it should not be a breaking change, but there is a rare case that workaround for metaclass introduced in `gym-0.25.0` is used:
```python
import abc, gym
class AbstractEnvMeta(type(gym.Env), abc.ABCMeta): pass
class SomeAbstractEnv(gym.Env, metaclass=AbstractEnvMeta): pass
``` 
Above code will crash if `issubclass(abc.ABCMeta, type(gym.Env))`, which is true without decorator metaclass introduced in `gym-0.25.0` since `type(gym.Env) is type`, because of [MRO conflict](https://stackoverflow.com/a/42567588/16613821). i.e., above code will crash on `gym<0.25.0` and after this change.
I want to note that following code won't crash on any version of `gym` or after this change:
```python
import abc, gym
class AbstractEnvMeta(abc.ABCMeta, type(gym.Env)): pass
class SomeAbstractEnv(gym.Env, metaclass=AbstractEnvMeta): pass
```

# Checklist:

### DIFF
--- a/gym/core.py
+++ b/gym/core.py
@@ -31,55 +31,36 @@ ActType = TypeVar("ActType")
 RenderFrame = TypeVar("RenderFrame")
 
 
-class _EnvDecorator(type):  # TODO: remove with gym 1.0
-    """Metaclass used for adding deprecation warning to the mode kwarg in the render method."""
+# TODO: remove with gym 1.0
+def _deprecate_mode(render_func):  # type: ignore
+    """Wrapper used for adding deprecation warning to the mode kwarg in the render method."""
+    render_return = Optional[Union[RenderFrame, List[RenderFrame]]]
 
-    def __new__(cls, name, bases, attr):
-        if "render" in attr.keys():
-            attr["render"] = _EnvDecorator._deprecate_mode(attr["render"])
+    def render(
+        self: object, *args: Tuple[Any], **kwargs: Dict[str, Any]
+    ) -> render_return:
+        if "mode" in kwargs.keys() or len(args) > 0:
+            deprecation(
+                "The argument mode in render method is deprecated; "
+                "use render_mode during environment initialization instead.\n"
+                "See here for more information: https://www.gymlibrary.ml/content/api/"
+            )
+        elif self.spec is not None and "render_mode" not in self.spec.kwargs.keys():  # type: ignore
+            deprecation(
+                "You are calling render method, "
+                "but you didn't specified the argument render_mode at environment initialization. "
+                "To maintain backward compatibility, the environment will render in human mode.\n"
+                "If you want to render in human mode, initialize the environment in this way: "
+                "gym.make('EnvName', render_mode='human') and don't call the render method.\n"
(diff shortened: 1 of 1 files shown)


younik (approved):
Hello, thanks for addressing it!
LGTM

(You will probably just need to fix `pre-commit hook`)

gym/core.py
```diff
@@ -106,6 +87,12 @@ class Env(Generic[ObsType, ActType], metaclass=decorator):
     Note: a default reward range set to :math:`(-\infty,+\infty)` already exists. Set it if you want a narrower range.
     """
 
+    def __init_subclass__(cls) -> None:
+        """Hook used for wrapping render method."""
+        super().__init_subclass__()
+        if "render" in vars(cls):
+            cls.render = _deprecate_mode(vars(cls)["render"])
```
RedTachyon, line 94: Any reason why this can't be `cls.render = _deprecate_mode(cls.render)`? If not, I'd prefer that for simplicity

YouJiacheng: Given `"render" in vars(cls)` is `True`, `vars(cls)["render"]` is the same as `cls.render` in 99.9% cases, except `__getattr__` being overridden.
I want to maximize the consistency between exist-check and access. Ideally I want to write `if "render" in vars(cls): vars(cls).update(render=vars(cls)["render"])`, but `vars(cls)` is read-only.
FWIW, the reason why `try...except` or `hasattr` check + `cls.render` cannot be used is that `cls.render` can come from base classes.

RedTachyon: I like the alternative you presented here, of course with the added wrapper

PR #3002, opened 2022-07-28
Lambda action and Lambda observation wrappers

gianlucadecola:
This PR adds the new lambda actions and observation wrappers, which use jumpy for compatibility with Jax and NumPy environments.

The PR is also related to the lambda wrappers PR (https://github.com/openai/gym/pull/2985) which is smaller and (I hope) easier to review, so I suggest reading that first.

Another suggestion I could give for reviewing it is to start from lambda_actions to get a grasp of the logic behind the wrappers, and then move to lambda_observation which involves pretty much more wrappers but with the same logic.

UPDATE:
Here an overview of the general PR structure:
Starting from dev_wrappers folder, lambda_actions.py and lambda_observations.py contains wrappers dedicated to altering the executed action/observation. They also typically modify the underlying action/observation space.

lambda_actions and lambda_observations wrappers work more or less in the same way. I will use lamba_action as an example.
LambdaActionV0 is the base lambda wrapper that wraps the environment and applies the provided function to the action.  Other wrappers inherit from LambdaActionV0 and typically perform also an update of the space (in terms of shape, bounds and so on)

In particular, every transformation to the space is handled inside `dev_wrappers/utils` using a functional approach. These functions return, where possible, the transformed space.

A requirement is to support `Dict` and `Tuple` spaces. Given the possibility to have potentially nested `Dict` and `Tuple` this is handled inside `dev_wrappers/utils/commons.py`. Here, the same functional approach allow to _explore_ the spaces and applying the required function to modify the space.

In `spaces/utils.py` a similar approach is used to actually apply to the action/observation the provided function. Since another requirements is to support highly parallelized environment with jax and numpy, the applied function are typically `jumpy` function.


gym/dev_wrappers/utils/filter_space.py
```diff
@@ -0,0 +1,39 @@
...
+from typing import Tuple as TypingTuple
+
+from gym.dev_wrappers import FuncArgType
+from gym.spaces import Box, Dict, Discrete, MultiBinary, MultiDiscrete, Space, Tuple
+
+
+@singledispatch
+def filter_space(space: Space, args: FuncArgType) -> Any:
+    """Filter space with the provided args."""
+
```
pseudo-rnd-thoughts, line 14: `raise NotImplementedError`

gym/dev_wrappers/utils/grayscale_space.py
```diff
@@ -0,0 +1,47 @@
...
+
+from gym.dev_wrappers import FuncArgType
+from gym.error import InvalidRGBShape
+from gym.spaces import Box, Discrete, MultiBinary, MultiDiscrete, Space
+
+
+@singledispatch
+def grayscale_space(space: Space, args: FuncArgType, fn: Callable) -> Any:
+    """Make observation space grayscale (i.e. flatten third dimension)."""
+
```
pseudo-rnd-thoughts, line 16: `raise NotImplementedError`

gym/dev_wrappers/utils/make_scale_args.py
```diff
@@ -0,0 +1,57 @@
...
+    """Compute args for rescaling action function.
+
+    Action space args needs to be extended in order
+    to correctly rescale the actions.
+    i.e. args before: {"body":{"left_arm": (-0.5,0.5)}, ...}
+    args after: {"body":{"left_arm": (-0.5,0.5,-1,1)}, ...}
+    where -1, 1 was the old action space bound.
+    old action space is needed to rescale actions.
+    """
+    ...
```
pseudo-rnd-thoughts, line 24: `raise NotImplementedError`

tests/dev_wrappers/test_lambda_actions/test_clip_actions.py
```diff
@@ -0,0 +1,287 @@
...
+    NEW_BOX_HIGH,
+    NEW_BOX_LOW,
+    NUM_ENVS,
+    SEED,
+    TUPLE_SPACE,
+    TUPLE_WITHIN_DICT_SPACE,
+)
+from tests.dev_wrappers.utils import TestingEnv
+
+try:
```
pseudo-rnd-thoughts, line 24: `pytest.importorskip` allows you to try import a module otherwise the file is skipped

pseudo-rnd-thoughts (changes requested):
I had a quick look through and cannot see anything particular wrong other than the comments made. However due to the complexity of the code. 
Could you update to the PR docstring or add a comment to explain the code structure the purpose of different parts so other people can understand the general style easier

tests/dev_wrappers/test_lambda_actions/test_clip_actions.py
gianlucadecola: With importorskip I can do this at module level:

```python
wrappers = pytest.importorskip("gym.wrappers")
```

But this does not seems to work:
```python
ClipActionsV0 = pytest.importorskip("gym.wrappers.ClipActionsV0")
```  

Another option I can think of is to do like this:  

``` python
pytest.importorskip("jumpy")
from gym.wrappers import ClipActionsV0 
```
The only caveat is that I need to add `# noqa` to the import because of pre-commit, what do you think it is better?

gianlucadecola:
> I had a quick look through and cannot see anything particular wrong other than the comments made. However due to the complexity of the code. Could you update to the PR docstring or add a comment to explain the code structure the purpose of different parts so other people can understand the general style easier

Yes, it is indeed quite a bulky PR, I'm going to update the PR docstring and explain the overall structure :+1:

RedTachyon:
Question - why do we actually need jumpy for this in the first place? Lambda wrappers feel like they should be the most framework-agnostic thing there is. Just give it a function, and it will do the rest.

gym/dev_wrappers/lambda_action.py
```diff
@@ -0,0 +1,174 @@
+"""Lambda action wrappers that uses jumpy for compatibility with jax (i.e. brax) and numpy environments."""
+
+from typing import Any, Callable
+from typing import Tuple as TypingTuple
```
RedTachyon, line 4: What is the point of this? There isn't any other `Tuple` here. If there was (e.g. from spaces), I think it'd be much clearer to just directly use `typing.Tuple`

gym/dev_wrappers/lambda_action.py
```diff
@@ -0,0 +1,174 @@
...
+        ... )
+        >>> obs, rew, done, info = env.step([np.float64(1.2), np.float64(1.2)])
+    """
+
+    def __init__(
+        self,
+        env: gym.Env,
+        func: Callable,
+        args: FuncArgType[Any],
+        action_space: gym.Space = None,
```
RedTachyon, line 57: Is it `gym.Space` or `Optional[gym.Space]`?

gianlucadecola:
> Question - why do we actually need jumpy for this in the first place? Lambda wrappers feel like they should be the most framework-agnostic thing there is. Just give it a function, and it will do the rest.

It was a requirement in the first place to support both numpy a jax and apply where possible jumpy functions (i.e. ClipActionsV0 which uses `jp.clip`). An alternative could be to leave out the generics LambdaActionV0 and LambdaObservationV0 wrappers from requiring jumpy

gym/dev_wrappers/lambda_action.py
gianlucadecola: Yes I've used it quite a lot for collision with Tuple and Dict spaces, typing.Tuple seems better :+1: 
I'm going to update it also in other part of the code

gianlucadecola:
Updated the PR description and done a bunch of updates according to review and to match the new changes made to gym (removed `new_step_api=True` from test arguments)

gianlucadecola:

Issue #3004, opened 2022-07-29
CarRacing environments can't be pickled


marho13:
### Question
Hi, I was trying to use Ray, and when it crashed Multi-processing for running multiple AI agents in gym.
But I received an error which was quite weird, namely that it couldn't find the track in the gym environment.
Looking into it more, it seems that the CarRacing environments are not serializable using the pickle library at least.
Is it supposed to be serializable?

To recreate the problem check the pastebin:
https://pastebin.com/1jw6XC0J

For an example, with multi-threaded environments, here is the pastebin for it:
https://pastebin.com/LNAMbrW2
The error I receive is shown in the image
![Multiprocessing_bug_report](https://user-images.githubusercontent.com/19758684/181731068-8d9eda08-79cb-4655-a7f3-213ca16a90dd.PNG)

Hope to hear from you soon

pseudo-rnd-thoughts:
The issue seems to be related to the Friction detector using a reference to the env where the track has been initialised yet
```python
env = gym.make("CarRacing-v2").unwrapped
print(id(env))
print(id(env.contactListener_keepref.env))

env = pickle.loads(pickle.dumps(gym.make("CarRacing-v2").unwrapped))
print(id(env))
print(id(env.contactListener_keepref.env))
```

Interestingly, bipedal walker and lunar lander have similar implementations but no one has reported issues with it before

marho13:
> The issue seems to be related to the Friction detector using a reference to the env where the track has been initialised yet
> 
> ```python
> env = gym.make("CarRacing-v2").unwrapped
> print(id(env))
> print(id(env.contactListener_keepref.env))
> # These are equal 
> 
> env = pickle.loads(pickle.dumps(gym.make("CarRacing-v2").unwrapped))
> print(id(env))
> print(id(env.contactListener_keepref.env))
> # These are different
> ```
> 
> The solution is probably to move the initialisation of the track to `init` and out of `reset` but this won't fix the issue that the detectors reference different environments
> 
> Interestingly, bipedal walker and lunar lander have similar implementations but no one has reported issues with it before

I think someone may have found the bug previously, as in gym 0.24.1 we find this code in Bipedal:
```python
self.world.contactListener_bug_workaround = ContactDetector(self)
self.world.contactListener = self.world.contactListener_bug_workaround
```
Something very similar is found in LunarLander as well, so I guess for now ill just implement it in the CarRacing environment and see if it works :P Thanks a lot for your help :)

[...]

PR #3006, opened 2022-07-29
Fix for pickling environments

pseudo-rnd-thoughts:
Issue #3004 noted that car racing cannot be pickled. 
This PR adds a test for all gym environments to check if the environment can be pickled and reset / step can be run

The tests also found that several mujoco environments failed due to the `EzPickle` misunderstand `kwargs` as actual keyword. 

- [ ] Fix the Car Racing environment

### DIFF
--- a/tests/envs/test_envs.py
+++ b/tests/envs/test_envs.py
@@ -4,7 +4,8 @@ import pytest
 
 import gym
 from gym.envs.registration import EnvSpec
-from gym.utils.env_checker import check_env
+from gym.utils import seeding
+from gym.utils.env_checker import check_env, data_equivalence
 from tests.envs.utils import (
     all_testing_env_specs,
     all_testing_initialised_envs,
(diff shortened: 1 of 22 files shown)


tests/envs/test_envs.py
```diff
@@ -120,3 +128,20 @@ def test_render_modes(spec):
...
     env.close()
+
+
+@pytest.mark.parametrize(
+    "env",
+    all_testing_initialised_envs,
+    ids=[env.spec.id for env in all_testing_initialised_envs],
+)
+def test_pickle_env(env: gym.Env):
+    env.np_random, _ = seeding.np_random(0)
```
RedTachyon, line 139: You're not resetting the env before pickling which is likely to mess things up (and I don't think it's supported behavior anyways, since you need to reset an env before doing literally anything with it)

pseudo-rnd-thoughts: I believe that pickling is used by RLLib vectorisation code so they make environments then pickle for making multiple copies.

pseudo-rnd-thoughts:
Closing in favor of https://github.com/openai/gym/pull/3025

Issue #3005, opened 2022-07-29
box2d envs with python 3.8, 3.9, 3.10


Xinyu-Wu-0000:
**Describe the bug**

Can't install gym[box2d] with python 3.8, 3.9, 3.10. I try to install it on ubuntu22.04 which shipped with python 3.10. I have to switch to python3.7.

**Code example**
```bash
conda create --prefix ./envs
conda activate ./envs
conda install pip
pip install gym[box2d]
```

**System Info**
ubuntu22.04 
python 3.8, 3.9, 3.10 not work
python 3.7 is ok

**Additional context**

error info:

```bash
Collecting gym[box2d]
  Using cached gym-0.25.1-py3-none-any.whl
Collecting numpy>=1.18.0
  Using cached numpy-1.23.1-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (17.1 MB)
Collecting importlib-metadata>=4.8.0
  Using cached importlib_metadata-4.12.0-py3-none-any.whl (21 kB)
Collecting cloudpickle>=1.2.0
  Using cached cloudpickle-2.1.0-py3-none-any.whl (25 kB)
Collecting gym-notices>=0.0.4
  Using cached gym_notices-0.0.7-py3-none-any.whl (2.7 kB)
Collecting pygame==2.1.0
  Using cached pygame-2.1.0-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (18.3 MB)
Collecting box2d-py==2.3.5
  Using cached box2d-py-2.3.5.tar.gz (374 kB)
  Preparing metadata (setup.py) ... done
Collecting zipp>=0.5
  Using cached zipp-3.8.1-py3-none-any.whl (5.6 kB)
Building wheels for collected packages: box2d-py
  Building wheel for box2d-py (setup.py) ... error
  error: subprocess-exited-with-error
  
  × python setup.py bdist_wheel did not run successfully.
  │ exit code: 1
  ╰─> [16 lines of output]
      Using setuptools (version 61.2.0).
      running bdist_wheel
      running build
      running build_py
      creating build
      creating build/lib.linux-x86_64-3.8
      creating build/lib.linux-x86_64-3.8/Box2D
      copying library/Box2D/__init__.py -> build/lib.linux-x86_64-3.8/Box2D
      copying library/Box2D/Box2D.py -> build/lib.linux-x86_64-3.8/Box2D
[...]

PR #3009, opened 2022-07-31
Add swig as a Box2D dependency

kir0ul:
# Description

This PR adds `swig` as a dependency when installing Gym with Box2D.

Fixes https://github.com/openai/gym/issues/3005#issuecomment-1199649526.

## Type of change

- [x] Dependency change

# Checklist:

### DIFF
--- a/py.Dockerfile
+++ b/py.Dockerfile
@@ -13,7 +13,6 @@ RUN apt-get -y update \
     xvfb \
     patchelf \
     ffmpeg cmake \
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts:
Would we want to pin the swig version?

kir0ul:
> Would we want to pin the swig version? Otherwise LGTM

I pinned it to major version 4.

PR #3011, opened 2022-08-01
Reduce memory consumption and hanging for Dict and Tuple spaces

### LINKED ISSUE #3010: [Bug Report] Action space seeding hangs due to memory constraints
**Describe the bug**
When seeding Dict action spaces, gym hangs.

**Code example**
```
from gym.spaces import Dict, Discrete, MultiDiscrete

space = Dict(a=Discrete(4), b=MultiDiscrete([6, 6, 6, 6, 6, 6]))

space.seed(1626)
```

**System Info**
 * Gym=0.21 (pip)
 * Windows 10 Enterprise H202
 * Python 3.9

**Additional context**
image

### Checklist

olipinski:
# Description

This PR fixes extensive memory consumption and hanging when seeding spaces.

- Iterates over length of subseeds and generates a seed from np_random.integers() instead of np_random.choice().
- Uses np.int32 instead of python int for portability.

Fixes #3010 

## Type of change

# Checklist:

### DIFF
--- a/gym/spaces/dict.py
+++ b/gym/spaces/dict.py
@@ -101,8 +101,15 @@ class Dict(Space[TypingDict[str, Space]], Mapping):
         )  # None for shape and dtype, since it'll require special handling
 
     def seed(self, seed: Optional[Union[dict, int]] = None) -> list:
-        """Seed the PRNG of this space and all subspaces."""
+        """Seed the PRNG of this space and all subspaces.
+
+        This method will generate a (mostly) unique seed for each subspace. If a truly unique seed is required use .seed(list).
+
+        Args:
+            seed: An optional list of ints or int to seed the (sub-)spaces.
+        """
         seeds = []
+
         if isinstance(seed, dict):
             for key, seed_key in zip(self.spaces, seed):
                 assert key == seed_key, print(
@@ -115,14 +122,9 @@ class Dict(Space[TypingDict[str, Space]], Mapping):
                 seeds += self.spaces[key].seed(seed[seed_key])
         elif isinstance(seed, int):
             seeds = super().seed(seed)
-            subseeds = []
-            assert (
-                len(self.spaces) <= np.iinfo(np.int32).max
-            ), f"Expected spaces.length <= 2147483647, got {len(self.spaces)}"
-            while len(subseeds) < len(self.spaces):
-                subseed = self.np_random.integers(low=0, high=np.iinfo(np.int32).max)
-                if subseed not in subseeds:
-                    subseeds.append(subseed)
+            subseeds = self.np_random.integers(
+                np.iinfo(np.int32).max, size=len(self.spaces)
+            )
             for subspace, subseed in zip(self.spaces.values(), subseeds):
                 seeds.append(subspace.seed(int(subseed))[0])
         elif seed is None:
--- a/gym/spaces/dict.py
+++ b/gym/spaces/dict.py
@@ -116,15 +116,13 @@ class Dict(Space[TypingDict[str, Space]], Mapping):
         elif isinstance(seed, int):
             seeds = super().seed(seed)
             subseeds = []
-            x = 0
             assert (
                 len(self.spaces) <= np.iinfo(np.int32).max
-            ), "Spaces too big to seed. Expected spaces.length <= 2147483647"
-            while x < len(self.spaces):
+            ), f"Expected spaces.length <= 2147483647, got {len(self.spaces)}"
+            while len(subseeds) < len(self.spaces):
                 subseed = self.np_random.integers(low=0, high=np.iinfo(np.int32).max)
                 if subseed not in subseeds:
                     subseeds.append(subseed)
-                    x += 1
             for subspace, subseed in zip(self.spaces.values(), subseeds):
                 seeds.append(subspace.seed(int(subseed))[0])
         elif seed is None:
--- a/gym/spaces/dict.py
+++ b/gym/spaces/dict.py
@@ -103,7 +103,10 @@ class Dict(Space[TypingDict[str, Space]], Mapping):
     def seed(self, seed: Optional[Union[dict, int]] = None) -> list:
         """Seed the PRNG of this space and all subspaces.
 
-        This method will generate a (mostly) unique seed for each subspace. If a truly unique seed is required use .seed(list).
+        Depending on the type of seed, the subspaces will be seeded differently
+        * None - All the subspaces will use a random initial seed
+        * Int - The integer is used to seed the `Dict` space that is used to generate seed values for each of the subspaces. Warning, this does not guarantee unique seeds for all of the subspaces.
+        * Dict - Using all the keys in the seed dictionary, the values are used to seed the subspaces. This allows the seeding of multiple composite subspaces (`Dict["space": Dict[...], ...]` with `{"space": {...}, ...}`).
 
         Args:
             seed: An optional list of ints or int to seed the (sub-)spaces.
--- a/gym/spaces/dict.py
+++ b/gym/spaces/dict.py
@@ -115,19 +115,16 @@ class Dict(Space[TypingDict[str, Space]], Mapping):
                 seeds += self.spaces[key].seed(seed[seed_key])
         elif isinstance(seed, int):
             seeds = super().seed(seed)
-            try:
-                subseeds = self.np_random.choice(
-                    np.iinfo(int).max,
-                    size=len(self.spaces),
-                    replace=False,  # unique subseed for each subspace
-                )
-            except ValueError:
-                subseeds = self.np_random.choice(
-                    np.iinfo(int).max,
-                    size=len(self.spaces),
-                    replace=True,  # we get more than INT_MAX subspaces
-                )
-
+            subseeds = []
+            x = 0
+            assert (
+                len(self.spaces) <= np.iinfo(np.int32).max
+            ), "Spaces too big to seed. Expected spaces.length <= 2147483647"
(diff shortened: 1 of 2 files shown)


gym/spaces/dict.py
```diff
@@ -115,19 +115,14 @@ def seed(self, seed: Optional[Union[dict, int]] = None) -> list:
...
-                subseeds = self.np_random.choice(
-                    np.iinfo(int).max,
-                    size=len(self.spaces),
-                    replace=True,  # we get more than INT_MAX subspaces
-                )
-
+            subseeds = []
+            assert (
+                len(self.spaces) <= np.iinfo(np.int32).max
+            ), f"Expected spaces.length <= 2147483647, got {len(self.spaces)}"
```
RedTachyon, line 121: We're introducing a new exception here (so de facto deprecating a new behavior), admittedly for an *extremely* niche use case, and if anyone encounters it I'd be concerned.

Still, it's better to maintain backwards compatibility than not. Can't we go the same way as in the previous version, so if `len(self.spaces) > np.iinfo(np.int32).max`, we just relax the uniqueness property?

gym/spaces/dict.py
```diff
@@ -115,19 +115,14 @@ def seed(self, seed: Optional[Union[dict, int]] = None) -> list:
...
+            while len(subseeds) < len(self.spaces):
```
RedTachyon, line 122: Can you do some profiling on this? Both for small and large numbers of subspaces, compared with the previous implementation. I'm slightly concerned this will be slow, since numpy tends to work better with vectorized operations, and here we're just doing a thing in a loop (which is generally a terrible idea in python)

olipinski: At least from a quick test like below, it seems to be slower indeed. The question is whether this is a big enough slow down, as this only appears for large number of subspaces.

Up to 1000 subspaces.
```
Loop: 38.9758081
Choice: 0.2712062999999958
```

Up to 10 subspaces.
```
Loop: 0.017434799999999973
Choice: 0.014714000000000005
```

Code:

```
import timeit

setup = """
from typing import Tuple, Optional, Any
from gym.spaces import Dict, Discrete, MultiDiscrete
import numpy as np
import timeit

    def __reduce__(self):
        _, init_args, *args = np.random.Generator.__reduce__(self)
        return RandomNumberGenerator._generator_ctor, init_args, *args

    @staticmethod
    def _generator_ctor(bit_generator_name="MT19937"):
        # Workaround method for RandomNumberGenerator pickling, see __reduce__ above.
        # Ported from numpy.random._pickle.__generator_ctor function.
        from numpy.random._pickle import BitGenerators

        if bit_generator_name in BitGenerators:
            bit_generator = BitGenerators[bit_generator_name]
        else:
            raise ValueError(
                f"{bit_generator_name} is not a known BitGenerator module."
            )
        return RandomNumberGenerator(bit_generator())


def np_random(seed: Optional[int] = None):
    seed_seq = np.random.SeedSequence(seed)
    rng = RandomNumberGenerator(np.random.PCG64(seed_seq))
    return rng

spaces = []
for i in range(1, 1000, 10):
    spaces.append(Dict({x:MultiDiscrete([10, 10, 10]) for x in range(i)}))

np_random = np_random(1626)
"""

method1 = """
for s in spaces:
    subseeds = []
    assert (
            len(s.spaces) <= np.iinfo(np.int32).max
    ), f"Expected spaces.length <= 2147483647, got {len(s.spaces)}"
    while len(subseeds) < len(s.spaces):
        subseed = np_random.integers(low=0, high=np.iinfo(np.int32).max)
        if subseed not in subseeds:
            subseeds.append(subseed)
"""

method2 = """
for s in spaces:
    try:
        subseeds = np_random.choice(
            np.iinfo(np.int32).max,
            size=len(s.spaces),
            replace=False,  # unique subseed for each subspace
        )
    except ValueError:
        subseeds = np_random.choice(
            np.iinfo(np.int32).max,
            size=len(s.spaces),
            replace=True,  # we get more than INT_MAX subspaces
        )
"""

print(timeit.timeit(setup=setup,
                    stmt=method1,
                    number=100))

print(timeit.timeit(setup=setup,
                    stmt=method2,
                    number=100))
```

gym/spaces/dict.py
olipinski: So instead of using the assert we would revert back to the old method of np.choice? Though I think even then, if `len(self.spaces) > np.iinfo(np.int32).max` the np.choice would hang, unless we expanded to np.int64?

pseudo-rnd-thoughts:
Doing some speed testing, the fastest non-choice-based solution had ~3-4x slow sample speeds for sizes up to 10,000 which could be possible in the near future with vectorisation.
However, this choice limits the availability to run vectorisation on some laptops and as `seed` is not frequently run during scripts, I would argue that we should change. 
image

Full script
```python
import timeit

import numpy as np

from gym.utils import seeding

# It is important to test upto 10_000 due to vectorisation
space_sizes = [5, 10, 15, 25, 100, 500, 750, 1_000, 1_500, 2_500, 5_000, 7_500, 10_000]
repeats = 20

np_random, _ = seeding.np_random(1626)


def list_single_sample(size):
    def f():
        subseeds = []
        while len(subseeds) < size:
            subseed = np_random.integers(low=0, high=np.iinfo(np.int32).max)
            if subseed not in subseeds:
                subseeds.append(subseed)
    return f


def list_multi_sample(size):
    def f():
        subseeds = list(set(np_random.integers(low=0, high=np.iinfo(np.int32).max, size=size)))
        while len(subseeds) < size:
            subseed = np_random.integers(low=0, high=np.iinfo(np.int32).max)
            if subseed not in subseeds:
                subseeds.append(subseed)
    return f


def set_multi_sample(size):
    def f():
        subseeds = set(np_random.integers(low=0, high=np.iinfo(np.int32).max, size=size))
        while len(subseeds) < size:
            subseed = np_random.integers(low=0, high=np.iinfo(np.int32).max)
            if subseed not in subseeds:
                subseeds.add(subseed)
        return list(subseeds)
    return f


def np_unique_multi_sample(size):
    def f():
        subseeds = list(np.unique(np_random.integers(low=0, high=np.iinfo(np.int32).max, size=size)))
        while len(subseeds) < size:
            subseed = np_random.integers(low=0, high=np.iinfo(np.int32).max)
            if subseed not in subseeds:
                subseeds.append(subseed)
    return f


def np_choice(size):
    def f():
        subseeds = np_random.choice(
            np.iinfo(np.int32).max,
            size=size,
            replace=False,
        )
    return f


fig, axs = plt.subplots(ncols=2, figsize=(10, 5))
axs[0].set_xscale('log')
axs[1].set_xscale('log')

algo_results = {
    algo.__name__: [
        timeit.repeat(algo(_size), number=repeats)
        for _size in space_sizes
    ]
    for algo in [
        # list_single_sample,
        list_multi_sample,
        # set_multi_sample,
        np_unique_multi_sample,
        np_choice
    ]
}


axs[0].set_title("Time taken for each algorithm")
for name, results in algo_results.items():
    axs[0].errorbar(x=space_sizes, y=np.mean(results, axis=1), yerr=np.std(results, axis=1), label=name)
axs[0].legend()

axs[1].set_title("Normalised time taken by np_choice")
for name, results in algo_results.items():
    axs[1].plot(space_sizes, np.mean(results, axis=1) / np.mean(algo_results["np_choice"], axis=1), label=name)
axs[1].legend()

plt.tight_layout()
plt.savefig("issue-3010.png")
plt.show()
```

gym/spaces/dict.py
RedTachyon: > So instead of using the assert we would revert back to the old method of np.choice?

Can't we just use `np_random.integers` or whatever variant of it? Just get random integers within the allowed range, of the required size

pseudo-rnd-thoughts: @RedTachyon Your original idea is interesting where we ignore the uniqueness property such that we just run `integer(seed=len(self.spaces))`

pseudo-rnd-thoughts:
In offline discussions with @RedTachyon, we don't understand the need for uniqueness when setting subspace seeds for an integer. In particular, if using `np.int32` then the chance of this happening is very low. 
Therefore, we propose changing the `Dict` and `Tuple` `.seed(int)` will use `np.integers(np.iinfo(np.int32).max, size=num_spaces)`. This will both improve the performance (about 50%) and reduce memory footprint.
For users who require unique seeds then we recommend users using `.seed(dict)` where every space can be given a unique seed.

gym/spaces/dict.py
```diff
@@ -101,8 +101,15 @@ def __init__(
         )  # None for shape and dtype, since it'll require special handling
 
     def seed(self, seed: Optional[Union[dict, int]] = None) -> list:
-        """Seed the PRNG of this space and all subspaces."""
+        """Seed the PRNG of this space and all subspaces.
+
+        This method will generate a (mostly) unique seed for each subspace. If a truly unique seed is required use .seed(list).
```
pseudo-rnd-thoughts, line 106: Could you replace with below and for Tuple
```
Depending on the type of seed, the subspaces will be seeded differently
* None - All of the subspaces will use a random initial seed
* Int - The integer is used to seed the `Dict` space that is used to generate seed values for each of the subspaces. Warnings, this does not guarantee unique seeds for all of the subspaces. 
* Dict - Using all of the keys in the seed dictionary, the values are used to seed the subspaces. This allows the seeding of multiple composite subspaces (`Dict["space": Dict[...], ...]` with `{"space": {...}, ...}`).

olipinski: Updated!

PR #3013, opened 2022-08-01
TimeAwareObservationV0 Wrapper

gianlucadecola:
This PR adds TimeAwareObservationV0 to the new set of dev_wrappers.

Compared to the old TimeAwareObservation this wrapper uses a Dict observation space with keys `time` and `obs`.  
The new wrappers support all types of spaces and vectorized environments.  

The wrapper supports only the new step API. If backward compatibility is required for dev_wrappers I will add it.


RedTachyon:
Can we maybe add a built-in flattening functionality? E.g. `TimeAwareObservationV0(env, flatten=True)`. Most implementations play well only with Box observations, so it's an extra hassle to convert from the dict

pseudo-rnd-thoughts:
I was thinking, if the environment obs space is a box and vector (len(shape) == 1) then append to the obs. Otherwise use the Dict space?

gianlucadecola:
> I was thinking, if the environment obs space is a box and vector (len(shape) == 1) then append to the obs. Otherwise use the Dict space?

I think this may be part of the `flatten=True` functionality proposed by @RedTachyon, so, if we want the default behavior the observation space is always `Dict` but if we set `flatten=True` it will always be the flattened `Box` with time appended at the end.
I found this to be less confusing than automatically switching to `Box` or `Dict` depending on the space shape.

gianlucadecola:
Updated with flattening functionality :+1:

gym/dev_wrappers/time_aware_observation.py
```diff
@@ -0,0 +1,105 @@
...
+    def observation(self, observation: ObsType):
+        """Adds to the observation with the current time step.
+
+        Args:
+            observation: The observation to add the time step to
+
+        Returns:
+            The observation with the time step appended to
+        """
+        observation = OrderedDict(obs=observation, time=self.t)
```
araffin, line 68: why not use the remaining time (as suggested by https://arxiv.org/abs/1712.00378 or done in https://sb3-contrib.readthedocs.io/en/master/common/wrappers.html#timefeaturewrapper)
unbounded observation sounds like a bad idea.

RedTachyon: I'm on board with that, either remaining time, or normalized by the max time.

pseudo-rnd-thoughts: Have another parameter, normalise time that is by default true? Otherwise returns the integer remaining time steps?

gianlucadecola:
I updated the wrapper, the only concern I'm having is that now we have introduced coupling between this wrapper and the `TimeLimit` wrapper by accessing the attribute `_max_episode_steps` but I think if we want to get bounded time this is the only way and probably not much of a problem since `TimeLimit` wraps the env by default.

gianlucadecola:
Updated PR to match new gym version by removing the `new_step_api=True` args

gianlucadecola:

PR #3015, opened 2022-08-02
Type hint mujoco_env.py and update pyright

pseudo-rnd-thoughts:
This PR removes `mujoco_env.py` from the set of ignored files such that every gym file is tested
In addition, this PR adds comments to the pyproject.toml file for the number of warnings that each pyright disable parameter would cause. 
This should help with the type hinting of the project in the future
I have removed unnecessary pyright argument that do not raise any new warnings or errors as the original pyright argument were just copied from another project will limited investigation into their effectiveness.


RedTachyon:
Looks good

Issue #2905, opened 2022-06-17
[Proposal] Switch to MoviePy for RecordVideo


younik:
### Proposal 

Remove `VideoRecorder` and use MoviePy for `RecordVideo` wrapper.

### Motivation

Right now, the `VideoRecorder` code is a bit messy and difficult to maintain. There are external libraries, such as `MoviePy` to which we can delegate the workflow to save a video starting from frames.

See also https://github.com/openai/gym/pull/2894

### Alternatives

Leave the code as it is.

PR #3016, opened 2022-08-02
Add save_video util and deprecate RecordVideo in favor of it

younik:
This PR adds a deprecation warning to RecordVideo, in favor of a cleaner util function https://github.com/younik/gym/blob/record-video/gym/utils/save_video.py

See also https://github.com/openai/gym/issues/2905

Sample usage:

```python
import gym
from gym.utils.save_video import save_video

env = gym.make("MyEnv", render_mode="rgb_array")

env.reset()
step_starting_index = 0
episode_index = 0
for step_index in range(199):
    _, _, done, _ = env.step(action)

    if done:
        save_video(
             env.render(), 
             "videos", 
             fps=env.metadata["render_fps"], 
             step_starting_index=step_starting_index, 
             episode_index=episode_index
        )
        step_starting_index = step_index + 1
        episode_index += 1
        env.reset()

env.close()
```


gym/utils/save_video.py
```diff
@@ -0,0 +1,98 @@
...
+        In case of multiple episode, the expected input is a List[List[RenderFrame]].
+        video_folder (str): The folder where the recordings will be stored
+        episode_trigger: Function that accepts an integer and returns ``True`` iff a recording should be started at this episode
+        step_trigger: Function that accepts an integer and returns ``True`` iff a recording should be started at this step
+        video_length (int): The length of recorded episodes. If -1, entire episodes are recorded.
+            Otherwise, snippets of the specified length are captured.
+        name_prefix (str): Will be prepended to the filename of the recordings.
+        episode_starting_index (int): The index of the first episode in frames.
+        step_starting_index (int): The step index of the first frame.
+        fps (float): Frame per second of the video.
```
pseudo-rnd-thoughts, line 60: Add kwargs

tests/utils/test_save_video.py
```diff
@@ -0,0 +1,91 @@
...
+    def thunk():
+        env = gym.make(gym_id, disable_env_checker=True, **kwargs)
+        env._max_episode_steps = 20
+        return env
+
+    return thunk
+
+
+def test_record_video_within_vector():
+    envs = gym.vector.SyncVectorEnv(
```
pseudo-rnd-thoughts, line 69: `gym.vector.make("CartPole-v1", nenv=2, "async", render_mode="rgb_array")`

pseudo-rnd-thoughts (changes requested):
Generally looks good, I added two comments for minor code changes

vwxyzjn:
How does this API work with vectorized environments?

In CleanRL currently we have something like this:

```python
def make_env(env_id, seed, idx, capture_video, run_name):
    def thunk():
        env = gym.make(env_id)
        env = gym.wrappers.RecordEpisodeStatistics(env)
        if capture_video:
            if idx == 0:
                env = gym.wrappers.RecordVideo(env, f"videos/{run_name}")
        env.seed(seed)
        env.action_space.seed(seed)
        return env

    return thunk
```
https://github.com/vwxyzjn/cleanrl/blob/df3ab2a1843b906c6e528d37e1eb0a34e831e5db/cleanrl/ppo.py#L82-L94

So we can only record videos for the first env in the vectorized environment. How does the proposed API support it?

Personally, I feel this is adding more code in the learning loop. The most common use case for `RecordVideo` is to record videos at a reasonable schedule, like recording videos for the 1st, 3rd, 9th, 27th, .... 100th episodes. Given the proposed API, the learning loops would need to maintain more boilerplate code to do it.

What is the motivation for this change? It should perhaps serve a practical purpose other than just making things look cleaner.

gym/utils/save_video.py
```diff
@@ -0,0 +1,109 @@
...
+    **kwargs,
+):
+    """Save videos from rendering frames.
+
+    This function extract video from a list of render frame episodes.
+
+    Args:
+        frames (List[RenderFrame]): A list of frames to compose the video.
+        video_folder (str): The folder where the recordings will be stored
+        episode_trigger: Function that accepts an integer and returns ``True`` iff a recording should be started at this episode
```
vwxyzjn, line 51: I am very confused - why would `episode_trigger` still be here? It seems strange when users execute `save_video` the videos might not be saved…

gym/utils/save_video.py
```diff
@@ -0,0 +1,109 @@
...
+        >>> env = gym.make("FrozenLake-v1", render_mode="rgb_array")
+        >>> env.reset()
+        >>> step_starting_index = 0
+        >>> episode_index = 0
+        >>> for step_index in range(199):
+        ...    action = env.action_space.sample()
+        ...    _, _, done, _ = env.step(action)
+        ...    if done:
+        ...       save_video(
+        ...          env.render(),
```
vwxyzjn, line 73: It looks like here only calls `env.render()` when the env is done. Does this mean when calling `env = gym.make("FrozenLake-v1", render_mode="rgb_array")` each frame is rendered and cached? If this is the case this could be an undesirable bottleneck — we might only keep the rendered frames of a few episodes.

younik:
> Thanks @younik for preparing the PR. What is the motivation for this change and why is using the `save_video` function cleaner? It looks to me `save_video` only adds more boilerplate code to the learning code: we need to maintain `step_starting_idx` and `episode_idx` and probably a scheduler logic to not record every episode.
> 
> Deprecating `RecordVideo` should perhaps serve a practical purpose.

Hello, yes, this adds more boilerplate to user code, on the other side, `save_video` function is clearer than RecordVideo and fits better with the new behavior of `rgb_array`.

gym/utils/save_video.py
younik: In the beginning, the expected type of frames was a List of episodes, where an episode is a List of frames. 
Talking with @pseudo-rnd-thoughts, we agree it is simpler as it is now, with frames as a List of frames of the same episode.

For how it is right now, yes, we can drop `episode_trigger` and the user can just do it with an if statement.
Notice that it still can happen that `save_video` doesn't save anything depending on `step_trigger`

gym/utils/save_video.py
younik: Yes, with the current Render API `env.render()` returns the list of frames of the whole episode when the mode is `rgb_array`; old-like behavior can be obtained with `single_rgb_array`.

When you call `render()` or `reset()`, the frame list is cleaned

vwxyzjn: Sorry, I am struggling to understand this. If I don’t want to render every frame, how can I achieve it with the current API? What is the old-like behavior with `single_rgb_array`?

younik: If you want to render a single frame representing the current state, you can achieve it with `single_rgb_array`, thus similar to before:

```python
env = gym.make("MyEnv", render_mode="single_rgb_array")
env.reset()

for _ in range(n_steps):
    env.step(...)
    env.render()  # single frame

env.close()
```

vwxyzjn:
Could we do both? I think having the `save_video` as a low-level API would be helpful, but `RecordVideo` should be a no-brainer utility to most researchers.

younik:
> Could we do both? I think having the `save_video` as a low-level API would be helpful, but `RecordVideo` should be a no-brainer utility to most researchers.

Sure, I updated `VideoRecorder` to use `moviepy` on the other PR, thus it is easy to see the differences. For me, we can keep both, or stick with one

pseudo-rnd-thoughts:
Personally, I think we should have both.
The wrapper for ease of use and a separate function for users who want more control

vwxyzjn (approved):
LGTM

PR #3022, opened 2022-08-12
removed inline RandomNumberGenerator implementation 

balisujohn:
# Description

Removed inline RandomNumberGenerator implementation and replaced with alias to np.random.Generator.

# Checklist:


gym/utils/seeding.py
```diff
@@ -31,201 +27,4 @@ def np_random(seed: Optional[int] = None) -> Tuple["RandomNumberGenerator", Any]
...
-        raise error.Error(f"Seed must be non-negative, not {bigint}")
-    elif bigint == 0:
-        return [0]
-
-    ints: List[int] = []
-    while bigint > 0:
-        bigint, mod = divmod(bigint, 2**32)
-        ints.append(mod)
-    return ints
+RandomNumberGenerator = np.random.Generator
```
XuehaiPan, line 30:
```suggestion
RNG = RandomNumberGenerator = np.random.Generator
```

I think we shall leave the short name convenience (line 137).

RedTachyon:
Is there actually anything preventing us from removing the `RandomNumberGenerator` abstraction completely, and instead jsut using `np.random.Generator`, which it is anyways? The extra layer of (trivial, but still) complexity doesn't seem necessary.

pseudo-rnd-thoughts:
I think for backward compatibility then we should keep it

RedTachyon:
In that case, let's keep the synonyms (`RNG = RandomNumberGenerator = np.random.Generator`), but let's not use them in the code itself. So e.g. the return type of `np_random` should be `Tuple[np.random.Generator, Any]` instead of `Tuple["RandomNumberGenerator", Any]`. This won't break any backwards compatibility because it's the same types, but it will keep the codebase a bit more sane.

gym/utils/seeding.py
balisujohn: Sounds good

pseudo-rnd-thoughts (approved):
LGTM

pickettgoogle (commented):
Removing hash_seed from utils/seeding.py breaks gym-retro, which looks for this function there.

pseudo-rnd-thoughts:
You will need to install an older version of gym

PR #3024, opened 2022-08-13
Consider key order in `spaces.Dict.__eq__`

### LINKED ISSUE #3023: [Bug Report] Total order of keys for `spaces.Dict`
If you are submitting a bug report, please fill in the following details and use the tag [bug].

**Describe the bug**

Pair of issue google/jax#11871. A total order is required for dictionary keys.

Equal `Dict` spaces `dict1 == dict2` do not imply `dict1.sample() == dict2.sample()` and `flatten_space(dict1) == flatten_space(dict2)`.

```python
In [1]: from gym import spaces

In [2]: dict1 = spaces.Dict({1: spaces.Box(0, 1, shape=(1,)), 'a': spaces.Box(0, 2, shape=(1,))})

In [3]: dict2 = spaces.Dict({'a': spaces.Box(0, 2, shape=(1,)), 1: spaces.Box(0, 1, shape=(1,))})

In [4]: dict1
Out[4]: Dict(1: Box(0.0, 1.0, (1,), float32), a: Box(0.0, 2.0, (1,), float32))

In [5]: dict2
Out[5]: Dict(a: Box(0.0, 2.0, (1,), float32), 1: Box(0.0, 1.0, (1,), float32))

In [6]: dict1 == dict2
Out[6]: True
```

Not equally seeded in subspaces:

```python
In [7]: dict1.seed(0)
Out[7]: [0, 2488343231644625808, 5874934615388537134]

In [8]: dict2.seed(0)
Out[8]: [0, 2488343231644625808, 5874934615388537134]

In [9]: dict1.sample() == dict2.sample()
Out[9]: False
```

Different order while flattening:

```python
In [10]: spaces.flatten_space(dict1)
Out[10]: Box(0.0, [1. 2.], (2,), float32)

In [11]: spaces.flatten_space(dict2)
Out[11]: Box(0.0, [2. 1.], (2,), float32)

In [12]: spaces.flatten_space(dict1) == spaces.flatten_space(dict2)
Out[12]: False
```

The order of keys is important when seeding the subspaces and flattening the space.

------

In `spaces.Dict.__init__` method, we always convert the inputs into an `OrderedDict`:

https://github.com/openai/gym/blob/8b744130bc25d0f4fe2eac8bce8e47527a8eee2c/gym/spaces/dict.py#L82-L92

However, function `sorted` will fail when sorting with uncomparable types (e.g. `int` vs. `str`):

```python
In [1]: d = {1: 1, 'a': 2}

In [2]: sorted(d.items())
TypeError: '<' not supported between instances of 'str' and 'int'
```

So we add a failback choice at line 88 in PR #2491. This means when the keys are not sortable, the keys are ordered by the insertion order (since Python 3.6). However, the order of keys is important when seeding the subspaces and flattening the space.

`seed`: (seed in order of `dict.spaces.values()`)

https://github.com/openai/gym/blob/8b744130bc25d0f4fe2eac8bce8e47527a8eee2c/gym/spaces/dict.py#L131-L135

`flatten_space`: (flatten in order of `dict.spaces.values()`)

https://github.com/openai/gym/blob/8b744130bc25d0f4fe2eac8bce8e47527a8eee2c/gym/spaces/utils.py#L326-L333

This will cause `dict1 == dict2` does not imply `flatten_space(dict1) == flatten_space(dict2)`.

**Code example**
Please try to provide a minimal example to reproduce the bug. Error messages and stack traces are also helpful.

See Description above.

**System Info**
 * Describe how Gym was installed: pip
 * What OS/version of Linux you're using: Ubuntu 20.04 LTS
 * Python version: 3.9

**Additional context**

See also:

- google/jax#4085
- google/jax#11871

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/openai/gym/issues) in the repo (**required**)

XuehaiPan:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

~Add type check for keys in `spaces.Dict`. Only a strict set of Python builtin types are allowed to be used as keys.~

```python
# Python builtin types (hashable) that can be used as keys in a Dict space.
# Note: bool is a subclass of int, so it is included here. Other hashable
# builtin types, such as range, enumerate, frozenset, and tuple are removed
# here because they are rarely used as keys and may cause errors.
ALLOWED_KEY_TYPES = (type(None), int, float, complex, str, bytes)
```

~This makes we can have a total order of the keys. And `spaces.Dict` is always sorted in the same order.~

~1. Add `key.__class.__qualname__` to sort keys to support comparing between different types.~
2. Consider key order in `spaces.Dict.__eq__` using `OrderedDict.__eq__`.

Fixes #3023

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

# Checklist:

- [X] I have run the [`pre-commit` checks](https://pre-commit.com/) with `pre-commit run --all-files` (see `CONTRIBUTING.md` instructions to set it up)
- [X] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/gym/spaces/dict.py
+++ b/gym/spaces/dict.py
@@ -81,19 +81,12 @@ class Dict(Space[TypingDict[str, Space]], Mapping):
 
         if spaces is None:
             spaces = spaces_kwargs
-        if isinstance(spaces, dict) and not isinstance(spaces, OrderedDict):
+        if isinstance(spaces, Mapping) and not isinstance(spaces, OrderedDict):
             try:
-                spaces = OrderedDict(
-                    # Add `key.__class__.__qualname__` to support sorting between
-                    # different types (e.g. `int` vs. `str`)
-                    sorted(
-                        spaces.items(),
-                        key=lambda kv: (kv[0].__class__.__qualname__, kv[0]),
-                    )
-                )
+                spaces = OrderedDict(sorted(spaces.items()))
             except TypeError:
-                # Non-sortable user-defined key types found. The keys are
-                # remaining in the insertion order.
+                # Incomparable types (e.g. `int` vs. `str`) or user-defined types found.
+                # The keys are remaining in the insertion order.
                 spaces = OrderedDict(spaces.items())
         if isinstance(spaces, Sequence):
             spaces = OrderedDict(spaces)
--- a/gym/spaces/dict.py
+++ b/gym/spaces/dict.py
@@ -193,14 +193,16 @@ class Dict(Space[TypingDict[str, Space]], Mapping):
 
     def __repr__(self) -> str:
         """Gives a string representation of this space."""
-        return "Dict(" + ", ".join([f"{k}: {s}" for k, s in self.spaces.items()]) + ")"
+        return (
+            "Dict(" + ", ".join([f"{k!r}: {s}" for k, s in self.spaces.items()]) + ")"
+        )
 
     def __eq__(self, other) -> bool:
         """Check whether `other` is equivalent to this instance."""
         return (
             isinstance(other, Dict)
             # Comparison of `OrderedDict`s is order-sensitive
-            and (self.spaces == other.spaces)
+            and self.spaces == other.spaces  # OrderedDict.__eq__
         )
 
     def to_jsonable(self, sample_n: list) -> dict:
--- a/gym/spaces/dict.py
+++ b/gym/spaces/dict.py
@@ -50,6 +50,12 @@ class Dict(Space[TypingDict[str, Space]], Mapping):
     implemented to deal with :class:`Dict` actions.
     """
 
+    # Python builtin types (hashable) that can be used as keys in a Dict space.
+    # Note: bool is a subclass of int, so it is included here. Other hashable
+    # builtin types, such as range, enumerate, frozenset, and tuple are removed
+    # here because they are rarely used as keys and may cause errors.
+    ALLOWED_KEY_TYPES = (type(None), int, float, complex, str, bytes)
+
     def __init__(
         self,
         spaces: Optional[TypingDict[str, Space]] = None,
@@ -82,10 +88,16 @@ class Dict(Space[TypingDict[str, Space]], Mapping):
         if spaces is None:
             spaces = spaces_kwargs
         if isinstance(spaces, dict) and not isinstance(spaces, OrderedDict):
-            try:
-                spaces = OrderedDict(sorted(spaces.items()))
-            except TypeError:  # raise when sort by different types of keys
-                spaces = OrderedDict(spaces.items())
+            assert all(
+                isinstance(key, self.ALLOWED_KEY_TYPES) for key in spaces
+            ), f"Dict keys must be one of the following types: {tuple(typ.__name__ for typ in self.ALLOWED_KEY_TYPES)}"
+            spaces = OrderedDict(
+                # Add `key.__class__.__qualname__` to support sorting between
+                # different types (e.g. `int` vs. `str`)
+                sorted(
+                    spaces.items(), key=lambda kv: (kv[0].__class__.__qualname__, kv[0])
+                )
+            )
         if isinstance(spaces, Sequence):
             spaces = OrderedDict(spaces)
 
--- a/gym/spaces/dict.py
+++ b/gym/spaces/dict.py
@@ -50,12 +50,6 @@ class Dict(Space[TypingDict[str, Space]], Mapping):
     implemented to deal with :class:`Dict` actions.
     """
 
-    # Python builtin types (hashable) that can be used as keys in a Dict space.
-    # Note: bool is a subclass of int, so it is included here. Other hashable
-    # builtin types, such as range, enumerate, frozenset, and tuple are removed
-    # here because they are rarely used as keys and may cause errors.
-    ALLOWED_KEY_TYPES = (type(None), int, float, complex, str, bytes)
-
     def __init__(
         self,
         spaces: Optional[TypingDict[str, Space]] = None,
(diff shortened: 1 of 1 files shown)


pseudo-rnd-thoughts:
Personally, I thought we should update the `__eq__` function instead of modifying the init

RedTachyon:
I'm very much not a fan of this solution, we're restricting functionality to fix a very weird (and in general debatable) edge case. I don't see any general reason to disallow arbitrary hashable types to be keys in a Dict, just like in a regular dictionary.

XuehaiPan:
> Personally, I thought we should update the __eq__ function instead of modifying the init

I'm fine with both solutions. But note that:

**Solution 1:** add total order of keys:

```python
     dict1 == dict2  # Python dict
<==> spaces.Dict(dict1) == spaces.Dict(dict2)
<==> flatten_space(spaces.Dict(dict1)) == flatten_space(spaces.Dict(dict2))
```

**Solution 2:** consider key order in `__eq__`:

```python
      dict1 == dict2  # Python dict
<=/=> spaces.Dict(dict1) == spaces.Dict(dict2)
```

and

```python
     spaces.Dict(dict1) == spaces.Dict(dict2)
<==> flatten_space(spaces.Dict(dict1)) == flatten_space(spaces.Dict(dict2))
```

XuehaiPan:
> I don't see any general reason to disallow arbitrary hashable types to be keys in a Dict, just like in a regular dictionary.

I removed the key type restriction but kept adding the sort keys to have the total order of Python builtin types. Having a total order and considering the key order in `__eq__` comparison are not in conflict with each other. So I kept both.

gym/spaces/dict.py
```diff
@@ -193,6 +202,14 @@ def __repr__(self) -> str:
         """Gives a string representation of this space."""
         return "Dict(" + ", ".join([f"{k}: {s}" for k, s in self.spaces.items()]) + ")"
 
+    def __eq__(self, other) -> bool:
+        """Check whether `other` is equivalent to this instance."""
+        return (
+            isinstance(other, Dict)
+            # Comparison of `OrderedDict`s is order-sensitive
```
pseudo-rnd-thoughts, line 209: Why are we not checking the keys are the same? 
Strange that we only checking the spaces

XuehaiPan: Note that `Dict.spaces` is always a `OrderedDict`. `OrderedDict.__eq__` checks both keys and values (implemented `dict.__eq__`). In addition to `dict.__eq__`, the key order is also required by `OrderedDict.__eq__` checks.

gym/spaces/dict.py
```diff
@@ -83,8 +83,17 @@ def __init__(
             spaces = spaces_kwargs
         if isinstance(spaces, dict) and not isinstance(spaces, OrderedDict):
             try:
-                spaces = OrderedDict(sorted(spaces.items()))
-            except TypeError:  # raise when sort by different types of keys
+                spaces = OrderedDict(
+                    # Add `key.__class__.__qualname__` to support sorting between
+                    # different types (e.g. `int` vs. `str`)
+                    sorted(
```
RedTachyon, line 89: I still don't like this, it actively restricts usage for no good reason, and is done via an extremely hacky method. Imo this should be kept as it was earlier, and the only change in this PR would be the `__eq__` method.

gym/spaces/dict.py
pseudo-rnd-thoughts: That makes sense

gym/spaces/dict.py
XuehaiPan: The semantic to change the `key` argument in the `sorted` function is to make:

Equal `dict`s imply equal `Dict` spaces output.

```python
dict1 == dict2 <==> spaces.Dict(dict1) == spaces.Dict(dict2)
```

If we don't allow sorting with different types but just update the `__eq__` method, we will get:

```python
# Python dict
{1: subspace1, 'a': subspace2} == {'a': subspace2, 1: subspace1}
```

but

```python
Dict(1: subspace1, 'a': subspace2) != Dict('a': subspace2, 1: subspace1)
```

Note that if the users always pass the keys in the same type, the new implementation here behaves just like the old one.

------

If the users want to get full control of the key order in the `spaces.Dict`, they should pass the argument as `OrderedDict`:

```python
spaces.Dict(spaces=OrderedDict([
    (key1, subspace1),
    ...
    (keyN, subspaceN),
]))
```

We only sort keys if the argument `spaces` is passed by vanilla Python `dict` or unordered `Mapping`. Which are:

```python
spaces.Dict(spaces={key1: subspace1, ..., keyN: subspaceN})  # arbitrary key type
spaces.Dict(key1=subspace1, ..., keyN=subspaceN)  # keys are `str` and sortable
```

The latter one is always sortable because the keys are `str`s.

pseudo-rnd-thoughts: Im uncertain about this PR. 
The downside is that this isn't backwards compatible which is only in a couple of very niche cases.
The upside is that this fixes a bug that could occur in those niche cases. 

I think we have two options.
1. We explicitly don't support multi different types of keys for Dict spaces, therefore won't need the PR
2. Support multi different types and accept this PR.

@XuehaiPan In what cases do the multi different types of keys occur? The only way that I can think of this occurring is by using custom objects as the keys.

RedTachyon:
My opinion is still that this PR should only override `__eq__` and that's all.

XuehaiPan:
I removed the `key` argument in `sorted` function for backward compatibility.

After this PR:

Equal `Dict` spaces imply equal flattened spaces:

```python
     dictspace1 == dictspace2
<==> flatten_space(dictspace1) == flatten_space(dictspace2)
```

But equal `dict` mappings do not imply equal `Dict` spaces:

```python
      dict1 == dict2  # Python unordered mapping
<=/=> spaces.Dict(dict1) == spaces.Dict(dict2)
```

gym/spaces/dict.py
```diff
@@ -191,7 +193,17 @@ def __len__(self) -> int:
 
     def __repr__(self) -> str:
         """Gives a string representation of this space."""
-        return "Dict(" + ", ".join([f"{k}: {s}" for k, s in self.spaces.items()]) + ")"
+        return (
+            "Dict(" + ", ".join([f"{k!r}: {s}" for k, s in self.spaces.items()]) + ")"
```
XuehaiPan, line 197: Use `repr` (`!r`) to get better representation for `str` keys:

Without `!r`, i.e., use `!s`:

```python
>>> str(Dict(a=subspaces1, b=subspace1))
Dict(a: subspaces1, b: subspace1)
```

With `!r`:

```python
>>> str(Dict(a=subspaces1, b=subspace1))
Dict('a': subspaces1, 'b': subspace1)
```

pseudo-rnd-thoughts: Good idea

pseudo-rnd-thoughts (approved):
LGTM

Issue #143, opened 2016-05-30
Box2d episode 5: Segfault strikes back


justheuristic:
A few ticks ago there was an [issue](https://github.com/openai/gym/issues/114) about 2 `BipedalWalker-v1` in parallel causing SegFault whence the first was reset() after the second got initialized.

Than there was a fix that finally allowed us to reset several Box2D `BipedalWalker-v2` 's, but unfortunately we still can't play them.

The new bone of contention looks thusly:
```
import gym
def segfault(env_name = "BipedalWalker-v2"):
    ll0 = gym.make(env_name)
    ll0.reset()
    ll1 = gym.make(env_name)
    ll1.reset()
    ll0.reset()
    ll1.reset()
    ll1.step(ll1.action_space.sample())
    ll0.step(ll1.action_space.sample())
    print "and nothing happened"
```

As usual, it works okay with LunarLander, but fails with BipedalWalker
```
>>> segfault("LunarLander-v2")
[2016-05-30 14:42:40,877] Making new env: LunarLander-v2
libGL error: failed to open drm device: No such file or directory
libGL error: failed to load driver: i965
[2016-05-30 14:42:43,056] Making new env: LunarLander-v2
and nothing happened
```

```
>>> segfault("BipedalWalker-v2")
[2016-05-30 14:43:07,731] Making new env: BipedalWalker-v2
[2016-05-30 14:43:07,769] Making new env: BipedalWalker-v2
Segmentation fault
```

A single bipedal walker works nice and stable in the same setup.

jietang:
@olegklimov, any ideas?

olegklimov:
I'll look within a few days.

justheuristic:
olegklimov@ any ideas so far?

olegklimov:
Right now I'm doing the other thing :(  The good news is you can start several environments using that other thing, each one in its own process. Maybe this will help you, to be out soon. But I'll return to this bug anyway, it should be fixed, obviously, crashes are not good.

olegklimov:
Okey, don't ask me why this fixes anything, because I don't know :)  But now I can run your code, in a loop.

jietang:
I don't really understand what the mechanism behind this fix is. It seems like there are some underlying issues behind how _reset() and contactListeners interact. For example it cropped up as https://github.com/openai/gym/issues/114, and it's now manifesting itself in this different way.

[...]

PR #3025, opened 2022-08-13
Fix unpickling Box2D and MuJoCo envs

RedTachyon:
# Description

Replaces #3006, fixing #3004, inspired by #187 which is a trivial fix for #143

I'll copy-paste my message from Discord explaining the underlying bug:
1. There's a collision detector which holds a pointer to the environment itself.
2. When unpickling an env, you first create an env `E1` which calls `__setstate__` provided by `EZPickle`. In `EZPickle`, you instantiate a new environment `E2` using the same arguments as originally. This includes the collision detector pointing to `E2`
3. After `E2` is instantiated, all its fields in `__dict__` are copied over to the `__dict__` of `E1`
4. It includes the collision detector which still points to `E2`
5. The environment that you have access to externally is `E1`
6. You try to reset the env. During the reset, you instantiate `self.track` inside `E1`. Later in the reset, the collision detector tries to use `env.track`, but its `env` still points to `E2`, which was never reset, and thus `self.track` doesn't exist
7. The sloppy coding and initializing `self.track` in `reset` is actually helpful here because it crashes instead of returning incorrect output

The fix does something which was already present in LunarLander and BipedalWalker to solve a bug from ancient history. It boils down to recreating the collision detector during `reset`, which is kinda ugly, but then again, it's Box2D envs, so not much hope for a prettier solution for now.

I included a fix for LunarLander and BipedalWalker which forget to properly handle `EzPickle` leading to unexpected consequences in certain scenarios, so that the tests actually pass.

### DIFF
--- a/gym/envs/box2d/bipedal_walker.py
+++ b/gym/envs/box2d/bipedal_walker.py
@@ -169,7 +169,7 @@ class BipedalWalker(gym.Env, EzPickle):
     }
 
     def __init__(self, render_mode: Optional[str] = None, hardcore: bool = False):
-        EzPickle.__init__(self)
+        EzPickle.__init__(self, render_mode, hardcore)
         self.isopen = True
 
         self.world = Box2D.b2World()
--- a/gym/envs/box2d/car_racing.py
+++ b/gym/envs/box2d/car_racing.py
@@ -200,7 +200,14 @@ class CarRacing(gym.Env, EzPickle):
         domain_randomize: bool = False,
         continuous: bool = True,
     ):
-        EzPickle.__init__(self)
+        EzPickle.__init__(
+            self,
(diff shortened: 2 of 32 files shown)


pseudo-rnd-thoughts:
To fix the CI issue, you need to add this to `Ezpickle.__init__`
```python
# If the environment has kwargs, they will be caught up as a keyword.
#   This remove the kwargs and re-adds all keywords as actual keywords for pickling
if "kwargs" in self._ezpickle_kwargs:
    env_kwargs = kwargs.pop("kwargs")
    for key, value in env_kwargs.items():
        self._ezpickle_kwargs[key] = value
```

pseudo-rnd-thoughts:
LGTM

PR #3027, opened 2022-08-14
Remove old Render API

younik:
Remove support for old render API (kwarg `mode`).

Alternatively to https://github.com/openai/gym/pull/3016, it keeps RecordVideo refactored using MoviePy.
We should decide if we want to keep the Wrapper, the function, or both.


pseudo-rnd-thoughts (commented):
I can't find any issues with this PR. Would be happy to merge, I just want @RedTachyon to quickly review as well.

tests/utils/test_play.py
```diff
@@ -22,6 +22,9 @@ class DummyEnvSpec:
 
 
 class DummyPlayEnv(gym.Env):
+    def __init__(self, render_mode=None):
```
RedTachyon, line 25: Type annotation for the argument? I know it's only a test, but let's keep it clean. I think it'd be `Optional[str]`, right?

RedTachyon (commented):
One minor comment, looks good otherwise

tests/utils/test_play.py
younik: yes, you are right, thanks

Issue #3021, opened 2022-08-11
[Bug Report] gym normalize wrappers are incompatible with envpool


vwxyzjn:
**Describe the bug**
`gym.wrappers.NormalizeObservation` and `gym.wrappers.NormalizeReward` are incompatible with envpool. See https://github.com/sail-sg/envpool/issues/185

**Code example**

```
import numpy as np
import envpool
import gym

envs = envpool.make(
    "HalfCheetah-v4",
    env_type="gym",
    num_envs=4,
)
envs.num_envs = 4
envs.single_action_space = envs.action_space
envs.single_observation_space = envs.observation_space
envs.is_vector_env = True
envs = gym.wrappers.ClipAction(envs)
envs = gym.wrappers.NormalizeObservation(envs)
envs = gym.wrappers.TransformObservation(envs, lambda obs: np.clip(obs, -10, 10))
envs = gym.wrappers.NormalizeReward(envs)
envs = gym.wrappers.TransformReward(envs, lambda reward: np.clip(reward, -10, 10))
obs = envs.reset()
envs.step(np.array([envs.action_space.sample() for _ in range(envs.num_envs)]))
```
```
Traceback (most recent call last):
  File "/home/costa/Documents/go/src/github.com/vwxyzjn/envpool-cleanrl/bug.py", line 22, in <module>
    envs.step(np.array([envs.action_space.sample() for _ in range(envs.num_envs)]))
  File "/home/costa/.cache/pypoetry/virtualenvs/envpool-cleanrl-uAHoRI5J-py3.9/lib/python3.9/site-packages/gym/core.py", line 532, in step
    step_returns = self.env.step(action)
  File "/home/costa/.cache/pypoetry/virtualenvs/envpool-cleanrl-uAHoRI5J-py3.9/lib/python3.9/site-packages/gym/wrappers/normalize.py", line 149, in step
    self.env.step(action), True, self.is_vector_env
  File "/home/costa/.cache/pypoetry/virtualenvs/envpool-cleanrl-uAHoRI5J-py3.9/lib/python3.9/site-packages/gym/core.py", line 493, in step
    step_returns = self.env.step(action)
  File "/home/costa/.cache/pypoetry/virtualenvs/envpool-cleanrl-uAHoRI5J-py3.9/lib/python3.9/site-packages/gym/wrappers/normalize.py", line 77, in step
    obs, rews, terminateds, truncateds, infos = step_api_compatibility(
[...]

PR #3028, opened 2022-08-16
Add testing for step api compatibility functions and wrapper

pseudo-rnd-thoughts:
Issue https://github.com/openai/gym/issues/3021 noted an issue with the step api compatibility. 
Investigating, I found that there was no full testing for the functions.
This PR adds testing with expected values and roundtripping for both conversion functions.

Using `is_vector_env` works for both `dict` and `list` based info

Changes
* For the old done step api, in all cases where termination or truncation is true then "TimeLimit.truncation" is added to info
* As the old done step api can only encode 3 termination and truncation states (we cannot include terminated=True and truncated=True), in cases where terminated=True and truncated=True, we favour terminated=True and set truncated=False.


gym/wrappers/step_api_compatibility.py
```diff
@@ -15,33 +18,33 @@ class StepAPICompatibility(gym.Wrapper):
 
     Args:
         env (gym.Env): the env to wrap. Can be in old or new API
-        new_step_api (bool): True to use env with new step API, False to use env with old step API. (False by default)
+        to_termination_truncation_api (bool): True to use env with new step API, False to use env with old step API. (False by default)
```
arjun-kg, line 21: I think it's better to use `to_terminated_truncated_api` to be consistent with the actual variable names

tests/utils/test_step_api_compatibility.py
```diff
@@ -0,0 +1,160 @@
...
+                0,
+                np.array([False, True]),
+                {},
+            ),
+            np.array([False, True]),
+            np.array([False, False]),
+        ),
+    ),
+)
+def test_to_done_step_api(
```
arjun-kg, line 57: The function names need to be interchanged. This is `test_to_terminated_truncated_step_api`

tests/utils/test_step_api_compatibility.py
```diff
@@ -0,0 +1,160 @@
...
+                np.array([False, True, False]),
+                np.array([False, False, True]),
+                [{}, {}, {}],
+            ),
+            np.array([False, True, True]),
+            np.array([False, False, True]),
+        ),
+    ),
+)
+def test_to_terminated_truncated_step_api(
```
arjun-kg, line 114: Same comment. Change function  name.

arjun-kg (changes requested):
Thank you especially for the super detailed test functions. Looks good except for some minor comments.

arjun-kg:
Also in `step_api_compatibility.py` inside utils, the function names haven't been updated, which seems to be what CI is complaining about.

vwxyzjn:
Thank you! I tested on my end, and the scripts run.

PR #3029, opened 2022-08-16
Corrected episode end description for mujoco/ant

raja-grewal:
Minor correction of episode end conditions for the ant environment as it is inconsistent with all the other mujoco environments.


pseudo-rnd-thoughts:
LGTM

PR #3037, opened 2022-08-22
Fix toy_text rendering on headless machines

### LINKED ISSUE #3032: [Bug Report] Cannot render toy_text envs in Google Colab
**Describe the bug**

The `env.render()` functionality is broken for toy text envs when running in google colab.

**Code example**

Below is a simple example to repro this bug, see also this [notebook](https://colab.research.google.com/drive/1aibrJhzxNN7UwbgdhGQR2YhlUWZjCSYU).

```python
%pip install 'gym==0.25.1' --quiet

import gym
print(gym.__version__)  # 0.25.1
```

```python
env = gym.make('FrozenLake-v1')
env.reset()
env.render()  # error: No available video device
```

The error is triggered when pygame tries to initialize the (nonexistent) display:
```python
pygame.init()          # ok
pygame.display.init()  # fail
```

**System Info**
This is running on google colab ([notebook](https://colab.research.google.com/drive/1aibrJhzxNN7UwbgdhGQR2YhlUWZjCSYU)).

**Additional context**

It looks like the backend for toy text envs was changed to pygame. What was the reason for that change?

RedTachyon:
# Description

Fixes #3032

The pygame code did stuff with the display even in `rgb_array` mode, which makes it impossible to run it on a headless machine. The relevant code has been moved into the `if` condition.

WIP because I need to check if it actually works on colab and on a head...ful? computer, and probably address potential similar problems in other envs.

### DIFF
--- a/gym/envs/toy_text/cliffwalking.py
+++ b/gym/envs/toy_text/cliffwalking.py
@@ -183,7 +183,7 @@ class CliffWalkingEnv(Env):
 
         if self.window_surface is None:
             pygame.init()
-            
+
             if mode == "human":
                 pygame.display.init()
                 pygame.display.set_caption("CliffWalking")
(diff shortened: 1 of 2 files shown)


KristianHolsheimer:
Thanks!

PR #3040, opened 2022-08-23
Rename "rgb_array" to "rgb_array_list" and "single_rgb_array" to "rgb_array"

pseudo-rnd-thoughts:
There was a miscommunication about the changes to render such that "rgb_array" returns list of rendered frames rather than the most recent.
This PR reverts back most of those changes with "rgb_array" returning the most recent frame and adds a new render_mode, "rgb_array_list" that returns a list of all rendered frames since reset.

### DIFF
--- a/tests/envs/test_envs.py
+++ b/tests/envs/test_envs.py
@@ -1,5 +1,6 @@
 import pickle
 
+import numpy as np
 import pytest
 
 import gym
@@ -9,7 +10,6 @@ from tests.envs.utils import (
     all_testing_env_specs,
     all_testing_initialised_envs,
     assert_equals,
-    gym_testing_env_specs,
 )
 
 # This runs a smoketest on each official registered env. We may want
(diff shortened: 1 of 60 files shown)


tests/envs/test_envs.py
```diff
@@ -118,21 +114,58 @@ def test_env_determinism_rollout(env_spec: EnvSpec):
...
+        check_rendered(rendered_frame, "rgb_array")
+    elif mode == "depth_array_list":
+        assert isinstance(rendered_frame, list)
+        for frame in rendered_frame:
+            check_rendered(frame, "depth_array")
+    elif mode == "depth_array":
+        assert isinstance(rendered_frame, np.ndarray)
+        assert len(rendered_frame.shape) == 2
+    else:
+        raise Exception(f"Unknown mode: {mode}")
```
younik, line 145: Do we want to raise an Exception for unknown modes or just ignore/show a warning?
Even if those are tests for in-gym envs, in the spirit we should just ignore it as long as it is specified in metadata (there is another test for it)

younik (approved):
I left some minor comments, but it looks good to me after we fix `test_render_modes`

tests/utils/test_play.py
```diff
@@ -10,31 +10,18 @@
...
-        return obs, rew, terminated, truncated, info
-
-    def reset(self, seed=None):
-        ...
-
-    def render(self):
-        return np.zeros((1, 1))
+PlayableEnv = partial(
+    GenericTestEnv,
+    render_modes=["rgb_array"],
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: Agreed. I will change

tests/envs/test_envs.py
pseudo-rnd-thoughts: I understand what you mean but I think it is more helpful to have errors such that new render modes can be added to test for

PR #3041, opened 2022-08-23
Updates the make and registration functions with better parameters and hints

pseudo-rnd-thoughts:
This PR originally started with fixing the type hints for `EnvSpec` entry_points where the value could be None by default however this won't work.
After this I looked through the whole file and updated / removed the following sections

1. Updated hint types / field for `entry_point` in `EnvSpec`
2. Remove the `__relocated__.py` file and related code. This error seems old and needed removing at some point
3. Updated `make` overloads with all environments. 
4. Removed `EnvRegistry` as noted to be removed at 1.0
5. Added all `EnvSpec` attributes as parameters to `gym.register`


tests/envs/test_register.py
```diff
@@ -143,7 +143,7 @@ def test_env_version_suggestions(
 def test_register_versioned_unversioned():
     # Register versioned then unversioned
     versioned_env = "Test/MyEnv-v0"
-    gym.register(versioned_env)
+    gym.register(versioned_env, None)
```
RedTachyon, line 146: Is this being type checked? Because you're simultaneously disallowing `entry_point=None` in the type system, and adding an explicit `entry_point=None` in the usage.

pseudo-rnd-thoughts: Agreed, I was confused about that as well. I think it is because we disabled `GeneralTypeIssues="none"`.  I will change it to "no-entry-points"

YouJiacheng: If you want more explicit semantic in the type system:
```python
from enum import Enum, auto
class GymEnum(Enum):
    no_entry_points = auto()

...
entry_point: Union[Callable, str, Literal[GymEnum.no_entry_points]]
```

pseudo-rnd-thoughts: This is purely for testing so I think that is unnecessary and could cause more issues for users

pseudo-rnd-thoughts:
@RedTachyon I have readded / kept the make overloads. I investigated testing that the overloads were correct. 
However, I found that you need to use the `typing_extensions.overload` rather than the `typing.overload` which pycharm doesn't recognise.

```python
def test_make_overloads():
    env_make_hints = {}
    for overloaded_func in typing_extensions.get_overloads(gym.make):
        make_type_hints = typing_extensions.get_type_hints(overloaded_func)
        assert (
            "id" in make_type_hints and "return" in make_type_hints
        ), f"Missing type hints for overloaded `gym.make` id and return, {make_type_hints.get('id', 'unknown env')}"
        if not isinstance(make_type_hints["id"], type):
            for env_id in typing_extensions.get_args(make_type_hints["id"]):
                obs_type, act_type = typing_extensions.get_args(
                    make_type_hints["return"]
                )
                env_make_hints[env_id] = (obs_type, act_type)

    assert (
        env_make_hints.keys() == gym.envs.registry.keys()
    ), f"Missing the make overloads envs: {set(gym.envs.registry.keys()) - set(env_make_hints.keys())}. Additional envs: {set(env_make_hints.keys()) - set(gym.envs.registry.keys())}"
    # todo: add type check on the obs and action type
```

tests/testing_env.py
```diff
@@ -49,7 +49,7 @@ def __init__(
         render_modes: Optional[List[str]] = None,
         render_fps: Optional[int] = None,
         render_mode: Optional[str] = None,
-        spec: EnvSpec = EnvSpec("TestingEnv-v0"),
+        spec: EnvSpec = EnvSpec("TestingEnv-v0", None),
```
arjun-kg, line 52: I assume this should also be "no-entry-point"?

arjun-kg (changes requested):
Looks good apart from a minor change

pzhokhov:
@pseudo-rnd-thoughts would it make sense to make a string constant, say, `EMPTY_ENTRY_POINT="no-entry-point"` instead of repeating the same string?

pseudo-rnd-thoughts:
Yes that probaby would have made sense but this is purely in testing and this file alone I believe

PR #3044, opened 2022-08-24
Fix#3043

### LINKED ISSUE #3043: [Bug Report] Mujoco envs offscreen rendering error for width > 640 or height > 480
Separately track issue "the size >500 making the environment crash" from https://github.com/openai/gym/issues/2989#issuecomment-1202214444 here, and describe it more precisely.

The cause is:
https://github.com/openai/gym/blob/6e2e921b5faaa9c356f589e4e938e718824c7d4a/gym/envs/mujoco/mujoco_rendering.py#L77-L78
https://github.com/openai/gym/blob/6e2e921b5faaa9c356f589e4e938e718824c7d4a/gym/envs/mujoco/mujoco_rendering.py#L63-L67
Precisely, `self.con.free()` will reset the struct(strangely, it will save and restore some fields) thus `con->offFBO` will be `0`, then
https://github.com/deepmind/mujoco/blob/090fe2db3f23ad608f833e3e8e9cb049e4daaa87/src/render/render_gl2.c#L37-L46
will set window buffer instead of offscreen buffer.
In principle, "[call ctx.free(), after which point it is the user’s responsibility to ensure that no further rendering calls are made on the context.](https://mujoco.readthedocs.io/en/latest/python.html#rendering)", we should create a new context after `free`.
## Fix
```python
  def update_offscreen_size(self, width, height):
      if width != self.con.offWidth or height != self.con.offHeight:
          self.model.vis.global_.offwidth = width
          self.model.vis.global_.offheight = height
          self.con.free()
          self.con = mujoco.MjrContext(self.model, mujoco.mjtFontScale.mjFONTSCALE_150)
          self._set_mujoco_buffers()
```
This snippet can fix this issue on `glfw`, but I wonder why `GLContext` with wrong `max_width` and `max_height`(since correct width and height is not passed to `_get_viewer`) doesn't cause any problem. (`egl` doesn't need it, but `glfw` need it)
@pseudo-rnd-thoughts

YouJiacheng:
I try to minimize diff in this PR, thus there are some boilerplate codes.
Fix #3043 
We might use `self._width` instead of `self.width` to inform user not modifying it.
Alternatively we can override `__setattr__` to ensure given attribute can only write once.

@rodrigodelazcano

### DIFF
--- a/gym/envs/mujoco/mujoco_env.py
+++ b/gym/envs/mujoco/mujoco_env.py
@@ -49,7 +49,9 @@ class BaseMujocoEnv(gym.Env):
         if not path.exists(self.fullpath):
             raise OSError(f"File {self.fullpath} does not exist")
 
-        self._initialize_simulation()
+        self.width = width
+        self.height = height
+        self._initialize_simulation()  # may use width and height
 
         self.init_qpos = self.data.qpos.ravel().copy()
         self.init_qvel = self.data.qvel.ravel().copy()
@@ -76,8 +78,6 @@ class BaseMujocoEnv(gym.Env):
         self.render_mode = render_mode
         render_frame = partial(
             self._render,
-            width=width,
-            height=height,
             camera_name=camera_name,
             camera_id=camera_id,
         )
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts (approved):
Looks good to merge
@YouJiacheng  Could you update the title to make it more helpful

PR #3046, opened 2022-08-28
Make moviepy an optional dependancy

pseudo-rnd-thoughts:
Currently, moviepy is an optional dependency in `setup.py` however as it is imported in the RecordWrapper then it is a required import.
This PR tries importing `moviepy` during `RecordVideo.__init__` such that the user doesn't get an error when the video is saved on `close`. 
Then when `moviepy` is required in `close`, the `ImageSequence` is imported locally.

### DIFF
--- a/gym/wrappers/monitoring/video_recorder.py
+++ b/gym/wrappers/monitoring/video_recorder.py
@@ -7,13 +7,6 @@ from typing import List, Optional
 
 from gym import error, logger
 
-try:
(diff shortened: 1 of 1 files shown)


wookayin:
The commit message sounds weird..

pseudo-rnd-thoughts:
What is weird?

wookayin:
"initial commit"

PR #3048, opened 2022-08-29
Add testing for Fundamental spaces with full coverage

pseudo-rnd-thoughts:
Update all of the fundamental spaces (`Box`, `Discrete`, `MultiDiscrete`, `MultiBinary`, `Text`) with full coverage
This is the second part of https://github.com/openai/gym/pull/2977 split into two

These tests now provide 100% coverage for all of the fundamental spaces

1. Update `Discrete.contains` to use `np.issubdtype` instead of `np.typecodes ` as this assumes to be a undocumented feature in numpy
2. Fix `Discrete.__setstate__` such that the state is set once
3. Add feature (/ bug fix) to `MultiBinary.sample` to have both 0, 1 and 2 for necessary 0 or 1 with 2 representing random
4. Make `MultiBinary.contains` to be positive (not negative) checks
5. Bug fix `MultiBinary.from_jsonable` with use the correct dtype
6. Fix type hints in `MultiDiscrete`
7. Change `MultiDiscrete.sample` to assume that the `nvec` is correct over the users mask
8. Make `MultiDiscrete.contains` to use positive (not negative) checks
9. Correctly implement `Text.sample` with masks
10. Add `Text` property changes for private variables: `_char_set`, `_char_list`, `_char_index`
11. Add util support for `Text` with the flattened character being the index in the character set.
12. Remove space-specific tests from `test_space` in favour of individual testing files. 
13. Add `tests/spaces/utils.py` where all testing spaces can be added such that new spaces added to `TESTING_SPACES` can be automatically added to all tests.

### DIFF
--- a/gym/spaces/box.py
+++ b/gym/spaces/box.py
@@ -164,7 +164,9 @@ class Box(Space[np.ndarray]):
         elif manner == "above":
             return above
         else:
-            raise ValueError("manner is not in {'below', 'above', 'both'}")
+            raise ValueError(
+                f"manner is not in {{'below', 'above', 'both'}}, actual value: {manner}"
+            )
 
     def sample(self, mask: None = None) -> np.ndarray:
         r"""Generates a single random sample inside the Box.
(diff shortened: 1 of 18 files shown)


younik (changes requested):
I cannot find any flaws in this big PR, except for the missing tests; thanks for addressing it.

PR #3050, opened 2022-08-30
Reduce the number of unexpected warnings during testing

pseudo-rnd-thoughts:
This PR aims to limit the number of unexpected warnings raised during testing to allow spotting of unexpecting errors / warnings

A majority of the changes are to add `with warnings.catch_warnings(record=True)` or `pytest.warns()` about the code that produces the warning


tests/envs/test_make.py
```diff
@@ -52,13 +53,14 @@ def test_make():
...
 
 def test_make_deprecated():
-    with pytest.raises(
-        gym.error.Error,
-        match=re.escape(
-            "Environment version v0 for `Humanoid` is deprecated. Please use `Humanoid-v4` instead."
-        ),
-    ):
-        gym.make("Humanoid-v0", disable_env_checker=True)
+    with warnings.catch_warnings(record=True):
```
Markus28, line 56: Don't we want to check whether the caught warnings can be safely ignored? As we do in /tests/envs/test_envs.py?

tests/envs/test_make.py
```diff
@@ -228,13 +230,16 @@ def test_make_render_mode():
...
 
     # Make sure that an additional error is thrown a user tries to use the wrapper on an environment with old API
-    with pytest.raises(
-        gym.error.Error,
-        match=re.escape(
-            "You passed render_mode='human' although test/NoHumanOldAPI-v0 doesn't implement human-rendering natively."
-        ),
-    ):
-        gym.make("test/NoHumanOldAPI-v0", render_mode="human", disable_env_checker=True)
+    with warnings.catch_warnings(record=True):
```
Markus28, line 233: Again, do we want to check the warnings?

tests/spaces/test_box.py
```diff
@@ -57,9 +58,11 @@ def test_box_shape_inference(box, expected_shape):
 )
 def test_box_values(value, valid):
     if valid:
-        with pytest.warns(None) as warnings:
+        with warnings.catch_warnings(record=True) as caught_warnings:
             Box(low=value, high=value)
-        assert len(warnings.list) == 0, tuple(warning.message for warning in warnings)
+        assert len(caught_warnings) == 0, tuple(
+            warning.message for warning in caught_warnings
+        )
```
Markus28, line 65: How about replacing this with
```python
with warnings.catch_warnings():
    warnings.simplefilter("error")
    Box(low=value, high=value)
```
Cf https://docs.pytest.org/en/7.0.x/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests

However, I believe that this should only be done in cases in which we do *not* also use `pytest.raises`, because otherwise `pytest.raises` might be satisfied with a warning that was promoted to an exception.

tests/utils/test_env_checker.py
```diff
@@ -47,10 +47,10 @@
 )
 def test_no_error_warnings(env):
     """A full version of this test with all gym envs is run in tests/envs/test_envs.py."""
-    with pytest.warns(None) as warnings:
+    with warnings.catch_warnings(record=True) as caught_warnings:
         check_env(env)
 
-    assert len(warnings) == 0, [warning.message for warning in warnings]
+    assert len(caught_warnings) == 0, [warning.message for warning in caught_warnings]
```
Markus28, line 53: Again, consider replacing it with 

```python
with warnings.catch_warnings():
    warnings.simplefilter("error")
    check_env(env)
```

tests/wrappers/test_pixel_observation.py
```diff
@@ -10,7 +10,7 @@
 
 
 class FakeEnvironment(gym.Env):
-    def __init__(self, render_mode=None):
+    def __init__(self, render_mode="single_rgb_array"):
```
Markus28, line 13: Why was this line changed?

Markus28 (approved):
Generally LGTM. CI needs to be fixed but this is probably not related to the PR.
I left some minor comments.

tests/envs/test_make.py
pseudo-rnd-thoughts: repeat of the same general issue

tests/spaces/test_box.py
pseudo-rnd-thoughts: As we are adding this to a limited number of cases then I dont think this should matter

tests/utils/test_env_checker.py
pseudo-rnd-thoughts: Personally, I like keeping warnings, warnings and errors, errors but this is possible solution

PR #3058, opened 2022-09-03
cliffwalking: remove annoying warnings

younik:
Remove `libpng warning: iCCP: known incorrect sRGB profile` warnings due to images profile


pseudo-rnd-thoughts:
LGTM

PR #3059, opened 2022-09-04
Correct the module scope of DeprecationWarning warning filter

wookayin:
# Description

This line mistakenly voids and nullifies all the previous warning filters for DeprecationWarning, registered by other packages or user programs via `warnings.simplefilter("ignore", DeprecationWarning, ...)` call which has an explicit intention of suppress some DeprecationWarning.

## Type of change

### Screenshots

N/A

# Checklist:

### DIFF
--- a/gym/logger.py
+++ b/gym/logger.py
@@ -14,7 +14,8 @@ DISABLED = 50
 min_level = 30
 
 
-warnings.simplefilter("once", DeprecationWarning)
+# Ensure DeprecationWarning to be displayed (#2685, #3059)
+warnings.filterwarnings("once", "", DeprecationWarning, module=r"^gym\.")
 
 
 def set_level(level: int):


wookayin:
This was added in #2685, which should not have been added. /cc @vwxyzjn

wookayin:
Or alternatively we could specify `module=gym` when adding a filter if we want a behavior described in #2685.

pseudo-rnd-thoughts:
I think it is better to not remove the warnings completely however Im not sure of what DeprecationWarnings the project still has. Would you be able to specify the module and keep the filter

wookayin:
Many of the deprecation warnings has gone since 3a8daafc, but there are some left (e.g., `output_truncation_bool=False` in `StepAPICompatibility`). One can grep with the keyword `deprecation`.

Indeed, we need a `filterwarnings` call for gym itself instead of completely removing in order to have the deprecation warnings from gym correctly displayed. I've changed it to specify the module.

FYI, here is a simple code one can reproduce this bug. I can add a formal tests (using pytest) -- let me work on it later (if needed).

```python
import PIL.Image
assert PIL.__version__.startswith('9')   # deprecated since 9.0+

import warnings
warnings.filterwarnings('ignore', message='.*will be removed in Pillow 10')

# With this import of gym (after some filterwarnings have been registered)
# will void the above filterwanings. (gym#2685)
import gym



# User code
# ---------

# Accessing to PIL will cause a DeprecationWarning:
# DeprecationWarning: NEAREST is deprecated and will be removed in Pillow 10
# (2023-07-01). Use Resampling.NEAREST or Dither.NONE instead.
assert PIL.Image.NEAREST == 0


# Using deprecated API in gym
# deprecation message will be printed only once, with the line
# 'warnings.simplefilter("once", DeprecationWarning) in gym/logger.py:17
for _ in range(2):
    gym.wrappers.StepAPICompatibility(gym.make("CartPole-v1"),
                                      output_truncation_bool=False)


# Expected output:
# no warnings for PIL.Image.NEAREST,
# deprecation warnings for "Initializing environment in old step API..."


# More context information for debugging
print ("""\n
----------- Registered filters -------------------------
(The first filter that matches only will have an effect)
""")
for f in warnings.filters:
    print(f)
```

pseudo-rnd-thoughts (approved):
LGTM @jkterry1

jkterry1:
@wookayin can you please fix tests?

PR #3060, opened 2022-09-04
environment agnostic {mode}_list render mode

younik:
After https://github.com/openai/gym/pull/3040, it makes sense to handle frame collection modes with a simple Wrapper (similar to HumanRendering).
This allows environments that can work with single rendering to support frame collection modes for free, while other environments can still define their own behavior by simply adding `rgb_array_list` (or whatever) to metadata.

This PR:
  - Removes Renderer util in favor of the new Wrapper `RenderCollection`
  - If an environment supports `rgb_array` (or another `mode`), it should not do anything to support `rgb_array_list` (or `mode_list`), because during making the environment will be wrapped with `RenderCollection` that will handle it
   - If an environment needs custom code for `rgb_array_list`, it can add `rgb_array_list` in metadata and the `RenderCollection` wrapper will not be applied. In this case, the environment should handle internally the mode `rgb_array_list`.


pseudo-rnd-thoughts:
LGTM

PR #3061, opened 2022-09-04
Fix pyright errors in envs and spaces folders

ldfrancis:
# Description
This PR aims to fix the `pyright` type errors that currently go unnoticed since the check is disabled in `pyproject.toml`

The proposed method of approach to accomplish this is to start with the `gym` folder and all sub-folders within. Then proceed to the `tests` folder. Once these errors are fixed, we then enable the check for type errors for the entire codebase.

At the moment, type errors are fixed for the folders;
`gym/envs` and `gym/spaces`

# Checklist:


gym/envs/toy_text/frozen_lake.py
```diff
@@ -61,11 +61,11 @@ def generate_random_map(size: int = 8, p: float = 0.8) -> List[str]:
         A random valid map
     """
     valid = False
-    board = []  # initialize to make pyright happy
+    board: List[List[str]] = []  # initialize to make pyright happy
```
pseudo-rnd-thoughts, line 64: `np.ndarray`? Though from experience typing with numpy is a pain

gym/envs/toy_text/frozen_lake.py
```diff
@@ -242,7 +242,7 @@ def update_probability_matrix(row, col, action):
         self.start_img = None
 
     def step(self, a):
-        transitions = self.P[self.s][a]
+        transitions = self.P[self.s][a]  # type: ignore
```
pseudo-rnd-thoughts, line 245: Im surprised this is an issue, is it due to numpy arrays?

pseudo-rnd-thoughts (approved):
Looks good, I just have two questions

gym/envs/toy_text/frozen_lake.py
ldfrancis: I can try it with a numpy array. It was initialized to a list because the validation function, `is_valid`, expects a `List[List[str]]`

gym/envs/toy_text/frozen_lake.py
ldfrancis: `self.P` is a dictionary of the form `Dict[int, Dict[int, list]]` 

The line fails the check for the following reasons:
1. The `step` method has no type annotation to indicate that the argument, `a`, is an `int`
2. `self.s` was not initialised in the constructor, it was done in the `reset` method, hence its type is not explicitly an int.

A walkaround would be to;
1. cast `self.s` and `a` to `int` before using as indices in `self.P`
2. initialize `self.s` in the constructor and type annotate the `step` method 
3. # type: ignore

ldfrancis: Casting `self.s` to `int` in the reset method also works.

gym/envs/toy_text/frozen_lake.py
ldfrancis: It works fine with a numpy array 🙂 
I had to change the `is_valid` function to accept `np.ndarray` instead of  `List[List[str]]`

pseudo-rnd-thoughts: Sorry, I thought they overwrote the board with a numpy. If they actually use a `list[list[str]]` keep it as that

gym/envs/toy_text/frozen_lake.py
pseudo-rnd-thoughts: I would just ignore the type for now.

gym/envs/toy_text/frozen_lake.py
ldfrancis: Okay

gym/envs/toy_text/frozen_lake.py
ldfrancis: Okay

pseudo-rnd-thoughts:
LGTM

gym/envs/mujoco/mujoco_env.py
```diff
@@ -391,9 +396,7 @@ def close(self):
             self.viewer.close()
         super().close()
 
-    def _get_viewer(
-        self, mode
-    ) -> Union["gym.envs.mujoco.Viewer", "gym.envs.mujoco.RenderContextOffscreen"]:
+    def _get_viewer(self, mode) -> Union[Viewer, RenderContextOffscreen]:
```
pseudo-rnd-thoughts, line 399: If you uninstall mujoco_py, then an error is raised due to this line I believe.
To replicate
`pip uninstall mujoco_py`
`import gym`
`gym.make("Ant-v4")`
`pip install "gym[mujoco_py]"`

ldfrancis: Yes, I've confirmed that it raises an error. Thanks for catching this. 
A solution would be to use;
```python
if TYPE_CHECKING:
    from gym.envs.mujoco import RenderContextOffscreen, Viewer
```
and 
```python
def _get_viewer(self, mode) -> Union["Viewer", "RenderContextOffscreen"]:
```
I'll now update this.

ldfrancis: There is a `ModuleNotFoundError` error when mujoco is uninstalled but it's not related to type checking.

pseudo-rnd-thoughts: That is because v4 environment use mujoco from while v2 and v3 environments use mujoco_py. Could you test with `Ant-v2`

ldfrancis: I get the same mujoco `ModuleNotFoundError`, coming from `gym\envs\mujoco\mujoco_rendering.py` when it is imported in `gym/envs/mujoco/__init__.py`

I can either remove the `mujoco_rendering` import in `__init__.py` or use a try-except in `mujoco_rendering.py` where mujoco is imported.

pseudo-rnd-thoughts: I think for now we don't worry about it and revert to the old implementation with type: ignores

ldfrancis: Okay

gym/envs/box2d/lunar_lander.py
```diff
@@ -489,11 +490,13 @@ def step(self, action):
...
 
         m_power = 0.0
-        if (self.continuous and action[0] > 0.0) or (
+
+        if (self.continuous and action[0] > 0.0) or (  # type: ignore
             not self.continuous and action == 2
         ):
             # Main engine
             if self.continuous:
+                assert isinstance(action, np.ndarray)
```
pseudo-rnd-thoughts, line 499: Im unsure about this, as pytorch or tensorflow Tensors would fail this

pseudo-rnd-thoughts:
@ldfrancis Hey, sorry I have been super busy with other stuff. could you fix the merge conflicts. I believe that #3072 will fix the mujoco issues that we were discussing and I have 1 comment to address

pseudo-rnd-thoughts (changes requested):

gym/envs/box2d/lunar_lander.py
ldfrancis: If a tensor is passed to the `step` method, the `np.clip` operation done previously ensures that it's converted to a numpy array. 
Moreover, the `step` method should only accept numpy arrays for the continuous case and an `int` for the discrete case, and users should ensure to supply the correct types (converting their tensors into numpy arrays).

gym/envs/box2d/car_racing.py
```diff
@@ -517,10 +517,11 @@ def reset(
             self.render()
         return self.step(None)[0], {}
 
-    def step(self, action: Union[np.ndarray, int]):
+    def step(self, action: Optional[Union[np.ndarray, int]]):
```
pseudo-rnd-thoughts, line 520: While I agree this is the correct typing, I don't think we want to encourage users to do this. Could we use type ignore on the reset

gym/envs/box2d/lunar_lander.py
pseudo-rnd-thoughts: Where is this `np.clip` happening? I can't see it in `step`

pseudo-rnd-thoughts (changes requested):

gym/envs/box2d/car_racing.py
ldfrancis: This is true. I would update it.

gym/envs/box2d/car_racing.py
```diff
@@ -517,10 +517,11 @@ def reset(
...
         assert self.car is not None
         if action is not None:
             if self.continuous:
+                assert isinstance(action, np.ndarray)
```
pseudo-rnd-thoughts, line 524: Sorry on the discussion on `np.clip` I was talking about Car Racing not Lunar Lander, my mistake

gym/envs/box2d/lunar_lander.py
pseudo-rnd-thoughts: Sorry, I was talking about the car racing not lunar lander I think this is good in lunar lander then

gym/envs/box2d/car_racing.py
ldfrancis: There is no call to `np.clip` in CarRacing. So passing a tensor would cause an error when the environment internals try to use the elements of the tensor. We can alleviate this by additionally calling `action = np.array(action)` which ensures tensor conversion to numpy arrays. 
Hence, as it is, the environment would not work with tensors, even without the assert that ensures actions are numpy arrays.

pseudo-rnd-thoughts: I would remove it for now

RedTachyon:
I'm slightly confused with how this PR works exactly. As far as I can tell, there are no changes to the pyproject.toml configuration, so there is no actual difference in which type checks now pass, correct? Since they weren't failing previously, and nothing new is enabled.

Given that, I'm strongly opposed to each of the `# type: ignore` annotations. They are band-aids which make sense in certain cases, but not as a blanket way to hide all errors/warnings - especially if we're not enabling any new type checks, **and** if there's so many of them.

There is a few other changes that I can see between the ignores, some of them are reasonable, some are potentially problematic (e.g. casting any `np.int` or a zero-dim `np.ndarray` to `int` in several places, I'd really rather not do that, and if pyright says it's not allowed, we should figure out why it's not allowed in the type system and fix that instead)

jkterry1:

PR #3063, opened 2022-09-04
fix: restore auto human rendering

younik:
Restore automatic rendering for human that was accidentally deleted with https://github.com/openai/gym/pull/3060


pseudo-rnd-thoughts:
LGTM but could you add a test that checks that all gym environments have an expected list of render modes, `rgb_array` and `human`

PR #3066, opened 2022-09-05
Add a complete compatibility wrapper

RedTachyon:
This should in principle make it so that any* legacy environment can be created and then converted into a valid modern environment.

WIP, I want to see if tests work for 3.6 (ugh) since I had to do a hack around protocols. I also still need to automatically integrate it into the `make` pipeline.

Brief explanation of the methodology: I implemented a [protocol](https://peps.python.org/pep-0544/) which will type check if something is a valid legacy env according to a minimal "mandatory" set of methods and fields. In python >= 3.8, this is available by default in `typing`. In 3.7, we need to use `typing_extensions`, which in turn does not support 3.6, for which the protocol check will be de facto disabled.

*Restrictions apply. Old environments expected to inherit the old `gym.Env` and will instead inherit from the new `gym.Env`, so there may be issues in some cases.

### DIFF
--- a/gym/wrappers/compatibility.py
+++ b/gym/wrappers/compatibility.py
@@ -101,6 +101,9 @@ class EnvCompatibility(gym.Env):
         """
         obs, reward, done, info = self.env.step(action)
 
+        if self.render_mode == "human":
+            self.render()
+
         return convert_to_terminated_truncated_step_api((obs, reward, done, info))
 
     def render(self) -> Any:
--- a/gym/wrappers/compatibility.py
+++ b/gym/wrappers/compatibility.py
@@ -52,7 +52,6 @@ class EnvCompatibility(gym.Env):
 
     Known limitations:
     - Environments that use `self.np_random` might not work as expected.
-    - `env.render_mode` cannot be set in `make` if the environment doesn't accept it as a parameter (e.g. via kwargs).
     """
 
     def __init__(self, old_env: LegacyEnv, render_mode: Optional[str] = None):
--- a/gym/wrappers/compatibility.py
+++ b/gym/wrappers/compatibility.py
@@ -114,3 +114,11 @@ class EnvCompatibility(gym.Env):
     def close(self):
         """Closes the environment."""
         self.env.close()
+
+    def __str__(self):
+        """Returns the wrapper name and the unwrapped environment string."""
+        return f"<{type(self).__name__}{self.env}>"
+
+    def __repr__(self):
+        """Returns the string representation of the wrapper."""
+        return str(self)
--- a/gym/wrappers/compatibility.py
+++ b/gym/wrappers/compatibility.py
@@ -88,6 +88,10 @@ class EnvCompatibility(gym.Env):
         if seed is not None:
             self.env.seed(seed)
         # Options are ignored
+
+        if self.render_mode == "human":
+            self.render()
+
         return self.env.reset(), {}
 
     def step(self, action: Any) -> Tuple[Any, float, bool, bool, Dict]:
--- a/gym/wrappers/compatibility.py
+++ b/gym/wrappers/compatibility.py
@@ -52,6 +52,7 @@ class EnvCompatibility(gym.Env):
 
     Known limitations:
     - Environments that use `self.np_random` might not work as expected.
+    - `env.render_mode` cannot be set in `make` if the environment doesn't accept it as a parameter (e.g. via kwargs).
 
     Args:
         old_env (gym.Env): the env to wrap. Can be in old or new API
--- a/gym/wrappers/compatibility.py
+++ b/gym/wrappers/compatibility.py
@@ -44,19 +44,15 @@ class LegacyEnv(Protocol):
 
 
 class EnvCompatibility(gym.Env):
-    r"""A wrapper which can transform an environment from new step API to old and vice-versa.
+    r"""A wrapper which can transform an environment from the old API to the new API.
 
-    Old step API refers to step() method returning (observation, reward, done, info)
-    New step API refers to step() method returning (observation, reward, terminated, truncated, info)
+    Old step API refers to step() method returning (observation, reward, done, info), and reset() only retuning the observation.
+    New step API refers to step() method returning (observation, reward, terminated, truncated, info) and reset() returning (observation, info).
     (Refer to docs for details on the API change)
 
     Known limitations:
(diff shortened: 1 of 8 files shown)


gym/wrappers/compatibility.py
```diff
@@ -0,0 +1,120 @@
...
+        Returns:
+            (observation, info)
+        """
+        if seed is not None:
+            self.env.seed(seed)
+        # Options are ignored
+        return self.env.reset(), {}
+
+    def step(self, action: Any) -> Tuple[Any, float, bool, bool, Dict]:
+        """Steps through the environment, returning 5 or 4 items depending on `apply_step_compatibility`.
```
pseudo-rnd-thoughts, line 98: Does `apply_step_compatibility` exist?

tests/envs/test_compatibility.py
```diff
@@ -0,0 +1,126 @@
...
+    env.unwrapped.render_mode = "rgb_array"
+    assert env.observation_space == Discrete(1)
+    assert env.action_space == Discrete(1)
+    assert env.reset() == (0, {})
+    assert env.reset(seed=0, options={"some": "option"}) == (0, {})
+    assert env.step(0) == (0, 0, False, False, {})
+    img = env.render()
+    assert isinstance(img, np.ndarray)
+    assert img.shape == (1, 1, 3)  # type: ignore
+    env.close()
```
pseudo-rnd-thoughts, line 111: Remove the environment from the registry

tests/envs/test_compatibility.py
```diff
@@ -0,0 +1,126 @@
...
+        id="LegacyTestEnv-v0",
+        entry_point=LegacyEnvExplicit,
+        apply_step_compatibility=True,
+    )
+    env = gym.make("LegacyTestEnv-v0")
+    env.unwrapped.render_mode = "rgb_array"
+    assert env.observation_space == Discrete(1)
+    assert env.action_space == Discrete(1)
+    assert env.reset() == (0, {})
+    assert env.reset(seed=0, options={"some": "option"}) == (0, {})
```
pseudo-rnd-thoughts, line 106: We don't have a test that reset actual calls seed correctly but Im not sure that matters

tests/envs/test_make.py
```diff
@@ -146,7 +140,6 @@ def test_apply_step_compatibility():
         max_episode_steps=3,
     )
     env = gym.make("testing-old-env")
-    assert has_wrapper(env, StepAPICompatibility)
```
pseudo-rnd-thoughts, line 149: Why is this assert removed?

pseudo-rnd-thoughts (changes requested):
Generally looks good. Do we not want to change the name in make and EnvSpec from`apply_step_compatibility` to `apply_env_compatibility`?

RedTachyon:
> Do we not want to change the name in make and EnvSpec from apply_step_compatibility to apply_env_compatibility?

Changed it to `apply_api_compatibility` to be a bit more informative, does that make sense?

gym/wrappers/compatibility.py
RedTachyon: Good catch, fixed

tests/envs/test_compatibility.py
RedTachyon: Done

tests/envs/test_make.py
RedTachyon: The old semantic for the compatibility layer was that it's a valid `gym.Wrapper`, because a `gym.Env` could be both in the old and new API. Now for an env to be a valid `gym.Env`, it has to follow the new API, which means that something written with the old API, isn't technically a `gym.Env`, i.e. it can't be the input to a `gym.Wrapper`.

The new compatibility layer is simply an implementation of a `gym.Env`, which happens to have a somewhat similar structure to a wrapper, but isn't one. So it used to be the case that an old-to-new environment is wrapped by StepAPICompatibility as a wrapper, which was checked in the test. Now that's not the case anymore.

pseudo-rnd-thoughts: Should the environment not come before the passive environment checker?

pseudo-rnd-thoughts (approved):
LGTM unless depending on if we want the passive environment checker to trigger for the old environments. Currently the warnings will occur.

Markus28:
I talked to @younik about auto human rendering today (i.e. not having to call `render` if render_mode is set to "human"; this feature was originally part of the new API, then accidentally removed, then added back in #3063). I think the compatibility wrapper should also implement this. 
With the current implementation, I believe that we need to keep calling `wrapped_env.render()` for human-rendering, which isn't in line with the new API.

RedTachyon:
I put the compatibility layer before the env checker, I agree it makes more sense this way. I think? I feel like some sort of an acknowledgment of "You shouldn't be doing this" is warranted (which would have been thrown up by the env checker), but at the same time, you have to enable this explicitly, so you know what you're doing.

---

About the rendering - I'm a bit out of loop with the details of those changes, as I understand we just want 
```python
if self.render_mode == "human":
    self.render()
```
inside `env.step`?

Markus28:
Yes, I think that's essentially right, though we would also want to add this to reset (and correspondingly, it should be at the very end of the step method).

RedTachyon:
In the last commit I added some logic to allow doing `gym.make("LegacyEnv", render_mode=...)`, does it make sense? It removes the `render_mode` argument from kwargs if we're using the compatibility layer, and redirects it to the wrapper.

RedTachyon:
(for anyone in the future wondering what's going on - I added an unrelated fix for #3068)

pseudo-rnd-thoughts:
LGTM

gym/wrappers/compatibility.py
```diff
@@ -0,0 +1,130 @@
...
+from typing import Any, Dict, Optional, Tuple
+
+import gym
+from gym.core import ObsType
+from gym.utils.step_api_compatibility import convert_to_terminated_truncated_step_api
+
+if sys.version_info >= (3, 8):
+    from typing import Protocol, runtime_checkable
+elif sys.version_info >= (3, 7):
+    from typing_extensions import Protocol, runtime_checkable
```
YouJiacheng, line 12: `typing_extension` does the version check internally, is it necessary to check here? pytorch seems to simply use `typing_extensions`.

pseudo-rnd-thoughts: Your right but this shouldn't change the implementation

PR #3081, opened 2022-09-09
Correctly change color of visited road tiles in CarRacing

foxik:
# Description

One of previous commit erroneously changed to code to overwrite `tile.color` instead of updating it in-place.

Specifically, commit c6deb81ad3b6 changed the lines
https://github.com/openai/gym/blob/07fd227a4fc0ecbd4fdac5563c0f48deeb5688ed/gym/envs/box2d/car_racing.py#L66-L68
into
https://github.com/openai/gym/blob/c6deb81ad3b6da7adbbffd4a2df2002a53861d8f/gym/envs/box2d/car_racing.py#L64-L65

However, the `tile.color` is used via reference in
https://github.com/openai/gym/blob/f4d668f9687ae417ff041cb1735966633d8ba759/gym/envs/box2d/car_racing.py#L449

## Type of change

### DIFF
--- a/gym/envs/box2d/car_racing.py
+++ b/gym/envs/box2d/car_racing.py
@@ -81,7 +81,7 @@ class FrictionDetector(contactListener):
             return
 
         # inherit tile color from env
-        tile.color = self.env.road_color / 255
+        tile.color[:] = self.env.road_color
         if not obj or "tiles" not in obj.__dict__:
             return
         if begin:


pseudo-rnd-thoughts:
Thanks for the PR, could you post an image of the original, the bugged version and the updated version?

foxik:
The original version (before the buggy update), commit 07fd227:
original

After the buggy update, commit c6deb81ad (the problem is the lighter road part under the car):
buggy

After the commit proposed here 846586fce5a5e4c160ec7fed7f1756b476def0c1:
fixed

pseudo-rnd-thoughts:
There is something weird with the right wheels of the car. Im not sure I see the difference very clearly but I guess it matters to a neural network

foxik:
The wheels "rotate" regularly (light gray rectangles move over them), which is unchanged in all the environments. You can see these gray parts also on the bottom of rear wheels in the first image; this the second one they are just "below" the wheels.

It is control by the `phase` variable
https://github.com/openai/gym/blob/f4d668f9687ae417ff041cb1735966633d8ba759/gym/envs/box2d/car_dynamics.py#L115
and you can see it being used in
https://github.com/openai/gym/blob/f4d668f9687ae417ff041cb1735966633d8ba759/gym/envs/box2d/car_dynamics.py#L297-L330

Overall, it is unrelated to the fix presented here.

pseudo-rnd-thoughts (approved):

RedTachyon:
I don't really see how this works. As I see, there are two changes in the PR:
1. You don't divide the color by 255 - this seems wrong, since the old representation of colors was in [0,1] (which is presumably what box2d expects), and self.env.road_color is in [0, 210], so we need to normalize it.
2. You assign the values by copying them. You say this is because later, the color is used by reference (is that right?). But I don't think this is a problem, since `tile.color = self.env.norm_road_color / 255` performs an implicit copy anyways due to the division.

Can you provide more explanation to what exactly is happening? In terms of the behavior (what color should it be in what situation and how is that wrong now), as well as the implementation.

The screenshot are helpful, but they should be more "standardized" - please use the same random seed, the same sequence of actions and the same timestep so that we can see what the difference really is

PR #3101, opened 2022-09-27
Feature/faster episode recorder

DavidSlayback:
# Description

This addresses two issues I've run into with RecordEpisodeStatistics

1) Silent list conversion in vectorized environments. Currently, RecordEpisodeStatistics converts terminated and truncated into lists, but obs and reward may still be arrays, and the new vectorized `info` interface uses arrays. I particularly run into this in my code if I do something like `dones.any()`. This PR leaves the underlying details of terminated/truncated alone by using NumPy array functionality.

2) Speed. Instead of looping through a potentially large batch of environments, this PR uses NumPy functions with their underlying vectorization to achieve the same task.

I considered building this with Jumpy, but I'm not sure on its status in the project, and it would also eat the cost of creating new arrays wherever I'm doing in-place modifications. 

I tested to make sure that this replicates the exact functionality of the original wrapper. To prove my speedup, I need to come up with some decent micro benchmarks; a couple attempts on Colab were fairly noisy.

## Type of change

- [x] Existing feature enhancement (non-breaking change which adds functionality)

# Checklist:

- [N/A] I have commented my code, particularly in hard-to-understand areas
- [N/A] I have made corresponding changes to the documentation
- [N/A] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/gym/wrappers/record_episode_statistics.py
+++ b/gym/wrappers/record_episode_statistics.py
@@ -92,7 +92,7 @@ class RecordEpisodeStatistics(gym.Wrapper):
                 "episode": {
                     "r": np.where(dones, self.episode_returns, 0),
                     "l": np.where(dones, self.episode_lengths, 0),
-                    "t": round(time.perf_counter() - self.t0, 6)
+                    "t": round(time.perf_counter() - self.t0, 6),
                 },
             }
             if self.is_vector_env:
--- a/gym/wrappers/record_episode_statistics.py
+++ b/gym/wrappers/record_episode_statistics.py
@@ -8,35 +8,6 @@ import numpy as np
 import gym
 
 
-def add_vector_episode_statistics(
-    info: dict, episode_info: dict, num_envs: int, env_num: int
-):
-    """Add episode statistics.
-
-    Add statistics coming from the vectorized environment.
-
-    Args:
-        info (dict): info dict of the environment.
-        episode_info (dict): episode statistics data.
-        num_envs (int): number of environments.
-        env_num (int): env number of the vectorized environments.
-
-    Returns:
-        info (dict): the input info dict with the episode statistics.
-    """
-    info["episode"] = info.get("episode", {})
-
-    info["_episode"] = info.get("_episode", np.zeros(num_envs, dtype=bool))
-    info["_episode"][env_num] = True
-
-    for k in episode_info.keys():
-        info_array = info["episode"].get(k, np.zeros(num_envs))
-        info_array[env_num] = episode_info[k]
-        info["episode"][k] = info_array
-
-    return info
-
-
 class RecordEpisodeStatistics(gym.Wrapper):
     """This wrapper will keep track of cumulative rewards and episode lengths.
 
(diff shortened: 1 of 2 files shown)


gym/wrappers/record_episode_statistics.py
```diff
@@ -114,38 +85,28 @@ def step(self, action):
...
-                else:
-                    infos = {**infos, **episode_info}
-                self.return_queue.append(episode_return)
-                self.length_queue.append(episode_length)
-                self.episode_count += 1
-                self.episode_returns[i] = 0
-                self.episode_lengths[i] = 0
+        dones = truncateds | terminateds
+        num_dones = np.sum(dones)
+        if num_dones:
```
pseudo-rnd-thoughts, line 90: Replace with `np.any(dones)`

tests/wrappers/test_record_episode_statistics.py
```diff
@@ -74,29 +73,3 @@ def test_wrong_wrapping_order():
 
     with pytest.raises(AssertionError):
         wrapped_env.step(wrapped_env.action_space.sample())
-
-
-def test_add_vector_episode_statistics():
```
pseudo-rnd-thoughts, line 79: Why have you removed this test? 
If it fails, could you fix it or replace it with an equivalent test

gym/wrappers/record_episode_statistics.py
```diff
@@ -8,35 +8,6 @@
 import gym
 
 
-def add_vector_episode_statistics(
```
pseudo-rnd-thoughts, line 11: Why have you removed this function?

pseudo-rnd-thoughts (changes requested):
Thanks for the PR, I think that the new code is nicer than the old code if it returns the same data. 
I have a couple of questions on the implement if you could ask them

gym/wrappers/record_episode_statistics.py
DavidSlayback: The function is only used for RecordEpisodeStatistics, and it undoes part of the point of this commit. 

Either way, when we finish an episode, we need to extend our `info` dictionary with the episode statistics in array form. 

In the original implementation, this was repeatedly called in a loop to produce statistics arrays and their corresponding masks, then insert them into the info dictionary

In this implementation, we create the exact same results by just using numpy indexing and the boolean done masks we already have. No need for an additional function that is used nowhere else

tests/wrappers/test_record_episode_statistics.py
DavidSlayback: If I remove the use of the function in RecordEpisodeStatistics, it is used nowhere else. If I delete the function, I suppose I could add a test that checks the returned EpisodeStatistics info object?

gym/wrappers/record_episode_statistics.py
DavidSlayback: I use the sum later to increment the episode count `self.episode_count += num_dones`

pseudo-rnd-thoughts: Sorry, I didn't see that. Ignore me

tests/wrappers/test_record_episode_statistics.py
pseudo-rnd-thoughts: Your right, I don't think we need to replace this function with anything

gym/wrappers/record_episode_statistics.py
pseudo-rnd-thoughts: Ok, that makes sense

pseudo-rnd-thoughts (approved):
Looks good to me

DavidSlayback:
Would it be more useful for me to move this PR over to gymnasium? Just saw that recently, didn't realize the devs were moving to a fork

pseudo-rnd-thoughts:
Yes, we were not accepting new features after Gymnasium goes live (soon). So feel free to close here and open on gymnasium if you want

PR #3111, opened 2022-10-03
Fix reset info being lost in vector environments

pseudo-rnd-thoughts:
As the new reset API returns `(obs, info)`. This was overwriting the `info` returned from the last `step` call

Reported: https://github.com/carlosluis/stable-baselines3/pull/1#issuecomment-1264295531

This PR fixes that issues through adding a new info key, "final_info" that contains all of the info in the last step in a similar way that "final_observation" contains the last observation


araffin (commented):
Look good but test and doc update are missing

pseudo-rnd-thoughts:
@araffin Tests added, documentation is not in this repo but I will add when also added to gymnasium

Issue #3108, opened 2022-10-02
gym env.render() went wrong with pygame.


jiapei100:
My code is as simple as like that:

```console
➜  ~ python
Python 3.10.6 (main, Aug 10 2022, 11:40:04) [GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import gym
>>> env = gym.make("Taxi-v3").env
>>> env.reset()
(121, {'prob': 1.0, 'action_mask': array([1, 1, 0, 1, 0, 0], dtype=int8)})
>>> env.render()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "~/.local/lib/python3.10/site-packages/gym/wrappers/order_enforcing.py", line 51, in render
    return self.env.render(*args, **kwargs)
  File "~/.local/lib/python3.10/site-packages/gym/wrappers/env_checker.py", line 53, in render
    return env_render_passive_checker(self.env, *args, **kwargs)
  File "~/.local/lib/python3.10/site-packages/gym/utils/passive_env_checker.py", line 316, in env_render_passive_checker
    result = env.render(*args, **kwargs)
  File "~/.local/lib/python3.10/site-packages/gym/envs/toy_text/taxi.py", line 284, in render
    return self._render_gui(self.render_mode)
  File "~/.local/lib/python3.10/site-packages/gym/envs/toy_text/taxi.py", line 302, in _render_gui
    assert (
AssertionError: Something went wrong with pygame. This should never happen.
>>> 
```

Looks like **pygame windows** is **NOT** initialized. Any suggestions?

**pygame** and **gym** version incompatibility?

```console
➜  ~ pip show pygame
Name: pygame
Version: 2.1.2
Summary: Python Game Development
Home-page: https://www.pygame.org
Author: A community project.
Author-email: pygame@pygame.org
License: LGPL
Location: /home/longervision/.local/lib/python3.10/site-packages
Requires: 
Required-by: 
➜  ~ pip show gym
Name: gym
Version: 0.26.1
Summary: Gym: A universal API for reinforcement learning environments
Home-page: https://www.gymlibrary.dev/
Author: Gym Community
Author-email: jkterry@umd.edu
License: MIT
Location: /home/longervision/.local/lib/python3.10/site-packages
Requires: cloudpickle, gym-notices, numpy
Required-by: 
➜  ~ 
```

pseudo-rnd-thoughts:
Render has been modified to use a new API.
[...]

PR #3112, opened 2022-10-03
ENH: add render warn for None

younik:
Add a warning when `render` is called without specifying `render_mode`, see https://github.com/openai/gym/issues/3108

### DIFF
--- a/gym/envs/box2d/bipedal_walker.py
+++ b/gym/envs/box2d/bipedal_walker.py
@@ -608,8 +608,8 @@ class BipedalWalker(gym.Env, EzPickle):
     def render(self):
         if self.render_mode is None:
             gym.logger.warn(
-                'You are calling render method without specifying any render mode. '
-                'You can specify the render_mode at initialization, '
+                "You are calling render method without specifying any render mode. "
+                "You can specify the render_mode at initialization, "
                 f'e.g. gym("{self.spec.id}", render_mode="rgb_array")'
             )
             return
--- a/gym/envs/box2d/bipedal_walker.py
+++ b/gym/envs/box2d/bipedal_walker.py
@@ -606,6 +606,12 @@ class BipedalWalker(gym.Env, EzPickle):
         return np.array(state, dtype=np.float32), reward, terminated, False, {}
 
     def render(self):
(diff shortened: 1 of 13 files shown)


gym/envs/box2d/bipedal_walker.py
```diff
@@ -608,8 +608,8 @@ def step(self, action: np.ndarray):
     def render(self):
         if self.render_mode is None:
             gym.logger.warn(
```
pseudo-rnd-thoughts, line 610: Should this be a warning or an error? I think it should be an error as otherwise users will ignore it and they shouldn't be ignoring it. 
Also I think, we should have the render_mode defaulting to human as that was the previous default used

younik: Since now in this if branch there is a return, calling render with `render_mode=None` simply results in doing nothing. Thus, I would not throw an error. I can see a user training an agent for a long time and render in the end resulting in an error while it can be just a warning since nothing bad happens. 

Regarding going back to `render_mode='human'` as default, if we want to do that, we would need again a deprecation period. Personally, I don't see the advantage of having `human` as default. I think rendering should be something that you explicitly turn on.

pseudo-rnd-thoughts: Sorry, about "human" I mean in the message instead of "rgb_array"
`gym.make("CartPole-v1", render_mode="human")`

younik: Ah, got it!
I didn't do that because with `render_mode="human"` you don't have to call `render`. It would be a bit confusing to suggest `human` as a warning because you called `render`. 
What do you think?

[Btw, right now, calling render with human works anyway without warning, it is just useless. We had a warning during the deprecation period and it is stated in doc.]

pseudo-rnd-thoughts: Agreed, but people most likely to get an issue is those calling render without an parameters which use to default to human. If users wanted a different render mode then they would have needed to enter a mode parameter and get an error prior to this warning. 

However, either mode would work

pseudo-rnd-thoughts (approved):

PR #3128, opened 2022-10-15
Change seeding error message to report seed type

### LINKED ISSUE #3127: [Bug Report] Error message after seeding with a non-int does not show incorrect seed type
**Description**
If the `seed` argument you pass to `env.reset(seed=seed)` is not an integer, the raised error is uninformative as it only includes the seed *value* not the seed *type*.

**Code example**
```
import numpy as np
import gym

rng = np.random.default_rng()
seed = rng.integers(2**32)

env = gym.make("MountainCarContinuous-v0")
env.reset(seed=seed)
```
raises
```
gym.error.Error: Seed must be a non-negative integer or omitted, not 1407292551
```

**System Info**
gym version 0.26.2, installed via pip

theo-brown:
# Description
Changed error message produced when `seed` argument of `env.reset(seed=seed)` is the incorrect type to include the *type* of the incorrect seed rather than the *value*. This is in line with most other Python error messages.

Fixes #3127

## Type of change
Bug fix (non-breaking change which fixes an issue)

# Example
```
import numpy as np
import gym

rng = np.random.default_rng()
seed = rng.integers(2**32)

env = gym.make("MountainCarContinuous-v0")
env.reset(seed=seed)
```
| Before | After |
|----------|--------|
| `gym.error.Error: Seed must be a non-negative integer or omitted, not 1407292551` | `gym.error.Error: Seed must be a non-negative integer or omitted, not <class 'numpy.int64'>` |

# Checklist:

**I have not followed the above checklist because I changed so little. Any problems with this PR exist externally in the openai/gym repo.**

### DIFF
--- a/gym/utils/seeding.py
+++ b/gym/utils/seeding.py
@@ -19,7 +19,7 @@ def np_random(seed: Optional[int] = None) -> Tuple[np.random.Generator, Any]:
         Error: Seed must be a non-negative integer or omitted
     """
     if seed is not None and not (isinstance(seed, int) and 0 <= seed):
-        raise error.Error(f"Seed must be a non-negative integer or omitted, not {seed}")
+        raise error.Error(f"Seed must be a non-negative integer or omitted, not {type(seed)}")
 
     seed_seq = np.random.SeedSequence(seed)
     np_seed = seed_seq.entropy
--- a/gym/utils/seeding.py
+++ b/gym/utils/seeding.py
@@ -19,7 +19,7 @@ def np_random(seed: Optional[int] = None) -> Tuple[np.random.Generator, Any]:
         Error: Seed must be a non-negative integer or omitted
     """
     if seed is not None and not (isinstance(seed, int) and 0 <= seed):
-        raise error.Error(f"Seed must be a non-negative integer or omitted, not {type(seed)}")
+        raise error.Error(f"Seed must be a non-negative integer or omitted, not {type(seed)} (got {seed})")
 
     seed_seq = np.random.SeedSequence(seed)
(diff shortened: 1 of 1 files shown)


balisujohn:
Thanks for your contribution! One small nitpick: I think it would be good to print both the type and the class. For reference, right now if I'm reading it right if I gave the reset function `seed = int(-1)`, I'd get the message

````
gym.error.Error: Seed must be a non-negative integer or omitted, not <class 'int'>
````

So maybe something like:

````
gym.error.Error: Seed must be a non-negative integer or omitted: current class: <class 'int'>, current value: -1
````

It seems like we already have adequate testing coverage for this function in [https://github.com/openai/gy/blob/master/tests/utils/test_seeding.py](https://github.com/openai/gym/blob/master/tests/utils/test_seeding.py) , so I don't think there is a super strong need to add a test.

**Comment by theo-brown** [comment]
I was following the [example TypeError from the Python docs](https://docs.python.org/3/tutorial/errors.html#exceptions), which doesn't print the value:
```
>>> '2' + 2
  File "<stdin>", line 1, in <module>
TypeError: can only concatenate str (not "int") to str
```

balisujohn:
LGTM :^)

pseudo-rnd-thoughts:
Hey, we just launched [gymnasium](https://github.com/Farama-Foundation/Gymnasium), a fork of Gym by the maintainers of Gym for the past 18 months where all maintenance and improvements will happen moving forward. Could you please move this over to the new repo?

If you'd like to read more about the story behind the backstory behind this and our plans going forward, click [here](https://farama.org/Announcing-The-Farama-Foundation).

